PR #24, opened 2017-10-02
Fix Travis CI failures / Upgrade Gradle to 4.2

trustin:
- Fix Travis CI failures
  - Unset _JAVA_OPTIONS which took too much heap (2GB) per JVM
- Upgrade Gradle to 4.2
  - Remove build-time-tracker plugin due to incompatibility
  - Add wrapper task setting so that *-all.zip is always used
  - Fix deprecation warnings


trustin:
gogradle plugin doesn't seem to work well on Windows unfortunately. Seems good otherwise.

minwoox:
Oops~ let me check~~

minwoox:
It seems like that gogradle does not handle `\` properly in the template used in build. So let's just remove it for now. could you add `.replaceAll("\\\\", "/")` after `binDir = "${goPath}/bin"` in the client-go build.gradle please?????

PR #27, opened 2017-10-10
Add the central dogma version to Go client automatically using gogradle

minwoox:
So, the releaser does not have to change the version again.


trustin:
Thanks, @minwoox !

PR #68, opened 2017-10-20
Lower the log level of the expected exceptions to DEBUG

minwoox:
Motivation:
Even though the expected exceptions during unit tests are literally expected, they are shown as
`WARN` level so can give some confuse to us.

Modification:
- Add `(expected exception)` and  lower the log level to `DEBUG` using `ExpectedExceptionAppender`

Result:
- Can distinguish the expected exceptions from unexpected ones

### DIFF
--- a/it/src/test/java/com/linecorp/centraldogma/it/GetFileTest.java
+++ b/it/src/test/java/com/linecorp/centraldogma/it/GetFileTest.java
@@ -20,6 +20,7 @@ import static com.linecorp.centraldogma.internal.thrift.ErrorCode.ENTRY_NOT_FOUN
 import static com.linecorp.centraldogma.internal.thrift.ErrorCode.PROJECT_NOT_FOUND;
 import static com.linecorp.centraldogma.internal.thrift.ErrorCode.QUERY_FAILURE;
 import static com.linecorp.centraldogma.internal.thrift.ErrorCode.REPOSITORY_NOT_FOUND;
+import static com.linecorp.centraldogma.testing.internal.ExpectedExceptionAppender.assertThatThrownByWithExpectedException;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 import java.util.concurrent.CompletionException;
@@ -31,6 +32,9 @@ import org.junit.Test;
 import com.linecorp.centraldogma.common.Query;
 import com.linecorp.centraldogma.common.Revision;
 import com.linecorp.centraldogma.internal.thrift.CentralDogmaException;
+import com.linecorp.centraldogma.server.internal.storage.project.ProjectNotFoundException;
+import com.linecorp.centraldogma.server.internal.storage.repository.EntryNotFoundException;
+import com.linecorp.centraldogma.server.internal.storage.repository.RepositoryNotFoundException;
 
 public class GetFileTest {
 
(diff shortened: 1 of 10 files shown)


testing-internal/src/main/java/com/linecorp/centraldogma/testing/internal/ExpectedExceptionAppender.java
```diff
@@ -0,0 +1,269 @@
...
+        }
+
+        @Override
+        public Object[] getArgumentArray() {
+            return event.getArgumentArray();
+        }
+
+        @Override
+        public Level getLevel() {
+            return Level.DEBUG;
```
trustin: How about returning `event.getLevel()` if it's lower than `Level.DEBUG`?

testing-internal/src/main/java/com/linecorp/centraldogma/testing/internal/ExpectedExceptionAppender.java
```diff
@@ -0,0 +1,269 @@
...
+     * {@code try-with-resources} block:
+     * <pre>{@code
+     * try (SafeCloseable ignored = ExpectedExceptionAppender.appendIfMatch(throwable, cause)) {
+     *     ...
+     *     // Expected exception happens.
+     *     ...
+     * }
+     * }</pre>
+     */
+    public static SafeCloseable appendIfMatch(Class<?> throwable, String cause) {
```
trustin: `appendIfMatch` makes me feel like the event will not be appended if not matches. How about: `withExpectedException()`? This also makes me think the following pattern may be better because `SafeCloseable` is unnecessary completely.

```java
withExpectedException(SomeException.class, "text", () -> {
    assertThatThrownBy(...);
});
```

minwoox: That's a good idea~!! Thank you~!

testing-internal/src/main/java/com/linecorp/centraldogma/testing/internal/ExpectedExceptionAppender.java
minwoox: Changed it~!!

testing-internal/src/main/java/com/linecorp/centraldogma/testing/internal/ExpectedExceptionAppender.java
```diff
@@ -0,0 +1,268 @@
...
+        } finally {
+            exceptionAndMessage.remove(throwable);
+        }
+    }
+
+    @Override
+    protected void append(ILoggingEvent eventObject) {
+        final IThrowableProxy exception = getException(eventObject);
+            for (String exceptionName : exceptionAndMessage.keySet()) {
+            if (exception != null && exception.getClassName().endsWith(exceptionName) &&
```
hyangtack: indent?

testing-internal/src/main/java/com/linecorp/centraldogma/testing/internal/ExpectedExceptionAppender.java
```diff
@@ -0,0 +1,268 @@
...
+     * Adds {@code "(expected exception)"} to the log message and lower its logging level to
+     * {@link Level#DEBUG} if the exception occurred from the {@code action} matches {@code throwable},
+     * and contains {@code message}.
+     */
+    public static void withExpectedException(String throwable, String message,
+                                             ThrowingRunnable action) throws Exception {
+        requireNonNull(throwable, "throwable");
+        requireNonNull(message, "message");
+        requireNonNull(action, "action");
+        exceptionAndMessage.put(throwable, message);
```
hyangtack: Just curious. Is there any chance to put messages with the same `throwable` at the same time?

minwoox: I think it's OK~! :-)
I was intending to use this only in the scope of `run()` in `ThrowingRunnable`. If the code emits more than one expected exception, then it must be separated in different `run()`. This is used only in unit tests and it is executed sequentially, so it's ok.

testing-internal/src/main/java/com/linecorp/centraldogma/testing/internal/ExpectedExceptionAppender.java
minwoox: oops~!!

it/src/test/java/com/linecorp/centraldogma/it/GetFileTest.java
```diff
@@ -42,34 +46,47 @@ public void testInvalidJsonPath() throws TException {
         assertThatThrownBy(() -> rule.client().getFile(
                 rule.project(), rule.repo1(), Revision.HEAD,
                 Query.ofJsonPath("/test/test2.json", "$.non_exist_path")).join())
-                .isInstanceOf(CompletionException.class).hasCauseInstanceOf(CentralDogmaException.class)
+                .isInstanceOf(CompletionException.class).hasCauseInstanceOf(
+                CentralDogmaException.class)
```
trustin: Seems unrelated?

it/src/test/java/com/linecorp/centraldogma/it/GetFileTest.java
```diff
@@ -42,34 +46,47 @@ public void testInvalidJsonPath() throws TException {
...
-    public void testInvalidFile() throws TException {
-        assertThatThrownBy(() -> rule.client().getFile(
-                rule.project(), rule.repo1(), Revision.HEAD,
-                Query.ofJsonPath("/test/non_existing_file.json", "$.a")).join())
-                .isInstanceOf(CompletionException.class).hasCauseInstanceOf(CentralDogmaException.class)
-                .matches(e -> ((CentralDogmaException) e.getCause()).getErrorCode() == ENTRY_NOT_FOUND);
+    public void testInvalidFile() throws Exception {
+        ExpectedExceptionAppender.withExpectedException(
+                EntryNotFoundException.class, "non_existing_file", () ->
+                        assertThatThrownBy(() -> rule.client().getFile(
```
trustin: It seems like it's fairly common to call a single `assertThatThrownBy()` with `withExpectedException`. How about adding this:

```java
import static ...ExpectedExceptionAppender.assertThatThrownByWithExpectedException;
assertThatThrownByWithExpectedException(
        EntryNotFoundException.class, "non_existing_file",
        () -> rule.client().....).isInstanceOf(...);
```

it/src/test/java/com/linecorp/centraldogma/it/WatchTest.java
```diff
@@ -79,14 +78,19 @@ public void testWatchRepository() throws Exception {
         final Revision rev2 = commit.revision();
 
         assertThat(rev2).isEqualTo(rev1.forward(1));
-        assertThat(future.get(3, TimeUnit.SECONDS)).isEqualTo(rev2);
+        ExpectedExceptionAppender.withExpectedException(
+                "AnnotatedConnectException", "Connection refused: /127.0.0.1:", () ->
```
trustin: 'Connection refused' and IP address are platform-dependent strings. Could we only check the exception type?

testing-internal/src/main/java/com/linecorp/centraldogma/testing/internal/ExpectedExceptionAppender.java
hyangtack: Ok! Thank you for explanation! :-)

it/src/test/java/com/linecorp/centraldogma/it/GetFileTest.java
minwoox: That's a brilliant idea~! Thank you~!!

trustin:
Thanks, @minwoox !

PR #70, opened 2017-10-20
Remove copying the go client binaries when gradlew startup

minwoox:
Motivation:
Every time the server starts using `gradlew startup`, it copies the go client binaries under dist folder.
Most of the tasks in gogradle does not support the input and output yet, it does not skip the tasks although there's no modified sources. Therefore, It takes a lot to startup.

Modifications:
- Remove copying the go client binaries when gradlew startup

Result:
- Starting ther server is faster

### DIFF
--- a/dist/build.gradle
+++ b/dist/build.gradle
@@ -75,20 +75,24 @@ task copyClientBinaries(group: 'Build',
     into "${project.ext.distDir}/bin/native"
 }
 
-task distDir(group: 'Build',
-             description: "Builds a distribution directory (${project.ext.relativeDistDir})",
-             dependsOn: [tasks.copyLicenses, tasks.copyLib, tasks.copyClientBinaries])
+task distDirForStartup(group: 'Build',
+        description: "Builds a distribution directory with libraries for startup task (${project.ext.relativeDistDir})",
+        dependsOn: [tasks.copyLicenses, tasks.copyLib])
 
 // Create the tasks that copy each directory under src/ into dist/
 ['bin', 'conf'].each { dirName ->
     def taskName = "copy${CaseFormat.LOWER_HYPHEN.to(CaseFormat.UPPER_CAMEL, dirName)}"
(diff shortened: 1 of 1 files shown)


dist/build.gradle
```diff
@@ -79,15 +79,21 @@ task distDir(group: 'Build',
...
 
+task distDirWithoutClientBinaries(group: 'Build',
+        description: "Builds a distribution directory without client binaries (${project.ext.relativeDistDir})",
+        dependsOn: [tasks.copyLicenses, tasks.copyLib])
+
 // Create the tasks that copy each directory under src/ into dist/
 ['bin', 'conf'].each { dirName ->
     def taskName = "copy${CaseFormat.LOWER_HYPHEN.to(CaseFormat.UPPER_CAMEL, dirName)}"
-    tasks.distDir.dependsOn(tasks.create(taskName, Copy) {
+    def copyTasks = tasks.create(taskName, Copy) {
```
trustin: - copyTasks -> copyTask
- With my earlier comment, you wouldn't need to define a local variable here.

dist/build.gradle
```diff
@@ -79,15 +79,21 @@ task distDir(group: 'Build',
              description: "Builds a distribution directory (${project.ext.relativeDistDir})",
              dependsOn: [tasks.copyLicenses, tasks.copyLib, tasks.copyClientBinaries])
```
trustin: You could make `distDir` depend on `distDirWithoutClientBinaries` to reduce duplication of dependencies.

dist/build.gradle
```diff
@@ -79,15 +79,21 @@ task distDir(group: 'Build',
...
 
+task distDirWithoutClientBinaries(group: 'Build',
```
trustin: How about `distDirServerOnly`?

minwoox:
A lot of tasks in gogradle does not support inputs and ouputs yet. https://github.com/gogradle/gogradle/issues/131

minwoox:
Replaced by #73

PR #73, opened 2017-10-24
Add inputs and outpus to gogradle

minwoox:
Motivation:
Do not need to compile and test again when there's no modified files.

Modification:
- Add inputs and outpus to test and build* tasks

### DIFF
--- a/client/go/build.gradle
+++ b/client/go/build.gradle
@@ -1,3 +1,5 @@
+import com.google.common.base.CaseFormat
+
 plugins {
     id 'com.github.blindpirate.gogradle' version '0.7.0'
 }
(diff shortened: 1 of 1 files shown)


client/go/build.gradle
```diff
@@ -48,3 +55,17 @@ build {
         ant.chmod(file: "$project.ext.binDir" + '/dogma.windows-amd64.exe', perm: '644')
     }
 }
+
+FileTree goSources = fileTree("${project.projectDir}").include('**/*.go').exclude('.gogradle/**')
+
+tasks.whenTaskAdded { task ->
+    def taskName = task.name
+    if (taskName.startsWith('build')) {
```
trustin: But it should not intercept the task 'build', right?

client/go/build.gradle
```diff
@@ -32,6 +34,11 @@ ext {
     binDir = "${goPath}/bin".replaceAll('\\\\', '/')
 }
 
+test {
+    inputs.files(fileTree("${project.projectDir}").include('**/*_test.go'))
+    outputs.dir("${project.projectDir}/.gogradle/reports/test")
+}
+
 // Gogradle puts several comman code check task into a check task. We need to explicitly specify this line
```
hyangtack: comman? common?

client/go/build.gradle
minwoox: Oops~

client/go/build.gradle
minwoox: Thank you~!

trustin:
Thanks, @minwoox

PR #75, opened 2017-10-24
Fix trivial typo

huydx:
Function signature

```
    default <T> Watcher<T> fileWatcher(String projectName, String repositoryName, Query<T> query) {
```


trustin:
Thanks, @huydx !

PR #81, opened 2017-10-27
Make CentralDogmaBeanFactory able to handle change callback

huydx:


client/java/src/main/java/com/linecorp/centraldogma/client/updater/CentralDogmaBeanFactory.java
```diff
@@ -93,22 +108,37 @@ public CentralDogmaBeanFactory(CentralDogma dogma, ObjectMapper objectMapper) {
...
      * @param beanType the type of {@code bean}
+     * @param changeListener {@link Consumer} of {@code beanType}, callback function when {@code bean} value has been updated
      * @param overrides the {@link CentralDogmaBeanConfig} whose properties will override the settings
      *                  specified by the {@link CentralDogmaBean} annotation
      *
      * @return a new Java bean whose getters return the latest known values mirrored from Central Dogma
      */
-    public <T> T get(T bean, Class<T> beanType, CentralDogmaBeanConfig overrides) {
+    public <T> T get(T bean, Class<T> beanType, Consumer<T> changeListener,
+                                                      CentralDogmaBeanConfig overrides) {
```
hyangtack: indent?

client/java/src/main/java/com/linecorp/centraldogma/client/updater/CentralDogmaBeanFactory.java
```diff
@@ -156,8 +188,8 @@ public CentralDogmaBeanFactory(CentralDogma dogma, ObjectMapper objectMapper) {
      *                              it got interrupted while waiting for the initial value
      */
     @SuppressWarnings("unchecked")
-    public <T> T get(T bean, Class<T> beanType, CentralDogmaBeanConfig overrides,
-                     long initialValueTimeout, TimeUnit initialValueTimeoutUnit)
+    public <T> T get(T bean, Class<T> beanType, Consumer<T> changeListener,
+                                                      CentralDogmaBeanConfig overrides, long initialValueTimeout, TimeUnit initialValueTimeoutUnit)
```
hyangtack: indent?

client/java/src/main/java/com/linecorp/centraldogma/client/updater/CentralDogmaBeanFactory.java
```diff
@@ -56,13 +58,26 @@
  * > public class Foo {
  * >     private int a;
  * >     private String b;
+ * >     private Revision rev;
  * >
  * >     public int getA() { return a; }
  * >     public void setA(int a) { this.a = a; }
  * >     public String getB() { return b; }
  * >     public void setB(String b) { this.b = b; }
+ * >     public Revision getRevision() { return this.rev; }
```
trustin: The field `rev` isn't used at all actually. We could just return null here, add some comment that this method is implemented by the factory and remove the field `rev`.

client/java/src/main/java/com/linecorp/centraldogma/client/updater/CentralDogmaBeanFactory.java
```diff
@@ -56,13 +58,26 @@
...
  * > }
  * }</pre>
+ *
+ * In the following example, callback will be called when new value is updated
```
trustin: new value is updated -> a property is updated with a new value

client/java/src/main/java/com/linecorp/centraldogma/client/updater/CentralDogmaBeanFactory.java
```diff
@@ -56,13 +58,26 @@
...
+ * <pre>{@code
+ *  CentralDogmaFactory factory = new CentralDogmaFactory(dogma, new ObjectMapper());
+ *  Consumer<Foo> fooUpdatedListener = (Foo f) -> {
+ *     System.out.println("foo has updated to: " + f);
+ *  }
```
trustin: - (Foo f) -> f
- Missing semicolon after }
- Unnecessary extra single space in every line?

  ```
  *  CentralDogmaFactory .. ->
  * CentralDogmaFactory ..
  ```

client/java/src/main/java/com/linecorp/centraldogma/client/updater/CentralDogmaBeanFactory.java
```diff
@@ -93,22 +108,38 @@ public CentralDogmaBeanFactory(CentralDogma dogma, ObjectMapper objectMapper) {
...
+        return get(bean, beanType, (T x) -> {
+        }, CentralDogmaBeanConfig.EMPTY);
+    }
+
+    /**
+     * Returns a newly-created bean instance with the settings specified by {@link CentralDogmaBean} annotation.
+     *
+     * @param bean a Java bean annotated with {@link CentralDogmaBean}
+     * @param beanType the type of {@code bean}
+     * @param changeListener {@link Consumer} of {@code beanType}, invoke when {@code bean} is updated
```
trustin: - `the {@link Consumer}`
- invoke -> invoked

client/java/src/main/java/com/linecorp/centraldogma/client/updater/CentralDogmaBeanFactory.java
```diff
@@ -93,22 +108,38 @@ public CentralDogmaBeanFactory(CentralDogma dogma, ObjectMapper objectMapper) {
...
+    public <T> T get(T bean, Class<T> beanType, Consumer<T> changeListener) {
+        return get(bean, beanType, changeListener, CentralDogmaBeanConfig.EMPTY);
     }
 
     /**
      * Returns a newly-created bean instance with some or all of its settings overridden.
      *
      * @param bean a Java bean annotated with {@link CentralDogmaBean}
      * @param beanType the type of {@code bean}
+     * @param changeListener {@link Consumer} of {@code beanType}, invoke when {@code bean} is updated
```
trustin: - `the {@link Consumer}`
- invoke -> invoked

client/java/src/main/java/com/linecorp/centraldogma/client/updater/CentralDogmaBeanFactory.java
```diff
@@ -120,6 +151,7 @@ public CentralDogmaBeanFactory(CentralDogma dogma, ObjectMapper objectMapper) {
...
```
trustin: - `the {@link Consumer}`
- invoke -> invoked

client/java/src/main/java/com/linecorp/centraldogma/client/updater/CentralDogmaBeanFactory.java
```diff
@@ -93,22 +108,38 @@ public CentralDogmaBeanFactory(CentralDogma dogma, ObjectMapper objectMapper) {
...
      * @param overrides the {@link CentralDogmaBeanConfig} whose properties will override the settings
      *                  specified by the {@link CentralDogmaBean} annotation
      *
      * @return a new Java bean whose getters return the latest known values mirrored from Central Dogma
      */
-    public <T> T get(T bean, Class<T> beanType, CentralDogmaBeanConfig overrides) {
```
trustin: Could you retain all old methods without the listeners?

it/src/test/java/com/linecorp/centraldogma/it/updater/CentralDogmaBeanTest.java
```diff
@@ -113,13 +120,19 @@ public void overrideSettings() throws Exception {
...
 
     @CentralDogmaBean(project = "a", repository = "b", path = "/c.json")
     static class TestProperty {
         int foo = 10;
         String bar = "20";
         List<String> qux = ImmutableList.of("x", "y", "z");
+        Revision revision;
+
+        public Revision getRevision() {
+            return revision;
```
trustin: The field `revision` isn't used at all actually. We could just return null here and remove the field `revision`.

client/java/src/main/java/com/linecorp/centraldogma/client/updater/CentralDogmaBeanFactory.java
```diff
@@ -56,13 +58,26 @@
...
+ * In the following example, callback will be called when new value is updated
+ * <pre>{@code
+ *  CentralDogmaFactory factory = new CentralDogmaFactory(dogma, new ObjectMapper());
+ *  Consumer<Foo> fooUpdatedListener = (Foo f) -> {
+ *     System.out.println("foo has updated to: " + f);
+ *  }
+ *  Foo mirroredFoo = factory.get(new Foo(), Foo.class, fooUpdatedListener);
+ * }
+ * </pre>
+ * </p>
```
trustin: You should not close the `<p>` tag because Javadoc is damn HTML4 :-D

client/java/src/main/java/com/linecorp/centraldogma/client/updater/CloseMethodDelegate.java
```diff
@@ -0,0 +1,33 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.linecorp.centraldogma.client.updater;
+
+import java.lang.reflect.Method;
+
+class CloseMethodDelegate implements MethodDelegate {
```
trustin: Very nice! @minwoox and I discussed about something similar yesterday actually. 👍

client/java/src/main/java/com/linecorp/centraldogma/client/updater/RevisionMethodDelegate.java
```diff
@@ -0,0 +1,38 @@
...
+        return thisMethod.getReturnType() == Revision.class &&
+               thisMethod.getParameterCount() == 0;
+    }
+
+    @Override
+    public <T> Object invoke(CentralDogmaBeanMethodHandler<T> handler, Object self, Method thisMethod,
+                             Method proceed, Object[] args) throws Throwable {
+        try {
+            return handler.watcher.latest().revision();
+        } catch (IllegalStateException ignore) { // gracefully handle "value not available yet" error
```
trustin: You could check if `watcher.initialValueFuture().isDone()` instead.

client/java/src/main/java/com/linecorp/centraldogma/client/updater/CentralDogmaBeanMethodHandler.java
```diff
@@ -18,24 +18,46 @@
...
     }
 
+    // note that below delegate will be match from top to bottom
+    // so if your matcher may duplicate with another one, please consider the order carefully
+    private static final List<MethodDelegate> methodDelegators = ImmutableList.of(
+            new RevisionMethodDelegate(),
+            new CloseMethodDelegate()
+    );
+
+    private static final MethodDelegate defaultDelegator = new DefaultMethodDelegate();
```
trustin: You could just move this into the last position of `methodDelegators`.

huydx: i just want to prevent case that some people forgot to see the comment and put default one into the middle of delegator chain, which may cause it to ignore everything else after.
HDYT?

trustin: Then how about doing this:

```java
methodDelegators = ... (
    ...
    // The last delegator must be DefaultMethodDelegator.
    new DefaultMethodDelegator());

for (d : methodDelegators) {
    ...
}

// Should never reach here.
throw new Error();
```

client/java/src/main/java/com/linecorp/centraldogma/client/updater/CloseMethodDelegate.java
minwoox: This is brilliant~!!

client/java/src/main/java/com/linecorp/centraldogma/client/updater/CentralDogmaBeanMethodHandler.java
huydx: sounds reasonable to me, applied

client/java/src/main/java/com/linecorp/centraldogma/client/updater/CentralDogmaBeanFactory.java
```diff
@@ -44,7 +45,8 @@
  * Creates a new bean instance that mirrors its properties from Central Dogma.
  *
  * <p>In the following example, {@code mirroredFoo.getA()} and {@code getB()} will return the latest known
- * values retrieved from {@code /myProject/myRepo/foo.json} in Central Dogma:
+ * values, along with latest revision retrieved from {@code /myProject/myRepo/foo.json} in Central Dogma.
```
minwoox: the latest revision?

client/java/src/main/java/com/linecorp/centraldogma/client/updater/CentralDogmaBeanFactory.java
```diff
@@ -44,7 +45,8 @@
...
+ * If value is not available yet, revision will be set to null.
```
minwoox: the value? the revision?

client/java/src/main/java/com/linecorp/centraldogma/client/updater/CentralDogmaBeanFactory.java
```diff
@@ -61,8 +63,21 @@
  * >     public void setA(int a) { this.a = a; }
  * >     public String getB() { return b; }
  * >     public void setB(String b) { this.b = b; }
+ * >     public Revision getRevision() { return null; }
  * > }
  * }</pre>
+ *
+ * <p>In the following example, callback will be called when a property is updated with a new value
```
minwoox: end with a dot?

client/java/src/main/java/com/linecorp/centraldogma/client/updater/CentralDogmaBeanFactory.java
```diff
@@ -61,8 +63,21 @@
...
+ *
+ * <pre>{@code
+ * > CentralDogmaFactory factory = new CentralDogmaFactory(dogma, new ObjectMapper());
```
minwoox: You don't have to use `>` here and following lines. It is used when the line starts with `@`

client/java/src/main/java/com/linecorp/centraldogma/client/updater/CentralDogmaBeanFactory.java
```diff
@@ -61,8 +63,21 @@
...
+ * > Consumer<Foo> fooUpdatedListener = (Foo f) -> {
+ * >    System.out.println("foo has updated to: " + f);
```
minwoox: indentation? (three spaces)

client/java/src/main/java/com/linecorp/centraldogma/client/updater/CentralDogmaBeanFactory.java
trustin: new value -> new value: (missing colon)

client/java/src/main/java/com/linecorp/centraldogma/client/updater/CentralDogmaBeanFactory.java
```diff
@@ -61,8 +63,21 @@
...
+ * > };
+ * >
+ * > Foo mirroredFoo = factory.get(new Foo(), Foo.class, fooUpdatedListener);
+ * > }
+ * </pre>
```
trustin: `}</pre>` ? (no need to break the line)

client/java/src/main/java/com/linecorp/centraldogma/client/updater/CentralDogmaBeanMethodHandler.java
```diff
@@ -18,24 +18,48 @@
...
 
 import com.linecorp.centraldogma.client.Watcher;
 
 import javassist.util.proxy.MethodHandler;
 
-class CentralDogmaBeanMethodHandler<T> implements MethodHandler {
-    private final Watcher<T> watcher;
-    private final T defaultValue;
+/**
+ * This class Handle method dispatch from original bean to the bean proxy.
```
minwoox: Handle -> handles? the original bean?

client/java/src/main/java/com/linecorp/centraldogma/client/updater/CentralDogmaBeanMethodHandler.java
```diff
@@ -18,24 +18,48 @@
...
         thisMethod.setAccessible(true);
-        return thisMethod.invoke(watcher.latestValue(defaultValue), args);
+        for (MethodDelegate md : methodDelegators) {
+            if (md.match(thisMethod)) {
+                return md.invoke(this, self, thisMethod, proceed, args);
+            }
+        }
+
+        // it must not reach here
+        throw new IllegalStateException("method is not handled");
```
trustin: Could you add class and method name to the exception message? i.e.

    "method not handled: " + thisMethod

client/java/src/main/java/com/linecorp/centraldogma/client/updater/CentralDogmaBeanMethodHandler.java
```diff
@@ -18,24 +18,48 @@
...
 
 import javassist.util.proxy.MethodHandler;
 
-class CentralDogmaBeanMethodHandler<T> implements MethodHandler {
-    private final Watcher<T> watcher;
-    private final T defaultValue;
+/**
+ * This class Handle method dispatch from original bean to the bean proxy.
+ *
+ * @param <T> type of bean will be handle of this class
```
minwoox: how about just `the type of the bean`?

client/java/src/main/java/com/linecorp/centraldogma/client/updater/CentralDogmaBeanFactory.java
```diff
@@ -61,8 +63,20 @@
...
  * >     public String getB() { return b; }
  * >     public void setB(String b) { this.b = b; }
+ * >     public Revision getRevision() { return null; }
  * > }
  * }</pre>
+ *
+ * <p>In the following example, callback will be called when a property is updated with a new value:
+ *
+ * <pre>{@code
+ * > CentralDogmaFactory factory = new CentralDogmaFactory(dogma, new ObjectMapper());
```
minwoox: You don't have to use `>` here. It needs when the line starts with `@`.

it/src/test/java/com/linecorp/centraldogma/it/updater/CentralDogmaBeanTest.java
```diff
@@ -87,6 +93,23 @@ public void test() throws Exception {
 
         assertThat(property.getBar()).isEqualTo("Y");
         assertThat(property.getQux()).containsExactly("0", "1");
+        assertThat(property.getRevision()).isEqualTo(new Revision("2"));
+        assertThat(called[0]).isEqualTo(1);
+
+        property.closeWatcher();
```
minwoox: Could you do this in the `overrideSettings()` as well please?

minwoox: I'm reaaaaaaaaaaaaaaaaaally sorry for this ambiguous word. I was just asking you to insert only the `property.closeWatcher()` to the `overrideSettings()`.
When there's no way to close the watcher, it was closed with an exception some times after  the `property` is created, and it polluted the logs from gradle build. That is why I asked you to close it explicitly. :-)

huydx: I see, thanks for the explaination, fixed :)

client/java/src/main/java/com/linecorp/centraldogma/client/updater/CentralDogmaBeanFactory.java
```diff
@@ -44,24 +45,37 @@
...
- * > public class Foo {
- * >     private int a;
- * >     private String b;
- * >
- * >     public int getA() { return a; }
- * >     public void setA(int a) { this.a = a; }
- * >     public String getB() { return b; }
- * >     public void setB(String b) { this.b = b; }
- * > }
+ *  CentralDogmaFactory factory = new CentralDogmaFactory(dogma, new ObjectMapper());
```
minwoox: Could you add `>` in front of lines from 52 to 67 and replace `{@literal @}` with `@` please?
I was asking you just remove the `>` from 73 to 75 lines.
The reason why we used `>` here is java-doc misinterpret when the line in `{@code }` block starts with `@`.
For example:
```
{@ code
...
@override
...
}
```
Although all of the codes in `{@code }` are displayed as they are, the above case is not working as we expect.
So we have decided to put `>` at the front of every line when there is a line which starts with `@` in the `{@code }` block.
The `>` is removed when the gradle builds, so we can see the the clean java-docs. :-)

client/java/src/main/java/com/linecorp/centraldogma/client/updater/CentralDogmaBeanFactory.java
```diff
@@ -44,24 +45,37 @@
...
+ *      public Revision getRevision() { return null; }
+ *  }
+ * }</pre>
+ *
+ * <p>In the following example, callback will be called when a property is updated with a new value:
+ *
+ * <pre>{@code
+ *  CentralDogmaFactory factory = new CentralDogmaFactory(dogma, new ObjectMapper());
+ *  Consumer<Foo> fooUpdatedListener = (Foo f) -> {
+ *     System.out.println("foo has updated to: " + f);
```
minwoox: three spaces here? or you can remove the space at the very front of lines(73, 74, 76).

huydx:
@minwoox fixed, thanks for your patience :)

client/java/src/main/java/com/linecorp/centraldogma/client/updater/CentralDogmaBeanFactory.java
```diff
@@ -180,6 +281,7 @@ public CentralDogmaBeanFactory(CentralDogma dogma, ObjectMapper objectMapper) {
                 buildQuery(settings),
                 jsonNode -> {
                     try {
+                        changeListener.accept(bean);
```
trustin: Could we surround this sentence with another `try-catch` block so that a user does not get a misleading exception message `Failed to convert...`?

huydx: nice catch, how about just move it outside to let it throw whatever exception received from accept?

trustin: That sounds fine.

huydx:
@trustin PTAL

trustin:
@huydx OK! Please ping us when it's ready.

huydx:
@trustin added, PTAL~

it/src/test/java/com/linecorp/centraldogma/it/updater/CentralDogmaBeanTest.java
```diff
@@ -87,6 +93,46 @@ public void test() throws Exception {
...
+        client.push("a", "b", Revision.HEAD, Author.SYSTEM, "Modify a.json",
+                    Change.ofJsonUpsert("/c.json",
+                                        '{' +
+                                        "  \"foo\": 50," +
+                                        "  \"bar\": \"Y2\"," +
+                                        "  \"qux\": [\"M\", \"N\"]" +
+                                        '}'))
+              .join();
+        // TODO(huydx): this test may be flaky, is there any better way?
+        Throwable thrown = catchThrowable(() -> await().atMost(2, TimeUnit.SECONDS)
```
trustin: I think it's fine. Perhaps we don't even need to specify the timeout (2 seconds)? The default timeout of 10 seconds should be OK.

trustin: Let me merge once you remove `.atMost()`. Great job!

huydx: this test will always await, so if we use default 10 seconds, each developer will need to wait 10sec everytime, is that ok?

trustin:
No, it has to finish ASAP.  I guess I misread the code. OK to merge, then!

trustin:
Thanks, @huydx!

PR #87, opened 2017-11-09
Use simpler JSON form for a Revision

trustin:
Motivation:

We do not and will not use the minor revision number in a Revision, and
thus there's no point of serializing the minor revision number if it's
zero.

Modification:

- Write the dedicated JsonSerializer and JsonDeserializer for Revision
  - A revision, whose minor is 0, is serialized into a number.
  - A revision with non-zero minor is serialized into an object just
    like before.
  - The deserializer understands various forms for backward
    compatibility.

Result:

Simpler JSON representation


trustin:
Thanks, @hyangtack @minwoox !

PR #89, opened 2017-11-16
Add author and timestamp property to all Commands

trustin:
Motivation:

We should keep who issued a command when in our API so that we can make
use of such information later.

Modifications:

- Add Command.author() and Command.timestamp()
- Replace creationTimeMillis() in some Command classes
  - Note that this isn't backward compatibility risk because we never
    released the version with this propery yet.
- Add backward compatibility tests

Result:

- Possibility of thorough audit


trustin:
Thanks for reviewing!

PR #90, opened 2017-11-16
Add a temporary option 'csrfTokenRequiredForThrift'

trustin:
.. which disables CSRF token verification completely for Thrift calls if
overridden to false. Given its security risk, this option is not listed
in the documentation or the example configuration, and thus it should be
used with caution and it may be removed in the future without any notice.


trustin:
Please do not merge yet. Will add WARN log messages for the hosts that send without the CSRF token, so we can keep track of such clients.

minwoox:
will review again when the log is added

PR #91, opened 2017-11-17
Add HTTP API version 1

minwoox:
Motivation:
The libthrift is not stable enough yet, we need two diffrent jars to compile.
If we use HTTP API, we can avoid that situation and have fewer depedencies for the client library.

Modifications:
- Revise admin REST APIs
- Add more APIs to cover all of the features in Thrift service

Result:
- We are ready to transfer from Thrift to HTTP API

Todos:
- Support adding and editing for multiple files
- Support preview diffs
- Adapt `MetaDataService`
- Create URLs with URI template parser
- Remove admin package or keep it minimum
- Remove thrift service ultimately


minwoox:
Please ignore the *Dtos for now. Since the fields for the response is not defined, I just used POJO without much thinking about it.

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/ContentServiceV1.java
```diff
@@ -0,0 +1,336 @@
...
+        }
+    }
+
+    /**
+     * PUT /projects/{projectName}/repos/{repoName}/contents{path}?revision={revision}
+     *
+     * <p>Adds or edit a file.
+     */
+    @Put("regex:/projects/(?<projectName>[^/]+)/repos/(?<repoName>[^/]+)/contents(?<path>/.*$)")
+    public CompletionStage<ContentDto> addOrEditFile(
```
minwoox: Because there's an input for revision, this is not idempotent. If I change this using `POST`,  it does not make sense to edit file. Do you have an idea for this?

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/HttpApiV1ResponseConverter.java
```diff
@@ -0,0 +1,73 @@
...
+
+            if (HttpMethod.POST == request.method()) {
+                final JsonNode jsonNode = Jackson.valueToTree(resObj);
+                final String url = jsonNode.get("url").asText();
+
+                // Remove the url field and send it with the LOCATION header.
+                ((ObjectNode) jsonNode).remove("url");
+                final HttpHeaders headers = HttpHeaders.of(HttpStatus.CREATED)
+                                                       .add(HttpHeaderNames.LOCATION, url)
+                                                       .addObject(HttpHeaderNames.CONTENT_TYPE,
```
minwoox: Do we need convenience methods for this on upstream? like `HttpResponse.of(HttpHeaders .....)`?

server/src/test/java/com/linecorp/centraldogma/server/internal/httpapi/ContentServiceV1Test.java
```diff
@@ -0,0 +1,369 @@
...
+        addFooFile();
+        addBarFile();
+        final String body =
+                '{' +
+                "   \"commitMessage\" : {" +
+                "       \"summary\" : \"Delete foo.json\"" +
+                "   }" +
+                '}';
+        final AggregatedHttpMessage res = httpClient
+                .execute(HttpHeaders.of(HttpMethod.DELETE, CONTENTS_PREFIX + "/foo.json?revision=2"),
```
minwoox: Do we need convenience methods for this on upstream? like `HttpClient.delete(uri, string body)`? It is not usual cases though.

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/ContentServiceV1.java
minwoox: Or change the logic which returns just 200 OK when there's no change between files regardless of the revision?

common/src/main/java/com/linecorp/centraldogma/internal/httpapi/v1/ContentDto.java
```diff
@@ -0,0 +1,86 @@
...
+
+package com.linecorp.centraldogma.internal.httpapi.v1;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
+
+import com.linecorp.centraldogma.common.Revision;
+
+@JsonInclude(Include.NON_DEFAULT)
+public class ContentDto {
```
minwoox: This class is removed

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/ContentServiceV1.java
```diff
@@ -0,0 +1,335 @@
...
+        }
+    }
+
+    /**
+     * PUT /projects/{projectName}/repos/{repoName}/contents{path}?revision={revision}
+     *
+     * <p>Adds or edit a file.
+     */
+    @Put("regex:/projects/(?<projectName>[^/]+)/repos/(?<repoName>[^/]+)/contents(?<path>/.*$)")
+    public CompletionStage<EntryDto> addOrEditFile(
```
minwoox: Because there's an input for revision, this is not idempotent. If I change this using POST, it does not make sense to edit file. Do you have an idea for this?
Or change the logic which returns just 200 OK when there's no change between files regardless of the revision?

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/HttpApiV1ResponseConverter.java
minwoox: Do we need convenience methods for this on upstream? like HttpResponse.of(HttpHeaders .....)?

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/ContentServiceV1.java
```diff
@@ -0,0 +1,334 @@
...
+            return listFiles(repository, path, new Revision(revision), ImmutableMap.of());
+        }
+    }
+
+    /**
+     * DELETE /projects/{projectName}/repos/{repoName}/contents{path}
+     *
+     * <p>Deletes a file.
+     */
+    @Delete("regex:/projects/(?<projectName>[^/]+)/repos/(?<repoName>[^/]+)/contents(?<path>/.*$)")
```
minwoox: Do we need convenience methods for this on upstream? like HttpClient.delete(uri, string body)? It is not a usual case though.

common/src/main/java/com/linecorp/centraldogma/common/Markup.java
```diff
@@ -46,4 +50,26 @@
     public String nameLowercased() {
         return nameLowercased;
     }
+
+    /**
+     * Returns the {@link Markup} from the specified {@code value}. If none of markup is matched,
+     * this will return {@link #UNKNOWN}.
+     */
+    public static Markup getMarkup(@Nullable String value) {
```
trustin: `parse` or `of` ?

common/src/main/java/com/linecorp/centraldogma/internal/Jackson.java
```diff
@@ -165,6 +167,15 @@ public static String writeValueAsPrettyString(Object value) throws JsonProcessin
         return compactMapper.convertValue(fromValue, toValueTypeRef);
     }
 
+    public static ObjectNode createObjectNode() {
+        compactMapper.createArrayNode();
+        return compactMapper.createObjectNode();
+    }
+
+    public static ArrayNode createArrayNode() {
+        return compactMapper.createArrayNode();
```
trustin: You could just use `JsonNodeFactory.instance.*Node()`?

common/src/main/java/com/linecorp/centraldogma/internal/Jackson.java
```diff
@@ -165,6 +167,15 @@ public static String writeValueAsPrettyString(Object value) throws JsonProcessin
         return compactMapper.convertValue(fromValue, toValueTypeRef);
     }
 
+    public static ObjectNode createObjectNode() {
+        compactMapper.createArrayNode();
+        return compactMapper.createObjectNode();
```
trustin: You could just use `JsonNodeFactory.instance.*Node()`?

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/ContentServiceV1.java
```diff
@@ -0,0 +1,334 @@
...
+        return repository.find(revision, pathPattern, options)
+                         .thenApply(
+                                 entries -> entries.values().stream()
+                                                   .filter(entry -> entry.type() != EntryType.DIRECTORY)
+                                                   .map(DtoConverter::convert).collect(toList()));
+    }
+
+    private void checkRepositoryExists(String projectName, String repoName) {
+        if (!projectManager().exists(projectName)) {
+            throw new HttpResponseException(HttpStatus.BAD_REQUEST);
```
trustin: `NOT_FOUND`?

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/ContentServiceV1.java
```diff
@@ -0,0 +1,334 @@
...
+        }
+
+        if (!projectManager().get(projectName).repos().exists(repoName)) {
+            throw new HttpResponseException(HttpStatus.BAD_REQUEST);
```
trustin: `NOT_FOUND`?

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/ContentServiceV1.java
trustin: I think it's OK to use POST here. Why doesn't it make sense to use POST for editing a file? We could 'post' a file modification request and the server could process it immediately?

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/ContentServiceV1.java
trustin: Like adding `Repository.deleteFiles(Revision revision, String pathPattern)`? Sounds good to me.

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/ContentServiceV1.java
```diff
@@ -0,0 +1,334 @@
...
+            final EntryType type =
+                    change.type().contentType().isAssignableFrom(JsonNode.class) ? EntryType.JSON
+                                                                                 : EntryType.TEXT;
+            return DtoConverter.convert(resultRevision, path, type, commitTimeMillis);
+        });
+    }
+
+    private static Change<?> getChangeOfPatch(String path, JsonNode jsonNode) {
+        if (path.charAt(path.length() - 1) == '/') {
+            throw new HttpResponseException(HttpStatus.BAD_REQUEST);
```
trustin: Perhaps we could rely on `Util.validateFilePath()` where necessary?

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/HttpApiV1ResponseConverter.java
trustin: `HttpRequest` seems to have one. It would be great to have something similar in `HttpResponse` as well. Please send a PR.

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/RepositoryServiceV1.java
```diff
@@ -0,0 +1,179 @@
...
+     * DELETE /projects/{projectName}/repos/{repoName}
+     *
+     * <p>Remove a repository.
+     */
+    @Delete("/projects/{projectName}/repos/{repoName}")
+    public CompletionStage<Object> removeRepository(@Param("projectName") String projectName,
+                                                      @Param("repoName") String repoName) {
+        checkProjectExists(projectName);
+        final long removeTimeMillis = System.currentTimeMillis();
+        return execute(Command.removeRepository(removeTimeMillis, AuthenticationUtil.currentAuthor(),
```
trustin: You can just use the factory method without timestamp parameter to let the upstream decide.

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/RepositoryServiceV1.java
```diff
@@ -0,0 +1,179 @@
...
+        try {
+            jsonNode = Jackson.readTree(message.content().toStringAscii()).get("name");
+        } catch (IOException e) {
+            throw new HttpResponseException(HttpStatus.BAD_REQUEST);
+        }
+        checkArgument(jsonNode != null, "repository name should be non-null");
+        final String repoName = jsonNode.asText();
+        checkArgument(!isNullOrEmpty(repoName), "repository name should be non-null");
+
+        return execute(Command.createRepository(System.currentTimeMillis(), AuthenticationUtil.currentAuthor(),
```
trustin: You can just use the factory method without timestamp parameter to let the upstream decide.

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/RepositoryServiceV1.java
```diff
@@ -0,0 +1,179 @@
...
+        } catch (IOException e) {
+            throw new HttpResponseException(HttpStatus.BAD_REQUEST);
+        }
+
+        // Currently, only the unremoveRepository patch is supported.
+        if (!jsonNode.equals(unremovePatch)) {
+            throw new HttpResponseException(HttpStatus.BAD_REQUEST);
+        }
+
+        return execute(Command.unremoveRepository(System.currentTimeMillis(),
```
trustin: You can just use the factory method without timestamp parameter to let the upstream decide.

common/src/main/java/com/linecorp/centraldogma/common/Markup.java
minwoox: Fixed. Thank you~!

common/src/main/java/com/linecorp/centraldogma/internal/Jackson.java
minwoox: Fixed. Thank you~!

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/ContentServiceV1.java
minwoox: Fixed. Thank you~!

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/ContentServiceV1.java
minwoox: I have read the spec again, and it's fine using POST here. :-)

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/ContentServiceV1.java
minwoox: Fixed. Thank you~!

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/RepositoryServiceV1.java
minwoox: I was planning to use the value to return the client at first. I will remove them for now, and bring it back if it needs. :-)

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/ContentServiceV1.java
```diff
@@ -0,0 +1,345 @@
...
+        }
+
+        return repository.find(revision, pathPattern, options)
+                         .thenApply(entries -> entries.values().stream()
+                                                      .filter(entry -> entry.type() != EntryType.DIRECTORY)
+                                                      .map(DtoConverter::convert).collect(toList()));
+    }
+
+    /**
+     * POST /projects/{projectName}/repos/{repoName}/contents{path}?revision={revision}
```
hyangtack: Looks different from the implementation. Which one is correct?

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/ContentServiceV1.java
```diff
@@ -0,0 +1,345 @@
...
+        final Markup markup = commitMessage.getMarkup();
+
+        return repository.normalize(revision).thenCompose(normalizedRevision -> execute(
+                Command.push(commitTimeMills, author, repository.parent().name(), repository.name(),
+                             normalizedRevision,
+                             summary, detail, markup, changes)));
+    }
+
+    /**
+     * GET /projects/{projectName}/repos/{repoName}/contents{path}?revision={revision}
```
hyangtack: `revision={revision}` -> `revision={revision}&`?

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/ContentServiceV1.java
minwoox: I didn't change the path when it's changed from PUT to POST, thank you~!

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/ContentServiceV1.java
```diff
@@ -0,0 +1,457 @@
...
+        return isNullOrEmpty(path) ? "/" : path;
+    }
+
+    /**
+     * POST /projects/{projectName}/repos/{repoName}/contents?revision={revision}
+     *
+     * <p>Adds or edits a file.
+     */
+    @Post("/projects/{projectName}/repos/{repoName}/contents")
+    public CompletionStage<EntryDto> addOrEditFile(@Param("projectName") String projectName,
```
hyangtack: nit: `EntryDto` -> `EntryDto<?>`?

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/WatchService.java
```diff
@@ -0,0 +1,133 @@
...
+
+    /**
+     * Awaits and retrieves the latest revision of the commit that changed the file that matches the specified
+     * {@code pathPattern} since the specified {@code lastKnownRevision}. This will wait until the specified
+     * {@code timeoutMillis} passes. If there's no change during the time, the returned future will be
+     * exceptionally completed with the {@link CancellationException}.
+     */
+    public CompletableFuture<Revision> watchRepository(Repository repo, Revision lastKnownRevision,
+                                                       String pathPattern, long timeoutMillis) {
+        if (isServerStopping()) {
```
hyangtack: Just curious. Is this kind of prevention required for every API calls? I mean that do we have something like a decorator which sends a failure response when the server starts stopping?

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/ContentServiceV1.java
```diff
@@ -0,0 +1,457 @@
...
+                                                   @Param("repoName") String repoName,
+                                                   @Param("revision") @Default("-1") String revision,
+                                                   AggregatedHttpMessage message) {
+        checkRepositoryExists(projectName, repoName);
+
+        final Entry<CommitMessageDto, Change<?>> commitMessageAndChange = commitMessageAndChange(message);
```
hyangtack: Try to use a request converter later! :-)

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/ContentServiceV1.java
```diff
@@ -0,0 +1,457 @@
...
+     * right away to the client with the {@link WatchResultDto}. {@link HttpStatus#NOT_MODIFIED} otherwise.
+     */
+    @Get("regex:/projects/(?<projectName>[^/]+)/repos/(?<repoName>[^/]+)/contents(?<path>(|/.*))$")
+    public CompletionStage<?> getFiles(@Param("projectName") String projectName,
+                                       @Param("repoName") String repoName,
+                                       @Param("path") String path,
+                                       @Param("revision") @Default("-1") String revision,
+                                       @Param("queryType") Optional<String> queryType,
+                                       @Param("expression") Optional<String> expression,
+                                       @Header("if-none-match") Optional<String> ifNoneMatch,
```
hyangtack: Try to use a request converter here, too. :-) I think it'd be better to make an object like `WatchRequest` so that we can move the codes like parsing timeout to the request converter.

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/CommitServiceV1.java
```diff
@@ -0,0 +1,142 @@
...
+                    }
+                    return changes.stream().map(DtoConverter::convert).collect(toImmutableList());
+                });
+    }
+
+    private Repository getRepository(String projectName, String repoName) {
+        return projectManager().get(projectName).repos().get(repoName);
+    }
+
+    private void checkRepositoryExists(String projectName, String repoName) {
```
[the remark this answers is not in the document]
hyangtack: Or, we may move this kind of codes to a request converter later.

common/src/main/java/com/linecorp/centraldogma/internal/httpapi/v1/CommitMessageDtoDeserializer.java
```diff
@@ -0,0 +1,49 @@
...
+    protected CommitMessageDtoDeserializer() {
+        super(CommitMessageDto.class);
+    }
+
+    @Override
+    public CommitMessageDto deserialize(JsonParser p, DeserializationContext ctxt) throws IOException {
+        final JsonNode jsonNode = p.readValueAsTree();
+        final JsonNode summary = jsonNode.get("summary");
+        if (summary.textValue() == null) {
+            ctxt.reportInputMismatch(CommitMessageDto.class, "commit message should hava summary");
```
hyangtack: nit: hava -> have

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/ContentServiceV1.java
```diff
@@ -0,0 +1,457 @@
...
+                         ImmutableMap.of(FindOption.FETCH_CONTENT, false));
+    }
+
+    private CompletionStage<List<EntryDto>> listFiles(Repository repository, String path,
+                                                      Revision revision, Map<FindOption<?>, ?> options) {
+        final String pathPattern = appendWildCardIfDirectory(path);
+        return repository.find(revision, pathPattern, options)
+                         .thenApply(entries -> entries.values().stream()
+                                                      .filter(entry -> entry.type() != EntryType.DIRECTORY)
+                                                      .map(DtoConverter::convert).collect(toList()));
```
trustin: toImmutableList()?

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/ContentServiceV1.java
```diff
@@ -0,0 +1,457 @@
...
+    }
+
+    private static String appendWildCardIfDirectory(String path) {
+        return path.endsWith("/") ? path + '*' : path;
+    }
+
+    // returns "/" if the specified path is null or empty
```
trustin: `/** ... */`?

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/ContentServiceV1.java
```diff
@@ -0,0 +1,457 @@
...
+        checkRepositoryExists(projectName, repoName);
+
+        final Entry<CommitMessageDto, Change<?>> commitMessageAndChange = commitMessageAndChange(message);
+        final Change<?> change = commitMessageAndChange.getValue();
+        final Repository repository = getRepository(projectName, repoName);
+        final Author author = AuthenticationUtil.currentAuthor();
+
+        final CompletableFuture<Revision> revisionFuture = repository.normalize(new Revision(revision));
+
+        // TODO (minwoox) add a logic that tries until it succeeds when the revision is HEAD or -1
```
trustin: I'm not sure if we need this feature. Perhaps it should be decided by a client?

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/ContentServiceV1.java
```diff
@@ -0,0 +1,457 @@
...
+
+    private CompletionStage<Revision> push(long commitTimeMills, Author author, Repository repository,
+                                           Revision revision, CommitMessageDto commitMessage,
+                                           Iterable<Change<?>> changes) {
+        final String summary = commitMessage.summary();
+        final String detail = commitMessage.detail();
+        final Markup markup = commitMessage.markup();
+
+        final CompletableFuture<Revision> normalizeFuture = repository.normalize(revision);
+        final CompletableFuture<Revision> headFuture = repository.normalize(Revision.HEAD);
```
trustin: If `revision` is `HEAD` then we should not call `normalize` twice, because two `normalize(HEAD)` calls may return different revisions on contention.

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/ContentServiceV1.java
```diff
@@ -0,0 +1,457 @@
...
+        final CompletableFuture<Revision> future = CompletableFutures.combine(
+                normalizeFuture, headFuture, (normalized, head) -> {
+
+                    // If the revision is not same with head revision, just throw an exception here with
+                    // HttpStatus.BAD_REQUEST. If the revision is same with head revision and if
+                    // there's a change while invoking this operation, then will throw an exception with
+                    // HttpStatus.CONFLICT.
+                    if (normalized.compareTo(head) != 0) {
+                        throw newHttpResponseException(HttpStatus.BAD_REQUEST,
+                                                       "must request with the latest revision");
```
trustin: I think we should respond with `CONFLICT` in this case. i.e. no need to check because `GitRespository` checks.

common/src/main/java/com/linecorp/centraldogma/internal/httpapi/v1/EntryDto.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    @JsonCreator
+    public EntryDto(String path, EntryType type, String projectName, String repoName, Revision revision,
+                    long commitTimeMillis) {
+        this.path = requireNonNull(path, "path");
+        this.type = requireNonNull(type, "type");
+        this.content = null;
+        this.revision = requireNonNull(revision, "revision");
+        url = PROJECTS_PREFIX + '/' + projectName + REPOS + '/' + repoName + path;
```
hyangtack: I think it'd be better to remove the logic which generates a URL from DTO class. :-)

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/ContentServiceV1.java
```diff
@@ -0,0 +1,457 @@
...
+
+        final Repository repository = getRepository(projectName, repoName);
+        if (query.isPresent()) {
+            // get a file
+            return repository.get(new Revision(revision), query.get())
+                             .thenApply(queryResult -> DtoConverter.convert(queryResult, path0))
+                             .exceptionally(thrown -> {
+                                 if (thrown instanceof EntryNotFoundException) {
+                                     throw newHttpResponseException(
+                                             HttpStatus.BAD_REQUEST, "entry not found: " + thrown.getMessage());
```
trustin: 404?

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/HttpApiV1Util.java
```diff
@@ -0,0 +1,105 @@
...
+    static HttpResponseException newHttpResponseException(HttpStatus status, String message) {
+        return HttpResponseException.of(newResponseWithErrorMessage(status, message));
+    }
+
+    /**
+     * Returns a newly created {@link HttpResponse} with the specified {@link HttpStatus} and {@code message}.
+     */
+    static HttpResponse newResponseWithErrorMessage(HttpStatus status, String message) {
+        // TODO(minwoox) refine the error message
+        final String content = "{\"message\":\"" + message + "\"}";
```
trustin: Better building with `ObjectNode`?

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/ProjectServiceV1.java
```diff
@@ -0,0 +1,160 @@
...
+        checkArgument(jsonNode.get("name") != null, "project name should be non-null");
+        final String name = jsonNode.get("name").textValue();
+        checkArgument(!isNullOrEmpty(name), "project name should be non-null");
+
+        return execute(Command.createProject(AuthenticationUtil.currentAuthor(), name))
+                .handle((unused, thrown) -> {
+                    if (thrown != null) {
+                        if (Throwables.getRootCause(thrown) instanceof StorageExistsException) {
+                            throw newHttpResponseException(HttpStatus.BAD_REQUEST,
+                                                           "project " + name + " already exists");
```
trustin: Perhaps `CONFLICT`?

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/ProjectServiceV1.java
```diff
@@ -0,0 +1,160 @@
...
+    @ConsumeType("application/json-patch+json")
+    @Patch("/projects/{projectName}")
+    public CompletionStage<ProjectDto> patchProject(
+            @Param("projectName") String name, AggregatedHttpMessage message) {
+        final JsonNode jsonNode = getJsonNode(message);
+
+        if (!jsonNode.equals(unremovePatch)) {
+            throw newHttpResponseException(HttpStatus.BAD_REQUEST,
+                                           "not supported JSON patch: " + message.content() +
+                                           " (expected: " + unremovePatch.toString() + ')');
```
trustin: Maybe `CONFLICT`?

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/RepositoryServiceV1.java
```diff
@@ -0,0 +1,182 @@
...
+                                                           AggregatedHttpMessage message) {
+        checkProjectExists(projectName);
+
+        final String repoName = getName(message);
+        return execute(Command.createRepository(AuthenticationUtil.currentAuthor(), projectName, repoName))
+                .handle((unused, thrown) -> {
+                    if (thrown != null) {
+                        if (Throwables.getRootCause(thrown) instanceof StorageExistsException) {
+                            throw newHttpResponseException(HttpStatus.BAD_REQUEST,
+                                                           "repository " + repoName + " already exists");
```
trustin: `CONFLICT`?

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/RepositoryServiceV1.java
```diff
@@ -0,0 +1,182 @@
...
+    public CompletionStage<RepositoryDto> patchRepository(@Param("projectName") String projectName,
+                                                          @Param("repoName") String repoName,
+                                                          AggregatedHttpMessage message) {
+        checkProjectExists(projectName);
+        final JsonNode jsonNode = getJsonNode(message);
+
+        if (!unremovePatch.equals(jsonNode)) {
+            throw newHttpResponseException(HttpStatus.BAD_REQUEST,
+                                           "not supported JSON patch: " + message.content() +
+                                           " (expected: " + unremovePatch.toString() + ')');
```
trustin: `CONFLICT`?

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/WatchService.java
trustin: > Just curious. Is this kind of prevention required for every API calls? I mean that do we have something like a decorator which sends a failure response when the server starts stopping?

Sounds like a good idea which is worth its own issue. Could you create one?

server/src/test/java/com/linecorp/centraldogma/server/internal/httpapi/CommitServiceV1Test.java
```diff
@@ -0,0 +1,283 @@
...
+import com.linecorp.armeria.common.AggregatedHttpMessage;
+import com.linecorp.armeria.common.HttpHeaderNames;
+import com.linecorp.centraldogma.testing.CentralDogmaRule;
+
+public class CommitServiceV1Test {
+
+    // TODO(minwoox) replace this unit test using nested structure in junit 5
+    // Rule is used instead of ClassRule because the listCommits is affected by other unit tests.
+    @Rule
+    public final CentralDogmaRule dogma = new CentralDogmaRule();
```
trustin: Alternatively, you can use a `static` field here and use a different repository for each test method, using the `TestName` rule.

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/ContentServiceV1.java
minwoox: I realized that we need to make `RequestObject` as optional for this. I'm not sure if it's good idea though. WDYT @hyangtack ?

common/src/main/java/com/linecorp/centraldogma/internal/Util.java
```diff
@@ -38,6 +38,8 @@
             "^(?:[-_0-9a-zA-Z](?:[-_\\.0-9a-zA-Z]*[-_0-9a-zA-Z])?)+$");
     private static final Pattern FILE_PATH_PATTERN = Pattern.compile(
             "^(?:/[-_0-9a-zA-Z](?:[-_\\.0-9a-zA-Z]*[-_0-9a-zA-Z])?)+$");
+    private static final Pattern JSON_FILE_PATH_PATTERN = Pattern.compile(
+            "^(?:/[-_0-9a-zA-Z](?:[-_\\.0-9a-zA-Z]*[-_0-9a-zA-Z])?)+\\.(?i)json$");
```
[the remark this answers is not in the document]
minwoox: I did not know. That was just my decision. I will remove the `(?i)` :-)

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/ProjectServiceV1.java
minwoox: Is this `CONFLICT`? This `patch` method only supports `unremovePatch` for now, so if the request is sent except the `patch`, I thought we need to give back `BAD_REQUEST`.

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/WatchService.java
hyangtack: Created! https://github.com/line/centraldogma/issues/107

trustin: Thanks, @hyangtack !

common/src/main/java/com/linecorp/centraldogma/internal/Util.java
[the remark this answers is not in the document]
trustin: Actually, it seems case-insensitive. In `EntryType`:

```
    /**
     * Guesses the {@link EntryType} from the specified {@code path}.
     */
    public static EntryType guessFromPath(String path) {
        requireNonNull(path, "path");
        if (path.isEmpty()) {
            throw new IllegalArgumentException("empty path");
        }

        if (path.charAt(path.length() - 1) == '/') {
            return DIRECTORY;
        }

        if (Ascii.toLowerCase(path).endsWith(".json")) {
            return JSON;
        }

        return TEXT;
    }
```

Sorry!

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/ProjectServiceV1.java
trustin: I think there are two cases involved here.

- The JSON document is a JSON patch but is not equal to `unremovePatch`, which would be `CONFLICT`.
- The JSON document is not a JSON patch, which would be `BAD_REQUEST`.

Could we send `BAD_REQUEST` when it's not a JSON patch or `CONFLICT` otherwise? You could try to deserialize the content into `JsonPatch` to test if it's a JSON patch.

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/ContentServiceV1.java
hyangtack: I think we can return `Optional<?>` from a request converter. Not possible?

common/src/main/java/com/linecorp/centraldogma/common/QueryType.java
```diff
@@ -46,4 +52,22 @@
...
+    /**
+     * Returns a {@link QueryType} from the specified {@code value}. If none of query type is matched,
+     * this will return {@link #IDENTITY}.
+     */
+    public static QueryType parse(@Nullable String value) {
+        if (isNullOrEmpty(value)) {
+            return IDENTITY;
+        }
+
+        final String queryType = Ascii.toUpperCase(value);
```
trustin: Seems like uppercasing here is unnecessary because you used `equalsIgnoreCase`?

common/src/main/java/com/linecorp/centraldogma/internal/httpapi/v1/ChangeDto.java
```diff
@@ -0,0 +1,73 @@
...
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.base.MoreObjects;
+import com.google.common.base.MoreObjects.ToStringHelper;
+
+import com.linecorp.centraldogma.common.ChangeType;
+
+@JsonInclude(Include.NON_DEFAULT)
```
hyangtack: Is this intentional? `@JsonInclude(Include.NON_NULL)`?

common/src/main/java/com/linecorp/centraldogma/common/QueryType.java
```diff
@@ -46,4 +52,22 @@
...
+
+        if ("JSON_PATH".equalsIgnoreCase(queryType)) {
+            return JSON_PATH;
+        }
+
+        return IDENTITY;
```
trustin: Shouldn't we raise an exception or return null?

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -344,12 +350,29 @@ private Server startServer(ProjectManager pm, CommandExecutor executor, Security
...
+            @Override
+            public Entry get(String path, @Nullable String contentEncoding) {
+                requireNonNull(path, "path");
+                return new ByteArrayEntry(path, MediaType.PLAIN_TEXT_UTF_8,
+                                          server.defaultHostname().getBytes(StandardCharsets.UTF_8));
+            }
+
+            @Override
+            public String meterTag() {
+                return "hostnameService";
```
trustin: Just `hostname`?

server/src/main/java/com/linecorp/centraldogma/server/internal/httpapi/WatchService.java
```diff
@@ -0,0 +1,123 @@
...
+import com.linecorp.centraldogma.common.QueryResult;
+import com.linecorp.centraldogma.common.Revision;
+import com.linecorp.centraldogma.server.internal.storage.repository.Repository;
+
+import io.netty.channel.EventLoop;
+
+/**
+ * A service class for watching repository or a file.
+ */
+public class WatchService {
```
trustin: nit: Probably OK to make `*Service` classes `final`?

common/src/main/java/com/linecorp/centraldogma/common/QueryType.java
minwoox: Oops~!!

trustin:
Thank you, @minwoox !

PR #98, opened 2017-12-06
Auto-generate replica ID

trustin:
Motivation:

Currently, a replica ID must be manually set in dogma.json. It
complicates deployment process of a server because it makes the content
of each dogma.json different.

Modifications:

- Remove replicaId property from ZooKeeperReplicationConfig
- Auto-generate a random UUID and store it in dataDir

Result:

- All replicas can have identical dogma.json. i.e. simpler deployment


trustin:
Thanks!

PR #99, opened 2017-12-08
Fix incorrect JSON patch generation, redundant upsert and repository format migration failure

trustin:
Motivation:

- DiffProcessor, which is responsible for generating a JSON patch,
  generates an incorrect 'move' operation if the moved value is an
  element of an array or an array's descendant.
- APPLY_TEXT_PATCH sometimes fails if old content does not end with an
  empty line.
- GitRepository.commit() allows a redundant upsert, which is not an
  expected behavior.

Modifications:

- Do not use 'move' operation at all.
- Do not count a redundant upsert so that a commit with redundant
  upserts are considered as empty.
- Sanitize TEXT entries so that all TEXT entries have UNIX line endings
  and the last line always ends with \n.
- Make sure GitRepository.diff() does not add an APPLY_JSON_PATCH change
  with empty content.
- Implement the workaround for the empty commits in GitRepository.cloneTo()
- Add progress listener to GitRepository.cloneTo()
  - Log repository format migration progress periodically.

Result:

- GitRepository.diff() produces correct JSON patch.
- GitRepository.commit() does not allow a redundant upsert.
- GitRepository.cloneTo() does not fail with ChangeConflictException.

### DIFF
--- a/common/src/main/java/com/linecorp/centraldogma/internal/Util.java
+++ b/common/src/main/java/com/linecorp/centraldogma/internal/Util.java
@@ -23,7 +23,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.StringReader;
 import java.nio.file.Files;
-import java.util.LinkedList;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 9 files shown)


trustin:
Thanks, reviewers!

PR #102, opened 2017-12-11
Clean up and optimize JsonPatch

trustin:
Motivation:

- The classes in internal.jsonpatch violates our Checkstyle rules.
- JsonPatchOperation implementations makes a deep copy of a whole JSON
  document on each application.
- DiffProcessor creates redundant intermediary DiffOperation objects.

Modifications:

- Reorganization
  - Remove DiffOperation.
  - Merge JsonDiff into JsonPatch
  - Move the classes in 'jsonpath.diff' and 'jsonpatch.utils' to
    'jsonpatch'.
  - Make all JsonPatchOperation classes package-local.
  - Extract some common logic in JsonPatchOperation implementations into
    the common methods in JsonPatchOperation and PathValueOperation.
- Optimization
  - Make a deep copy of a JSON document only once per JsonPatch.apply().
  - JsonPatch.unchangedValues() is invoked only when really necessary.
  - Optimize JsonPatchOperation implementations so that they do not
    create unnecessary deep copies.
- Un-exclude the 'internal.jsonpatch' from Checkstyle suppression.

Result:

- JsonPatch.apply() and JsonPatch.generate() performs better.


common/src/main/java/com/linecorp/centraldogma/internal/jsonpatch/AddOperation.java
```diff
@@ -115,19 +113,18 @@ private JsonNode addToArray(final JsonPointer path, final JsonNode node)
             throw new JsonPatchException("not an index: " + rawToken + " (expected: a non-negative integer)");
         }
 
-        if (index < 0 || index > size)
+        if (index < 0 || index > size) {
```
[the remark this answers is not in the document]
trustin: The condition is correct but the exception message is not. Fixed the incorrect message. :-)

common/src/main/java/com/linecorp/centraldogma/internal/jsonpatch/CopyOperation.java
```diff
@@ -25,7 +40,7 @@
 import com.fasterxml.jackson.databind.JsonNode;
 
 /**
- * JSON Patch {@code copy} operation
+ * JSON Patch {@code copy} operation.
  *
  * <p>For this operation, {@code from} is the JSON Pointer of the value to copy,
  * and {@code path} is the destination where the value should be copied.</p>
```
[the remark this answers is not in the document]
trustin: Fixed

common/src/main/java/com/linecorp/centraldogma/internal/jsonpatch/DiffProcessor.java
```diff
@@ -0,0 +1,122 @@
...
+    void valueRemoved(final JsonPointer pointer) {
+        diffs.add(new RemoveOperation(pointer));
+    }
+
+    void valueAdded(final JsonPointer pointer, final JsonNode value) {
+        final JsonPatchOperation op;
+        if (value.isContainerNode()) {
+            // Use copy operation only for container nodes.
+            final JsonPointer ptr = findUnchangedValue(value);
+            op = ptr != null ? new CopyOperation(ptr, pointer)
```
hyangtack: Just curious. Do we need to use `CopyOperation` here? How about using `AddOperation` only without making `unchangedValues` map?

trustin: Yes, a copy operation generates a smaller patch than an add operation when the value being copied is a container node. Imagine adding vs. copying an array with 100 elements or an object with 100 fields.

trustin:
Thanks for reviewing!

common/src/main/java/com/linecorp/centraldogma/internal/jsonpatch/DiffProcessor.java
hyangtack: Ok, thanks! :-)

PR #103, opened 2017-12-11
Remove runspace-related code completely

trustin:
Motivation:

We never exposed runspace-related RPC operations in our project history,
and we never will.

Modifications:

- Remove the code on the execution path which is evaluated only when
  Revision.minor is 0.

Result:

- Less bloated


trustin:
Thanks for reviewing!

PR #104, opened 2017-12-12
Use dedicated commit ID database instead of Git tags

### LINKED ISSUE #100: Store revision-commitId mapping in a dedicated store
`GitRepository` currently uses a tag to maintain the mappings between revisions and commitIds. It has the following problems:

- When using the default refstorage implementation (refdir), it creates a symbolic link for each revision, which is slow and which creates too many files in the long term.
- When using the 'reftree' refstorage implementation, the disk usage of the repository increases way too much (more than 2x). Also, it becomes impossible to browse the repository using vanilla `git` command because reftree is a jGit-only extension.

I'd like to experiment with storing the revision-commitId mappings in a dedicated store like B-tree.

Related: https://github.com/lmdbjava/lmdbjava

trustin:
Motivation:

GitRepository currently uses a tag to maintain the mappings between
revisions and commitIds. It has the following problems:

- When using the default refstorage implementation (refdir), it creates
  a symbolic link for each revision, which is slow and which creates too
  many files in the long term.
- When using the 'reftree' refstorage implementation, the disk usage of
  the repository increases way too much (more than 2x). Also, it becomes
  impossible to browse the repository using vanilla `git` command
  because reftree is a jGit-only extension.

Modifications:

- Replace Git tags with CommitIdDatabase which is better suited for
  storing and accessing Revision-to-ObjectId mappings
- Implement automated migration

Result:

- Fixes #100

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/CommitIdDatabase.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/CommitIdDatabase.java
@@ -1,0 +1,264 @@
+/*
+ * Copyright 2017 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.linecorp.centraldogma.server.internal.storage.repository.git;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkState;
+import static com.linecorp.centraldogma.server.internal.storage.repository.git.GitRepository.R_HEADS_MASTER;
+
+import java.io.EOFException;
+import java.io.File;
+import java.io.IOException;
+import java.nio.ByteBuffer;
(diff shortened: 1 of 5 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/CommitIdDatabase.java
```diff
@@ -0,0 +1,264 @@
...
+                    throw new StorageException("mismatching revision: " + gitRepo.getDirectory() +
+                                               " (actual: " + currentRevision.major() +
+                                               ", expected: " + expectedRevision.major() + ')');
+                }
+
+                put(currentRevision, currentId, false);
+                previousRevision = currentRevision;
+            }
+
+            // All commit IDs but the head have been stored. Store the head finally.
```
minwoox: I feel that only the head is stored... how about just using except instead of but?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/CommitIdDatabase.java
```diff
@@ -0,0 +1,264 @@
...
+        } catch (IOException e) {
+            throw new StorageException("failed to read the commit ID database: " + path, e);
+        }
+
+        buf.flip();
+
+        final int actualRevision = buf.getInt();
+        if (actualRevision != revision.major()) {
+            throw new StorageException("incorrect revision number in the commit ID database: " + path +
+                                       "(actual: " + actualRevision + ", expected: " + revision.major());
```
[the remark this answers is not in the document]
trustin: Fixed

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/CommitIdDatabase.java
trustin: Replaced but with except for clarity.

trustin:
Thanks for reviewing!

PR #108, opened 2017-12-18
Suppress IllegalArgumentException if a token does not start with ‘app…

hyangtack:
…Token-‘

Modifications:
- Return `null` if TokenService cannot find a token. Do not raise an exception anymore.


trustin:
Thanks, @hyangtack !

PR #124, opened 2018-01-05
Support addition and modification for multiple files on HTTP API

minwoox:
Motivation:
Currently, a user can add or modify only one file using HTTP API. If he or she would like to add
multiple files, one should call the operation as much as the number of entries.

Modifications:
- Support addition and modification for multiple files
- Support rename operation

Result:
- HTTP API is covering most of the range of RPC API


server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
```diff
@@ -127,62 +130,39 @@ private static String rootDirIfEmpty(String path) {
...
-    public CompletionStage<EntryDto<?>> addOrEditFile(@Param("revision") @Default("-1") String revision,
-                                                      @RequestObject Repository repository,
-                                                      @RequestObject JsonNode node,
-                                                      @RequestObject Author author) {
-        final Entry<CommitMessageDto, Change<?>> commitMessageAndChange = commitMessageAndChange(node);
-        final Change<?> change = commitMessageAndChange.getValue();
-        final CompletableFuture<Revision> revisionFuture = repository.normalize(new Revision(revision));
-
-        return revisionFuture.thenCompose(normalizedRevision -> {
+    public CompletionStage<?> addOrEditFiles(@Param("revision") @Default("-1") String revision,
```
trustin: I wonder if we can just rename this to `commit`?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
```diff
@@ -127,62 +130,39 @@ private static String rootDirIfEmpty(String path) {
...
-                    changes -> push(commitTimeMillis, author, repository, normalizedRevision,
-                                    commitMessage, changes.values()));
-
-            final EntryType type = entryTypeFromChange(change);
-            return resultRevisionFuture.thenApply(result -> DtoConverter.convert(
-                    result, repository.parent().name(), repository.name(), change.path(), type,
-                    commitTimeMillis));
+                    previewDiffs -> {
+                        if (previewDiffs.size() == 0) {
+                            throw new IllegalArgumentException("the contents are the same as the old ones");
```
trustin: I wonder if we can reuse `RedundantChangeException` here.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
```diff
@@ -196,6 +176,25 @@ private static EntryType entryTypeFromChange(Change<?> change) {
                              summary, detail, markup, changes)));
     }
 
+    private static String aggregatePath(Iterable<Change<?>> changes) {
```
trustin: How about `joinPaths`?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/converter/CommitMessageRequestConverter.java
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * A request converter that converts to {@link CommitMessageDto}.
+ */
+public final class CommitMessageRequestConverter implements RequestConverterFunction {
+
+    @Override
+    public Object convertRequest(ServiceRequestContext ctx, AggregatedHttpMessage request,
+                                 Class<?> expectedResultType) throws Exception {
+        if (expectedResultType.isAssignableFrom(CommitMessageDto.class)) {
```
trustin: - Shouldn't this be reversed? i.e. `CommitMessageDto.class.isAssignableFrom(expectedResultType)`

  ```
  $ jjs
  jjs> java.lang.String.class.isAssignableFrom(java.lang.String.class)
  true
  jjs> java.lang.String.class.isAssignableFrom(java.lang.Object.class)
  false
  jjs> java.lang.Object.class.isAssignableFrom(java.lang.String.class)
  true
  ```
- Please review all `RequestConverterFunction` implementations just in case the same mistakes were made.

trustin: Just realized that simply reversing `isAssignableFrom()` does not make sense because we won't be able to cast `CommitMessageDto` object to `SubtypeOfCommitMessageDto`.

common/src/main/java/com/linecorp/centraldogma/common/ChangeType.java
```diff
@@ -68,4 +72,15 @@
     public Class<?> contentType() {
         return contentType;
     }
+
+    /**
+     * Returns a {@link ChangeType} from the specified {@code value} case-insensitively.
+     */
+    public static ChangeType parse(String value) {
+        checkArgument(!isNullOrEmpty(value), "ChangeType should be non-null");
```
trustin: - `requireNonNull(value, "value"); checkArgument(!value.isEmpty(), ...);`

common/src/main/java/com/linecorp/centraldogma/common/ChangeType.java
```diff
@@ -68,4 +72,15 @@
...
+
+        final String type = Ascii.toUpperCase(value);
+
+        return ChangeType.valueOf(type);
```
trustin: `return ChangeType.valueOf(Ascii.toUpperCase(value));` for simplicity?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
```diff
@@ -196,6 +170,25 @@ private static EntryType entryTypeFromChange(Change<?> change) {
                              summary, detail, markup, changes)));
     }
 
+    private static String joinPath(Iterable<Change<?>> changes) {
```
trustin: `joinPaths`? (plural)

server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
```diff
@@ -281,73 +274,53 @@ private Object handleWatchFailure(Throwable thrown) {
...
+                                          @Param("path") @Default("/**") String path,
+                                          @Param("to") Optional<String> to,
+                                          @RequestObject Repository repository) {
+        final Revision fromRevision;
+        final Revision toRevision;
+
+        // 1. only the "revision" is specified:       get the "revision"
+        // 2. only the "to" is specified:             get from "HEAD" to "to"
+        // 3. the "revision" and "to" is specified:   get from the "revision" to "to"
+        // 4. nothing is specified:                   get from "HEAD" to "INIT"
```
trustin: The case 2 and 4 would beg for memory spike or OOME with the current `Repository` API, which does not stream commits. Could we limit the number of commits retrieved?

minwoox: So just return the first `N` commits where `N` is configurable?
Also, I feel strongly that we need to implement streaming commits using `Relation` since we are going to return just the part of the request.

trustin: Yes and yes.

common/src/main/java/com/linecorp/centraldogma/common/ChangeType.java
```diff
@@ -68,4 +72,14 @@
     public Class<?> contentType() {
         return contentType;
     }
+
+    /**
+     * Returns a {@link ChangeType} from the specified {@code value} case-insensitively.
+     */
+    public static ChangeType parse(String value) {
+        requireNonNull(value, "value");
+        checkArgument(!value.isEmpty(), "the value for ChangeType should not be empty");
```
trustin: Please end with a period if it's a sentence.

common/src/main/java/com/linecorp/centraldogma/internal/api/v1/CommitMessageDtoDeserializer.java
```diff
@@ -38,7 +38,7 @@ protected CommitMessageDtoDeserializer() {
     public CommitMessageDto deserialize(JsonParser p, DeserializationContext ctxt) throws IOException {
         final JsonNode jsonNode = p.readValueAsTree();
         final JsonNode summary = jsonNode.get("summary");
-        if (summary.textValue() == null) {
+        if (summary == null || summary.textValue() == null) {
             ctxt.reportInputMismatch(CommitMessageDto.class, "commit message should have a summary");
```
trustin: Please end with a period if it's a sentence.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
```diff
@@ -281,73 +277,67 @@ private Object handleWatchFailure(Throwable thrown) {
...
 
-        return push(commitTimeMillis, author, repository, new Revision(revision),
-                    commitMessage, ImmutableList.of(Change.ofRemoval(path0)))
-                .handle(HttpApiUtil::throwUnsafelyIfNonNull);
+        return repository
+                .history(fromRevision, toRevision, path)
+                .thenApply(commits -> {
+                    final boolean toList = isNullOrEmpty(revision) || "/".equalsIgnoreCase(revision) ||
+                                           to.isPresent();
+                    return objectOrList(commits, toList, DtoConverter::convert);
```
hyangtack: According to the `javadoc`, it should return the list of commits. But I think there's chance to return an object. Which one is correct?

minwoox:
Thanks reviewers~!!

PR #126, opened 2018-01-10
Add Repository.normalizeNow()

minwoox:
Motivation:
`Repository.normalize()` does not have to be asynchronous. But it's hard to change the API
because it is used from everywhere.
Therefore, I decided to add `Repository.normalizeNow()` which returns the normalized `Revision`
right away instead. The `Repository.normalize()` will be removed when we remove the RPC API.

Also in our current `Repository` API, fetching many commits can cause memory spike or OOME.
We need a mechanism to restrict the maximum number of commits at a time.

Modifications:
- Limit the maximum number of commits
- Add `Repository.normalizeNow()` that returns `Revision` immediately

Result:
- Less memory spike

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
@@ -70,6 +70,7 @@ import com.linecorp.centraldogma.server.internal.storage.project.ProjectManager;
 import com.linecorp.centraldogma.server.internal.storage.repository.FindOption;
 import com.linecorp.centraldogma.server.internal.storage.repository.RedundantChangeException;
 import com.linecorp.centraldogma.server.internal.storage.repository.Repository;
+import com.linecorp.centraldogma.server.internal.storage.repository.Repository.NormalizedFromToRevision;
 
 /**
  * Annotated service object for managing and watching contents.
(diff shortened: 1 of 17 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
```diff
@@ -133,25 +134,22 @@ private static String rootDirIfEmpty(String path) {
...
-            final CompletableFuture<Revision> resultRevisionFuture = changesFuture.thenCompose(
-                    previewDiffs -> {
-                        if (previewDiffs.size() == 0) {
-                            throw new RedundantChangeException("changes did not change anything: " + changes);
-                        }
-                        return push(commitTimeMillis, author, repository, normalizedRevision,
-                                    commitMessage, previewDiffs.values());
-                    });
+            if (previewDiffs.size() == 0) {
+                throw new RedundantChangeException("changes did not change anything: " + changes);
```
trustin: This will generate a large string is `changes` is large. How about using the default constructor?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/Repository.java
```diff
@@ -94,10 +97,57 @@ default Author author() {
...
+
+        return new NormalizedFromToRevision(normalizedFrom, normalizedTo, false);
+    }
+
+    /**
+     * Returns a {@link NormalizedFromToRevision} which contains an absolute {@code from} {@link Revision} and
+     * an absolute {@code to} {@link Revision}. If the specified {@code ascending} is {@code true},
+     * the {@link Revision#major} value of the {@code from} in the {@link NormalizedFromToRevision} will be
+     * lower or equal to the {@link Revision#major} value of the {@code to} in the
+     * {@link NormalizedFromToRevision}.
```
trustin: How about:

- Removing `ascending` parameter and returning without switching `from` and `to`
- Adding the following methods to `NormalizedFromToRevision`:
  - `toAscending()`
  - `toDescending()`

..? (Because a user might want to preserve the original ordering)

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/Repository.java
```diff
@@ -388,4 +435,46 @@ default Author author() {
     default <T> CompletableFuture<QueryResult<T>> watch(Revision lastKnownRevision, Query<T> query) {
         return RepositoryUtil.watch(this, lastKnownRevision, query);
     }
+
+    /**
+     * A wrapper class which contains absolute {@code from} and {@code to} {@link Revision}s.
+     * This class is created by {@link #normalizeFromToRevision(Revision, Revision)}
+     * or {@link #normalizeFromToRevision(Revision, Revision, boolean)}.
+     */
+    final class NormalizedFromToRevision {
```
trustin: How about:

- extracting to the top level
- renaming to `RevisionRange`
- moving to the `common` package for future use
- removing `swapped` unless it's really necessary
- allowing to hold relative revisions as well so that it can be used for more use cases
- adding `isRelative()` which would be like `return from.isRelative() || to.isRelative()`

..?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/Repository.java
```diff
@@ -94,10 +97,57 @@ default Author author() {
...
+     */
+    Revision normalizeNow(Revision revision);
+
+    /**
+     * Returns a {@link NormalizedFromToRevision} which contains an absolute {@code from} {@link Revision} and
+     * an absolute {@code to} {@link Revision}.
+     *
+     * @throws RevisionNotFoundException it the specified {@code from} or {@code to} is not found
+     */
+    default NormalizedFromToRevision normalizeFromToRevision(Revision from, Revision to) {
```
trustin: Probably OK to rename to `normalizeNow`?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/Repository.java
```diff
@@ -94,10 +97,57 @@ default Author author() {
     }
 
     /**
-     * Validates the specified {@link Revision} and converts it into an absolute {@link Revision}.
+     * Returns the {@link CompletableFuture} whose value is the absolute {@link Revision} of the
+     * specified {@link Revision}. This will validate the specified {@link Revision} before converting it.
+     *
+     * @see #normalizeNow(Revision)
      */
     CompletableFuture<Revision> normalize(Revision revision);
```
trustin: Could be a default method like: 

```java
try {
    return CompletableFuture.completedFuture(normalizeNow(revision));
} catch (Exception e) {
    return CompletableFutures.exceptionallyCompletedFuture(e);
}
```

and then we can remove `GitRepository.normalize()`

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/cache/CachingRepository.java
```diff
@@ -136,12 +138,15 @@ public Author author() {
...
         requireNonNull(query, "query");
 
-        final CompletableFuture<Object> future = normalizeAndCompose(
-                from, to,
-                (normalizedFrom, normalizedTo) ->
-                        cache.get(new CacheableSingleDiffCall(repo, normalizedFrom, normalizedTo, query)));
+        final NormalizedFromToRevision normalizedFromToRevision;
+        try {
+            normalizedFromToRevision = normalizeFromToRevision(from, to, true);
+        } catch (RevisionNotFoundException e) {
```
trustin: Global comment: `Exception` just in case we had a bug in here?

minwoox:
Let me close and send the PR again :-)

PR #127, opened 2018-01-16
Add Repository.normalizeNow()

minwoox:
Motivation:
`Repository.normalize()` does not have to be asynchronous. But it's hard to change the API
because it is used from everywhere.
Therefore, I decided to add `Repository.normalizeNow()` which returns the normalized `Revision`
right away instead. The `Repository.normalize()` will be removed when we remove the RPC API.

Also in our current `Repository` API, fetching many commits can cause memory spike or OOME.
We need a mechanism to restrict the maximum number of commits at a time.

Modifications:
- Limit the maximum number of commits
- Add `Repository.normalizeNow()` that returns `Revision` immediately

Result:
- Less memory spike

### DIFF
--- a/common/src/main/java/com/linecorp/centraldogma/common/RevisionRange.java
+++ b/common/src/main/java/com/linecorp/centraldogma/common/RevisionRange.java
@@ -1,0 +1,131 @@
+/*
+ * Copyright 2018 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.linecorp.centraldogma.common;
+
+import static java.util.Objects.requireNonNull;
+
+import com.google.common.base.MoreObjects;
+
+/**
+ * A class which contains {@link Revision}s of {@code from} and {@code to}.
+ */
+public class RevisionRange {
+
+    private final Revision from;
+
+    private final Revision to;
+
+    /**
+     * Creates a new instance with the specified integer value of {@code from} and {@code to}.
+     */
+    public RevisionRange(int from, int to) {
+        this(new Revision(from), new Revision(to));
+    }
+
+    /**
+     * Creates a new instance with the specified {@code from} {@link Revision} and {@code to} {@link Revision}.
+     */
+    public RevisionRange(Revision from, Revision to) {
+        this.from = requireNonNull(from, "from");
+        this.to = requireNonNull(to, "to");
+    }
+
+    public Revision from() {
+        return from;
+    }
+
+    public Revision to() {
+        return to;
+    }
+
+    /**
+     * Returns a {@link RevisionRange} whose major value of {@code from} {@link Revision} is lower than
+     * or equal to the major value of {@code to} {@link Revision}.
+     *
+     * @throws UnsupportedOperationException if {@link #isRelative()} returns {@code true}
+     */
+    public RevisionRange toAscending() {
+        if (isAscending() || from.equals(to)) {
+            return this;
+        }
+
+        return new RevisionRange(to, from);
+    }
+
+    /**
+     * Returns a {@link RevisionRange} whose major value of {@code from} {@link Revision} is greater than
+     * or equal to the major value of {@code to} {@link Revision}.
+     *
+     * @throws UnsupportedOperationException if {@link #isRelative()} returns {@code true}
+     */
+    public RevisionRange toDescending() {
+        if (isAscending()) {
+            return new RevisionRange(to, from);
+        }
+
+        return this;
+    }
+
+    /**
+     * Returns {@code true} if the major value of {@code from} {@link Revision} is lower than the major
+     * value of {@code to} {@link Revision}.
+     *
+     * @throws UnsupportedOperationException if {@link #isRelative()} returns {@code true}
+     */
+    public boolean isAscending() {
+        if (isRelative()) {
+            throw new UnsupportedOperationException("contains relative revision. from: " +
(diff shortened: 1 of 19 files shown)


common/src/main/java/com/linecorp/centraldogma/common/RevisionRange.java
```diff
@@ -0,0 +1,131 @@
...
+        return this;
+    }
+
+    /**
+     * Returns {@code true} if the major value of {@code from} {@link Revision} is lower than the major
+     * value of {@code to} {@link Revision}.
+     *
+     * @throws UnsupportedOperationException if {@link #isRelative()} returns {@code true}
+     */
+    public boolean isAscending() {
```
hyangtack: nit: Just curious. How about returning `true` or `false` here instead of throwing an exception? Then, throw the exception when this range is relative in `toAscending()` and `toDescending`? (adding some private method like `ensureAbsoluteRange()`)

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/DefaultMetaRepository.java
```diff
@@ -86,7 +86,7 @@ public DefaultMetaRepository(Repository repo) {
         mirrorLock.lock();
         try {
             // TODO(trustin): Asynchronization
```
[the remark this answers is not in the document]
minwoox: I also found that the lock in blockingCommit() does not need anymore. Can I remove it?

common/src/main/java/com/linecorp/centraldogma/common/RevisionRange.java
minwoox: I think it can confuse the user when it returns `false`. If it returns `false`, a user might think that this is descending or equal revision. WDYT @hyangtack?

common/src/main/java/com/linecorp/centraldogma/common/RevisionRange.java
```diff
@@ -0,0 +1,131 @@
...
+
+    /**
+     * Creates a new instance with the specified {@code from} {@link Revision} and {@code to} {@link Revision}.
+     */
+    public RevisionRange(Revision from, Revision to) {
+        this.from = requireNonNull(from, "from");
+        this.to = requireNonNull(to, "to");
+    }
+
+    public Revision from() {
```
trustin: Missing Javadoc

common/src/main/java/com/linecorp/centraldogma/common/RevisionRange.java
```diff
@@ -0,0 +1,131 @@
...
+        return from;
+    }
+
+    public Revision to() {
```
trustin: Missing Javadoc

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/Repository.java
```diff
@@ -94,9 +97,35 @@ default Author author() {
...
+            return CompletableFuture.completedFuture(normalizeNow(revision));
+        } catch (Exception e) {
+            return CompletableFutures.exceptionallyCompletedFuture(e);
+        }
+    }
+
+    /**
+     * Returns the absolute {@link Revision} of the specified {@link Revision}.
+     *
+     * @throws RevisionNotFoundException it the specified {@link Revision} is not found
```
trustin: it -> if

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/Repository.java
```diff
@@ -94,9 +97,35 @@ default Author author() {
...
+     *
+     * @see #normalize(Revision)
+     */
+    Revision normalizeNow(Revision revision);
+
+    /**
+     * Returns a {@link RevisionRange} which contains the absolute {@link Revision}s of the specified
+     * {@code from} and {@code to}.
+     *
+     * @throws RevisionNotFoundException it the specified {@code from} or {@code to} is not found
```
trustin: it -> if

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/Repository.java
```diff
@@ -94,9 +97,35 @@ default Author author() {
...
      */
-    CompletableFuture<Revision> normalize(Revision revision);
+    RevisionRange normalizeNow(Revision from, Revision to);
```
trustin: Would we need `CompletableFuture<RevisionRange> normalize(from, to)` as well?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
```diff
@@ -378,6 +371,13 @@ private Revision blockingNormalize(Revision revision) {
         }
     }
 
+    @Override
+    public RevisionRange normalizeNow(Revision from, Revision to) {
+        final int baseMajor = cachedHeadRevision().major();
+
+        return new RevisionRange(normalizeNow(from, baseMajor), normalizeNow(to, baseMajor));
```
trustin: You'll have to acquire a read lock:

```java
readLock.lock();
try {
    final int baseMajor = ...
    return new RevisionRange(normalizeNow(...), normalizeNow(...));
} finally {
    readLock.unlock();
}
```

minwoox:
Fixed. PTAL @trustin, @hyangtack  :)

common/src/main/java/com/linecorp/centraldogma/common/RevisionRange.java
```diff
@@ -0,0 +1,140 @@
...
+        return to;
+    }
+
+    /**
+     * Returns the {@link RevisionRange} whose major value of {@code from} {@link Revision} is lower than
+     * or equal to the major value of {@code to} {@link Revision}.
+     *
+     * @throws IllegalStateException if the {@code from} and {@code to} {@link Revision}s are in the
+     *                               different state. they should be either absolute or relative
```
hyangtack: nit: I'm not sure..  `they`->`They` and adding a dot(`.`) to the end of the sentence?

minwoox: Thank you for approving this.
This is one of the hardest stuff that I cannot make a decision.
We do not start with an uppercase and do not end with dot when there is just only one sentence.
But how can I do when there are two sentences? Do I need to use uppercase and dot? Do we have a convention for this? :)

trustin:
Thanks, @minwoox !

PR #130, opened 2018-01-29
Annotate all public packages with ParametersAreNonnullByDefault

trustin:
Motivation:

Some languages such as Kotlin assumes all Java API are nullable unless
annotated with `@Nonnull`. Our API assumes non-null unless annotated
with `@Nullable`.  As a result, Kotlin complains about Armeria usage.

Modifications:

- Add and require `@ParametersAreNonnullByDefault` annotation for all
  packages so that Kotlin assumes non-nullness of Armeria API
- Miscellaneous
  - Remove STATIC_INIT token from EmptyLineSeparator checkstyle rule
    - See https://github.com/line/armeria/issues/700

Result:

- Better Kotlin interop


trustin:
Superceded by #212

PR #131, opened 2018-02-05
Return null on an invalid credential

trustin:
Motivation:

SearchFirstActiveDirectory throws an exception when a user does not
exist or a user provided an invalid password. Shiro will translate this
into a different exception than InvalidCredentialsException.

Modifications:

- Return null instead of throwing an exception when authentication has
  failed due to invalid credentials

Result:

- A user does not get 500 Internal Server Error when he or she entered
  invalid credentials


trustin:
Thanks, reviewers :-)

Issue #59, opened 2017-10-11
Remove Thrift support in favor of RESTful API


trustin:
.. because:

- libthrift isn't stable enough yet, e.g. 0.10.0 introduced breaking changes since 0.9.x. We do not want to ship two client library JARs - one for Thrift 0.9 and the other for 0.10.0.
- A client library can have fewer dependencies, especially without libthrift and Armeria.

To make this come true, we need to:

- Revise our RESTful API & implementation so that it covers all features provided by the current Thrift service.
- Rewrite the client library so it uses the RESTful API.

PR #141, opened 2018-03-05
Provide Go client library

minwoox:
Motivation:
If we provide Go client library, more people can adapt Central Dogma.

Modifications:
- Add Go client library

Result:
- More customers

Todos:
- Repackage the original sources
- Adapt CLI using this library
- Add the watcher


common/src/main/java/com/linecorp/centraldogma/internal/api/v1/AccessToken.java
```diff
@@ -0,0 +1,70 @@
...
+
+    private final String accessToken;
+
+    private final String tokenType = BEARER;
+
+    private final long expiresIn;
+
+    private final String refreshToken;
+
+    //TODO(minwoox) add scope when it needs
```
trustin: Add scope if needed. ?

common/src/main/java/com/linecorp/centraldogma/internal/api/v1/AccessToken.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @JsonProperty("access_token")
+    public String accessToken() {
+        return accessToken;
+    }
+
+    @JsonProperty("expires_in")
+    public long expiresIn() {
+        return expiresIn;
```
trustin: I wonder if we should use `expires_at` so that a client knows when exactly a token will be expired. Or is this a part of OAUTH2 specification?

server/src/main/java/com/linecorp/centraldogma/server/internal/admin/authentication/LoginService.java
```diff
@@ -106,5 +123,56 @@ protected void doPost(ServiceRequestContext ctx, HttpRequest req,
...
+    }
+
+    private static UsernamePasswordToken usernamePassword(List<String> usernames, List<String> passwords) {
+        if (usernames != null && passwords != null) {
+            final String username = usernames.get(0);
+            final String password = passwords.get(0);
+            return new UsernamePasswordToken(username, password);
+        }
+
+        return throwResponseException("invalid_request", "request was missing the username or password.");
```
trustin: `request must contain username and password.`?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/HttpApiUtil.java
```diff
@@ -55,18 +56,35 @@
...
+    public static HttpResponseException newHttpResponseException(HttpStatus status, String message) {
         return HttpResponseException.of(newResponseWithErrorMessage(status, message));
     }
 
+    /**
+     * Returns a newly created {@link HttpResponseException} with the specified {@link HttpStatus} and
+     * {@link JsonNode}.
+     */
+    public static HttpResponseException newHttpResponseException(HttpStatus status, JsonNode node) {
+        return HttpResponseException.of(newResponseWithJson(status, node));
```
trustin: Perhaps we should provide a way to specify an `HttpStatus` with a response object in Armeria?

server/src/test/java/com/linecorp/centraldogma/server/internal/admin/authentication/LoginAndLogoutTest.java
```diff
@@ -50,7 +56,29 @@ static AggregatedHttpMessage login(CentralDogmaRule rule, String username, Strin
...
+
+    static AggregatedHttpMessage loginWithBasicAuth(
+            CentralDogmaRule rule, String username, String password) {
+        return rule.httpClient().execute(
+                HttpHeaders.of(HttpHeaderNames.METHOD, "POST",
+                               HttpHeaderNames.PATH, "/api/v0/authenticate",
+                               HttpHeaderNames.AUTHORIZATION, "basic " + encoder.encodeToString(
+                                       (USERNAME + ':' + PASSWORD).getBytes(StandardCharsets.US_ASCII)),
+                               HttpHeaderNames.CONTENT_TYPE, MediaType.FORM_DATA.toString()),
+                "grant_type=client_credentials",
```
trustin: Shouldn't this be `password`?

minwoox: From the Golang oauth2 client, I thought it was `"grant_type=client_credentials"`.
But I checked again the [spec](https://tools.ietf.org/html/rfc7617#section-2.1), and it actually does not require the `grant_type`.
So I will remove it :-)

common/src/main/java/com/linecorp/centraldogma/internal/api/v1/AccessToken.java
minwoox: Yes, it is part of the [spec](https://auth0.com/docs/tokens/access-token). :-)

common/src/main/java/com/linecorp/centraldogma/internal/api/v1/AccessToken.java
minwoox: As a part of the response, there is a scope.
```
scope (optional) If the scope the user granted is identical to the scope the
app requested, this parameter is optional. If the granted scope is different
from the requested scope, such as if the user modified the scope, then
this parameter is required.
```
in [here](https://www.oauth.com/oauth2-servers/access-tokens/access-token-response/)
But I don't think we need it right now, so just left a comment.

trustin: I actually meant to update the comment to 'Add scope if needed.' rather than asking to add it in this PR. :-D

server/src/test/java/com/linecorp/centraldogma/server/internal/admin/authentication/LoginAndLogoutTest.java
trustin: I thought I deleted this review comment but seems like I forgot. :-$

common/src/main/java/com/linecorp/centraldogma/internal/api/v1/AccessToken.java
minwoox: Oops~ got it.~! Thank you~!

server/src/main/java/com/linecorp/centraldogma/server/internal/api/HttpApiUtil.java
minwoox: I was think about adding methods whose signatures are same as `HttpResponse.of(...)`, but I was not sure which methods I needed to add to `HttpResponseException`, because there are a lot. :-)
To handle above case, I need a method like this.
```java
public static HttpResponseException of(HttpStatus status, MediaType mediaType, byte[] content) {
...
}
or 
public static HttpResponseException of(HttpHeaders headers, byte[] content) {
...
}
```

client/go/dogma/dogma.go
```diff
@@ -0,0 +1,332 @@
...
+	"strings"
+)
+
+const (
+	DefaultScheme     = "http"
+	DefaultHostName   = "localhost"
+	DefaultPort       = 36462
+	DefaultPathPrefix = "api/v1/"
+	PathPrefixV0      = "api/v0/"
+	DefaultBaseURL    = "http://localhost:36462/"
```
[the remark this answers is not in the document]
minwoox: Fixed. Thank you ~!
I thought I couldn't make it because `strconv.Itoa(port)` didn't work. But I found that I can simply do `string(port)`.

client/go/dogma/dogma.go
```diff
@@ -0,0 +1,332 @@
...
+
+type service struct {
+	client *Client
+}
+
+func NewClientWithCredential(baseURL string, config clientcredentials.Config) (*Client, error) {
+	return NewClient(baseURL, config.Client(context.Background()))
+}
+
+func NewClient(baseURL string, client *http.Client) (*Client, error) {
```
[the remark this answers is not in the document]
minwoox: Fixed. Thank you~!

server/src/main/java/com/linecorp/centraldogma/server/internal/api/HttpApiUtil.java
trustin: Yeah we should add those methods so that a user can send an error response conveniently. However, I guess a user will find it weird to throw an exception when sending a successful response.

minwoox: That's right. Do we need to check the status here?

trustin: I guess not. You know, some peculiar API uses 200 OK for error responses ;-)

minwoox: Yeah, we talked about it in this morning. lol

client/go/dogma/content_service_test.go
```diff
@@ -0,0 +1,306 @@
...
+		"path":     "",
+		"oldValue": "bar",
+		"value":    "baz"})
+
+	query := &Query{Path: "/a.json", QueryType: "json_path", Expressions: []string{"$.a"}}
+	_, _, _ = c.GetDiff(context.Background(), "foo", "bar", "3", "4", query)
+	_ = &Change{Path: "/a.json", Type: "APPLY_JSON_PATCH", Content: content}
+
+	// TODO(minwoox) fix this to work
+	//if !reflect.DeepEqual(entry, want) {
```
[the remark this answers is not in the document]
minwoox: Fixed~!!

client/go/dogma/dogma.go
```diff
@@ -0,0 +1,332 @@
...
+)
+
+const (
+	DefaultScheme     = "http"
+	DefaultHostName   = "localhost"
+	DefaultPathPrefix = "api/v1/"
+	PathPrefixV0      = "api/v0/"
+	DefaultBaseURL    = DefaultScheme + "://" + DefaultHostName + ":36462/"
+
+	defaultPort         = 36462
```
trustin: I actually meant making all constants private, except `DefaultPort`. :-)

minwoox: Oops~~ let me fix this~

client/go/dogma/dogma.go
```diff
@@ -0,0 +1,375 @@
...
+	}
+
+	var buf io.ReadWriter
+	if body != nil {
+		if str, ok := body.(string); ok {
+			buf = bytes.NewBufferString(str)
+		} else {
+			buf = new(bytes.Buffer)
+			enc := json.NewEncoder(buf)
+			//enc.SetEscapeHTML(true)
```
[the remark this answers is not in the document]
minwoox: It's necessary. Forgot to bring it back. Thank you!

server/src/main/java/com/linecorp/centraldogma/server/internal/admin/authentication/LoginService.java
```diff
@@ -60,14 +71,13 @@ public LoginService(CentralDogmaSecurityManager securityManager, CommandExecutor
     protected HttpResponse doPost(ServiceRequestContext ctx, HttpRequest req) throws Exception {
         final CompletableFuture<HttpResponse> future = new CompletableFuture<>();
         req.aggregate().thenAccept(aMsg -> {
```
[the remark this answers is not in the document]
minwoox: Fixed! Thank you~!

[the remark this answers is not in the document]
minwoox: Oops I realized that this class inherits the `AbstractHttpService` and the method overrides, so I can not change :-)

server/src/main/resources/webapp/scripts/components/auth/provider/auth.session.service.js
```diff
@@ -12,9 +12,13 @@ angular.module('CentralDogmaAdmin')
                      headers: {
                        'Content-Type': 'application/x-www-form-urlencoded'
                      }
-                   }).then(function (sessionId) {
-                     if (sessionId !== null) {
-                       $window.localStorage.setItem('sessionId', sessionId);
+                   }).then(function (accessToken) {
+                     var sessionId;
```
hyangtack: I'm not sure, but I think we need to check `access_token` property. How about this?
```
 if (accessToken === null ||
     Angular.isUndefined(accessToken.access_token) ||
     accessToken.access_token === null) {
   return null;
 }
 $window.localStorage.setItem('sessionId', accessToken.access_token);
 return accessToken.access_token;
```

huydx:
Curious question before dig into detail implement review: is there any reason to use REST instead of directly thrift?

minwoox:
As described in #59, we are going to remove Thrift :)

client/go/dogma/content_service.go
```diff
@@ -0,0 +1,320 @@
...
+			urlQuery += "&"
+		}
+		urlQuery += fmt.Sprintf("jsonpath=%v", jsonPath)
+	}
+	// the urlQuery starts with '?'
+	return urlQuery, nil
+}
+
+func (con *contentService) getFiles(ctx context.Context,
+	projectName, repoName, revision, pathPattern string) ([]*Entry, *http.Response, error) {
```
huydx: is there any chance that user would care about http.Response? Adding response will make the API couple to http package.

client/go/dogma/content_service.go
```diff
@@ -0,0 +1,320 @@
...
+	u := fmt.Sprintf("%vprojects/%v/repos/%v/contents%v", defaultPathPrefix, projectName, repoName, pathPattern)
+
+	if len(revision) != 0 {
+		u += "?revision=" + revision
+	}
```
huydx: how about use [url#Values pacakge](https://golang.org/pkg/net/url/#Values) to format

client/go/dogma/content_service.go
```diff
@@ -0,0 +1,320 @@
...
+	return entries, res, nil
+}
+
+func (con *contentService) getHistory(ctx context.Context,
+	projectName, repoName, from, to, pathPattern string) ([]*Commit, *http.Response, error) {
+	u := fmt.Sprintf("%vprojects/%v/repos/%v/commits/%v", defaultPathPrefix, projectName, repoName, from)
+
+	urlQuery := ""
+	if len(pathPattern) != 0 {
```
huydx: ditto (using url#Values instead)

client/go/dogma/content_service.go
```diff
@@ -0,0 +1,320 @@
...
+	res, err := con.client.do(ctx, req, &changes)
+	if err != nil {
+		return nil, res, err
+	}
+	return changes, res, nil
+}
+
+type Push struct {
+	CommitMessage *CommitMessage `json:"commitMessage"`
+	Changes       []Change       `json:"changes"`
```
huydx: how about use []*Change instead for uniformity

client/go/dogma/dogma.go
```diff
@@ -0,0 +1,375 @@
...
+			err = nil
+		}
+	}
+
+	return res, err
+}
+
+// CreateProject creates a project.
+func (c *Client) CreateProject(ctx context.Context, name string) (*Project, *http.Response, error) {
+	return c.project.create(ctx, name)
```
huydx: Not sure what is better, but compare between below 2 interfaces, I prefer former one. HDYT?

```
// expose service object
dogma.Project.Create(...)
```

vs

```
// current approach
dogma.CreateProject(...)
```

client/go/dogma/dogma.go
```diff
@@ -0,0 +1,375 @@
...
+
+	projects, res, err := client.ListProjects(context.Background())
+
+Or, you can create a client specifying an http.Client which handles authentication
+for you.
+
+	import "golang.org/x/oauth2/clientcredentials"
+
+	func Client() {
+		config := clientcredentials.Config{ClientID: clientID,
```
huydx: we support oauth only and not basic auth?

client/go/dogma/content_service.go
minwoox: That's a good idea. 
Is there any convenient way to convert the `[]Change` to `[]*Change`?

minwoox:
> One more question: as a user I want something like Watcher in java client, is it possible without Thrift or we need to make some REST API for polling?

I am planning to make the `Watcher` in different PR. I'm still trying to figure out which is the best way to implement it. :-)

client/go/dogma/content_service.go
minwoox: That's a good question~! This is the one that I barely made a decision.

Some go client libraries such as [go-github]("https://github.com/google/go-github") that I referred, returns the `*http.Response`.
I was thinking that a user may want to take a look at the response when something unexpected happens. 
And I made the constructor for client which takes the `*http.Client` as an argument. So I though t is was OK. If the user does not want it, the user can just ignore it with `_`.
But I think I can take that out as well. :-)
Do you think it is too tight with the http package?

client/go/dogma/content_service.go
huydx: I'm afraird not :(

client/go/dogma/dogma.go
huydx: > same method name as in Java client 
this sound reasonable, so let's keep compability for emthod naming rule

client/go/dogma/content_service.go
minwoox: Oops, I totally misunderstood the API. Thank you for letting me know that.

client/go/dogma/content_service.go
minwoox: I just changed the method parameter to take `[]*Change`.
I realized that I don't have to take the argument as varargs because I need at least one `Change`.
Thank you~!

minwoox:
@huydx , Fixed the issues, please take a look.

client/go/dogma/content_service.go
```diff
@@ -0,0 +1,316 @@
...
+
+type contentService service
+
+// Query specifies a query on a file.
+type Query struct {
+	Path string
+	// QueryType can be "identity" or "json_path". "identity" is used to retrieve the content as it is.
+	// "json_path" applies a series of JSON path to the content.
+	// See https://github.com/json-path/JsonPath/blob/master/README.md
+	QueryType   string
```
huydx: sorry I missed that at first round, do you think string is better than iota? 
Using iota may need some more boilerplate code to generate equivalent string, but for users it's easier to use because they know that what type of QueryType is supported, and prevent typo as well.

client/go/dogma/content_service.go
```diff
@@ -0,0 +1,316 @@
...
+	Expressions []string
+}
+
+// Entry represents an entry in the repository.
+type Entry struct {
+	Path       string      `json:"path"`
+	Type       string      `json:"type"` // can be JSON, TEXT or DIRECTORY
```
huydx: same with above comment about iota

client/go/dogma/content_service.go
```diff
@@ -0,0 +1,316 @@
...
+	Detail  string `json:"detail,omitempty"`
+	Markup  string `json:"markup,omitempty"`
+}
+
+// Change represents a change to commit in the repository.
+type Change struct {
+	Path string `json:"path"`
+
+	// can be UPSERT_JSON, UPSERT_TEXT, REMOVE, RENAME, APPLY_JSON_PATCH or APPLY_TEXT_PATCH
+	Type    string      `json:"type"`
```
huydx: same with above comment about iota

client/go/dogma/dogma.go
```diff
@@ -0,0 +1,362 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	config := clientcredentials.Config{ClientID: clientID, ClientSecret: clientSecret,
+		TokenURL: normalizedURL.String() + pathAuthenticate}
+	return newClientWithHTTPClient(normalizedURL.String(), config.Client(context.Background()))
+}
+
+// newClientWithHTTPClient returns a Central Dogma client withe the specified baseURL and client.
```
huydx: typo withe

client/go/dogma/dogma.go
```diff
@@ -0,0 +1,362 @@
...
+	repository *repositoryService
+	content    *contentService
+}
+
+type service struct {
+	client *Client
+}
+
+// NewClient returns a Central Dogma client with the specified baseURL, clientID and clientSecret.
+func NewClient(baseURL string, clientID, clientSecret string) (*Client, error) {
```
huydx: how about 2 separate constructors like

```
NewClient(baseUrl string) *Client, error {
}

NewSecureClient(baseURL string, clientID, clientSecret string) *Client, error {
}
```

And in inner implement we will check if clientID and clientSecret is empty, we will use secure API and opposite?
Not sure about the spec of CD server right now.

huydx:
@minwoox added second round comments

client/go/dogma/content_service.go
minwoox: `iota` should be much better~! Thank you~!

client/go/dogma/content_service.go
minwoox: This field is used to store the JSON response. If I use this as `iota`, I need a custom deserializer  to convert the sting to corresponding iota number.
Can I implement the custom deserializer or is it better to use predefined string constants?

client/go/dogma/dogma.go
minwoox: Thanks~!

client/go/dogma/dogma.go
minwoox: I think having just one constructor is fine for now so that the client does not have to handle the authentication by itself. Let's add it when we think it really needs :-)

client/go/dogma/content_service.go
huydx: As a user, I would prefer iota constant. IMHO it's fine to add deserialize function like `func (o *Type) UnmarshalJSON(b []byte) error {}`

minwoox: Got it. I will try the way you suggest. Thang you!

minwoox:
@huydx Please take a look again~. I was trying to use some library like [enumer](https://github.com/alvaroloes/enumer) to convert the iota values, but it seems like I coudln't not have different names for the enum. For example, we use `UPSERT_JSON` in the Central Dogma server, but 
If I name the values as golang style like `UpsertJSON`, I cannot convert the value using that library.
So, I decide to just declare maps and arrays to convert that.

minwoox:
Thanks @huydx. Let's wait a little bit in case @trustin, @hyangtack want to leave a comment.

minwoox:
I misunderstood the Oauth2 spec. I removed the client credentials part. :-)

client/go/dogma/dogma.go
```diff
@@ -77,16 +78,20 @@ type service struct {
 	client *Client
 }
 
-// NewClient returns a Central Dogma client with the specified baseURL, clientID and clientSecret.
-func NewClient(baseURL string, clientID, clientSecret string) (*Client, error) {
+// NewClient returns a Central Dogma client with the specified baseURL, username and password.
+func NewClient(baseURL string, username, password string) (*Client, error) {
```
hyangtack: nit: Just curious. `username` -> `username string`? Is the type unnecessary?

minwoox: Oops~! Thank you for pointing it out. `(baseURL, username, password string)` is correct because all of the three parameters are string, I can just put one `string` at the end. (does not have to put type for every parameter, if they are same)

minwoox:
Thanks reviews~!

PR #144, opened 2018-03-06
Choose the last matching client profile

trustin:
Motivation:

- Spring Boot uses the last matching profile, which is opposed to
  CentralDogmaBuilder.profile(), which chooses the first matching one.

Modifications:

- Choose the last matching client profile.
- Make profile matching more robust by ignoring problematic profiles
  instead of raising an exception

Result:

- Works better with Spring Boot
- Deals better with poorly written profile


trustin:
Thanks, reviewers!

PR #147, opened 2018-03-08
Expose TLS configuration

### LINKED ISSUE #146: Support TLS
Currently Central Dogma instance does not support TLS. We need to add a property to its configuration for supporting TLS.

hyangtack:
Motivation:
Some user may feel anxiety when sending their login password via cleartext protocol such as http and h2c.
So we need to provide a way to protect their data from the network.

Modifications:
- Add `tls` property to `dogma.json` for specifying paths of certificates file and a key file.

Result:
- Closes #146

### DIFF
--- a/client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java
+++ b/client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java
@@ -89,6 +89,59 @@ public interface CentralDogma {
     }
 
     /**
+     * Returns a newly-created {@link CentralDogma} instance which connects to the given host at port 36462
+     * using the default {@link ClientFactory}. This {@link CentralDogma} instance connects to the server
+     * with TLS enabled.
+     *
+     * @param host the host name or IP address of the Central Dogma server
+     */
+    static CentralDogma forTlsHost(String host) {
+        return new CentralDogmaBuilder().host(host).useTls().build();
+    }
+
+    /**
+     * Returns a newly-created {@link CentralDogma} instance which connects to the given host and port
+     * using the default {@link ClientFactory}. This {@link CentralDogma} instance connects to the server
+     * with TLS enabled.
+     *
+     * @param host the host name or IP address of the Central Dogma server
+     * @param port the port number of the Central Dogma server
+     */
+    static CentralDogma forTlsHost(String host, int port) {
+        return new CentralDogmaBuilder().host(host, port).useTls().build();
+    }
+
+    /**
+     * Returns a newly-created {@link CentralDogma} instance which connects to the given host at port 36462
+     * using the specified {@link ClientFactory}. This {@link CentralDogma} instance connects to the server
+     * with TLS enabled.
+     *
+     * @param clientFactory the {@link ClientFactory} that will manage the connections
+     * @param host the host name or IP address of the Central Dogma server
+     */
+    static CentralDogma forTlsHost(ClientFactory clientFactory, String host) {
+        return new CentralDogmaBuilder().clientFactory(clientFactory)
+                                        .host(host)
+                                        .useTls()
+                                        .build();
+    }
+
+    /**
+     * Returns a newly-created {@link CentralDogma} instance which connects to the given host and port
+     * using the specified {@link ClientFactory}. This {@link CentralDogma} instance connects to the server
+     * with TLS enabled.
+     *
+     * @param clientFactory the {@link ClientFactory} that will manage the connections
+     * @param host the host name or IP address of the Central Dogma server
+     */
+    static CentralDogma forTlsHost(ClientFactory clientFactory, String host, int port) {
+        return new CentralDogmaBuilder().clientFactory(clientFactory)
+                                        .host(host, port)
+                                        .useTls()
+                                        .build();
+    }
+
+    /**
      * Returns a newly-created {@link CentralDogma} instance with the given profile names and the default
      * {@link ClientFactory}.
      *
--- a/client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaBuilder.java
+++ b/client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaBuilder.java
@@ -65,6 +65,7 @@ public class CentralDogmaBuilder {
 
     private ClientFactory clientFactory = ClientFactory.DEFAULT;
     private List<Endpoint> endpoints = new ArrayList<>();
+    private boolean useTls;
     private String selectedProfile;
     private ArmeriaClientConfigurator clientConfigurator = cb -> {
     };
@@ -119,6 +120,21 @@ public class CentralDogmaBuilder {
 
         checkState(selectedProfile == null, "profile() and host() cannot be used together.");
         endpoints.add(Endpoint.parse(host + ':' + port));
+        return this;
+    }
+
+    /**
+     * Sets the client to use TLS.
+     */
+    public CentralDogmaBuilder useTls() {
+        return useTls(true);
+    }
+
+    /**
+     * Sets whether the client uses TLS or not.
+     */
+    public CentralDogmaBuilder useTls(boolean useTls) {
+        this.useTls = useTls;
         return this;
     }
 
@@ -301,7 +317,8 @@ public class CentralDogmaBuilder {
             endpoint = Endpoint.ofGroup(groupName);
         }
 
-        final String uri = "tbinary+http://" + endpoint.authority() + "/cd/thrift/v1";
+        final String scheme = "tbinary+" + (useTls ? "https" : "http") + "://";
+        final String uri = scheme + endpoint.authority() + "/cd/thrift/v1";
         final ClientBuilder builder = new ClientBuilder(uri)
                 .factory(clientFactory)
                 .decorator(RpcRequest.class, RpcResponse.class,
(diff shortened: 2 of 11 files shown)


site/src/sphinx/setup-configuration.rst
```diff
@@ -17,9 +17,14 @@ defaults:
             "host": "*",
             "port": 36462
           },
-          "protocol": "http"
+          "protocol": "https"
```
trustin: How about keeping this code block identical with the default `dogma.json` and adding a dedicated section for configuring TLS instead?

site/src/sphinx/setup-configuration.rst
```diff
@@ -200,9 +221,14 @@ Once you have an access to a ZooKeeper cluster, update the ``replication`` secti
...
```
trustin: Same here. Let's add a dedicated section for TLS configuration and keep using `"protool": "http"` and `"tls": null`

server/src/main/java/com/linecorp/centraldogma/server/TlsConfig.java
```diff
@@ -0,0 +1,42 @@
...
+package com.linecorp.centraldogma.server;
+
+import java.io.File;
+
+import javax.annotation.Nullable;
+
+/**
+ * TLS configuration.
+ */
+public interface TlsConfig {
```
trustin: Is there any reason you split TLS configuration into an interface and an implementation?

trustin:
What do you think about generating a self-signed certificate automatically if a user did not specify `TlsConfig` and enabled `https` protocol? (We could log at WARN about it when we're doing this.)

server/src/main/java/com/linecorp/centraldogma/server/TlsConfig.java
hyangtack: At first time, they were slightly different due to the properties for specifying file path. But now I realized they became same, so I think we don't need to split them. I'll fix it. :-)

site/src/sphinx/setup-configuration.rst
hyangtack: OK!

hyangtack:
> What do you think about generating a self-signed certificate automatically if a user did not specify TlsConfig and enabled https protocol? (We could log at WARN about it when we're doing this.)

That's good suggestion. I will add code for that. Thanks!

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaBuilder.java
```diff
@@ -301,7 +317,8 @@ public CentralDogma build() {
             endpoint = Endpoint.ofGroup(groupName);
         }
 
-        final String uri = "tbinary+http://" + endpoint.authority() + "/cd/thrift/v1";
+        final String scheme = useTls ? "tbinary+https://" : "tbinary+http://";
```
trustin: nit: `"tbinary+" + (useTls ? "https" : "http") + "://" ?

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -321,8 +324,29 @@ private CommandExecutor startCommandExecutor(
...
+        }
+        if (requiresTls) {
+            try {
+                final TlsConfig tlsConfig = cfg.tls();
+                if (tlsConfig != null) {
+                    sb.sslContext(SessionProtocol.HTTPS,
+                                  tlsConfig.keyCertChainFile(), tlsConfig.keyFile(), tlsConfig.keyPassword());
+                } else {
+                    logger.warn(
+                            "Missing TLS configuration. Generating self-signed certificates for TLS support.");
```
trustin: a self-signed certificate?

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -321,8 +324,29 @@ private CommandExecutor startCommandExecutor(
...
             sb.port(p);
+            if (p.protocol().isTls()) {
+                requiresTls = true;
+            }
+        }
+        if (requiresTls) {
+            try {
+                final TlsConfig tlsConfig = cfg.tls();
+                if (tlsConfig != null) {
+                    sb.sslContext(SessionProtocol.HTTPS,
```
trustin: Now we can use the `tls(...)` method.

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -321,8 +324,29 @@ private CommandExecutor startCommandExecutor(
...
+                                  tlsConfig.keyCertChainFile(), tlsConfig.keyFile(), tlsConfig.keyPassword());
+                } else {
+                    logger.warn(
+                            "Missing TLS configuration. Generating self-signed certificates for TLS support.");
+                    final SelfSignedCertificate ssc = new SelfSignedCertificate();
+                    sb.sslContext(SessionProtocol.HTTPS, ssc.certificate(), ssc.privateKey());
```
trustin: Now we can use the `tls(...)` method.

site/src/sphinx/setup-configuration.rst
```diff
@@ -261,3 +268,81 @@ Once you have an access to a ZooKeeper cluster, update the ``replication`` secti
...
+      "numMirroringThreads": null,
+      "maxNumFilesPerMirror": null,
+      "maxNumBytesPerMirror": null,
+      "accessLogFormat": "common"
+    }
+
+- ``tls``
+
+  - the configuration for TLS support. It will be applied to the port which is configured with ``https``
+    protocol. Specify ``null`` to disable TLS.
```
trustin: This needs an update.

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java
```diff
@@ -88,6 +88,59 @@ static CentralDogma forHost(ClientFactory clientFactory, String host, int port)
                                         .build();
     }
 
+    /**
+     * Returns a newly-created {@link CentralDogma} instance which connects to the given host at port 36462
+     * using the default {@link ClientFactory}. This {@link CentralDogma} instance connects to the server
+     * with TLS layer.
```
trustin: How about 'with TLS enabled`?

site/src/sphinx/setup-configuration.rst
```diff
@@ -66,7 +67,12 @@ Core properties
 
   - ``protocol`` (string)
 
-    - the protocol. ``http`` is the only supported protocol at the moment.
+    - the protocol. ``http`` and ``https`` are supported.
+
+- ``tls``
+
+  - the configuration for Transport Layer Security(TLS) support. Specify ``null`` to disable TLS.
+    It will be explained later in this page.
```
trustin: Could you use hyperlink? See `concepts.rst` and `client-java.rst` for an example: https://github.com/line/centraldogma/search?l=reStructuredText&q=concepts&type=&utf8=%E2%9C%93

site/src/sphinx/setup-configuration.rst
```diff
@@ -268,3 +275,81 @@ Once you have an access to a ZooKeeper cluster, update the ``replication`` secti
 
   -  the minimum allowed age of log items before they are removed from ZooKeeper. If ``null`` the default
      value of '3600000 milliseconds' (1 hour) is used.
+
+Configuring TLS
+---------------
+Central Dogma supports TLS for its API and web pages. To enable TLS, a user may configure ``tls`` property
+in ``dogma.json`` as follows. Note that Central Dogma does not support Server Name Indication(SNI) extension
+at the moment.
```
trustin: - How about adding a space before `(`?
- Would there any use case of SNI for Central Dogma ?

site/src/sphinx/setup-configuration.rst
```diff
@@ -268,3 +275,81 @@ Once you have an access to a ZooKeeper cluster, update the ``replication`` secti
...
+
+    - the path to the certificate chain file.
+
+  - ``keyFile`` (string)
+
+    - the path to the private key file.
+
+  - ``keyPassword`` (string)
+
+    - the password of the private key file. Specify ``null`` if no password is set.
```
trustin: How about adding:

    Note that `null` (no password) and `"null"` (password is 'null') are different.

site/src/sphinx/setup-configuration.rst
hyangtack: > Would there any use case of SNI for Central Dogma ?

I think that there's not so much use case of SNI. I guess most of our users may run their Central Dogma under a single service domain. Would it be better to remove the sentence?

site/src/sphinx/setup-configuration.rst
```diff
@@ -73,7 +73,7 @@ Core properties
 - ``tls``
 
   - the configuration for Transport Layer Security(TLS) support. Specify ``null`` to disable TLS.
-    It will be explained later in this page.
+    It will be explained in :ref:`tls`.
```
trustin: How about: 

    See :ref:`tls` for more information.

site/src/sphinx/setup-configuration.rst
```diff
@@ -147,7 +147,7 @@ Core properties
   - ``method`` (string)
 
     - the replication method. ``NONE`` indicates 'standalone mode' without replication. ZooKeeper-based
-      multi-master replication will be explained later in this page.
+      multi-master replication will be explained in :ref:`replication`.
```
trustin: Oh, thanks! How about then

    See :ref:`replication` to learn how to configure ZooKeeper-based multi-master replication.

hyangtack:
Thanks, reviewers!

PR #149, opened 2018-03-09
Update Armeria to 0.59.0

trustin:
- Also removed some usage of deprecated API
- Use `Instant` instead of `Date` or `ZonedDateTime`


trustin:
Thanks, reviewers!

PR #165, opened 2018-03-15
Fix a condition to do not create duplicated armeriaClient

julnamoo:
Motivations:

`CentralDogmaConfiguration` always creates `clientFactory` bean due to
the missing condition checking when retrieved `methodMetadata` is not a type of `StandardMethodMetadata `

Modifications:

- Rename to `CentralDogmaClientAutoConfiguration`
- Move the configuration condition
- Rename `armeriaClientFactory` to avoid bean overriding due to the
common name

Result:

- Do not duplicate `@ForCentralDogma` annotated armeria client factory
- Easy to understand the role of the configuration class
- Enable to load the configuration context when it is necessary only

### DIFF
--- a/client/java-spring-boot-autoconfigure/src/main/java/com/linecorp/centraldogma/client/spring/CentralDogmaClientAutoConfiguration.java
+++ b/client/java-spring-boot-autoconfigure/src/main/java/com/linecorp/centraldogma/client/spring/CentralDogmaClientAutoConfiguration.java
@@ -17,7 +17,6 @@ package com.linecorp.centraldogma.client.spring;
 
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
-import java.lang.reflect.Method;
 import java.util.Optional;
 
 import javax.inject.Qualifier;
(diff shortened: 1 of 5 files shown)


client/java-spring-boot-autoconfigure/src/main/java/com/linecorp/centraldogma/client/spring/CentralDogmaClientAutoConfiguration.java
```diff
@@ -118,11 +116,8 @@ private static boolean hasQualifier(ConfigurableListableBeanFactory beanFactory,
                 if (beanDef instanceof AnnotatedBeanDefinition) {
                     final AnnotatedBeanDefinition abd = (AnnotatedBeanDefinition) beanDef;
                     final MethodMetadata fmm = abd.getFactoryMethodMetadata();
-                    if (fmm instanceof StandardMethodMetadata) {
-                        final Method method = ((StandardMethodMetadata) fmm).getIntrospectedMethod();
-                        if (AnnotationUtils.getAnnotation(method, ForCentralDogma.class) != null) {
-                            return true;
-                        }
+                    if (fmm.isAnnotated(ForCentralDogma.class.getName())) {
+                        return true;
```
trustin: Thanks for simplification.

minwoox:
Thanks @julnamoo

PR #168, opened 2018-03-20
Add vectorized logo

trustin:
`central_dogma.svg` was not really a vector image but just an SVG file
with a rasterized PNG image embedded. This commit adds the truly
vectorized AI file and removes the offending SVG file.

Note: I tried to export the AI file into SVG but it seems like meshes
are not converted correctly by Inkscape. The PNG image exported from
Inkscape-generated SVG contains grid-ish artifacts in the logo.


trustin:
Had to remove the .svg file because Inkscape does not seem to convert correctly.

PR #170, opened 2018-03-22
Add /api/v1/login and /api/v1/logout

trustin:
Motivation:

The login/logout API is currently only under /api/v0. We should provide
same service under /api/v1, so we can remove /api/v0 later.

Modifiations:

Bind login/logout service at /api/v1/login and /api/v1/logout as well.

Result:

Easier to drop the old HTTP API


trustin:
Thanks!

PR #181, opened 2018-04-02
Allow specifying multiple protocols on a single port

hyangtack:
Modifications:
- Change `protocol` field to `protocols` in `dogma.json`. Its type is also changed to an array.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaConfig.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaConfig.java
@@ -49,7 +49,6 @@ import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.fasterxml.jackson.databind.node.JsonNodeType;
 import com.fasterxml.jackson.databind.util.StdConverter;
-import com.google.common.base.Ascii;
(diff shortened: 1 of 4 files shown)


server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaConfig.java
```diff
@@ -358,8 +360,8 @@ public ServerPort deserialize(JsonParser p, DeserializationContext ctx) throws I
                 return fail(ctx, root);
             }
 
-            final JsonNode proto = root.get("protocol");
-            if (proto == null || proto.getNodeType() != JsonNodeType.STRING) {
```
trustin: How about retaining this for backward compatibility?

hyangtack: Good point. Fixed!

site/src/sphinx/setup-configuration.rst
```diff
@@ -67,9 +69,9 @@ Core properties
 
       - the port number. 36462 is preferred.
 
-  - ``protocol`` (string)
+  - ``protocols`` (set of string)
```
minwoox: set of strings?

hyangtack: Fixed. Thanks!

hyangtack:
Thanks, reviewers!

PR #182, opened 2018-04-03
Add 500ms delay when refreshing web UI after metadata modification

hyangtack:
Motivation:
A user may get a dirty result after modifying metadata due to replication lag. Adding some delay before refreshing may help to make the situation less often.

Modifications:
- Add 500ms delay before refreshing web UI.


trustin:
Thanks, @hyangtack !

PR #183, opened 2018-04-03
Make web UI better

hyangtack:
Modifications:
- Change icons: `pencil` -> `cog` and `edit`, `close` -> `trash`
- Update docs


site/src/sphinx/auth.rst
```diff
@@ -129,6 +129,9 @@ Anyone who is logged into the Central Dogma can create a new ``Application Token
 for everyone. So any owner of a project can add any token to their project. However only both the token
 creator and the administrator are allowed to deactivate and/or remove the token.
 
+There are two levels of a token, which are ``Admin`` and ``User``. ``Admin`` level token can be created by
+only the administrators. It is one of the administrators, so it is also allowed to administrator-level APIs.
```
trustin: allowed to -> allowed to access

site/src/sphinx/auth.rst
```diff
@@ -129,6 +129,10 @@ Anyone who is logged into the Central Dogma can create a new ``Application Token
 for everyone. So any owner of a project can add any token to their project. However only both the token
 creator and the administrator are allowed to deactivate and/or remove the token.
 
+There are two levels of a token, which are ``Admin`` and ``User``. ``Admin`` level token can be created by
+only the administrators. It is one of the administrators, so it is also allowed to access administrator-level
```
minwoox: How about changing to `A user who uses the token is allowed to ...`

hyangtack: @minwoox 
Fixed as follows:
`A client who sends a request with the token is allowed to access administrator-level APIs.`

I think that `client` is better than `user` in this section. Please let me know if you have an another idea!

trustin:
Thanks, @hyangtack !

PR #185, opened 2018-04-04
Add Caffeine cache for session information

### LINKED ISSUE #177: Response with the same token in LoginService if the user already logged in
Currently, if a user logins with his or her `username` and `password` consecutively, `LoginService` creates sessions for each login and responses with different tokens. If the token is not expired, we do not have to issue a new token again because the previous token is still valid. Also, it can be a burden because the server has to send `CreateToken` `Command` to all the replicas to create a token.

minwoox:
Motivation:
Currently, if a user logins with his or her `username` and `password` consecutively,
`LoginService` creates sessions for each login and responses with different tokens.
If the token is not expired, we do not have to issue a new token again because the
previous token is still valid. Also, it can be a burden because the server has to
send `CreateToken` Command to all the replicas to create a token.

Modifications:
- Add `sessionCacheSpec`
- Rename `cacheSpec` to `repositoryCacheSpec` to distinguish
- Add `Caffeine` cache to `LoginService`
- Change the cache inside `SecurityManager` to `Caffeine` as well

Result:
- Close #177

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaBuilder.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaBuilder.java
@@ -16,6 +16,7 @@
 
 package com.linecorp.centraldogma.server;
 
+import static com.linecorp.centraldogma.server.internal.storage.repository.cache.RepositoryCache.validateCacheSpec;
 import static java.util.Objects.requireNonNull;
 
 import java.io.File;
@@ -26,6 +27,7 @@ import java.util.Collections;
 import java.util.List;
 
 import org.apache.shiro.config.Ini;
+import org.apache.shiro.session.Session;
 
 import com.github.benmanes.caffeine.cache.CaffeineSpec;
 import com.google.common.collect.ImmutableSet;
@@ -33,7 +35,7 @@ import com.google.common.collect.ImmutableSet.Builder;
 
 import com.linecorp.armeria.common.SessionProtocol;
 import com.linecorp.armeria.server.ServerPort;
-import com.linecorp.centraldogma.server.internal.storage.repository.cache.RepositoryCache;
+import com.linecorp.centraldogma.server.internal.storage.repository.Repository;
 
 /**
  * Builds a {@link CentralDogma} server.
@@ -55,9 +57,15 @@ public final class CentralDogmaBuilder {
     static final int DEFAULT_NUM_MIRRORING_THREADS = 16;
     static final int DEFAULT_MAX_NUM_FILES_PER_MIRROR = 8192;
     static final long DEFAULT_MAX_NUM_BYTES_PER_MIRROR = 32 * 1048576; // 32 MiB
-    static final String DEFAULT_CACHE_SPEC = "maximumWeight=134217728," + // Cache up to apx. 128-megachars
-                                             "expireAfterAccess=5m";      // Expire on 5 minutes of inactivity
+
     static final long DEFAULT_WEB_APP_SESSION_TIMEOUT_MILLIS = 604800000;   // 7 days
+    static final String DEFAULT_REPOSITORY_CACHE_SPEC =
+            "maximumWeight=134217728," + // Cache up to apx. 128-megachars.
+            "expireAfterAccess=5m";      // Expire on 5 minutes of inactivity.
+
+    static final String DEFAULT_SESSION_CACHE_SPEC =
+            // Expire after the duration of session timeout.
+            "maximumSize=8192,expireAfterWrite=" + (DEFAULT_WEB_APP_SESSION_TIMEOUT_MILLIS / 1000) + "s";
 
     // Armeria properties
     // Note that we use nullable types here for optional properties.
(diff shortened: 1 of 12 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/admin/authentication/LoginService.java
```diff
@@ -161,6 +176,18 @@ private UsernamePasswordToken usernamePassword(AggregatedHttpMessage req) {
         return throwResponseException("invalid_request", "request must contain username and password.");
     }
 
+    @Nullable
+    private AccessToken currentUserTokenIfPresent(UsernamePasswordToken usernamePassword) {
+        securityManager.authenticate(usernamePassword);
+        // Because securityManager.authenticate does not throw any Exception, the user is authenticated.
+        final AccessToken currentUserToken = cache.getIfPresent(usernamePassword.getUsername());
+        if (currentUserToken != null && currentUserToken.deadline() > System.currentTimeMillis()) {
```
hyangtack: nit: Do we need to add the minimum remaining period for reusing the existing token? e.g.
```
currentUserToken.deadline() > System.currentTimeMillis() + 3600000
```

server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaBuilder.java
```diff
@@ -357,9 +390,10 @@ private CentralDogmaConfig buildConfig() {
 
         return new CentralDogmaConfig(dataDir, ports, tls, numWorkers, maxNumConnections,
                                       requestTimeoutMillis, idleTimeoutMillis, maxFrameLength,
-                                      numRepositoryWorkers, cacheSpec, gracefulShutdownTimeout,
-                                      webAppEnabled, webAppSessionTimeoutMillis,
-                                      mirroringEnabled, numMirroringThreads,
+                                      // TODO(minwoox) remove one repositoryCacheSpec
+                                      numRepositoryWorkers, repositoryCacheSpec, repositoryCacheSpec,
```
hyangtack: It might be OK to replace the second `repositoryCacheSpec` with `null`, isn't it?

server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaConfig.java
```diff
@@ -226,9 +238,25 @@ int numRepositoryWorkers() {
         return numRepositoryWorkers;
     }
 
+    /**
+     * Returns the {@code repositoryCacheSpec}.
+     *
+     * @deprecated Use {@link #repositoryCacheSpec()} instead.
```
trustin: `instead` is redundant.

server/src/main/java/com/linecorp/centraldogma/server/internal/admin/authentication/LoginService.java
```diff
@@ -161,6 +176,18 @@ private UsernamePasswordToken usernamePassword(AggregatedHttpMessage req) {
...
+            return new AccessToken(currentUserToken.accessToken(),
+                                   currentUserToken.deadline() - System.currentTimeMillis());
```
trustin: Could call `currentTimeMillis()` once and reuse.

server/src/main/java/com/linecorp/centraldogma/server/internal/admin/authentication/LoginService.java
trustin: An hour sounds OK to me. Just a minute should be OK?

server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaBuilder.java
```diff
@@ -55,9 +57,15 @@
...
-                                             "expireAfterAccess=5m";      // Expire on 5 minutes of inactivity
+
     static final long DEFAULT_WEB_APP_SESSION_TIMEOUT_MILLIS = 604800000;   // 7 days
+    static final String DEFAULT_REPOSITORY_CACHE_SPEC =
+            "maximumWeight=134217728," + // Cache up to apx. 128-megachars.
+            "expireAfterAccess=5m";      // Expire on 5 minutes of inactivity.
+
+    static final String DEFAULT_SESSION_CACHE_SPEC =
+            // Expire after the duration of session timeout.
+            "maximumSize=4096,expireAfterWrite=" + (DEFAULT_WEB_APP_SESSION_TIMEOUT_MILLIS / 1000) + "s";
```
trustin: Probably OK to use a larger value? Even if we assume a token takes 1 KiB, it will only be 4 MiB. How about 8192, which should cover almost all employees in most engineering organizations?

site/src/sphinx/setup-configuration.rst
```diff
@@ -110,14 +111,21 @@ Core properties
...
     access cache. Refer to `the Caffeine API documentation
     <https://static.javadoc.io/com.github.ben-manes.caffeine/caffeine/2.5.5/index.html?com/github/benmanes/caffeine/cache/CaffeineSpec.html>`_
     for more information. Note that the weight of the cache has been tuned to be roughly proportional to its
     memory usage.
 
+- ``sessionCacheSpec`` (string)
+
+  - the cache specification string which determines the capacity and behavior of the cache for the session
+    information of the server. Refer to `the Caffeine API documentation
+    <https://static.javadoc.io/com.github.ben-manes.caffeine/caffeine/2.5.5/index.html?com/github/benmanes/caffeine/cache/CaffeineSpec.html>`_
```
trustin: Instead of repeating the same URL twice, you could extract it into the beginning of the `.rst` file.

minwoox:
All fixed. Thank you~!

trustin:
Thanks, @minwoox !

PR #186, opened 2018-04-04
Try again quietly when LDAP server is unavailable

hyangtack:


server/src/main/java/com/linecorp/centraldogma/server/support/shiro/SearchFirstActiveDirectoryRealm.java
```diff
@@ -101,6 +102,16 @@ protected void setSearchTimeoutMillis(int searchTimeoutMillis) {
     @Override
     protected AuthenticationInfo queryForAuthenticationInfo(
             AuthenticationToken token, LdapContextFactory ldapContextFactory) throws NamingException {
+        try {
+            return queryForAuthenticationInfo0(token, ldapContextFactory);
+        } catch (ServiceUnavailableException e) {
+            // It might be a temporary failure, so try again.
```
minwoox: Just wondering if we need to wait a little bit.

hyangtack: Good point! But here is not a good place to wait because we have no other option except for `Thread.sleep()` here. As I discussed with @trustin, `LoginService` is better place to wait and retry. Let me merge this PR first then refactor later because most of cases I saw seem like 'connection closed by peer', so the connection might be recovered immediately by reconnecting, I guess. :-)

hyangtack:
Thanks, reviewers!
And I filed an issue regarding `retrying with a delay`: #188

PR #189, opened 2018-04-04
Make searching available

hyangtack:
Motivation:
A user cannot search his or her files from web UI.

Modifications:
- Remove buggy code which messes the specified search term.

Result:
A user can search his or her files from web UI.


hyangtack:
Thanks, reviewers!

PR #192, opened 2018-04-06
Fix RejectedExecutionException when mirroring workers are busy

trustin:
Motivation:

DefaultMirroringService uses a SynchronousQueue as the task queue of
mirroring worker thread pool so that the worker queue does not grow
infinitely when the workers cannot catch up the schedule.

However, ThreadPoolExecutor will fail to submit a mirroring task with a
RejectedExecutionException when all workers are busy, because a
SynchronousQueue does not have capacity.

When a RejectedExecutionException is triggered, DefaultMirroringService
will stop schedule mirroring completely.

Modifications:

- Specify RejectedExecutionHandler so that RejectedExecutionExeception
  is not triggered and the scheduler simply waits until the task is accepted
- Add a test case

Result:

MirroringService does not terminate when it cannot catch up.

### DIFF
--- a/server/src/test/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirroringServiceTest.java
+++ b/server/src/test/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirroringServiceTest.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2018 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


server/src/test/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirroringServiceTest.java
```diff
@@ -0,0 +1,89 @@
...
+import java.net.URI;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+import com.cronutils.model.Cron;
+import com.cronutils.model.CronType;
+import com.cronutils.model.definition.CronDefinition;
```
minwoox: Check style fail~

trustin: Oops!

trustin:
Thank you for reviewing!

PR #194, opened 2018-04-06
Normalize the path parameter in HTTP API v1

minwoox:
Motivations:
Currently, it does not return direcotries when calling `listFiles` or `getFiles`.
And there are a lot of chances that a user requests with a `path` that is not following
the pattern completely  such as a direcotry without ending slash.

Modifications:
- Normalizes the path so that it can handle most of cases
- Changes to return `CompletebleFuture` in HTTP API V1 for consistency

Result:
- Can use HTTP API v1

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
@@ -83,6 +83,9 @@ import com.linecorp.centraldogma.server.internal.storage.repository.Repository;
 @ExceptionHandler(HttpApiExceptionHandler.class)
 public class ContentServiceV1 extends AbstractService {
 
+    private static final Map<FindOption<?>, ?> FIND_WITHOUT_CONTENT =
+            ImmutableMap.of(FindOption.FETCH_CONTENT, false);
+
     private final WatchService watchService;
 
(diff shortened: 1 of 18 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
```diff
@@ -90,40 +92,65 @@ public ContentServiceV1(ProjectManager projectManager, CommandExecutor executor,
...
-    }
-
-    private static String appendWildCardIfDirectory(String path) {
-        return path.endsWith("/") ? path + '*' : path;
+    private static void listFiles(Repository repository, String pathPattern, Revision normalizedRevision,
+                                  Map<FindOption<?>, ?> options, CompletableFuture<List<EntryDto<?>>> result) {
+        repository.find(normalizedRevision, pathPattern, options).handle(voidFunction((entries, thrown) -> {
+            if (thrown != null) {
+                result.completeExceptionally(thrown);
+            }
```
hyangtack: Do we need “return” here?

minwoox: yes! Thank you!

trustin:
Thanks, @minwoox !

PR #196, opened 2018-04-09
Add a test case that ensures 'from/toRevision' is renamed to 'from/to'

trustin:
.. so that we don't break TTEXT backward compatibility again.

Related: #160 #1


trustin:
Thanks, reviewers!

PR #199, opened 2018-04-10
Allow accessing `metadata.json` to only administrators

### LINKED ISSUE #197: Only administrator should access meta/metadata.json
Currently, project owners can access `meta/metadata.json`. We need to permit that to only administrators. Also we need to prevent `meta` repository from removing by owners.

hyangtack:
Modifications:
- Store `metadata.json` in `dogma` repository, instead of `meta` repository.
- Rename internal repository name of the internal project from `main` to `dogma` for consistency.
- Enhance `MigrationUtil` to support migration from 0.23.0.
- Miscellaneous
  - Invalidate session cache when a user is logged out.

Result:
- Closes #197

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/metadata/MigrationUtil.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/metadata/MigrationUtil.java
@@ -48,6 +48,7 @@ import com.linecorp.centraldogma.server.internal.command.CommandExecutor;
 import com.linecorp.centraldogma.server.internal.storage.project.Project;
 import com.linecorp.centraldogma.server.internal.storage.project.ProjectManager;
 import com.linecorp.centraldogma.server.internal.storage.project.SafeProjectManager;
+import com.linecorp.centraldogma.server.internal.storage.repository.Repository;
 
 /**
  * A utility class for upgrading Central Dogma servers.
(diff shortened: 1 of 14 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/metadata/MigrationUtil.java
```diff
@@ -97,99 +93,150 @@ public static synchronized void migrate(ProjectManager projectManager, CommandEx
...
+
+        Tokens tokens = null;
+        // Check whether "dogma/main/tokens.json" exists or not in case of upgrading from 0.23.0.
+        // The difference between 0.23.0 and now is only the repository name.
+        if (project.repos().exists("main")) {
+            try {
+                tokens = tokensRepo.fetch(INTERNAL_PROJECT_NAME, "main", TOKEN_JSON)
+                                   .thenApply(HolderWithRevision::object)
+                                   .join();
+            } catch (Throwable ignore) {
```
trustin: We should probably peel the exception and swallow only when it's `RepositoryNotFoundException` or `EntryNotFoundException`?

hyangtack: We checked whether the repository exists before fetching. So it might not be raised. :-)

hyangtack:
> How about removing the old files after migration is finished?
Fixed! No rollback after migration! 😄

hyangtack:
Thanks!

PR #202, opened 2018-04-24
Add maxCommits url query to the REST API

minwoox:
Motivation:
A user may not know how many history is in the repository and may have get an error
to get history using relative revision number (e.g, from `-1` to `-10` when there're
only three history).
If the user uses the `maxCommits` url query parameter (e.g `maxCommits=10`  with from `-1` to `1`),
there's no likely chance to fail and he or she is able to get the history no matter how many history
is is.

Modifications:
- Add `maxCommits` url query to the REST API
- Modify Go library to use it

Result:
- Better user experience


trustin:
Thanks, @minwoox !

PR #203, opened 2018-04-25
Do not allow removing meta repository

hyangtack:


trustin:
How about adding a test case?

hyangtack:
Thanks, reviewers!

PR #204, opened 2018-04-25
Do not allow any operations when server is shutting down or a reposit…

trustin:
…ory is removed

Motivation:

When a Repository is closed, then pending operations for the Repository
may fail with an unexpected exception because CommitIdDatabase or jGit
Repository will be closed.

Modifications:

- Make sure a Repository is open before performing any operations.
- Change the signature of the close() methods in the following classes
  so that we can tell why a Repository or a Project has been closed, 1)
  server is shutting down or 2) the Repository or the Project has been
  removed from its RepositoryManager or ProjectManager.
  - GitRepository
  - StorageManager and all its subtypes
- Move the shutdown handling code in WatchService to GitRepository for
  consistency
- Change the shutdown sequence of the server components so that an RPC
  server does not close the connection before sending the SHUTTING_DOWN
  responses.
  - Also replace the nested try-finally blocks with catch(Throwable) for
    simplicity

Result:

- A user is never able to perform an operation on:
  - removed repositories
  - removed projects
  - server that's shutting down

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
@@ -36,13 +36,18 @@ import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.StringJoiner;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executor;
-import java.util.concurrent.locks.Lock;
-import java.util.concurrent.locks.ReentrantLock;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.ReadWriteLock;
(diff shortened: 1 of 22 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
```diff
@@ -1224,15 +1274,21 @@ static void doRefUpdate(org.eclipse.jgit.lib.Repository jGitRepository, RevWalk
         final CompletableFuture<Revision> future = new CompletableFuture<>();
 
         normalize(lastKnownRevision).thenAccept(normLastKnownRevision -> {
```
hyangtack: Not related to your fix. How about using `normalizeNow()` here?

trustin: It's actually more convenient to use `normalize()` here because otherwise we will need additional `try-catch` block here.

PR #207, opened 2018-05-02
Add watcher to Go library

minwoox:
Motivation:
If we provide `watcher` in Go library, a user can utlize Central Dogma more.

Modifications:
- Add `Watcher` to Go library
- Add github.com/sirupsen/logrus/ in order to log by the level

Result:
- Better usage of Go library


client/go/dogma/dogma_util.go
```diff
@@ -0,0 +1,82 @@
...
+			nextDelay = calculatedDelay
+		}
+	}
+	minJitter := int64(float64(nextDelay) * (1 - jitterRate))
+	maxJitter := int64(float64(nextDelay) * (1 + jitterRate))
+	bound := maxJitter - minJitter + 1
+	random := random(bound)
+	result := saturatedAdd(minJitter, random)
+	if result < 0 {
+		return 0
```
trustin: Is this converted to `time.Duration` implicitly?

client/go/dogma/dogma_util.go
```diff
@@ -0,0 +1,82 @@
...
+		}
+	}
+	return result
+}
+
+// This code is from Guava library.
+func saturatedAdd(a, b int64) int64 {
+	naiveSum := a + b
+	if a^b < 0 || a^naiveSum >= 0 {
+		//If a and b have different signs or a has the same sign as the result then there was no overflow.
```
trustin: Space after `//`

client/go/dogma/watch_service.go
```diff
@@ -0,0 +1,378 @@
...
+}
+
+// Close stops watching the file specified in the Query or the pathPattern in the repository.
+func (w *Watcher) Close() {
+	atomic.StoreInt32(&w.state, stopped)
+	select {
+	case latest := <-w.initialChan:
+		w.initialChan <- latest
+	case <-time.After(100 * time.Millisecond):
+		w.initialChan <- &Latest{Err: errors.New("watcher is closed")}
```
trustin: What's the rationale for 100ms timeout? Shouldn't we just make sure no `&Latest` will be sent after `&Latest{Error: ...}` is sent with some concurrency construct?

client/go/dogma/watch_service.go
```diff
@@ -0,0 +1,378 @@
...
+
+	w := newWatcher(projectName, repoName, query.Path)
+	w.doWatchFunc = func(lastKnownRevision int) <-chan *WatchResult {
+		return ws.watchFile(w.watchCTX, projectName, repoName, strconv.Itoa(lastKnownRevision),
+			query, watchTimeout)
+	}
+	w.convertingResultFunc = func(result *WatchResult) *Latest {
+		var value interface{}
+		if convertingValueFunc != nil {
+			value = convertingValueFunc(result.Commit.Entries[0].Content)
```
trustin: What should we do when something goes wrong with this user-provided function?

client/go/dogma/watch_service.go
```diff
@@ -0,0 +1,378 @@
...
+	w := newWatcher(projectName, repoName, pathPattern)
+	w.doWatchFunc = func(lastKnownRevision int) <-chan *WatchResult {
+		return ws.watchRepo(w.watchCTX, projectName, repoName, strconv.Itoa(lastKnownRevision),
+			pathPattern, watchTimeout)
+	}
+	w.convertingResultFunc = func(result *WatchResult) *Latest {
+		var value interface{}
+		revision := result.Commit.Revision
+		if convertingValueFunc != nil {
+			value = convertingValueFunc(revision)
```
trustin: Ditto

client/go/dogma/watch_service.go
```diff
@@ -0,0 +1,378 @@
...
+		// Do not notify after stopped.
+		return
+	}
+
+	latest := w.Latest()
+	w.listenersMutex.Lock()
+	defer w.listenersMutex.Unlock()
+
+	go func() {
+		for _, listener := range w.updateListeners {
```
trustin: Doesn't this goroutine get executed later after `listernsMutext.Unlock()` is called and `notifyListeners()` return? Shouldn't we make a copy of the listeners to make this really thread-safe?

trustin:
We don't have a CLI command for this yet, right?

client/go/dogma/watch_service.go
minwoox: Oops I will make a copy of it. Originally there was no Goroutine that executes the `listeners`.
But I thought that using Goroutine here is better in case something happens while executing `listerens` such as long time job.
So I put it and forgot to make a copy of it. :-)

client/go/dogma/watch_service.go
minwoox: I was thinking about removing this `convertingValueFunc` and passing the `WatchResult` as an argument so that a user can do whatever he or she wants in the `listener`.
Golang does not have generic and I have to use `interface{}` as the type to express the content and it's sort of annoying. WDYT?

trustin: OK

client/go/dogma/cmd/watch_cmd.go
```diff
@@ -0,0 +1,81 @@
...
+				return
+			}
+		}
+
+		fmt.Printf("Content:\n%s\n", content)
+	}
+
+	fw.Watch(listener)
+
+	// TODO(minwoox) implement shutdown gracefully by the input from CLI.
```
trustin: What happens when a user presses CTRL+C?

client/go/dogma/cmd/watch_cmd.go
```diff
@@ -0,0 +1,81 @@
...
+
+func (wc *watchCommand) execute(c *cli.Context) error {
+	repo := wc.repo
+	client, err := newDogmaClient(c, repo.remoteURL)
+	if err != nil {
+		return err
+	}
+
+	query := createQuery(repo.path, wc.jsonPaths)
+	fw, err := client.FileWatcher(repo.projName, repo.repoName, query)
```
trustin: I think we should not use `FileWatcher` here but just the low-level watch API, so that a user can specify `--last-known-rev` from CLI.

minwoox: Because I added a `streaming` option, I left the `FileWatcher` for less codes. But I can change using the low-level watch API when the `streaming` option is not specified.

trustin: SGTM.

client/go/dogma/cmd/cli_cmds.go
```diff
@@ -50,6 +50,11 @@ var maxCommitsFlag = cli.IntFlag{
 	Usage: "Specifies the number of maximum commits to fetch",
 }
 
+var streamingFlag = cli.BoolFlag{
+	Name:  "streaming, s",
+	Usage: "Specifies to keep watching the file",
```
trustin: Specifies whether to keep watching the file

client/go/dogma/watch_service.go
```diff
@@ -0,0 +1,358 @@
...
+	} else {
+		req.Header.Set("if-none-match", "-1")
+	}
+	if timeout != 0 {
+		req.Header.Set("prefer", fmt.Sprintf("wait=%v", timeout.Seconds()))
+	}
+
+	go func() {
+		commit := new(Commit)
+		res, err := ws.client.do(ctx, req, commit)
```
huydx: nits: just thought if we make `client.do` return a channel of struct (which compose repsonse and error) and make it async from inside, so every future use will not need to manually create new routine.

client/go/dogma/watch_service.go
```diff
@@ -0,0 +1,358 @@
...
+		if err != nil {
+			watchResult <- &WatchResult{Commit: nil, Res: res, Err: err}
+		} else {
+			watchResult <- &WatchResult{Commit: commit, Res: res, Err: nil}
+		}
+	}()
+}
+
+const watchTimeout = 1 * time.Minute
```
huydx: user couldn't change timeout value?

client/go/dogma/watch_service.go
```diff
@@ -0,0 +1,358 @@
...
+const (
+	initial int32 = iota
+	started
+	stopped
+)
+
+// Watcher watches the changes of a repository or a file.
+type Watcher struct {
+	state              int32
+	initialChan        chan *Latest // channel whose buffer is 1.
```
huydx: initialValueCh instead? (I was thought initialChan is a channel for state management)

client/go/dogma/watch_service.go
```diff
@@ -0,0 +1,358 @@
...
+			w.scheduleWatch(numAttemptsSoFar + 1)
+			return
+		}
+
+		newLatest := w.convertingResultFunc(watchResult)
+		if w.isInitialChanSet == 0 && atomic.CompareAndSwapInt32(&w.isInitialChanSet, 0, 1) {
+			// The initial latest is set for the first time. So write the value to initialChan as well.
+			w.initialChan <- newLatest
+		}
+		atomic.StorePointer(&w.latestPtr, (unsafe.Pointer)(newLatest))
```
huydx: I'm not sure but could we use `atomic.Value.store/load` instead?

client/go/dogma/watch_service.go
```diff
@@ -0,0 +1,358 @@
...
+	started
+	stopped
+)
+
+// Watcher watches the changes of a repository or a file.
+type Watcher struct {
+	state              int32
+	initialChan        chan *Latest // channel whose buffer is 1.
+	isInitialChanSet   int32        // 0 is false, 1 is true
+	scheduleCTX        context.Context
```
huydx: just question, if `scheduleCTX` and `watchCTX` mean to be called in `Close()` only, why do we need a lot of contexts? What will happen if we have single `watcherCTX` for a single watcher and handle everything (closed state, close signal dispatcher...) through that?

client/go/dogma/watch_service.go
minwoox: I didn't make it because the Java version does not have the functionality. Do you think we need it? Then I think I have to change the Java version as well. :)

client/go/dogma/watch_service.go
minwoox: Fixed. Thank you~!

client/go/dogma/watch_service.go
minwoox: I didn't know that I can use in that way. Thank you~!

client/go/dogma/watch_service.go
minwoox: Then the other APIs have to use the channel as return value which I think overkill. And I heard that netpoller does the asynchronous job inside of it so we don't have to use channel for the other APIs in dogma.go.

client/go/dogma/watch_service.go
minwoox: You are right~! I don't have to use two different context here. Thank you~!

minwoox:
@huydx All fixed and answered. Plz take a look~! :)

PR #209, opened 2018-05-09
Use Highlight.js instead of Pygments for syntax highlighting

trustin:
Motivation:

I can't find a nice theme for Pygments.

Modifications:

- Use Highlight.js instead of Pygments for syntax highlighting
- Increase the font size of code snippets

Result:

- Code snippets look better.


trustin:
Thanks!

PR #211, opened 2018-05-09
Compress API responses if requested

trustin:
Motivation:

An API response can be as big as the size of the entries stored in
Central Dogma. A user may want to reduce the traffic by sending his or
her request with a 'accept-encoding: deflate' (or gzip) header.

Modifications:

- Apply HttpEncodingService to Thrift and HTTP API endpoints

Result:

A user can choose to get a compressed response.

### DIFF
--- a/server/src/test/java/com/linecorp/centraldogma/server/ContentCompressionTest.java
+++ b/server/src/test/java/com/linecorp/centraldogma/server/ContentCompressionTest.java
@@ -1,0 +1,130 @@
+/*
+ * Copyright 2018 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


server/src/test/java/com/linecorp/centraldogma/server/ContentCompressionTest.java
```diff
@@ -0,0 +1,130 @@
...
+    public void http() throws Exception {
+        final HttpClient client = new HttpClientBuilder(
+                "http://127.0.0.1:" + rule.serverAddress().getPort())
+                .setHttpHeader(HttpHeaderNames.AUTHORIZATION, "bearer " + CsrfToken.ANONYMOUS)
+                .setHttpHeader(HttpHeaderNames.ACCEPT_ENCODING, "deflate")
+                .build();
+
+        final String contentPath = HttpApiV1Constants.PROJECTS_PREFIX + '/' + PROJ +
+                                   HttpApiV1Constants.REPOS + '/' + REPO +
+                                   "/contents" + PATH;
```
minwoox: You can use `HttpApiV1Constants.CONTENTS` if you want.

trustin:
Thanks for reviewing.

PR #214, opened 2018-05-11
Cache the fast-path check of Repository.watch()

trustin:
Motivation:

A client that watches a file usually starts watching a file with:

  Repository.watch(Revision.INIT, pathPattern);

Repository will then traverse the diff between HEAD and INIT to check if
there are any entries that match the specified `pathPattern`. This can
be a time-consuming task and thus could be cached.

Modifications:

- Add Repository.findLatestRevision(lastKnownRevision, pathPattern)
  which returns the latest known revision that contains the changes
  matched by `pathPattern` since `lastKnownRevision`.
- Modify GitRepository.watch(lastKnownRevision, pathPattern) to use
  findLatestRevision()
- Add caching for findLatestRevision() to CachingRepository
- Modify CachingRepository.watch(lastKnownRevision, pathPattern) to
  leverage the cached findLatestRevision()

Result:

- Faster watch() when clients send the same watch() calls


server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/Repository.java
```diff
@@ -390,6 +390,16 @@ default Author author() {
      */
     CompletableFuture<List<Commit>> history(Revision from, Revision to, String pathPattern, int maxCommits);
 
+    /**
+     * Returns whether there's any commit that contains an {@link Entry} matched by the specified
```
minwoox: How about `Returns the latest {@link Revision} if there's`?

PR #216, opened 2018-05-14
Fix watcher leaks

trustin:
Related: e0fcb62507679ce85f9b2eb7f71e6911407d9e15

Motivation:

CachingRepository.watch(lastKnownRevision, pathPattern) leaks watchers
because it does not cancel the watcher when a user cancels the future
returned by CachingRepository.

Modification:

- Propagate the state between the two involved futures so that its
  cancellation request does not leak the futures.
- Reuse the future returned by findLatestRevision() if possible.

Result:

Leak is gone


trustin:
Thanks, @hyangtack !

PR #217, opened 2018-05-14
Add jitters to mirroring jobs

### LINKED ISSUE #215: Add jitters to mirroring schedules
We obviously do not want the server to get busy suddenly because everyone scheduled their Git-to-CD mirrors to run every minute.

trustin:
Motivation:

Most users schedule their mirroring jobs to run every minute.
MirroringService runs those jobs when 'second' is 0, we see a spike in
system activity every minute.

Modifications:

- Calculate a jitter beteen -1 and 1 minute for each Mirror so that the
  system load is spread.
  - Add Mirror.nextExecutionTime()
  - Make Mirror.schedule() visible only for testing
- Miscellaneous:
  - Rename lastRunTime to lastExecutionTime
  - Increase the keep-alive time of mirroring threads from 60 to 90
    seconds, so that they do not die too often even if there is only one
    mirroring job that runs every minute.

Result:

- Less number of load spikes
- Fixes #215

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirroringService.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirroringService.java
@@ -33,7 +33,6 @@ import java.util.stream.Stream;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.cronutils.model.time.ExecutionTime;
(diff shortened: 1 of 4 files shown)


trustin:
Now ready for a review.

trustin:
Will fix the build failure due to mocking issues tomorrow. Please review anyway.

trustin:
Thanks, reviewers!

PR #218, opened 2018-05-15
Update metadata when creating/removing a project/repository by Thrift…

### LINKED ISSUE #210: Update metadata when Thrift APIs add/remove a project/repository
There's a bug in Central Dogma's Thrift-side code where permission-related metadata is not populated when a Thrift client creates a repository/project.

hyangtack:
… API

Motivation:
There's a bug in Central Dogma's Thrift-side code where permission-related metadata is not populated when a Thrift client creates a repository/project.

Modifications:
- Update metadata for the following RPC methods:
  - removeProject
  - unremoveProject
  - createRepository
  - removeRepository
  - unremoveRepository

Result:
- Closes #210

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/thrift/CentralDogmaServiceImpl.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/thrift/CentralDogmaServiceImpl.java
@@ -98,6 +98,17 @@ public class CentralDogmaServiceImpl implements CentralDogmaService.AsyncIface {
         }
     }
 
+    private static void handleAsVoidResult(CompletableFuture<?> future, AsyncMethodCallback resultHandler) {
+        future.handle((res, cause) -> {
+            if (cause != null) {
+                resultHandler.onError(convert(cause));
+            } else {
+                resultHandler.onComplete(null);
+            }
+            return null;
+        });
+    }
+
     @Override
     public void createProject(String name, AsyncMethodCallback resultHandler) {
         // ProjectInitializingCommandExecutor initializes a metadata for the specified project.
@@ -106,6 +117,7 @@ public class CentralDogmaServiceImpl implements CentralDogmaService.AsyncIface {
 
     @Override
     public void removeProject(String name, AsyncMethodCallback resultHandler) {
+        // Metadata must be updated first because it cannot be updated if the project is removed.
         handle(mds.removeProject(SYSTEM, name)
                   .thenCompose(unused -> executor.execute(Command.removeProject(SYSTEM, name))),
                resultHandler);
@@ -113,10 +125,10 @@ public class CentralDogmaServiceImpl implements CentralDogmaService.AsyncIface {
 
     @Override
     public void unremoveProject(String name, AsyncMethodCallback resultHandler) {
-        handle(executor.execute(Command.unremoveProject(SYSTEM, name))
-                       .thenCompose(unused -> mds.restoreProject(SYSTEM, name))
-                       .thenApply(revision -> null),
-               resultHandler);
+        // Restore the project first then update its metadata as 'active'.
+        handleAsVoidResult(executor.execute(Command.unremoveProject(SYSTEM, name))
+                                   .thenCompose(unused -> mds.restoreProject(SYSTEM, name)),
+                           resultHandler);
     }
 
     @Override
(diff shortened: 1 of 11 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/thrift/CentralDogmaServiceImpl.java
```diff
@@ -121,22 +136,34 @@ public void listRemovedProjects(AsyncMethodCallback resultHandler) {
     }
 
     @Override
-    public void createRepository(
-            String projectName, String repositoryName, AsyncMethodCallback resultHandler) {
-        handle(executor.execute(Command.createRepository(SYSTEM, projectName, repositoryName)),
+    public void createRepository(String projectName, String repositoryName,
+                                 AsyncMethodCallback resultHandler) {
+        handle(mds.addRepo(SYSTEM, projectName, repositoryName)
+                  .thenCompose(unused -> executor.execute(
```
trustin: We call `mds` before `CommandExecutor` for `createRepository` and `removeProject` but not for others (`unremoveProject`, `unremoveRepository` and `removeRepository`). Do we need some sort of consitency in call order?

server/src/test/java/com/linecorp/centraldogma/server/internal/metadata/ThriftBackwardCompatibilityTest.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    @Test
+    public void shouldBackwardCompatible() throws Exception {
+        final String repo1 = "repo1";
+
+        dogma.client().createProject(projectName).join();
+        dogma.client().createRepository(projectName, repo1).join();
+        dogma.client().push(projectName, repo1, Revision.HEAD, "",
+                            Change.ofTextUpsert("/sample.txt", "test")).join();
```
trustin: Could we use raw Thrift client instead of `dogma.client()`, like we did in https://github.com/line/centraldogma/blob/57a45d824e83d41673b4ca27d6c62e7045688f97/server/src/test/java/com/linecorp/centraldogma/server/ContentCompressionTest.java#L81 ? We may switch to an HTTP client later.

server/src/main/java/com/linecorp/centraldogma/server/internal/thrift/CentralDogmaServiceImpl.java
hyangtack: I made their call order the same as the call order in HTTP v1 API services. When I applied metadata to HTTP v1 API services first, I thought the call order had to be different for each case. But now I'm thinking that calling `CommandExecutor` before updating metadata is reasonable for all cases. Let me fix them.

server/src/test/java/com/linecorp/centraldogma/server/internal/metadata/ThriftBackwardCompatibilityTest.java
hyangtack: Got it. :-)

server/src/main/java/com/linecorp/centraldogma/server/internal/thrift/CentralDogmaServiceImpl.java
hyangtack: For `removeProject`, we should update metadata first, because we cannot update metadata if the project is removed.

server/src/test/java/com/linecorp/centraldogma/server/internal/metadata/ThriftBackwardCompatibilityTest.java
```diff
@@ -0,0 +1,119 @@
...
+    private static final Revision head = new Revision(-1, 0);
+    private static final Author author = new Author(SYSTEM.name(), SYSTEM.email());
+
+    private static final String projectName = "foo";
+
+    @BeforeClass
+    public static void init() {
+        final InetSocketAddress serverAddress = dogma.dogma().activePort().get().localAddress();
+        httpClient = new HttpClientBuilder("http://" + serverAddress.getHostString() +
+                                           ':' + serverAddress.getPort())
```
trustin: How about just using `127.0.0.1` just in case `/etc/hosts` is misconfigured?

hyangtack:
Thanks, reviewers!

PR #222, opened 2018-05-16
Apply jitter to watch timeout

trustin:
Related: #217

Motivation:

A Central Dogma server can get a large number of watch requests if
clients use the same watch timeout value, a.k.a thundering herd problem.

Modifications:

- Apply 20% jitter to all watch timeouts
- Miscellaneous
  - Reject zero or negative timeout values explicitly

Result:

Less spikes in server load

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/api/WatchService.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/api/WatchService.java
@@ -22,6 +22,7 @@ import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.TimeUnit;
 
 import com.linecorp.armeria.common.RequestContext;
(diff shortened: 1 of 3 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/api/WatchService.java
```diff
@@ -61,11 +64,29 @@
         return result;
     }
 
+    /**
+     * Awaits and retrieves the latest revision of the commit that changed the file that matches the specified
+     * {@link Query} since the specified {@code lastKnownRevision}. This will wait until the specified
+     * {@code timeoutMillis} passes. If there's no change during the time, the returned future will be
+     * exceptionally completed with the {@link CancellationException}.
+     */
+    public <T> CompletableFuture<Entry<T>> watchFile(Repository repo, Revision lastKnownRevision,
```
trustin: This was just moved up.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/converter/WatchRequestConverter.java
```diff
@@ -63,16 +63,25 @@ public Object convertRequest(ServiceRequestContext ctx, AggregatedHttpMessage re
...
+            return rejectPreferHeader(preferHeader);
         }
+
+        if (timeoutSeconds <= 0) {
+            return rejectPreferHeader(preferHeader);
+        }
+        return TimeUnit.SECONDS.toMillis(timeoutSeconds);
+    }
+
+    private static long rejectPreferHeader(String preferHeader) {
```
minwoox: private static void?

minwoox: Ah! didn't think about that.

trustin:
Thanks for reviewing.

PR #224, opened 2018-05-18
Refactor the enums stringer

RuniVN:
Motivation:
This doesn't look good.
```
type ChangeType int

const (
	UpsertJSON ChangeType = iota + 1
	UpsertText
	Remove
	Rename
	ApplyJSONPatch
	ApplyTextPatch
)

var changeTypeMap = map[string]ChangeType{
	"UPSERT_JSON":      UpsertJSON,
	"UPSERT_TEXT":      UpsertText,
	"REMOVE":           Remove,
	"RENAME":           Rename,
	"APPLY_JSON_PATCH": ApplyJSONPatch,
	"APPLY_TEXT_PATCH": ApplyTextPatch,
}

var changeTypeValues = [...]string{
	"UNKNOWN",
	"UPSERT_JSON",
	"UPSERT_TEXT",
	"REMOVE",
	"RENAME",
	"APPLY_JSON_PATCH",
	"APPLY_TEXT_PATCH",
}
```
- We have to maintain 2 ChangeType lookup data structures that were being coupled together. If we add one more value for `changeTypeMap`, `changeTypeValues` must be update, as follows.
- Similar to EntryType

Modifications:
- Add method `String()` for enums `ChangeType` and `EntryType`. 
- Simply call `type.String()` instead looking up from `changeTypeValues[type]`.

Result:

- Better look
- Easier to extend

### DIFF
--- a/client/go/dogma/content_service.go
+++ b/client/go/dogma/content_service.go
@@ -60,7 +60,7 @@ func (c *Entry) MarshalJSON() ([]byte, error) {
 		Type string `json:"type"`
 		*Alias
 	}{
-		Type:  entryTypeValues[c.Type],
+		Type:  c.Type.String(),
 		Alias: (*Alias)(c),
 	})
 }
@@ -110,7 +110,7 @@ func (c *Change) MarshalJSON() ([]byte, error) {
 		Type string `json:"type"`
 		*Alias
 	}{
-		Type:  changeTypeValues[c.Type],
+		Type:  c.Type.String(),
 		Alias: (*Alias)(c),
 	})
 }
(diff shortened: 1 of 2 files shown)


hyangtack:
@RuniVN I'm not familiar with Go, so I'm not sure if this is good or not.. How about using bidirectional map? Is there any good implementation in Go?

minwoox:
@RuniVN Thanks a lot for your first contribution! 😆
It makes the code look much better.
By the way, could you sign the ICLA please? https://github.com/line/centraldogma/blob/master/CONTRIBUTING.md#contributor-license-agreement

RuniVN:
@hyangtack Actually no. Go prefers using 2 maps or reverse a map to get the reversal instead bimap.

There are some Go bimap implementations in github but kind of awkward.

RuniVN:
@minwoox I've signed the ICLA.

trustin:
Please do not merge yet - I'm checking internally if ICLA has been signed. (Having a permission issue browsing ICLA signature list.)

trustin:
Confirmed the ICLA. Thanks for your patience, @RuniVN !

PR #228, opened 2018-05-23
Automatic client-side load balancing using DNS lookup

trustin:
Motivation:

Distribution of server addresses via `.properties` file is not enough
because it's static and a user has to upgrade manually when the cluster
configuration changes. This can be disatrous when a replica in a cluster
crashes in an irrecoverable way.

Motivations:

- Use Armeria's DnsAddressEndpointGroup to dynamically query the DNS
  servers and update the endpoint list.
  - An administrator can simply update his or her DNS A/AAAA records to
    change the cluster members.
- Use HttpHealthCheckedEndpointGroup so that the requests are not sent
  to an unhealthy replica
- Extract health check service path to `HEALTH_CHECK_PATH`

Result:

- Less chance of client-side errors when a replica goes down.
- A lot more flexible addition and removal of a cluster member.

### DIFF
--- a/client/go/dogma/watch_service.go
+++ b/client/go/dogma/watch_service.go
@@ -139,8 +139,8 @@ func newWatcher(projectName, repoName, pathPattern string) *Watcher {
 	return &Watcher{state: initial, initialValueCh: make(chan *Latest, 1),
 		watchCTX: watchCTX, watchCancelFunc: watchCancelFunc,
 		listenersMutex: &sync.Mutex{},
-		projectName: projectName,
-		repoName: repoName, pathPattern: pathPattern}
+		projectName:    projectName,
+		repoName:       repoName, pathPattern: pathPattern}
 }
 
 // AwaitInitialValue awaits for the initial value to be available.
--- a/client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/AbstractArmeriaCentralDogmaBuilder.java
+++ b/client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/AbstractArmeriaCentralDogmaBuilder.java
@@ -19,7 +19,11 @@ import static com.google.common.base.Preconditions.checkState;
 import static java.util.Objects.requireNonNull;
 
 import java.net.InetSocketAddress;
+import java.net.UnknownHostException;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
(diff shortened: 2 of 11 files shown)


client/go/dogma/watch_service.go
```diff
@@ -139,8 +139,8 @@ func newWatcher(projectName, repoName, pathPattern string) *Watcher {
 	return &Watcher{state: initial, initialValueCh: make(chan *Latest, 1),
 		watchCTX: watchCTX, watchCancelFunc: watchCancelFunc,
 		listenersMutex: &sync.Mutex{},
-		projectName: projectName,
-		repoName: repoName, pathPattern: pathPattern}
+		projectName:    projectName,
+		repoName:       repoName, pathPattern: pathPattern}
```
trustin: Auto-formatted by gofmt during the build process

client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/AbstractArmeriaCentralDogmaBuilder.java
```diff
@@ -101,24 +119,58 @@ protected final Endpoint endpoint() {
...
+                } catch (Exception e) {
+                    final UnknownHostException cause = new UnknownHostException(
+                            "failed to resolve any of: " + hosts);
+                    cause.initCause(e);
+                    throw cause;
+                }
+            }
+
+            if (healthCheckEnabled) {
+                group = new HttpHealthCheckedEndpointGroupBuilder(group, "/monitor/l7check")
```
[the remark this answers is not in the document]
trustin: Added `HttpApiV1Constants.HEALTH_CHECK_PATH`.

client/go/dogma/watch_service.go
minwoox: Thank you!

client/java-armeria/src/test/resources/centraldogma-profile-alice.properties
```diff
@@ -14,5 +14,5 @@
 # under the License.
 #
 
-centraldogma.hosts.0=alice.com
-centraldogma.hosts.1=bob.com:8080
+centraldogma.hosts.0=1.2.3.4.xip.io
```
minwoox: Don't we need to change the name of this file? :)

trustin: Sure, why not? :-D

PR #231, opened 2018-05-25
Fix a bug where 36462 port is appened when the scheme is HTTPS

minwoox:


minwoox:
> How about changing "http" to defaultScheme here?

I used `http` intentionally because it is checking if the scheme is `http` or `https`.
If I use `!strings.HasPrefix(baseURL, defaultScheme)`, I thought it would be confusing because someone could feel if it's only checking `http`.

minwoox:
Thanks reviewers~!

PR #232, opened 2018-05-28
Allow using different ports for cleartext and TLS when using a client…

trustin:
… profile

Motivation:

It is currently impossible to specify different ports for cleartext and
TLS connections when using a client profile.

Modifications:

Define two new property keys - `centraldogma.httpPorts.*` and
`centraldogma.httpsPorts.*` - so that a client profile can specify
different port numbers.

Result:

Flexibility


trustin:
Will send a reworked pull request.

PR #233, opened 2018-05-29
Allow using different ports for cleartext and TLS when using a client profile

trustin:
Motivation:

It is currently impossible to specify different ports for cleartext and
TLS connections when using a client profile.

Modifications:

- Replace the `.properties` format with a new JSON format which is more
  expressive and thus can specify different ports for different protocols

Result:

- Flexibility
- Old `.properties` files will not work anymore.


client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogmaBuilder.java
```diff
@@ -214,53 +224,65 @@ public final B profile(ClassLoader classLoader, Iterable<String> profiles) {
...
-                }
-            } catch (IOException e) {
-                throw new IllegalArgumentException("failed to load: " + path, e);
+            final ImmutableSet.Builder<InetSocketAddress> newHostsBuilder = ImmutableSet.builder();
+            candidate.hosts().stream()
+                     .filter(e -> (useTls ? "https" : "http").equals(e.protocol()))
+                     .forEach(e -> newHostsBuilder.add(newEndpoint(e.host(), e.port())));
+
+            final ImmutableSet<InetSocketAddress> newHosts = newHostsBuilder.build();
+            if (!newHosts.isEmpty()) {
```
hyangtack: How about adding a warning log message like what we did before when `newHosts` is empty?

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogmaBuilder.java
```diff
@@ -214,53 +224,65 @@ public final B profile(ClassLoader classLoader, Iterable<String> profiles) {
         checkState(selectedProfile == null, "profile cannot be loaded more than once.");
         checkState(hosts.isEmpty(), "profile() and host() cannot be used together.");
 
-        String selectedProfile = null;
-        ImmutableSet<InetSocketAddress> hosts = null;
-        boolean profilesIsEmpty = true;
-        for (String p : profiles) {
-            checkNotNull(p, "profiles contains null: %s", profiles);
-            profilesIsEmpty = false;
+        final URL resourceUrl = findFirstResource(
```
hyangtack: Just curious. What if we don't use `TEST_PROFILE_RESOURCE_PATH`? Or, how about adding a new method to this builder which sets `profileResourcePath` so that a user can specify his or her resource?

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogmaBuilder.java
trustin: I'm not sure if it's a good idea because it can be triggered even when `hosts` is not empty but there's no matching protocol.

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogmaBuilder.java
trustin: If `centraldogma-profiles-test.json` is missing, it will automatically fall back into `centraldogma-profiles.json`, just like Logback does for `logback-test.xml` and `logback.xml`. I believe `-test.*` has precedence so that a user puts `-test.*` into `src/test/resources` so that it does not pick up `src/main/resources/`.

I'm curious if we need an option to specify the `InputStream`, `URL` or `File` along with `profile()`. Do we?

trustin: Added `profileResources(String... paths)` to the builder class.

trustin:
Will update the Javadoc tomorrow.

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogmaBuilder.java
hyangtack: I can see now what you intend. :-) I was just curious what if we provide a way to specify user resource path instead of using a fixed path of `centraldogma-profiles.json` and `centraldogma-profiles-test.json`. But your way is reasonable to me, so please forget my previous comment. Thanks!

client/java/src/main/java/com/linecorp/centraldogma/client/ClientProfile.java
```diff
@@ -0,0 +1,110 @@
...
+        @JsonCreator
+        Entry(@JsonProperty(value = "host", required = true) String host,
+              @JsonProperty(value = "protocol", required = true) String protocol,
+              @JsonProperty(value = "port", required = true) Integer port) {
+            this.host = requireNonNull(host, "host");
+            checkArgument(!host.isEmpty(), "hostname is empty.");
+            this.protocol = requireNonNull(protocol, "protocol");
+            checkArgument(!protocol.isEmpty(), "protocol is empty.");
+            this.port = requireNonNull(port, "port");
+            checkArgument(port > 0 && port < 65536, "port: %s (expected: 0..65535)", port);
```
minwoox: 1..65535?

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogmaBuilder.java
hyangtack: I misread the previous code. Thanks! :-)

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogmaBuilder.java
```diff
@@ -128,18 +141,38 @@ protected final boolean isUseTls() {
...
+
+    /**
+     * Sets the paths to look for to read the {@code .json} file that contains the client profiles.
+     * The paths are tried in the order of iteration. The default value of this property is
+     * <code>[ {@value #TEST_PROFILE_RESOURCE_PATH}, {@value #PROFILE_RESOURCE_PATH} ]</code>, which means
+     * the builder will check if {@value #TEST_PROFILE_RESOURCE_PATH} exists first and will try
+     * {@value #PROFILE_RESOURCE_PATH} only if {@value #TEST_PROFILE_RESOURCE_PATH} is missing.
+     */
+    public final B profileResources(Iterable<String> paths) {
+        final List<String> newPaths = ImmutableList.copyOf(requireNonNull(paths, "paths"));
```
hyangtack: nit: How about checking `selectedProfile == null` and `hosts.isEmpty()`? (or resetting them because profile resources are to be changed here?)

trustin: Good point. Added a check for `selectedProfile == null`. I didn't check if `hosts.isEmpty()` though because a user is allowed add an individual host using `host()` at any time.

client/java/src/main/java/com/linecorp/centraldogma/client/ClientProfile.java
trustin: Oops. Fixed :smile:

PR #239, opened 2018-06-04
Fix flaky tests in ArmeriaCentralDogmaBuilderTest

trustin:
.. which often occurs due to domain name resolution timing.


trustin:
Thanks.

PR #240, opened 2018-06-05
Fix a bug where ContentServiceV1.getFiles() does not apply JSON path query for watch responses

trustin:
Motivation:

When a user attempts to watch a file with a JSON path query using the
HTTP API, the entry content of the watch result should be the JSON path
query result rather than the full content.

Modifications:

- Split ContentServiceV1.handleWatchSuccess() into handleWatchFileSuccess()
  and handleWatchRepositorySuccess() so that they are handles differently.
- Add a test case for watching a file with a JSON path query

Result:

- Correct response


trustin:
Thanks, @hyangtack !

minwoox:
LGTM 👍

PR #242, opened 2018-06-07
Reduce the amount of information returned by HTTP API

### LINKED ISSUE #241: Reduce the amount of information returned by HTTP API
TL;DR - We are returning more information than necessary for some operations.

### Pushing a commit

I find `author`, `commitMessage` and `entries` are redundant information because `commitMessage` and `entries` are given by the client and the client should know who it is.

```json
{
  "revision": ...,
  "author": ...,
  "pushedAt": ...,
  "commitMessage": ...,
  "entries": ...
}
```

Ideally, the response could be:

```json
{
  "revision": ...,
  "pushedAt": ...
}
```

### Getting the commit list

`entries` is unnecessary. A user should send an additional query to retrieve the snapshot of a revision or the diff between two revisions based on what he or she needs.

```json
[
    {
        "revision": ...,
        "author": ...,
        "pushedAt": ...,
        "commitMessage": ...,
        "entries": [ ... ]
    },
    ...
]
```

### Watching a file

`entries` always contains a single entry. Perhaps we should just replace `entries` with `entry`?

```json
{
   "revision": ...,
   "author": ...,
   "pushedAt": ...,
   "commitMessage":, ...
   "entries": [ ... ]
}
```

### Watching a repository

`entries` could be removed, given that a user knows what files will be in the list given that he or she specified `pathPattern` parameter? Also, `entries` can increase the amount of the traffic.

trustin:
Motivation:

- When pusing a commit, I find `author`, `commitMessage` and `entries`
  are redundant information because `commitMessage` and `entries` are
  given by the client and the client should have that information
  already.
- When getting the commit list, `entries` is unnecessary. A user should
  send an additional query to retrieve the snapshot of a revision or the
  diff between two revisions based on what he or she needs.
- When watching a file, `entries` always contains a single entry.
  Perhaps we should just replace `entries` with `entry`?
- When watching a repository, `entries` could be removed, given that a
  user knows what files will be in the list given that he or she
  specified `pathPattern` parameter? Also, `entries` can increase the
  amount of the traffic.

Modifications:

- Change `CentralDogma.getHistory()` return `List<Commit>` rather than
  `List<CommitAndChanges<?>>`
  - Remove `CommitAndChanges` and `CommitAndChangeConverter`
- Change `CentralDogma.push()` return `PushResult` rather than `Commit`
  - Add `PushResult` and `PushResultConverter`
  - Add `PushResultDto`
- Split `CommitDto` into `CommitDto` and `WatchResultDto` so that
  `CommitDto` does not contain any entries

Result:

- Fixes #241


server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
```diff
@@ -302,7 +297,7 @@ private static String joinPaths(Iterable<Change<?>> changes) {
                          .thenCombine(historyFuture, (entryMap, commits) -> {
                              final ImmutableList<EntryDto<?>> entryDtos = entryDtos(repository, entryMap);
```
minwoox: We can remove this line :)

minwoox: And we don't have to do `repository.find()` here because we don't need it :)

trustin: Oops! Fixed

trustin:
Thanks for reviewing :-)

PR #243, opened 2018-06-07
Fix a bug where jumping to a revision in file list view did not work

trustin:
Modifications:

- Ensure `?revision=<target_revision>` is passed to the GET request sent
  by `getTree()`


trustin:
Thanks for reviewing.

PR #244, opened 2018-06-07
Change Go client library corrsponding to the changes in HTTP API

minwoox:
Motivation:
This is related to #242.

Modification:
- Add `PushResult` struct
- Remove `entries` in `Commit` struct
- Add `entry` into `WatchResult` struct

Result:
- Compatible with #242


minwoox:
Thanks for reviewing!

PR #246, opened 2018-06-08
Provide error information in Go client

minwoox:
Motivation:
If we provide error information to the client, it will be helpful to debug.

Modification:
- Provide an error message if it exists

Result:
- Better for debuging


trustin:
Thanks, @minwoox !

PR #248, opened 2018-06-12
Consistent HTTP API error response

trustin:
Motivation:

The current HTTP API error response:

- has fewer information than our Java client requires, such as:
  - whether a commit has failed because the changeset is redundant or
    the changeset really has a conflict
- sometimes returns a JSON document with different fields, e.g.
  TokenService

Modifications:

- Revamp the error response factory methods in `HttpApiUtil`
  - Include the exception type if possible
  - Always set 'message' field.
- Use the error response factory methods in `HttpApiUtil` wherever
  possible

Result:

- Consistency
- Clients have more information about why the operation has failed,
  which is sometimes not enough with an HTTP status code.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/api/AdministrativeService.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/api/AdministrativeService.java
@@ -86,7 +86,8 @@ public final class AdministrativeService extends AbstractService {
         if (writableNode.asBoolean()) {
             if (!oldStatus.writable) {
                 // TODO(trustin): Implement exiting read-only mode.
-                throw HttpStatusException.of(HttpStatus.NOT_IMPLEMENTED);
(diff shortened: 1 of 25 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/api/AdministrativeService.java
```diff
@@ -86,7 +86,8 @@ public ServerStatus updateStatus(@RequestObject JsonNode patch) throws Exception
         if (writableNode.asBoolean()) {
             if (!oldStatus.writable) {
                 // TODO(trustin): Implement exiting read-only mode.
-                throw HttpStatusException.of(HttpStatus.NOT_IMPLEMENTED);
+                return HttpApiUtil.throwResponse(HttpStatus.NOT_IMPLEMENTED,
+                                                 "Please consider sending a pull request. ;-)");
```
hyangtack: 😆

PR #251, opened 2018-06-25
Ignore non-file objects such as a submodule while mirroring a Git rep…

trustin:
…ository

Motivation:

The mirroring service fails with a `Missing unknown <sha1>` error when a
Git repository being mirrored contains a submodule.

Modification:

- Skip an object that's neither a file nor a tree while walking a Git
  tree
- Miscellaneous:
  - Log the detailed information about the mirror on an unexpected
    failure

Result:

- Git mirroring does not fail on a repository with a submodule.


it/src/test/java/com/linecorp/centraldogma/it/mirror/git/GitMirrorTest.java
```diff
@@ -292,6 +292,38 @@ public void remoteToLocal_merge() throws Exception {
                 .containsExactlyInAnyOrder(expectedMirrorState, expectedAlphabets);
     }
 
+    @Test
+    public void remoteToLocal_submodule() throws Exception {
+        pushMirrorSettings(null, null);
+
+        // Create a new repository for a submodule.
+        final File gitSubmoduleWorkTree = new File(gitRepoDir.getRoot(), testName.getMethodName() + ".submodule").getAbsoluteFile();
```
minwoox: nit: this line is too long?

trustin: Thanks. Fixed :-)

PR #255, opened 2018-07-09
Make log messages related with replication failures more detailed

trustin:
Motivation:

We internally had some consistency issues that made our replicas enter
read-only mode and we wished we had more information in our log message
to pinpoint which repository had issues, because we have lots of
repositories.

Modifications:

- Make the replication error message more detailed
- Add more replication error log messages
- Make a replica enter read-only mode for more failure cases
- Make the message of RedundantChangeException more detailed

Result:

- Hopefully more information on a next inconsistency
- Fail fast and enter read-only mode for more failure scenarios


trustin:
Thanks for reviewing :-)

PR #257, opened 2018-07-12
Add 'ID' field to credentials

trustin:
Motivations:

Sometimes, a user has to specify different credentials for different
repositories that reside in the same host. In this case, we cannot
select the credentials based on hostname patterns, because their
hostnames are same.

Modifications:

- Add `id` field to `credentials.json`
- Add `credentialId` and `defaultCredentialId` field to `mirrors.json`
- Update the documentation

Result:

Can specify different credentials for different repositories that reside
in the same host

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/DefaultMetaRepository.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/DefaultMetaRepository.java
@@ -24,11 +24,14 @@ import java.net.URI;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.regex.Matcher;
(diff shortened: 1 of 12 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/DefaultMetaRepository.java
```diff
@@ -180,16 +183,26 @@ public DefaultMetaRepository(Repository repo) {
...
-                if (c.matches(remoteUri)) {
-                    credential = c;
-                    break;
+        static MirrorCredential findCredential(Iterable<MirrorCredential> credentials, URI remoteUri,
+                                               @Nullable String credentialId) {
+            if (credentialId != null) {
+                // Find by credential ID.
+                for (MirrorCredential c : credentials) {
+                    final Optional<String> id = c.id();
+                    if (id.isPresent() && credentialId.equals(id.get())) {
```
trustin: This is obviously a sequential search but I guess it's fine for usual use cases. Will optimize later if really necessary.

trustin:
Thanks for.reviewing :-)

PR #261, opened 2018-07-17
Do not use framed version of Javadoc

trustin:
Motivation:

Since Java 9, Javadoc defaults to non-framed version. We should be
consistent across our web site for hyperlinks to Javadoc.

Modifications:

- Use hyperlinks to non-framed versions of Javadoc

Result:

Consistency


trustin:
Thanks, @minwoox :-)

Issue #268, opened 2018-08-07
tail command fails with 'unrecognized file system type' error in Docker image


trustin:
See: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=810669

PR #271, opened 2018-08-20
Change base image to 'openjdk:8'

hyangtack:
Modifications:
- Do not use 'java:openjdk-8-jre' anymore which is deprecated
- Copy missing binaries to the docker image

### DIFF
--- a/dist/Dockerfile
+++ b/dist/Dockerfile
@@ -1,29 +1,30 @@
 #
 # CentralDogma Dockerfile
 #
-FROM java:openjdk-8-jre
+FROM openjdk:8
 
 # Environment variables.
 ENV CENTRALDOGMA_HOME "/opt/centraldogma"
 ENV CENTRALDOGMA_OPTS "-nodetach"
-ENV JAVA_HOME /usr/lib/jvm/java-8-openjdk-amd64/
 ENV JAVA_OPTS "$CENTRALDOGMA_OPTS"
 
 # Install CentralDogma binaries and configurations.
-RUN mkdir -p                    \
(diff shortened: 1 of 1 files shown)


dist/Dockerfile
```diff
@@ -1,29 +1,30 @@
 #
 # CentralDogma Dockerfile
 #
-FROM java:openjdk-8-jre
+FROM openjdk:8
 
 # Environment variables.
 ENV CENTRALDOGMA_HOME "/opt/centraldogma"
 ENV CENTRALDOGMA_OPTS "-nodetach"
-ENV JAVA_HOME /usr/lib/jvm/java-8-openjdk-amd64/
```
hyangtack: Use the default java home where is `/docker-java-home`.

hyangtack:
This PR is related to #267 and #268

dist/Dockerfile
```diff
@@ -1,29 +1,30 @@
...
 ENV JAVA_OPTS "$CENTRALDOGMA_OPTS"
 
 # Install CentralDogma binaries and configurations.
-RUN mkdir -p                    \
-  "$CENTRALDOGMA_HOME"/bin      \
-  "$CENTRALDOGMA_HOME"/conf     \
-  "$CENTRALDOGMA_HOME"/lib      \
-  "$CENTRALDOGMA_HOME"/licenses \
+RUN mkdir -p                      \
+  "$CENTRALDOGMA_HOME"/bin        \
```
minwoox: Because the option is -p, perhaps we don't need this line?

hyangtack: Yes, right. But I wanted to clearly specify the directories to be created.

PR #272, opened 2018-08-21
Fix frontend-side route conflicts

### LINKED ISSUE #270: Repository name 'new' does not work in web UI
When there's a repository named 'new', the link for the repository view is `/#/{project}/new`.
But, we cannot open the view because the link makes us jump to the view of creating a new repository.

trustin:
Motivation:

Our frontend-side router has conflicts like:

- `/:projectName/new` vs. `/:projectName/:repoName` when `repoName` is `new`
- `/projects` vs. `/:projectName` when `projectName` is `projects`

Modifications:

- Make sure all entity paths have the form of
  `/projects/:projectName/repos/:repoName`
- Clarify the type of the entity being created
  - `new` -> `new_project`, `new_repo` or `new_file`

Result:

- Fixes #270
- A user can create and access a repository whose name is `new`, `list`,
  `files`, etc, from the web UI.
- A user can create and access a project whose name is `projects`,
  `metadata`, etc, from the web UI.
- URLs of the entities in the web UI have been changes in an incompatible way.
  - Note: The server-side HTTP API remains unaffected.

/cc @occho

### DIFF
--- a/server/src/main/resources/webapp/scripts/app/entities/metadata/metadata.project.html
+++ b/server/src/main/resources/webapp/scripts/app/entities/metadata/metadata.project.html
@@ -1,6 +1,6 @@
 <div ng-cloak>
   <h2>
-    <a ng-href="#">Projects</a> / {{project.name}}
+    <a ng-href="#/projects">Projects</a> / {{project.name}}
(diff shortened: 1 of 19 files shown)


trustin:
Thanks, reviewers.

PR #284, opened 2018-10-02
Use a dedicated repository for Go client

trustin:
Motivation:

The current way of our using gogradle plugin to build a Go client has
various issues including:

- non-standard layout as a Go project
- incompatibility with many Go tools due to the symbolic links generated
  by gogradle
- unnecessarily long import path

Modifications:

- Move the Go client to https://github.com/line/centraldogma-go
- Replace `client/go` with `cli`
  - Use gogradle only for downloading Go distributions

Result:

- Go developers will find a dedicated repository for the Go client,
  which follows Go's convention.
- Go developers now have much shorter import path: `go.linecorp.com/centraldogma`


trustin:
Thanks for reviewing.

PR #285, opened 2018-10-04
Make it possible to keep replicating in read-only mode

trustin:
Motivation:

A user may want to keep the replication turned on while prohibiting any
direct writes to a replica. This behavior is useful when a user needs to
make sure that all replicas caught up all replication logs, which is
unachievable if entered read-only mode with replication turned off
because the read-only replica will not be able to catch up the replication
logs created by other replicas after it went read-only.

Modifications:

- Add `CommandExecutor.writable` property so that it is possible to
  enter read-only mode without stopping replication.
- Update `AdministrativeService` to expose `replicating` property as well
  as `writable` in the server status.
- `AdministrativeService` now allows leaving read-only mode.

Result:

- It is much easier for a user to ensure that all replicas contain the
  same data during maintenance.
- A user can leave read-only mode without restarting a replica.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/api/AdministrativeService.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/api/AdministrativeService.java
@@ -16,6 +16,8 @@
 
 package com.linecorp.centraldogma.server.internal.api;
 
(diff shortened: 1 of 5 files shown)


trustin:
Thanks for reviewing

PR #288, opened 2018-10-11
Add JVM and Gradle options to gradle.properties

trustin:
- Add `-Xmx1280m` JVM option
- Enable on-demand project configuration


trustin:
Thanks, @hyangtack !

PR #289, opened 2018-10-16
fix JSON file name

LiYing2010:
Fix some incorrect JSON filename in document


trustin:
Thank you! Forgot to approve before merging but :-)

PR #293, opened 2018-10-23
Provide a way to merge multiple JSON files into one

### LINKED ISSUE #260: Provide a way to integrate multiple json files into one file when fecthing
Let's say that there are two configuration files:
```
global_settings.json
{
  "foo": "foo in global",
  "bar": "bar in global"
}

local_settings.json
{
  "foo": "foo in local",
  "baz": "baz in local"
}
```

A user might want to aggregate those two files and get as the sole file, overwriting same keys so the file will be:
```
aggregated_settings.json
{
  "foo": "foo in local",
  "bar": "bar in global",
  "baz": "baz in local"
}
```

minwoox:
Motivation
A user might want to fetch merged JSON files

Modifications
- Add the REST API to fetch merged JSON file
- Add the `aggregateEntries` method to thrift service
- Add `AggregatedEntry` class to represent the data

Result:
- Robust API
- Close #260


common/src/main/java/com/linecorp/centraldogma/common/AggregatedEntry.java
```diff
@@ -0,0 +1,147 @@
...
+    private String contentAsText;
+    @Nullable
+    private String contentAsPrettyText;
+
+    /**
+     * Creates a new instance.
+     */
+    private AggregatedEntry(EntryType type, T content) {
+        this.type = requireNonNull(type, "type");
+        this.content = requireNonNull(content, "content");
```
trustin: Check if `content` is instance of `type.type()`.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
```diff
@@ -368,4 +369,13 @@ private static Object handleWatchFailure(Throwable thrown) {
         }
         return converter.apply(Iterables.getOnlyElement(collection));
     }
+
+    @Get("/projects/{projectName}/repos/{repoName}/aggregation")
+    public CompletableFuture<?> aggregateEntries(@Param("revision") @Default("-1") String revision,
+                                                 @RequestObject Repository repository,
+                                                 @RequestObject(PathsRequestConverter.class)
+                                                                Iterable<String> paths) {
```
trustin: Can we make `I` in `Iterable` placed in the same column with `@` above?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
```diff
@@ -368,4 +369,13 @@ private static Object handleWatchFailure(Throwable thrown) {
         }
         return converter.apply(Iterables.getOnlyElement(collection));
     }
+
+    @Get("/projects/{projectName}/repos/{repoName}/aggregation")
```
trustin: `aggregated` or `aggregate`? (because we are retrieving an aggregated entry)

common/src/main/java/com/linecorp/centraldogma/common/AggregatedEntry.java
minwoox: Fixed. Thank you~!

server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
minwoox: I like `aggregated`. Thank you~!

server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
minwoox: Fixed. Thank you

minwoox:
> How about adding a new common interface...

That's a great idea. Thank you~!

> Could we allow a user specify JSON paths like we can to retrieve a single file?

I think it's okay even if a user specify only one JSON file. Is there any concern?

common-legacy/src/main/java/com/linecorp/centraldogma/internal/thrift/PathAndOptionalConverter.java
```diff
@@ -0,0 +1,46 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.linecorp.centraldogma.internal.thrift;
+
+        import com.google.common.base.Converter;
```
hyangtack: nit: indentation

common/src/main/java/com/linecorp/centraldogma/common/AggregatedEntry.java
```diff
@@ -0,0 +1,117 @@
...
+    public String contentAsPrettyText() {
+        if (contentAsPrettyText == null) {
+            contentAsPrettyText = ContentHolder.super.contentAsPrettyText();
+        }
+        return contentAsPrettyText;
+    }
+
+    @Override
+    public int hashCode() {
+        return (type.hashCode()) * 31 + content.hashCode();
```
hyangtack: nit: unnecessary parentheses

common/src/main/java/com/linecorp/centraldogma/common/PathAndOptional.java
```diff
@@ -0,0 +1,81 @@
...
+ */
+
+package com.linecorp.centraldogma.common;
+
+import static java.util.Objects.requireNonNull;
+
+import com.google.common.base.MoreObjects;
+
+/**
+ * A holder that contains a {@code path} and {@code optional} which the path is required.
```
hyangtack: nit: `{@code isOptional}`?

common/src/main/java/com/linecorp/centraldogma/internal/Jackson.java
```diff
@@ -257,6 +261,15 @@ public static String textValue(JsonNode node, String defaultValue) {
         return node != null && node.getNodeType() == JsonNodeType.STRING ? node.textValue() : defaultValue;
     }
 
+    public static JsonNode extractTree(JsonNode jsonNode, Iterable<String> jsonPaths) {
+        final int size = Iterables.size(jsonPaths);
+        for (int i = 0; i < size; i++) {
+            final String p = Iterables.get(jsonPaths, i);
+            jsonNode = Jackson.extractTree(jsonNode, p);
```
hyangtack: nit: `Jackson` can be removed?

server/src/main/java/com/linecorp/centraldogma/server/internal/ServiceUtil.java
```diff
@@ -0,0 +1,115 @@
...
+        requireNonNull(paths, "paths");
+        requireNonNull(expressions, "expressions");
+
+        // Only JSON files can currently be aggregated.
+        paths.forEach(path -> validateJsonFilePath(path.getPath(), "path"));
+
+        final List<CompletableFuture<Entry<?>>> entryFutures = new ArrayList<>(Iterables.size(paths));
+        paths.forEach(path -> entryFutures.add(repository.get(normalizedRevision, path.getPath())));
+        final CompletableFuture<Void> allDoneFuture = CompletableFuture.allOf(
+                entryFutures.toArray(new CompletableFuture[0]));
```
hyangtack: nit: how about making `new CompletableFuture[0]` as a constant?

server/src/main/java/com/linecorp/centraldogma/server/internal/ServiceUtil.java
```diff
@@ -0,0 +1,115 @@
...
+        final Builder<JsonNode> builder = new Builder<>();
+        for (int i = 0; i < entryFutures.size(); i++) {
+            try {
+                final Entry<?> entry = entryFutures.get(i).get();
+                builder.add(entry.contentAsJson());
+            } catch (Exception e) {
+                final PathAndOptional path = Iterables.get(paths, i);
+                if (!path.isOptional()) {
+                    if (e instanceof JsonParseException) {
+                        throw new CompletionException("failed to parse the content as JSON. path: " +
```
hyangtack: How about using another exception? Because it might be peeled at several points in our code, so its message(ex, path) might be lost.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/converter/PathAndOptionalRequestConverter.java
```diff
@@ -0,0 +1,64 @@
...
+
+    private static final Splitter keyValueSplitter = Splitter.on('=').trimResults().omitEmptyStrings();
+
+    @Nullable
+    @Override
+    public Object convertRequest(ServiceRequestContext ctx, AggregatedHttpMessage request,
+                                 Class<?> expectedResultType) throws Exception {
+        final String queryString = ctx.query();
+        if (queryString != null) {
+            final String decodeString = QueryStringDecoder.decodeComponent(queryString);
```
hyangtack: How about using `QueryStringDecoder.parameters()`?

minwoox: I have to parse it by myself because I have to keep the order of the `path` and   `optional_path`.
For example, when a user specify the query such as
```
path=1.json&optional_path=2.json&path=3.json
```
the aggregation will be applied in the order of `1.json`, `2.json`, `3.json`. But, it's hard to keep the original order if I use `parameters()`. :)

server/src/main/java/com/linecorp/centraldogma/server/internal/ServiceUtil.java
minwoox: Changed the exception to `IllegalStateException` because the server stores a JSON file content which is not a JSON document(which is not possible). Thank you~!

client/java-armeria-legacy/src/main/java/com/linecorp/centraldogma/client/armeria/legacy/LegacyCentralDogma.java
```diff
@@ -230,6 +233,38 @@
         });
     }
 
+    @Override
+    public CompletableFuture<AggregatedEntry<?>> aggregateEntries(String projectName, String repositoryName,
+                                                                  Revision revision,
+                                                                  List<PathAndOptional> paths,
+                                                                  List<String> expressions) {
```
trustin: How about introducing `AggregationQuery<T>` so both paths and expressions are presented in a single type, just like we do in `getFile()` and `Query<T>`? We could also add a shortcut method similar to `getFile(projectName, repositoryName, revision, path)`.

common-legacy/build.gradle
```diff
@@ -16,6 +16,7 @@ tasks.compileThrift.doLast {
         def content = oldContent
                 .replaceAll('fromRevision', 'from')
                 .replaceAll('toRevision', 'to')
+                .replaceAll('isIsOptional', 'isOptional')
```
trustin: I don't think we need this because we are adding a new property. There's no need for backward compatibility and we can just use `isIsOptional()` internally. A user is not supposed to use Thrift-generated classes.

common/src/main/java/com/linecorp/centraldogma/common/AggregatedEntry.java
```diff
@@ -0,0 +1,117 @@
...
+        if (this == o) {
+            return true;
+        }
+        if (!(o instanceof AggregatedEntry)) {
+            return false;
+        }
+
+        @SuppressWarnings("unchecked")
+        final AggregatedEntry<T> that = (AggregatedEntry<T>) o;
+
```
trustin: nit: perhaps better removing an empty line.

common-legacy/src/main/thrift/CentralDogma.thrift
```diff
@@ -339,6 +349,15 @@ service CentralDogmaService {
...
                             5: Query query) throws (1: CentralDogmaException e),
 
+    /**
+     * Retrieves the aggregated entry of the specified paths at the specified revision. If you specify
+     * the expressions, they will be applied to the content of the aggregated entry. Only JSON entry
+     * aggregation and JSON path expressions are currently supported.
+     */
+    AggregatedEntry aggregateEntries(1: string projectName, 2: string repositoryName,
+                                     3: Revision revision, 4: list<PathAndOptional> paths,
+                                     5: list<string> expressions) throws (1: CentralDogmaException e),
```
trustin: Ditto. Could be replaced with `AggregationQuery` as well.

server/src/main/java/com/linecorp/centraldogma/server/internal/ServiceUtil.java
```diff
@@ -0,0 +1,118 @@
...
+        allDoneFuture.whenComplete((unused1, unused2) -> {
+            JsonNode result;
+            try {
+                result = aggregateJson(paths, entryFutures);
+            } catch (Exception e) {
+                future.completeExceptionally(e);
+                return;
+            }
+
+            if (Iterables.size(expressions) > 0) {
```
trustin: Could use `!Iterables.isEmpty()`

server/src/main/java/com/linecorp/centraldogma/server/internal/ServiceUtil.java
```diff
@@ -0,0 +1,118 @@
...
+        for (int i = 0; i < entryFutures.size(); i++) {
+            try {
+                final Entry<?> entry = entryFutures.get(i).get();
+                builder.add(entry.contentAsJson());
+            } catch (Exception e) {
+                final PathAndOptional path = Iterables.get(paths, i);
+                if (!path.isOptional()) {
+                    if (e instanceof JsonParseException) {
+                        // Should never happen because the server has a JSON file whose content is not JSON.
+                        throw new IllegalStateException("failed to parse the content as JSON. path: " +
```
trustin: If it never happens we don't really need to check this because we rethrow in `else` anyway. How about removing the `if` block and just rethrowing `e`?

server/src/main/java/com/linecorp/centraldogma/server/internal/ServiceUtil.java
```diff
@@ -0,0 +1,118 @@
...
+        requireNonNull(repository, "repositoy");
+        requireNonNull(paths, "paths");
+        requireNonNull(expressions, "expressions");
+
+        // Only JSON files can currently be aggregated.
+        paths.forEach(path -> validateJsonFilePath(path.getPath(), "path"));
+
+        final List<CompletableFuture<Entry<?>>> entryFutures = new ArrayList<>(Iterables.size(paths));
+        paths.forEach(path -> entryFutures.add(repository.get(normalizedRevision, path.getPath())));
+        final CompletableFuture<Void> allDoneFuture = CompletableFuture.allOf(
```
trustin: Could use `CompletableFutures.allAsList()`

server/src/main/java/com/linecorp/centraldogma/server/internal/ServiceUtil.java
```diff
@@ -0,0 +1,118 @@
...
+public final class ServiceUtil {
+
+    private static final CompletableFuture[] EMPTY_COMPLETABLE_FUTURE = new CompletableFuture[0];
+
+    /**
+     * Aggregates the JSON files in the specified paths and returns the {@link AggregatedEntry}.
+     */
+    public static CompletableFuture<AggregatedEntryDto<?>> aggregatedEntries(
+            Revision normalizedRevision, Repository repository,
+            Iterable<PathAndOptional> paths, Iterable<String> expressions) {
```
trustin: Ditto, could make use of `AggregationQuery`

common-legacy/src/main/thrift/CentralDogma.thrift
```diff
@@ -196,6 +196,16 @@ struct Query {
     3: required list<string> expressions,
 }
 
+enum AggregationQueryType {
```
trustin: - Could reuse `QueryType`?
- `IDENTITY` is missing?

common/src/main/java/com/linecorp/centraldogma/common/AggregationJsonQuery.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.List;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.google.common.base.MoreObjects;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Streams;
+
+import com.linecorp.centraldogma.internal.Util;
+
+final class AggregationJsonQuery implements AggregationQuery<JsonNode> {
```
trustin: `JsonAggregationQuery`? Which sounds better?

common/src/main/java/com/linecorp/centraldogma/common/AggregationJsonQuery.java
```diff
@@ -0,0 +1,84 @@
...
+        }
+        @SuppressWarnings("unchecked")
+        final AggregationJsonQuery that = (AggregationJsonQuery) o;
+        return pathAndOptionals().equals(that.pathAndOptionals()) && expressions().equals(that.expressions());
+    }
+
+    @Override
+    public String toString() {
+        return MoreObjects.toStringHelper(this)
+                          .add("pathAndOptionals", pathAndOptionals())
```
trustin: Could add `type`?

common/src/main/java/com/linecorp/centraldogma/common/AggregationQuery.java
```diff
@@ -0,0 +1,76 @@
...
+    /**
+     * Returns a newly-created {@link AggregationQuery} that aggregates the JSON contents in the specified
+     * {@code pathAndOptionals}. Then, the specified
+     * <a href="https://github.com/json-path/JsonPath/blob/master/README.md">JSON path expressions</a>
+     * are applied to the content.
+     *
+     * @param pathAndOptionals the paths of JSON files being aggregated and whether it is optional
+     * @param jsonPaths the JSON path expressions to apply
+     */
+    static AggregationQuery<JsonNode> ofAggregationJson(Iterable<PathAndOptional> pathAndOptionals,
```
trustin: I think we need to be consistent with `Query` API:

- `ofJson()` - identity aggregation query (which could be just a shortcut of `ofJsonPath()` with empty expressions unlike `Query` though.)
- `ofJsonPath()` - JSON aggregation query

common/src/main/java/com/linecorp/centraldogma/common/AggregationQueryType.java
```diff
@@ -0,0 +1,44 @@
...
+package com.linecorp.centraldogma.common;
+
+import java.util.Collections;
+import java.util.EnumSet;
+import java.util.Set;
+
+/**
+ * The type of an {@link AggregationQuery}.
+ */
+public enum AggregationQueryType {
```
trustin: Is there a reason we cannot reuse `QueryType`?

common/src/main/java/com/linecorp/centraldogma/common/JsonPathQuery.java
```diff
@@ -21,49 +21,37 @@
...
-import com.jayway.jsonpath.JsonPath;
 
 import com.linecorp.centraldogma.internal.Jackson;
+import com.linecorp.centraldogma.internal.Util;
 
 final class JsonPathQuery implements Query<JsonNode> {
 
     private final String path;
     private final List<String> jsonPaths;
+    @Nullable
```
trustin: It's never nullable.

common/src/main/java/com/linecorp/centraldogma/internal/Util.java
```diff
@@ -102,6 +104,24 @@ public static boolean isValidJsonFilePath(String path) {
                JSON_FILE_PATH_PATTERN.matcher(path).matches();
     }
 
+    public static String validateJsonPath(String jsonPath, String paramName) {
+        requireNonNull(jsonPath, paramName);
+        if (isValidJsonPath(jsonPath)) {
+            return jsonPath;
+        }
+
+        throw new IllegalArgumentException(paramName + " is invalid JSON path: " + jsonPath);
```
trustin: How about: is not a valid JSON path

server/src/main/java/com/linecorp/centraldogma/server/internal/ServiceUtil.java
```diff
@@ -46,17 +45,17 @@
     private static final CompletableFuture[] EMPTY_COMPLETABLE_FUTURE = new CompletableFuture[0];
 
     /**
-     * Aggregates the JSON files in the specified paths and returns the {@link AggregatedEntry}.
+     * Aggregates the JSON files in the specified {@link AggregationQuery} and returns the
+     * {@link AggregatedEntry}.
```
trustin: and returns the ... -> nil ? (Seems like redundant information)

server/src/main/java/com/linecorp/centraldogma/server/internal/ServiceUtil.java
```diff
@@ -75,7 +74,8 @@
                 return;
             }
 
-            if (Iterables.size(expressions) > 0) {
+            final List<String> expressions = query.expressions();
+            if (!Iterables.isEmpty(expressions)) {
```
trustin: Now that `expressions` is a `List`, we can just use `expressions.isEmpty()`.

common/src/main/java/com/linecorp/centraldogma/internal/Jackson.java
```diff
@@ -281,6 +294,18 @@ public static String escapeText(String text) {
         return new String(enc.quoteAsString(text));
     }
 
+    public static JsonNode aggregateJsonNodes(Iterable<JsonNode> jsonNodes) {
+        final ObjectNode node = compactMapper.createObjectNode();
+        for (JsonNode jsonNode : jsonNodes) {
+            final Iterator<Entry<String, JsonNode>> fields = jsonNode.fields();
```
trustin: Could we fail if `jsonNode` is not an object?

common/src/main/java/com/linecorp/centraldogma/internal/Jackson.java
```diff
@@ -281,6 +294,18 @@ public static String escapeText(String text) {
         return new String(enc.quoteAsString(text));
     }
 
+    public static JsonNode aggregateJsonNodes(Iterable<JsonNode> jsonNodes) {
+        final ObjectNode node = compactMapper.createObjectNode();
```
trustin: We could just use `JsonNodeFactory.instance.objectNode()`

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java
```diff
@@ -164,6 +167,36 @@
     CompletableFuture<Map<String, Entry<?>>> getFiles(String projectName, String repositoryName,
                                                       Revision revision, String pathPattern);
 
+    /**
+     * Retrieves the merged entry of the specified {@link PathAndOptional}s at the specified revision.
+     * Only JSON entry merger is currently supported.
```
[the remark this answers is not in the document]
minwoox: Fixed~!

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java
```diff
@@ -164,6 +167,36 @@
...
+     */
+    default <T> CompletableFuture<MergedEntry<?>> mergeFiles(
```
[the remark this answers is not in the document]
minwoox: Fixed~!

common/src/main/java/com/linecorp/centraldogma/common/JsonMergerQuery.java
```diff
@@ -0,0 +1,85 @@
...
+final class JsonMergerQuery implements MergerQuery<JsonNode> {
+
+    private final List<PathAndOptional> pathAndOptionals;
+
+    private final List<String> jsonPaths;
+
+    JsonMergerQuery(Iterable<PathAndOptional> pathAndOptionals, Iterable<String> jsonPaths) {
+        this.pathAndOptionals = ImmutableList.copyOf(requireNonNull(pathAndOptionals, "pathAndOptionals"));
+        this.jsonPaths = Streams.stream(requireNonNull(jsonPaths, "jsonPaths"))
+                                .peek(jsonPath -> Util.validateJsonPath(jsonPath, "jsonPath"))
```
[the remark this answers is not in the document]
minwoox: Fixed~!

common/src/main/java/com/linecorp/centraldogma/internal/Jackson.java
```diff
@@ -281,6 +295,76 @@ public static String escapeText(String text) {
...
+                final JsonNode updateValue = update.get(fieldName);
+
+                if (baseValue == null || baseValue.isNull() || updateValue.isNull()) {
+                    if (isMerging) {
+                        baseObject.set(fieldName, updateValue);
+                    }
+                    continue;
+                }
+
+                if (baseValue.getNodeType() == updateValue.getNodeType()) {
```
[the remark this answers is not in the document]
minwoox: I couldn't think about it. Thank you for reducing my codes. :-)

server/src/main/java/com/linecorp/centraldogma/server/internal/api/converter/MergerQueryRequestConverter.java
```diff
@@ -0,0 +1,71 @@
...
+                }
+                if ("path".equals(Iterables.get(keyValue, 0))) {
+                    builder.add(new PathAndOptional(Iterables.get(keyValue, 1), false));
+                } else if ("optional_path".equals(Iterables.get(keyValue, 0))) {
+                    builder.add(new PathAndOptional(Iterables.get(keyValue, 1), true));
+                }
+            }
+
+            final List<String> jsonPaths = new QueryStringDecoder(queryString, false).parameters()
+                                                                                     .get("jsonpath");
```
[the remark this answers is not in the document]
minwoox: Fixed~!

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java
```diff
@@ -164,6 +168,59 @@
...
+        final CompletableFuture<MergedEntry<?>> future =
+                (CompletableFuture<MergedEntry<?>>) (CompletableFuture<?>) mergeFiles(
+                        projectName, repositoryName, revision, MergerQuery.ofJsonPath(pathAndOptionals));
+        return future;
+    }
+
+    /**
+     * Retrieves the merged entry of the specified {@link MergerQuery} at the specified revision.
+     * Only JSON entry merger is currently supported. The JSON files are merged sequentially in the order
+     * of specified in the {@link MergerQuery}.
```
trustin: in the order of ... -> as specified in the MergerQuery ?

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java
```diff
@@ -164,6 +168,59 @@
...
+     * @return the {@link MergedEntry} which contains the merged JSON files in the
+     *         specified {@code pathAndOptionals}
+     */
+    default CompletableFuture<MergedEntry<?>> mergeFiles(
+            String projectName, String repositoryName,
+            Revision revision, Iterable<PathAndOptional> pathAndOptionals) {
+        @SuppressWarnings("unchecked")
+        final CompletableFuture<MergedEntry<?>> future =
+                (CompletableFuture<MergedEntry<?>>) (CompletableFuture<?>) mergeFiles(
+                        projectName, repositoryName, revision, MergerQuery.ofJsonPath(pathAndOptionals));
```
trustin: `MergerQuery.ofJson`

common-legacy/src/main/thrift/CentralDogma.thrift
```diff
@@ -178,6 +188,7 @@ struct Project {
 enum QueryType {
     IDENTITY = 1,
     JSON_PATH = 2,
+    JSON_MERGER = 3,
```
trustin: I don't think we need this. `QueryType` of `mergeFiles` should be:

- `IDENTITY` if no JSON paths are given.
- `JSON_PATH` if one or more JSON paths are given.

common/src/main/java/com/linecorp/centraldogma/common/PathAndOptional.java
```diff
@@ -0,0 +1,80 @@
...
+        @SuppressWarnings("unchecked")
+        final PathAndOptional that = (PathAndOptional) o;
+        return path.equals(that.path) && isOptional == that.isOptional;
+    }
+
+    @Override
+    public String toString() {
+        return MoreObjects.toStringHelper(this)
+                          .add("path", path)
+                          .add("isOptional", isOptional)
```
trustin: `optional`?

common/src/main/java/com/linecorp/centraldogma/common/PathAndOptional.java
```diff
@@ -0,0 +1,80 @@
...
+package com.linecorp.centraldogma.common;
+
+import static java.util.Objects.requireNonNull;
+
+import com.google.common.base.MoreObjects;
+
+/**
+ * A holder that contains a {@code path} and {@code isOptional} which indicates whether the path is required.
+ */
+public class PathAndOptional {
```
trustin: I'm not sure I like this name... How about `MergeSource`?

common/src/main/java/com/linecorp/centraldogma/internal/Jackson.java
```diff
@@ -281,6 +295,66 @@ public static String escapeText(String text) {
         return new String(enc.quoteAsString(text));
     }
 
+    public static JsonNode mergeJsonNodes(JsonNode... jsonNodes) {
+        return mergeJsonNodes(ImmutableList.copyOf(requireNonNull(jsonNodes, "jsonNodes")));
+    }
+
+    public static JsonNode mergeJsonNodes(Iterable<JsonNode> jsonNodes) {
```
trustin: How about `mergeTree` for consistency with `extractTree`?

common/src/main/java/com/linecorp/centraldogma/internal/Jackson.java
```diff
@@ -281,6 +295,66 @@ public static String escapeText(String text) {
...
+        requireNonNull(jsonNodes, "jsonNodes");
+        final int size = Iterables.size(jsonNodes);
+        checkArgument(size > 0, "jsonNodes size: %s (expected: > 0)", size);
```
trustin: size is always zero, so you can simplify this: `jsonNodes is empty.`

common/src/main/java/com/linecorp/centraldogma/internal/Jackson.java
```diff
@@ -281,6 +295,66 @@ public static String escapeText(String text) {
...
+        final JsonNode first = Iterables.get(jsonNodes, 0);
+        JsonNode merged = first.deepCopy();
+
+        StringBuilder builder = new StringBuilder("/");
+        for (int i = 1; i < size; i++) {
+            final JsonNode addition = Iterables.get(jsonNodes, i);
```
trustin: This can be seriously inefficient. You can:

```java
final Iterable<JsonNode> it = jsonNodes.iterator();
final JsonNode first = it.next();
JsonNode merged = first.deepCopy();

while (it.hasNext()) {
    final JsonNode addition = it.next();
    ...
}
```

common/src/main/java/com/linecorp/centraldogma/internal/Jackson.java
```diff
@@ -281,6 +295,66 @@ public static String escapeText(String text) {
...
+        return merged;
+    }
+
+    private static JsonNode traverse(JsonNode base, JsonNode update, StringBuilder builder, boolean isMerging) {
+        if (base.getNodeType() != update.getNodeType() && (!base.isNull() || !update.isNull())) {
+            throw new MismatchedValueException(builder + " type: " + update.getNodeType() +
+                                               " (expected: " + base.getNodeType() + ')');
+        }
+
+        if (base.isObject() && update.isObject()) {
```
trustin: Could we fail if this is not the case at the root level?

common/src/main/java/com/linecorp/centraldogma/internal/MismatchedValueException.java
```diff
@@ -0,0 +1,50 @@
...
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.linecorp.centraldogma.internal;
+
+/**
+ * A {@link RuntimeException} that is raised when merging JSON files is failed due to mismatched value.
+ */
+public final class MismatchedValueException extends RuntimeException {
```
trustin: How about more generic name like `MergeConflictException` or `EntryMergeConflictException`, because otherwise we will have to create an exception for every case where merge operation fails?

common/src/main/java/com/linecorp/centraldogma/internal/api/v1/MergedEntryDto.java
```diff
@@ -0,0 +1,54 @@
...
+package com.linecorp.centraldogma.internal.api.v1;
+
+import static java.util.Objects.requireNonNull;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.base.MoreObjects;
+
+import com.linecorp.centraldogma.common.EntryType;
+
+public class MergedEntryDto<T> {
```
trustin: How about adding `revision` as well?

server/src/main/java/com/linecorp/centraldogma/server/internal/ServiceUtil.java
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * Utility methods that are useful when implementing a service.
+ */
+public final class ServiceUtil {
+
+    /**
+     * Merges the JSON files in the specified {@link MergerQuery}.
+     */
+    public static <T> CompletableFuture<MergedEntryDto<JsonNode>> mergeFiles(
```
trustin: Could you move this to `internal.storage.repository.Repository`?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
```diff
@@ -369,4 +371,13 @@ private static Object handleWatchFailure(Throwable thrown) {
         }
         return converter.apply(Iterables.getOnlyElement(collection));
     }
+
+    @Get("/projects/{projectName}/repos/{repoName}/merger")
```
trustin: `/merged` or `/merge`?

server/src/main/java/com/linecorp/centraldogma/server/internal/thrift/CentralDogmaServiceImpl.java
```diff
@@ -322,7 +333,7 @@ public void watchRepository(
     private static void handleWatchRepositoryResult(
             CompletableFuture<com.linecorp.centraldogma.common.Revision> future,
             AsyncMethodCallback resultHandler) {
-        future.handle(voidFunction((res, cause) -> {
+        future.whenComplete((res, cause) -> {
```
trustin: (!)

common/src/main/java/com/linecorp/centraldogma/common/MergerQuery.java
```diff
@@ -0,0 +1,96 @@
...
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.google.common.collect.ImmutableList;
+
+/**
+ * A merger query on files.
+ *
+ * @param <T> the content type of files being queried
+ */
+public interface MergerQuery<T> {
```
trustin: Google search shows me that 'merge query' is a fairly common term. How about renaming to `MergeQuery`?

common/src/main/java/com/linecorp/centraldogma/common/PathAndOptional.java
```diff
@@ -0,0 +1,80 @@
...
+
+    private final String path;
+
+    private final boolean isOptional;
+
+    /**
+     * Creates a new instance.
+     */
+    public PathAndOptional(String path, boolean isOptional) {
```
trustin: How about making this private and adding two public static factory methods? `ofRequired` and `ofOptional`

common/src/main/java/com/linecorp/centraldogma/common/Query.java
```diff
@@ -80,7 +81,7 @@
      * @param jsonPaths the JSON path expressions to apply
      */
     static Query<JsonNode> ofJsonPath(String path, String... jsonPaths) {
-        return new JsonPathQuery(path, jsonPaths);
+        return ofJsonPath(path, ImmutableList.copyOf(requireNonNull(jsonPaths, "jsonPaths")));
```
trustin: Perhaps OK to revert?

common/src/main/java/com/linecorp/centraldogma/common/QueryType.java
```diff
@@ -32,7 +32,12 @@
      * Applies a series of <a href="https://github.com/json-path/JsonPath/blob/master/README.md">JSON path
      * expressions</a> to the content.
      */
-    JSON_PATH(EnumSet.of(EntryType.JSON));
+    JSON_PATH(EnumSet.of(EntryType.JSON)),
+
+    /**
+     * Retrieves the merged JSON content.
+     */
+    JSON_MERGER(EnumSet.of(EntryType.JSON));
```
trustin: I'm not sure why we need a new `QueryType` value.

server/src/main/java/com/linecorp/centraldogma/server/internal/ServiceUtil.java
```diff
@@ -0,0 +1,108 @@
...
+
+            JsonNode result = Jackson.mergeJsonNodes(builder.build());
+            final List<String> expressions = query.expressions();
+            if (!expressions.isEmpty()) {
+                try {
+                    result = Jackson.extractTree(result, expressions);
+                } catch (Exception e) {
+                    // A user sent the request with a invalid JSON path or pointing to the property which
+                    // the merged JSON doesn't have.
+                    future.completeExceptionally(new IllegalArgumentException(e));
```
trustin: `QueryExecutionException`?

common/src/main/java/com/linecorp/centraldogma/internal/MismatchedValueException.java
trustin: Alternatively, maybe we could just use `QueryExecutionException` and do not add a new exception at all?

common-legacy/src/main/thrift/CentralDogma.thrift
minwoox: I found while I'm fixing this that the supported entry types for `IDENTITY` are `TEXT` and `JSON`. Currently, we only support `JSON` for merge. Is it OK to use `IDENTITY`?

common/src/main/java/com/linecorp/centraldogma/internal/Jackson.java
minwoox: Renamed to `mergeTree`. Thank you~!

common/src/main/java/com/linecorp/centraldogma/internal/MismatchedValueException.java
minwoox: Just used `QueryExecutionException`.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
minwoox: Changed to `merge`. Thank you~!

common/src/main/java/com/linecorp/centraldogma/common/MergeQuery.java
```diff
@@ -0,0 +1,111 @@
...
+     * {@code mergeSources}. Then, the specified
+     * <a href="https://github.com/json-path/JsonPath/blob/master/README.md">JSON path expressions</a>
+     * are applied to the content.
+     *
+     * @param type the type of the {@link MergeQuery}
+     * @param mergeSources the paths of JSON files being merged and indicates whether it is optional
+     * @param jsonPaths the JSON path expressions to apply
+     */
+    static MergeQuery<JsonNode> of(QueryType type, Iterable<MergeSource> mergeSources,
+                                   Iterable<String> jsonPaths) {
```
trustin: Could we make this general like `Query.of()` just in case we introduce merge queries to other formats?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/Repository.java
```diff
@@ -424,4 +434,62 @@ default Author author() {
...
+                    builder.add(entry.contentAsJson());
+                } catch (JsonParseException e) {
+                    future.completeExceptionally(e);
+                    return null;
+                }
+            }
+
+            JsonNode result;
+            try {
+                result = Jackson.mergeTree(builder.build());
```
trustin: I wonder if we should pass an empty list if all merge sources are non-existent. Should we raise a `QueryExecutionException` rather than `IllegalArgumentException`?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/cache/CacheableMergeQueryCall.java
```diff
@@ -0,0 +1,97 @@
...
+            weight += value.contentAsText().length();
+        }
+        return weight;
+    }
+
+    @Override
+    CompletableFuture<MergedEntry<?>> execute() {
+        @SuppressWarnings("unchecked")
+        final CompletableFuture<MergedEntry<?>> future =
+                (CompletableFuture<MergedEntry<?>>) (CompletableFuture<?>) repo.mergeFiles(revision, query);
```
trustin: This will prevent the cached-fetches for individual merge sources. Could we modify this a little bit so that `Repository.mergeFiles()` (the default implementation) is invoked here instead of `CachedRepository.mergeFiles()`?

common-legacy/src/main/thrift/CentralDogma.thrift
```diff
@@ -339,6 +356,13 @@ service CentralDogmaService {
                             3: Revision fromRevision, 4: Revision toRevision,
                             5: Query query) throws (1: CentralDogmaException e),
 
+    /**
+     * Retrieves the merged entry of the specified query at the specified revision. Only JSON entry
+     * merge is currently supported. The JSON files are merged sequentially as specified in the mergeQuery.
+     */
```
hyangtack: nit: How about adding the following description for consistency?
```
     * <p>Note that only {@link ObjectNode} is recursively merged traversing the children. Other node types are
     * simply replaced.

```

minwoox:
> Would it be useful for a user if we include the list of the entry paths which participated into the query? For example, a user could specify three optional merge sources and want to know which merge sources were used actually.

Yeah, it will be very useful. Will add it. Thank you~!

common/src/main/java/com/linecorp/centraldogma/common/Entry.java
```diff
@@ -174,101 +165,58 @@ public void ifHasContent(Consumer<? super T> consumer) {
     }
 
     /**
-     * Returns the content of this {@link Entry}.
+     * Returns a new {@link Entry} which applied the specified {@link Query} to the {@link #content()}.
```
trustin: Could you rephrase this a little bit? It's not an entry who applied the query. How about:

```
Applies the specified {@link Query} to the {@link #content()} of this {@link Entry} and returns
the query result.
```

common/src/main/java/com/linecorp/centraldogma/common/Entry.java
```diff
@@ -174,101 +165,58 @@ public void ifHasContent(Consumer<? super T> consumer) {
...
      *
      * @throws IllegalStateException if this {@link Entry} is a directory
-     */
+     * @throws QuerySyntaxException if the syntax of specified {@link Query} is invalid
+     * @throws QueryExecutionException if an {@link Exception} is raised while applying the specified
+     *                                 {@link Query} to the {@link #content()}
+     */
+    public Entry<T> applyQuery(Query<T> query) {
```
trustin: Could you explain why you had to introduce this method instead of using `Query.apply()`?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/converter/MergeQueryRequestConverter.java
```diff
@@ -0,0 +1,68 @@
...
+
+    private static final Splitter querySplitter = Splitter.on('&').trimResults().omitEmptyStrings();
+
+    @Nullable
+    @Override
+    public Object convertRequest(ServiceRequestContext ctx, AggregatedHttpMessage request,
+                                 Class<?> expectedResultType) throws Exception {
+        final String queryString = ctx.query();
+        if (queryString != null) {
+            final String decodedString = QueryStringDecoder.decodeComponent(queryString);
```
trustin: Could you put some comments about why you had to decode this by yourself?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/RepositoryUtil.java
```diff
@@ -20,32 +20,109 @@
...
+            return null;
+        });
+        return unsafeCast(future);
+    }
+
+    private static String concatenatePaths(Iterable<MergeSource> mergeSources) {
+        return Streams.stream(mergeSources).map(MergeSource::path).collect(Collectors.joining(","));
+    }
+
+    public static <T> void completeQueryExecutionException(CompletableFuture<MergedEntry<T>> future,
```
trustin: Could you name this better to explain what this method does more precisely?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/cache/CacheableMergeQueryCall.java
```diff
@@ -0,0 +1,97 @@
...
+        }
+        if (value != null) {
+            weight += value.contentAsText().length();
+        }
+        return weight;
+    }
+
+    @Override
+    CompletableFuture<MergedEntry<?>> execute() {
+        throw new UnsupportedOperationException();
```
trustin: I'm curious if we could keep the result as a field and return it here, i.e. memoization, so we don't need to call `AsyncLoadingCache.get(key, function)` but just call `get(key)` like a usual cache lookup.

server/src/test/java/com/linecorp/centraldogma/server/internal/api/MergeFileTest.java
```diff
@@ -0,0 +1,171 @@
...
+        assertThatJson(actualJson).isEqualTo(expectedJson);
+
+        queryString = "path=/foo.json" + '&' +
+                      "path=/foo1.json" + '&' +
+                      "path=/foo2.json" + '&' +
+                      "path=/foo3.json";
+        aRes = httpClient().get("/api/v1/projects/myPro/repos/myRepo/merge?" + queryString).aggregate()
+                           .join();
+        assertThat(aRes.status()).isEqualTo(HttpStatus.NOT_FOUND);
+        assertThat(aRes.content().toStringUtf8()).contains("Entry '/foo3.json (4)' does not exist.");
```
trustin: Global comment: Could you make sure the failure is handled by `HttpApiExceptionHandler` so that the error response is also JSON?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/Repository.java
```diff
@@ -442,54 +433,32 @@ default Author author() {
         requireNonNull(revision, "revision");
         requireNonNull(query, "query");
 
-        final List<MergeSource> paths = query.mergeSources();
+        final List<MergeSource> mergeSources = query.mergeSources();
         // Only JSON files can currently be merged.
-        paths.forEach(path -> validateJsonFilePath(path.path(), "path"));
+        mergeSources.forEach(path -> validateJsonFilePath(path.path(), "path"));
 
         final Revision normalizedRevision = normalizeNow(revision);
```
hyangtack: Just curious. Isn't necessary catching an exception and returning completed future if an exception is caught, like the other code?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/cache/CachingRepository.java
```diff
@@ -272,8 +323,64 @@ public RevisionRange normalizeNow(Revision from, Revision to) {
...
+            return CompletableFutures.exceptionallyCompletedFuture(e);
+        }
+
+        final CacheableMergeQueryCall key = new CacheableMergeQueryCall(repo, normalizedRevision, query);
+        final CompletableFuture<Object> value = cache.getIfPresent(key);
+        if (value != null) {
+            return unsafeCast(value);
+        }
+
+        final List<MergeSource> mergeSources = query.mergeSources();
```
hyangtack: I'm curious what if we just call `Repository.mergeFiles()` here, instead of L344-375, like as follows:
```
return Repository.super.mergeFiles(revision, query).thenApply(mergedEntry -> {
    cache.get(key, unused -> mergedEntry);
    return mergedEntry;
});
```

hyangtack: I realized that this may be related to @trustin 's comment: https://github.com/line/centraldogma/pull/293#discussion_r236561229.
Forget this comment if you can resolve his comment. :-)

server/src/main/java/com/linecorp/centraldogma/server/internal/api/converter/MergeQueryRequestConverter.java
```diff
@@ -40,9 +40,11 @@ public Object convertRequest(ServiceRequestContext ctx, AggregatedHttpMessage re
                                  Class<?> expectedResultType) throws Exception {
         final String queryString = ctx.query();
         if (queryString != null) {
+            // Decode queryString here so that the original order of "path" and "optional_path" is preserved.
```
trustin: here -> manually

server/src/main/java/com/linecorp/centraldogma/server/internal/api/converter/MergeQueryRequestConverter.java
```diff
@@ -40,9 +40,11 @@ public Object convertRequest(ServiceRequestContext ctx, AggregatedHttpMessage re
...
+            // For example if a user specify the query as "path=/a.json&optional_path=b.json&path=c.json",
```
trustin: specifies

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/RepositoryUtil.java
```diff
@@ -113,14 +116,32 @@ private static String concatenatePaths(Iterable<MergeSource> mergeSources) {
...
+            throw new QueryExecutionException(query + " does not support this entry type: " + entryType);
         }
 
-        future.completeExceptionally(new QueryExecutionException(cause));
+        if (queryType == IDENTITY) {
+            return entry;
+        } else if (queryType == JSON_PATH) {
+            return Entry.of(entry.revision(), query.path(), entryType, query.apply(entry.content()));
+        } else {
+            throw new QueryExecutionException(query + " does not support this entry type: " + entryType);
```
trustin: How about:

`"Unsupported entry type: " + entryType + " (query: " + query + ')'`

because the query string may be pretty long.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/RepositoryUtil.java
```diff
@@ -113,14 +116,32 @@ private static String concatenatePaths(Iterable<MergeSource> mergeSources) {
...
+     *                                 {@link Query} to the {@link Entry#content()}
+     */
+    public static <T> Entry<T> applyQuery(Entry<T> entry, Query<T> query) {
+        requireNonNull(query, "query");
+        entry.content(); // Ensure that content is not null.
+        final EntryType entryType = entry.type();
+
+        final QueryType queryType = query.type();
+        if (!queryType.supportedEntryTypes().contains(entryType)) {
+            throw new QueryExecutionException(query + " does not support this entry type: " + entryType);
```
trustin: Ditto

common/src/main/java/com/linecorp/centraldogma/common/MergeQuery.java
```diff
@@ -0,0 +1,111 @@
...
+     * {@code mergeSources}. Then, the specified
+     * <a href="https://github.com/json-path/JsonPath/blob/master/README.md">JSON path expressions</a>
+     * are applied to the content.
+     *
+     * @param type the type of the {@link MergeQuery}
+     * @param mergeSources the paths of JSON files being merged and indicates whether it is optional
+     * @param jsonPaths the JSON path expressions to apply
+     */
+    static MergeQuery<?> of(QueryType type, Iterable<MergeSource> mergeSources,
+                            Iterable<String> jsonPaths) {
```
trustin: jsonPaths -> expressions, just in case we add more `QueryType`s.

common/src/main/java/com/linecorp/centraldogma/common/MergeQuery.java
```diff
@@ -0,0 +1,111 @@
...
+    static MergeQuery<JsonNode> ofJsonPath(Iterable<MergeSource> mergeSources,
+                                           Iterable<String> jsonPaths) {
+        return new JsonMergeQuery(JSON_PATH, mergeSources, jsonPaths);
+    }
+
+    /**
+     * Returns a newly-created {@link MergeQuery} that merges the JSON contents as specified in the
+     * {@code mergeSources}. Then, the specified
+     * <a href="https://github.com/json-path/JsonPath/blob/master/README.md">JSON path expressions</a>
+     * are applied to the content.
```
trustin: Please generalize this just in case we add more `QueryType`s than just JSON path.

common/src/main/java/com/linecorp/centraldogma/common/MergeQuery.java
```diff
@@ -0,0 +1,111 @@
...
+     *
+     * @param mergeSources the paths of JSON files being merged and indicates whether it is optional
+     * @param jsonPaths the JSON path expressions to apply
+     */
+    static MergeQuery<JsonNode> ofJsonPath(Iterable<MergeSource> mergeSources,
+                                           String... jsonPaths) {
+        return ofJsonPath(mergeSources, ImmutableList.copyOf(requireNonNull(jsonPaths, "jsonPaths")));
+    }
+
+    static MergeQuery<JsonNode> ofJsonPath(Iterable<MergeSource> mergeSources,
```
trustin: Missing Javadoc

common/src/main/java/com/linecorp/centraldogma/common/MergeSource.java
```diff
@@ -0,0 +1,95 @@
...
+
+    private final String path;
+
+    private final boolean optional;
+
+    /**
+     * Creates a new instance.
+     */
+    private MergeSource(String path, boolean optional) {
+        this.path = requireNonNull(path, "path");
```
trustin: Perhaps we should use `Util.validateFilePath()`?

common/src/main/java/com/linecorp/centraldogma/common/MergedEntry.java
```diff
@@ -0,0 +1,154 @@
...
+        this.type = requireNonNull(type, "type");
+        requireNonNull(content, "content");
+        final Class<?> entryType = type.type();
+        checkArgument(entryType.isAssignableFrom(content.getClass()),
+                      "content type: %s (expected: %s)", content.getClass(), entryType);
+        this.content = content;
+        this.paths = ImmutableList.copyOf(requireNonNull(paths, "paths"));
+    }
+
+    // TODO(minwoox) Add this method to ContentHolder when we include the revision in Entry as well.
```
trustin: Could you add Javadoc meanwhile?

common/src/main/java/com/linecorp/centraldogma/internal/Util.java
```diff
@@ -102,6 +104,24 @@ public static boolean isValidJsonFilePath(String path) {
                JSON_FILE_PATH_PATTERN.matcher(path).matches();
     }
 
+    public static String validateJsonPath(String jsonPath, String paramName) {
+        requireNonNull(jsonPath, paramName);
+        if (isValidJsonPath(jsonPath)) {
```
trustin: How about negating the condition and throwing first?

server/src/main/java/com/linecorp/centraldogma/server/internal/metadata/MetadataRepository.java
```diff
@@ -42,14 +42,14 @@
 import com.linecorp.centraldogma.server.internal.storage.project.ProjectManager;
 import com.linecorp.centraldogma.server.internal.storage.repository.Repository;
 
-final class RepositoryUtil<T> {
+final class MetadataRepository<T> {
```
trustin: How about `MetadataAccess`? It's not a `Repository`.

hyangtack: How about some other name such as `RepositoySupport`? Because it's not for metadata only.

trustin:
Nice work. Looking forward to the documentation update! :-)

minwoox:
Thanks, @trustin, and @hyangtack for the great reviews!

PR #296, opened 2018-10-24
Load client profiles from all resources

### LINKED ISSUE #292: Look for more than one `centraldogma-profiles.json` files in class path
Currently, Central Dogma client uses only the first found `centraldogma-profiles.json` file to load client profiles. It would be useful to load all `centraldogma-profiles.json` files in the class path so that it is possible for a user to add custom profiles on top of the official profiles provided by administrators.

~We are going to handle the case of profile name conflicts strictly though.~ A user may want to override an official profile, so we will have to use the first-found profile in case of conflict.

trustin:
Motivation:

Currently, Central Dogma client uses only the first found
`centraldogma-profiles.json` file to load client profiles. It would be
useful to load all `centraldogma-profiles.json` files in the class path
so that it is possible for a user to add custom profiles on top of the
official profiles provided by administrators.

Modifications:

- Add `priority' property to a client profile
- Load all profile JSONs so that all profiles are considered.
- Make sure the profile that has higher priority or appeared first in
  the classpath wins when there are more than one profile with the same
  name.

Result:

- Fixes #292


trustin:
Thanks for reviewing!

PR #307, opened 2018-11-11
Add the path to Central Dogma bin directory to PATH environment variable

matsumana:
I think it's more helpful for the bin directory to be added to the PATH environment variable, when trying the official documentation's Command-line client examples with Docker.

# As-Is

```
root@59898138fb3a:/# dogma -v
bash: dogma: command not found
```

# To-Be

```
root@66eda13670d0:/# dogma -v
Central Dogma version 0.32.2-SNAPSHOT (d0f922e)
```


trustin:
Thank you!

PR #308, opened 2018-11-13
Update Javadoc based on customer feed back

trustin:
- Add descriptions about return values
- Mention the known issues with `watchRepository()`
- Other miscellaneous cleanup
- Undeprecate `getFile(...String pathPattern)` with fixed return type.
  - Suggested using `getFile(...Query)` instead


trustin:
Thanks, reviewers :)

PR #309, opened 2018-11-14
Specify maxparallel in chmod

minwoox:
Motivation:

Some OS limit the number of files which a `chmod` can change in once.
So we have to limit the number not to raise an error.


trustin:
Thanks, @minwoox !

PR #313, opened 2018-11-23
Copy dependencies for saml and shiro modules to lib directory

hyangtack:
Motivation:
Copying dependencies for saml and shiro modules to `dist/lib` was missing.
So a user cannot start a server with authentication config.

Modifications:
- Add their dependencies to the copy target.


hyangtack:
Thanks, reviewers!

PR #325, opened 2018-12-06
Adjust timeout for HTTP `watch` requests according to `wait` parameter

hyangtack:
Motivation:
Currently Central Dogma server does not adjust timeout for `watch` requests.
So every `watch` request would be timed out after the default timeout in the server even if the request contains `wait` value which is greater than the default timeout. It causes that the server responds `503 Service Unavailable` to the client.

Modifications:
- Adjust timeout for `watch` requests in `WatchRequestConverter`.
- Change the max timeout from `Long.MAX_VALUE` to `TimeUnit.DAYS.toMillis(1)` (1 day).
- Fix test cases.

Result:
- A user can adjust the timeout of a `watch` request.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/api/converter/WatchRequestConverter.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/api/converter/WatchRequestConverter.java
@@ -55,6 +55,8 @@ public final class WatchRequestConverter implements RequestConverterFunction {
             } else {
                 timeoutMillis = DEFAULT_TIMEOUT_MILLIS;
             }
+            // Update timeout according to the watch API specifications.
+            ctx.setRequestTimeoutMillis(timeoutMillis);
             return Optional.of(new WatchRequest(lastKnownRevision, timeoutMillis));
         }
         return Optional.empty();
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/api/converter/WatchRequestConverter.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/api/converter/WatchRequestConverter.java
@@ -29,6 +29,7 @@ import com.linecorp.armeria.common.HttpHeaderNames;
 import com.linecorp.armeria.server.ServiceRequestContext;
 import com.linecorp.armeria.server.annotation.RequestConverterFunction;
 import com.linecorp.centraldogma.common.Revision;
+import com.linecorp.centraldogma.internal.api.v1.WatchTimeout;
 
 /**
  * A request converter that converts to {@link WatchRequest} when the request contains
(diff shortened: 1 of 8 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/api/converter/WatchRequestConverter.java
```diff
@@ -55,6 +55,8 @@ public Object convertRequest(ServiceRequestContext ctx, AggregatedHttpMessage re
             } else {
                 timeoutMillis = DEFAULT_TIMEOUT_MILLIS;
             }
+            // Update timeout according to the watch API specifications.
+            ctx.setRequestTimeoutMillis(timeoutMillis);
```
trustin: - How about using `LongMath.saturatedAdd(timeoutMillis, ctx.requestTimeoutMillis())` ?
- How about enforcing upper bound to the timeout? We don't want to respect a client that specifies a timeout longer than 1 year. You might also want to update `CentralDogmaTimeoutScheduler` as well for consistent behavior.

server/src/test/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1TestBase.java
```diff
@@ -26,12 +26,20 @@
...
 
 public class ContentServiceV1TestBase {
 
     @Rule
-    public final CentralDogmaRule dogma = new CentralDogmaRule();
+    public final CentralDogmaRule dogma = new CentralDogmaRule() {
+        @Override
+        protected void configure(CentralDogmaBuilder builder) {
+            // Shorten the default request timeout here, in order to do a test that a watch request overrides
+            // the default request timeout.
```
trustin: How about mentioning `watchRepositoryTimeout()` and `watchFileTimeout()` so people knows where to look?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/converter/WatchRequestConverter.java
hyangtack: Added [WatchTimeout](https://github.com/line/centraldogma/pull/325/files#diff-cb3f0f652cf2ec37af3f5db13a135b0d) class. Please take a look again. :-)

hyangtack:
This PR changes the behavior when a user wants to set a timeout which exceeds the upper bound.
- AS-IS: the max value is `Long.MAX_VALUE`. If the timeout exceeds that, `0` will be set, which means 'unlimited'.
- TO-BE: the max value is `TimeUnit.DAYS.toMillis(365)`. If the timeout exceeds that, the max will be set.

hyangtack:
Had a chat with @trustin about that the max timeout. We agreed that 1 day is enough. So I changed it from 365 days to 1 day.

trustin:
Thanks, @hyangtack !

PR #328, opened 2018-12-11
Redirect to the first requested URL after SAML SSO performed

hyangtack:
Motivation:
Currently the web browser is redirected to the root(`/`). It'd be better if we can redirect it to the first requested URL.

Modifications:
- Add `ref` parameter when redirecting an unauthenticated request to `/link/auth/login`.
  - The parameter will be sent to an identity provider and get back via `relayState` parameter.


trustin:
Thanks, @hyangtack !

PR #330, opened 2018-12-11
Write a log when HTTP API returns 500 Internal Server Error

hyangtack:
Modifications:
- Write a log when the API returns `409 Conflict` with `ChangeConflictException` as well.


server/src/main/java/com/linecorp/centraldogma/server/internal/api/HttpApiUtil.java
```diff
@@ -153,6 +158,18 @@ private static HttpResponse newResponse0(HttpStatus status,
 
         node.put("message", message != null ? message : status.reasonPhrase());
 
+        // TODO(hyangtack) Need to introduce a new field such as 'stackTrace' in order to return
+        //                 the stack trace of the cause to the trusted client.
+        if (status == HttpStatus.INTERNAL_SERVER_ERROR) {
+            if (cause != null) {
+                logger.warn("Returning an internal server error with a cause:", cause);
+            } else if (message != null) {
+                logger.warn("Returning an internal server error with a message: {}", message);
```
trustin: - What if `message != null && cause != null`?
- Could we also add `RequestContext` in the beginning of the log messages?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/HttpApiUtil.java
```diff
@@ -153,6 +158,18 @@ private static HttpResponse newResponse0(HttpStatus status,
...
+            }
+        } else if (status == HttpStatus.CONFLICT && cause instanceof ChangeConflictException) {
+            logger.warn("Returning a conflict with a cause:", cause);
+        }
```
minwoox: Could you add when the exception is `QueryExecutionException`, please?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/HttpApiUtil.java
```diff
@@ -153,6 +158,18 @@ private static HttpResponse newResponse0(HttpStatus status,
...
 
+        // TODO(hyangtack) Need to introduce a new field such as 'stackTrace' in order to return
+        //                 the stack trace of the cause to the trusted client.
+        if (status == HttpStatus.INTERNAL_SERVER_ERROR) {
+            if (cause != null) {
+                logger.warn("Returning an internal server error with a cause:", cause);
+            } else if (message != null) {
+                logger.warn("Returning an internal server error with a message: {}", message);
+            }
+        } else if (status == HttpStatus.CONFLICT && cause instanceof ChangeConflictException) {
```
trustin: How about `!(cause instanceof RedundantChangeException)`?

PR #340, opened 2018-12-19
Cache in find() when getFile() is called with the query whose type is…

minwoox:
… the IDENTITY

Motivation:
Currently, getFile() is cached using `CacheableQueryCall`.
`mergeFiles()` calls CachingRespository.find() internally to cache the entries.
So if we change to cache the result in the one place using `CachableFindCall`, the chances of cache hit will be increased.

Modifications:
- Change to call `find()` in `CachingRepository.getOrNull()` when the query type is `IDENTITY`

Result:
- Better cache hit ratio


minwoox:
Thanks for reviewing!

PR #348, opened 2019-02-28
Remove commons-daemon from dependencies

### LINKED ISSUE #277: Replace commons-daemon and jsvc with something else.
The current version of jsvc does not support Java 9+

hyangtack:
Motivation:
The current version of `jsvc` does not support Java 9+.
If we can remove `commons-daemon` from dependencies, we would have less blocker for supporting latest Java version.

Modifications:
- Removed `commons-daemon` from dependencies.
- Removed `jsvc` binaries from `dist/src/bin`.
- The server now creates a pidfile when it starts up and deletes it when it is terminated.
  - `startup`/`shutdown` waits until the pidfile is created/deleted.
- Updated `Dockerfile`:
  - Using `openjdk:lastest` instead of `openjdk:8`.
  - Set the entry point in order to correctly handle a signal from a user.
    `docker stop` command, which sends `TERM` signal to the container, will work.

Result:
- Less dependencies.
- Support Java 9+.
- Closes #277


server/src/main/java/com/linecorp/centraldogma/server/Main.java
```diff
@@ -199,35 +207,60 @@ public void run() {
...
         if (main.state != State.INITIALIZED) {
             System.exit(1);
             return;
         }
 
-        main.start();
+        try {
+            main.start();
+        } catch (Throwable cause) {
+            logger.warn("Failed to start the Central Dogma:", cause);
```
trustin: I guess this should be logged at `ERROR` level because it will prevent Central Dogma from working at all. (Irrecoverable)

server/src/main/java/com/linecorp/centraldogma/server/Main.java
```diff
@@ -199,35 +206,50 @@ public void run() {
...
+                throw new IllegalStateException("Failed to create a PID file. A file already exists: " +
+                                                file.getPath());
+            }
+
+            final int pid = POSIXFactory.getPOSIX().getpid();
+            final Path temp = Files.createTempFile("central-dogma", ".tmp");
+            Files.write(temp, Integer.toString(pid).getBytes());
+            Files.move(temp, file.toPath(), StandardCopyOption.ATOMIC_MOVE);
+
+            logger.debug("A PID file created: " + file.getPath());
```
trustin: `logger.debug("A PID file has been created: {}", file);`

server/src/main/java/com/linecorp/centraldogma/server/Main.java
```diff
@@ -199,35 +206,50 @@ public void run() {
...
         }
 
-        @Override
-        public String[] getArguments() {
-            return args;
+        void destroy() throws IOException {
+            if (Files.deleteIfExists(file.toPath())) {
+                logger.debug("Successfully deleted the PID file: " + file.getPath());
```
trustin: Could use `{}` substitution.

PR #349, opened 2019-02-28
Show configured title on the web-based administrative console

hyangtack:
Motivation:
Currently, there's no easy way to know if the CD instance he or she is accessing belongs to BETA or RELEASE zone. We currently show the host name of the instance, but it's not clear enough. We could add a simple text value option whose value is appended to the web page title. e.g. `Central Dogma (BETA, {hostname})`

Modifications:
- Changed to send both of hostname and title from `/hostname` API.
  - The title will be displayed on the navigation bar of the web console.
- Updated the document about the configuration.

Result:
- Make web console users less confused.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
@@ -461,9 +461,7 @@ public class CentralDogma implements AutoCloseable {
                                           .setHeader(HttpHeaderNames.CONTENT_TYPE, MediaType.JSON_UTF_8)
                                           .build();
                 } catch (JsonProcessingException e) {
-                    return HttpFileBuilder.of(HttpData.ofUtf8("{\"hostname\":\"" + s.defaultHostname() + "\"}"))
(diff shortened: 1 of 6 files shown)


server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -447,9 +448,23 @@ private Server startServer(ProjectManager pm, CommandExecutor executor,
...
+                final String title = cfg.webAppTitle();
+                if (!isNullOrEmpty(title)) {
+                    b.put("title", title);
+                }
+                try {
+                    return HttpFileBuilder.of(HttpData.ofUtf8(Jackson.writeValueAsString(b.build())))
+                                          .setHeader(HttpHeaderNames.CONTENT_TYPE, MediaType.JSON_UTF_8)
+                                          .build();
+                } catch (JsonProcessingException e) {
+                    return HttpFileBuilder.of(HttpData.ofUtf8("{\"hostname\":\"" + s.defaultHostname() + "\"}"))
```
trustin: I guess we will never reach here. How about just throwing an `Error`?

hyangtack: Sounds good to me. :-)

PR #351, opened 2019-03-04
Use react-scripts instead of react-scripts-ts for webapp in server-auth

hyangtack:
Motivation:
Github.com told us that `webpack-dev-server` before `3.1.11` is vulnerable:
```
An issue was discovered in lib/Server.js in webpack-dev-server before 3.1.11. 
Attackers are able to steal developer's code because the origin of requests is 
not checked by the WebSocket server, which is used for HMR (Hot Module Replacement). 
Anyone can receive the HMR message sent by the WebSocket server via a 
ws://127.0.0.1:8080/ connection from any origin.
```
So we need to update it to the version of `3.1.11` or later.

Modifications:
- Removed `react-scripts-ts` which depends on `webpack-dev-server` and added `react-scripts` instead.
  - Referred to `https://vincenttunru.com/migrate-create-react-app-typescript-to-create-react-app/` for migration.

Result:
- Provides safer webapp.

### DIFF
--- a/server-auth/webapp/images.d.ts
+++ b/server-auth/webapp/images.d.ts
@@ -1,3 +1,0 @@
-declare module '*.svg'
-declare module '*.png'
-declare module '*.jpg'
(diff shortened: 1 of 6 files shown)


trustin:
I believe our webapp itself is unaffected by this issue, though.

trustin:
Thanks, @hyangtack !

PR #356, opened 2019-03-11
Use the new HTTP-based client for Spring Boot integration

trustin:
Motivation:

Spring Boot integration currently uses `LegacyCentralDogma` which has
been deprecated.

Modifications:

- Added `CentralDogmaSettings` so that a user can specify:
  - hosts or profile
  - access token
  - whether to use TLS
  - health check request interval
- Updated the documentation.

Result:

- A user can specify a Central Dogma client profile explicitly,
  without using Spring Boot profiles.
- A user can configure all properties of a Central Dogma client when
  using Spring Boot integration.
- (Breaking) A user must specify an access token if the target Central
  Dogma server has access control enabled.

### DIFF
--- a/client/java-spring-boot-autoconfigure/build.gradle
+++ b/client/java-spring-boot-autoconfigure/build.gradle
@@ -1,7 +1,8 @@
 dependencies {
     compile project(':client:java-armeria-legacy')
-    compile 'javax.validation:validation-api'
+    compileOnly 'javax.validation:validation-api'
     compile 'org.springframework.boot:spring-boot-starter'
+    annotationProcessor 'org.springframework.boot:spring-boot-configuration-processor'
 
     testCompile project(':client:java-armeria')
     testCompile 'org.springframework.boot:spring-boot-starter-test'
--- a/client/java-spring-boot-autoconfigure/src/main/java/com/linecorp/centraldogma/client/spring/CentralDogmaClientAutoConfiguration.java
+++ b/client/java-spring-boot-autoconfigure/src/main/java/com/linecorp/centraldogma/client/spring/CentralDogmaClientAutoConfiguration.java
@@ -18,6 +18,7 @@ package com.linecorp.centraldogma.client.spring;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.net.UnknownHostException;
+import java.util.List;
 import java.util.Optional;
 
 import javax.inject.Qualifier;
@@ -29,6 +30,7 @@ import org.springframework.beans.factory.config.BeanDefinition;
 import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
 import org.springframework.beans.factory.support.AbstractBeanDefinition;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Condition;
 import org.springframework.context.annotation.ConditionContext;
@@ -39,16 +41,19 @@ import org.springframework.core.env.Environment;
 import org.springframework.core.type.AnnotatedTypeMetadata;
 import org.springframework.core.type.MethodMetadata;
 
+import com.google.common.net.HostAndPort;
+
 import com.linecorp.armeria.client.ClientFactory;
 import com.linecorp.centraldogma.client.CentralDogma;
+import com.linecorp.centraldogma.client.armeria.ArmeriaCentralDogmaBuilder;
 import com.linecorp.centraldogma.client.armeria.ArmeriaClientConfigurator;
-import com.linecorp.centraldogma.client.armeria.legacy.LegacyCentralDogmaBuilder;
 
 /**
  * Spring bean configuration for {@link CentralDogma} client.
  */
 @Configuration
 @ConditionalOnMissingBean(CentralDogma.class)
+@EnableConfigurationProperties(CentralDogmaSettings.class)
 public class CentralDogmaClientAutoConfiguration {
 
     /**
@@ -76,15 +81,48 @@ public class CentralDogmaClientAutoConfiguration {
     @Bean
     public CentralDogma client(
             Environment env,
+            CentralDogmaSettings settings,
             @ForCentralDogma ClientFactory clientFactory,
             Optional<ArmeriaClientConfigurator> armeriaClientConfigurator) throws UnknownHostException {
 
-        return new LegacyCentralDogmaBuilder()
-                .clientFactory(clientFactory)
-                .profile(env.getActiveProfiles())
-                .clientConfigurator(cb -> armeriaClientConfigurator.ifPresent(
-                        configurator -> configurator.configure(cb)))
-                .build();
+        final ArmeriaCentralDogmaBuilder builder = new ArmeriaCentralDogmaBuilder();
+
+        builder.clientFactory(clientFactory);
+        builder.clientConfigurator(cb -> armeriaClientConfigurator.ifPresent(
+                configurator -> configurator.configure(cb)));
+
+        // Set health check interval.
+        final Long healthCheckIntervalMillis = settings.getHealthCheckIntervalMillis();
+        if (healthCheckIntervalMillis != null) {
+            builder.healthCheckIntervalMillis(healthCheckIntervalMillis);
+        }
+
+        // Enable or disable TLS.
+        final Boolean useTls = settings.getUseTls();
+        if (useTls != null) {
+            builder.useTls(useTls);
+        }
+
+        // Set profile or hosts.
+        final String profile = settings.getProfile();
+        final List<String> hosts = settings.getHosts();
+        if (profile != null) {
+            builder.profile(CentralDogmaClientAutoConfiguration.class.getClassLoader(), profile);
+        } else if (hosts != null) {
+            for (String h : hosts) {
+                final HostAndPort hostAndPort = HostAndPort.fromString(h);
+                if (hostAndPort.hasPort()) {
+                    builder.host(hostAndPort.getHost(), hostAndPort.getPort());
+                } else {
+                    builder.host(hostAndPort.getHost());
+                }
+            }
+        } else {
+            // Use the currently active Spring Boot profiles if neither profile nor hosts was specified.
+            builder.profile(env.getActiveProfiles());
+        }
+
+        return builder.build();
     }
 
     /**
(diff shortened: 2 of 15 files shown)


trustin:
@yuinacor @imasahiro PTAL

client/java-spring-boot-autoconfigure/src/test/java/com/linecorp/centraldogma/client/spring/CentralDogmaClientAutoConfigurationHostsTest.java
```diff
@@ -0,0 +1,59 @@
...
+@RunWith(SpringRunner.class)
+@SpringBootTest(classes = TestConfiguration.class)
+@ActiveProfiles({ "local", "useHosts", "confTest" })
+public class CentralDogmaClientAutoConfigurationHostsTest {
+    @Configuration
+    @Import(CentralDogmaClientAutoConfiguration.class)
+    public static class TestConfiguration {}
+
+    @Inject
+    CentralDogma client;
```
imasahiro: nit. Can add `private`?

site/src/sphinx/client-java.rst
```diff
@@ -515,17 +515,73 @@ client profile JAR again.
 Spring Boot integration
 -----------------------
 If you are using `Spring Framework <https://spring.io/>`_, you can inject :api:`com.linecorp.centraldogma.client.CentralDogma`
-client very easily.
+client very easily. First, add ``centraldogma-client-spring-boot-autoconfigure`` into your dependencies.
```
hyangtack: nit: How about using `centraldogma-client-spring-boot-starter`? They are same but we have `-starter` module too, so I think it would be better to use `-starter` module.

client/java-spring-boot-autoconfigure/src/main/java/com/linecorp/centraldogma/client/spring/CentralDogmaSettings.java
```diff
@@ -0,0 +1,189 @@
...
+public class CentralDogmaSettings {
+    /**
+     * The Central Dogma client profile name. If not specified, {@code "hosts"} must be specified.
+     */
+    @Nullable
+    private String profile;
+
+    /**
+     * The list of Central Dogma hosts. e.g. {@code "foo.example.com"} or {@code "foo.example.com:36462"}.
+     * If not specified, {@code "profile"} must be specified.
```
yuinacor: IMO, This restriction can be expressed by JSR validator.
If so, spring boot will print human friendly error report.

client/java-spring-boot-autoconfigure/src/main/java/com/linecorp/centraldogma/client/spring/CentralDogmaSettings.java
```diff
@@ -0,0 +1,189 @@
...
+ * <p>Note that {@code "healthCheckIntervalMillis"} and {@code "useTls"} are optional.</p>
+ */
+@ConfigurationProperties(prefix = "centraldogma")
+@Validated
+public class CentralDogmaSettings {
+    /**
+     * The Central Dogma client profile name. If not specified, {@code "hosts"} must be specified.
+     */
+    @Nullable
+    private String profile;
```
yuinacor: From my use case, consider not only spring boot profile but also centraldogma profile seems confuse.

Spring boot supports including existing profiles, so may good to depend on framework feature.

client/java-spring-boot-autoconfigure/src/main/java/com/linecorp/centraldogma/client/spring/CentralDogmaClientAutoConfiguration.java
```diff
@@ -76,15 +81,48 @@ public ClientFactory dogmaClientFactory() {
...
+                final HostAndPort hostAndPort = HostAndPort.fromString(h);
+                if (hostAndPort.hasPort()) {
+                    builder.host(hostAndPort.getHost(), hostAndPort.getPort());
+                } else {
+                    builder.host(hostAndPort.getHost());
+                }
+            }
+        } else {
+            // Use the currently active Spring Boot profiles if neither profile nor hosts was specified.
+            builder.profile(env.getActiveProfiles());
```
yuinacor: It magically works, but hard to recognize why it work...

Is this to support backward compatibility?

client/java-spring-boot-autoconfigure/src/main/java/com/linecorp/centraldogma/client/spring/CentralDogmaSettings.java
```diff
@@ -0,0 +1,189 @@
...
+     */
+    @Nullable
+    private String accessToken;
+
+    /**
+     * The number of milliseconds between each health-check requests to Central Dogma servers.
+     * {@code 0} disables the health check requests.
+     */
+    @Nullable
+    private Long healthCheckIntervalMillis;
```
yuinacor: How about support default value or `@NotNull`?

client/java-spring-boot-autoconfigure/src/main/java/com/linecorp/centraldogma/client/spring/CentralDogmaSettings.java
yuinacor: Good to see.
https://spring.io/blog/2018/03/28/property-binding-in-spring-boot-2-0#origin-support

client/java-spring-boot-autoconfigure/build.gradle
```diff
@@ -1,7 +1,8 @@
 dependencies {
     compile project(':client:java-armeria-legacy')
-    compile 'javax.validation:validation-api'
+    compileOnly 'javax.validation:validation-api'
```
yuinacor: Looks `centraldogma` uses gradle `5.2.1`.
Then, how about `api` instead of `compileOnly`?

https://docs.gradle.org/current/userguide/java_library_plugin.html

client/java-spring-boot-autoconfigure/src/test/java/com/linecorp/centraldogma/client/spring/CentralDogmaClientAutoConfigurationHostsTest.java
trustin: Sure. Fixes them all :-)

client/java-spring-boot-autoconfigure/src/main/java/com/linecorp/centraldogma/client/spring/CentralDogmaClientAutoConfiguration.java
trustin: Yes, and documented in `client-java.rst`.

client/java-spring-boot-autoconfigure/src/main/java/com/linecorp/centraldogma/client/spring/CentralDogmaSettings.java
trustin: Default value is handled in `CentralDogmaBuilder` implementations so I did not want to duplicate it here.

client/java-spring-boot-autoconfigure/build.gradle
trustin: line/gradle-scripts does not support the new configuration model yet, so we have to stick to the old model.

client/java-spring-boot-autoconfigure/src/main/java/com/linecorp/centraldogma/client/spring/CentralDogmaClientAutoConfiguration.java
trustin: Let me add some logging so that a user knows where the profile name came from.

site/src/sphinx/client-java.rst
```diff
@@ -515,17 +515,73 @@ client profile JAR again.
...
+If you prefer using client profiles as described in :ref:`using_client_profiles`, use the ``profile`` property:
+
+.. code-block:: yaml
+
+    centraldogma:
+      profile: beta
+
+If neither ``hosts`` nor ``profile`` property is specified, currently active
+`Spring Boot profile <https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-profiles.html>`_
+will be used as the client profile. When more than one Spring Boot profile are active, the last matching one
```
minwoox: When ... is active? Because the subject is when not the profiles?

site/src/sphinx/client-java.rst
trustin: You're right. Fixed.

client/java-spring-boot-autoconfigure/src/main/java/com/linecorp/centraldogma/client/spring/CentralDogmaSettings.java
trustin: I agree that the name 'profile' may confuse some users, but it is indeed a 'profile', so let me add some note about this in the `client-java.rst` rather than renaming it.

client/java-spring-boot-autoconfigure/src/main/java/com/linecorp/centraldogma/client/spring/CentralDogmaSettings.java
```diff
@@ -0,0 +1,148 @@
...
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.validation.annotation.Validated;
+
+import com.google.common.base.MoreObjects;
+import com.google.common.collect.ImmutableList;
+
+/**
+ * Settings for a Central Dogma client.
+ *
+ * <h3>Example with {@code "host"} option</h3>
```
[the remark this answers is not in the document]
trustin: Fixed

site/src/sphinx/client-java.rst
```diff
@@ -515,17 +515,73 @@ client profile JAR again.
...
+
+.. code-block:: yaml
 
-A new :api:`com.linecorp.centraldogma.client.CentralDogma` client will be created and injected using your
-`Spring Boot profile <https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-profiles.html>`_.
-When more than one profile is active, the last matching one will be used from ``/centraldogma-profiles-test.json``
-or ``/centraldogma-profiles.json``.
+    centraldogma:
+      profile: staging
+      use-tls: true
```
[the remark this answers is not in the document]
trustin: Ah, yes. Fixed :-)

client/java-spring-boot-autoconfigure/src/main/java/com/linecorp/centraldogma/client/spring/CentralDogmaSettings.java
trustin: Just added a check for `hosts` and `profile` in the auto configuration implementation.

trustin:
Thank you, reviewers :smile:

PR #358, opened 2019-03-12
[WIP] Pluggable service support

### LINKED ISSUE #353: Pluggable service running on the leader node
It would be useful if we can provide a pluggable service which is started when the node is elected as a leader and stopped when its role is changed to a follower. 
MirroringService is one of the candidates which can be implemented as a plugin.

hyangtack:
Motivation:
It would be useful if we can provide a pluggable service which is started when the node is elected as a leader and stopped when its role is changed to a follower.
MirroringService is one of the candidates which can be implemented as a plugin.

Modifications:
- Added `c.l.c.server.pluggable` and `c.l.c.server.internal.pluggable` packages.
  - Added `PluggableService` interface to start/stop the plug-in service.
  - Added `LeaderService` interface to start/stop the plug-in service when the node has taken the leadership or is about to release the leadership.

Result:
- Closes #353

To-Dos:
- Migrate `MirroringService` to `LeaderService`.
- Decide whether to remove `PluggableService` or not.
  - Currently, we only need `LeaderService`, but is there any requirement about the service which is running with the same life-cycle of a node?


server/src/main/java/com/linecorp/centraldogma/server/pluggable/LeaderService.java
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * Invoked when the node is about to release the leadership.
+     */
+    CompletionStage<Void> onReleaseLeadership();
+
+    /**
+     * Returns a {@link PluggableService} which wraps this {@link LeaderService}.
+     */
+    default PluggableService toPluggableService() {
```
trustin: I find it may be confusing because the life cycle of `PluggableService` and `LeaderService` are different from each other.

Perhaps we could:

- define some common interface between `PluggableService` and `LeaderService` instead of adapting, or
- add a method like `target()` to `PluggableService` and remove `LeaderService`?

For example:

```java
public interface PluggableService {
    PluggableServiceTarget target();
    CompletionStage<Void> start();
    CompletionStage<Void> stop();
}

public enum PluggableServiceTarget {
    ALL_REPLICAS,
    LEADER_ONLY
}
```

server/src/main/java/com/linecorp/centraldogma/server/internal/pluggable/PluggableServices.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.util.List;
+import java.util.ServiceLoader;
+import java.util.function.Function;
+
+import com.google.common.collect.ImmutableList.Builder;
+
+import com.linecorp.centraldogma.server.pluggable.PluggableService;
+
+public final class PluggableServices {
```
trustin: Just curious, did you choose this name because there is `Plugin` already? If so, you can ignore it and name this as `Plugin`.

hyangtack: Yes, right! Let me rename it. :-)

server/src/main/java/com/linecorp/centraldogma/server/pluggable/LeaderService.java
hyangtack: Sounds good to me. Let me handle them. :-)

hyangtack:
Will send a PR again once it's ready for review. :-)

PR #365, opened 2019-03-19
Update dependencies

trustin:
- Armeria 0.81.0 -> 0.82.0
  - Made use of `SystemInfo.hostname()` and `pid()`.
  - Added the constant for `Prefer` HTTP header name because it was
    removed from `HttpHeaderNames` by mistake.
- jGit 5.2.1 -> 5.3.0
  - Set `false` to `commit.gpgSign` so that a commit does not fail when
    a user's `.gitconfig` has set `true` to `commit.gpgSign`.
- Removed JNR in favor of Armeria's `SystemUtil`.
- Build:
  - json-unit 2.4.0 -> 2.5.0
  - AssertJ 3.12.1 -> 3.12.2
  - Mockito 2.25.0 -> 2.5.1
  - gradle-node-plugin 1.2.0 -> 1.3.1


server/src/main/java/com/linecorp/centraldogma/server/internal/api/converter/WatchRequestConverter.java
```diff
@@ -31,25 +32,31 @@
...
+    // TODO(trustin): Replace with HttpHeaderNames.PREFER.
+    @VisibleForTesting
+    static final AsciiString HEADER_NAME_PREFER = HttpHeaderNames.of("prefer");
+
     private static final long DEFAULT_TIMEOUT_MILLIS = TimeUnit.SECONDS.toMillis(120);
 
     /**
      * Converts the specified {@code request} to {@link Optional} which contains {@link WatchRequest} when
-     * the request has {@link HttpHeaderNames#IF_NONE_MATCH}. {@link Optional#EMPTY} otherwise.
+     * the request has {@link HttpHeaderNames#IF_NONE_MATCH}. {@link Optional#empty()} otherwise.
```
minwoox: Thanks!

PR #367, opened 2019-03-21
Send a JSON response for authentication failures

trustin:
Motivation:

Central Dogma server sends a response without content when authorization
fails. We could respond with JSON so that the clients knows the exact
cause of the failure.

Modifications:

- Use `HttpApiUtil` to build an authentication failure response.
- Do not use `HttpStatus.reasonPhrase()` when exception message is
  missing, because it often mismatches with the actual exception type,
  e.g. `ShuttingDownException` having `Internal Server Error` message.

Result:

- Proper response even on authentication failure
- `message` field in an error response is empty if an exception does not
  have a message.


trustin:
Thanks for reviewing.

PR #372, opened 2019-03-25
Do not traverse the whole tree when comparing two trees.

trustin:
Motivation:

Unless a user requests a full diff, we do not really need to traverse
the whole tree. We can stop comparing as soon as we found what we're
looking for.

Modifications:

- Fork jGit's `DiffFormatter` and `DiffEntry` so that we can stop the
  tree traversal loop while comparing two trees.
- Stop the tree traversal loop in `blockingFindLatestRevision()` and
  `notifyWatchers`.

Result:

- Watching takes less time in a repository with many files.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/DiffScanner.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/DiffScanner.java
@@ -61,8 +61,8 @@ package com.linecorp.centraldogma.server.internal.storage.repository.git;
 import java.io.IOException;
 import java.util.function.Predicate;
 
+import org.eclipse.jgit.attributes.Attribute;
 import org.eclipse.jgit.dircache.DirCacheIterator;
-import org.eclipse.jgit.lib.AbbreviatedObjectId;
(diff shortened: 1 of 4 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/DiffScanner.java
```diff
@@ -133,38 +128,46 @@ private static boolean scan(ObjectReader reader, AbstractTreeIterator a, Abstrac
...
+                final DiffEntry entry = new DiffEntry(path, DEV_NULL, diffAttribute, oldMode, newMode,
+                                                      ChangeType.DELETE, oldId, newId);
                 if (matcher.test(entry)) {
                     return true;
                 }
-            } else if (!entry.oldId.equals(entry.newId)) {
-                entry.changeType = ChangeType.MODIFY;
-                if (sameType(entry.oldMode, entry.newMode)) {
+            } else if (!oldId.equals(newId)) {
+                final DiffEntry entry = new DiffEntry(path, path, diffAttribute, oldMode, newMode,
```
hyangtack: Could we make this entry only when the `sameType(oldMode,newMode)` is true?

PR #375, opened 2019-03-27
Cache the comparison result between two Git trees / Fail the timed-ou…

trustin:
…t requests fast

Related commits:

- 47764e9f098f4abc2de811962bdc3294e8540c50
- 422cec92ab1b4999e52277097883bacd4c15292e
- 674d4ab03c0aef95448d64e565acd9b4e21339e5

Motivation:

- Downgrading jGit from 5.3.0 to 5.2.1 did not help.
- We should reuse the `RepositoryCache` for caching the comparison result
  between two Git trees.
- We should fail more timed-out requests fast.

Modifications:

- Updated jGit from 5.2.1 to 5.3.0.
- Moved `server.internal.storage.cache.RepositoryCache` and `CacheableCall`
  up to `server.internal.storage`.
- `GitRepository` now depends on `RepositoryCache` optionally.
- Added `CacheableCompareTreesCall` and took advantage of the cache in
  `GitRepository.findLatestRevision()`.
- Removed the forked `DiffScanner` and `DiffEntry`.
- Added `FailFastUtil` and used it in most `GitRepository` operations.
- Modified `GitRepository.commit0()` to return the `List<DiffEntry>`
  produced during validation, so that `notifyWatchers()` does not need
  to compare the trees but reuse what's returned by `commit0()`.
- Used `StampedLock` instead of `ReadWriteLock` to make `watch()`
  completely async.
  - Added `ReadLockStamp` to implement reentrancy.
- Miscellaneous:
  - Added more test cases to `WatchTest`.
  - Added `RepositoryCache.get()` and `getIfPresent()` to reduce unsafe
    casts on the caller side.
  - Removed unnecessary `try { ... } catch (Exception) { ... }` blocks.
  - Renamed `GitRepository.toTreeId()` to `toTree()` and changed it to
    return `RevTree` instead of `ObjectId`.
  - De-duplicated anonymous `PathEdit` implementations into inner classes.

Result:

- The comparison result between two Git trees are now stored in
  `RepositoryCache` which can be configured by a user.
- All `GitRepository` operations fail fast when timed out.


server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/ReadLockStamp.java
```diff
@@ -0,0 +1,76 @@
...
+
+        checkArgument(readLockStamp.lock == lock, "mismatching lock");
+        final int oldDepth = depthUpdater.getAndIncrement(readLockStamp);
+        checkState(oldDepth > 0, "readLockStamp unlocked already");
+        return readLockStamp;
+    }
+
+    private final StampedLock lock;
+    @VisibleForTesting
+    final long readLockStamp;
```
minwoox: How about naming as `value`, `stamp`, `stampValue` or `readLockStampValue`?
Because it equals to the name of the class, it's a little confusing while I'm looking at the test code. 😄

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
```diff
@@ -651,43 +651,66 @@ private static Commit toCommit(RevCommit revCommit) {
...
-            readUnlock();
+    private static TreeFilter pathPatternFilterOrTreeFilter(@Nullable String pathPattern) {
+        final TreeFilter filter;
+        if (pathPattern != null) {
+            final PathPatternFilter pathPatternFilter = PathPatternFilter.of(pathPattern);
+            filter = pathPatternFilter.matchesAll() ? TreeFilter.ALL : pathPatternFilter;
+        } else {
+            filter = TreeFilter.ALL;
         }
+        return filter;
```
minwoox: How about simplifying this?
```java
private static TreeFilter pathPatternFilterOrTreeFilter(@Nullable String pathPattern) {
    if (pathPattern == null) {
        return TreeFilter.ALL;
    }
    ...
```

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
```diff
@@ -1292,109 +1262,112 @@ static void doRefUpdate(org.eclipse.jgit.lib.Repository jGitRepository, RevWalk
...
-                    commitWatchers.add(normLastKnownRevision, pathPattern, future);
-                }
+                final CompletableFuture<Revision> future = new CompletableFuture<>();
+                commitWatchers.add(normLastKnownRevision, pathPattern, future);
+                return future;
             } finally {
-                readUnlock();
+                readLockStamp.getAndSet(null).unlock();
+            }
+        }).whenComplete((unused1, unused2) -> {
```
hyangtack: nit: How about using `handle` instead of `whenComplete` for performance?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
trustin: Good idea. Fixed

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
trustin: Switched to `handle()`

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/ReadLockStamp.java
trustin: Renamed to `value`

trustin:
Thanks for reviewing.

PR #378, opened 2019-04-03
Add `revision` property to `EntryDto`

### LINKED ISSUE #310: Include the revision number in the getFile response
We currently do not return the revision of an entry. The reasoning behind this behavior is that the response would contain duplicate 'revision' fields if a user requested more than one entry. Because of that, there's a problem that a user could not notice quickly which version of a file is adapted to his/her service.
To solve this issue, there're two ways as @trustin suggested:
- Allow some level of duplicate information.
- Separate the API into two which are for getting a single file and multiple files. 

I personaly prefer the second one because I don’t want to include duplicate information and I think there’s not so much people who are using it, so it’s not a big deal.

trustin:
Motivation:

We currently do not return the revision of an entry. The reasoning
behind this behavior is that the response would contain duplicate
'revision' fields if a user requested more than one entry. Because of
that, there's a problem that a user could not notice quickly which
version of a file is loaded into his/her service.

Modifications:

- Added `revision` property to `EntryDto`
- Miscellaneous:
  - Used `HttpHeaderNames.PREFER` which reappeared in Armeria 0.83.0
  - Used `HttpData.toStringUtf8()` instead of `.content().toStringUtf8()`

Result:

- Closes #310

### DIFF
--- a/client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogma.java
+++ b/client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogma.java
@@ -99,12 +99,7 @@ import com.linecorp.centraldogma.internal.Jackson;
 import com.linecorp.centraldogma.internal.Util;
 import com.linecorp.centraldogma.internal.api.v1.WatchTimeout;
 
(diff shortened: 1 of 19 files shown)


minwoox:
👍

PR #384, opened 2019-04-05
Wait until process is terminated in stop script / Add visual feedback while waiting

trustin:
Motivation:

Our `shutdown` shell script sometimes exists even before the JVM
process is fully terminated, which can cause race conditions when
restarting a server.

Also, it'd be nice if there is some visual feedback which tells a user
that the shell scripts are waiting for something.

Modifivations:

- Updated `shutdown` script to wait until the JVM process is fully
  terminated.
- Updated `startup` and `shutdown` script to print a dot (.) every
  second while waiting for a process or a PID file.
- Miscellaneous:
  - Added back `APP_LOGBACK_APPENDER`

Result:

- Server restart does not fail anymore.
- A user can wait more patiently.

### DIFF
--- a/dist/src/bin/common.sh
+++ b/dist/src/bin/common.sh
@@ -2,36 +2,39 @@
 #
 # This script checks if the following environment variables exist and sets their default values if necessary.
 # -----------------------------------------------------------------------------------------------------------
-# APP_HOME        (Optional) The base directory of the application.
-#                 If not set, the parent directory of this script is used.
+# APP_HOME             (Optional) The base directory of the application.
+#                      If not set, the parent directory of this script is used.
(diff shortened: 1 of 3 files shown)


trustin:
Example output:

```
Starting up centraldogma .. 
Waiting for PID file creation up to 60 second(s) ..... done 
Started up centraldogma successfully: 26669 

Shutting down centraldogma (26669) .. 
Waiting for PID file deletion up to 60 second(s) ...... done 
Waiting for process 26669 to terminate up to 60 second(s) .. done 
Shut down centraldogma successfully. 
```

trustin:
See #386 instead.

PR #386, opened 2019-04-05
Wait until process is terminated in stop script / Add visual feedback while waiting

trustin:
Motivation:

Our `shutdown` shell script sometimes exists even before the JVM
process is fully terminated, which can cause race conditions when
restarting a server.

Also, it'd be nice if there is some visual feedback which tells a user
that the shell scripts are waiting for something.

Modifivations:

- Updated `shutdown` script to wait until the JVM process is fully
  terminated.
- Updated `startup` and `shutdown` script to print a dot (.) every
  second while waiting for a process or a PID file.
- Miscellaneous:
  - Added back `APP_LOGBACK_APPENDER`

Result:

- Server restart does not fail anymore.
- A user can wait more patiently.

### DIFF
--- a/dist/src/bin/shutdown
+++ b/dist/src/bin/shutdown
@@ -21,21 +21,56 @@ else
   exit 100
 fi
 
+
 # Shut the application down
 set +e
 eecho "Shutting down $APP_NAME ($(cat "$APP_PID_FILE")) .."
 
 kill -TERM $PID
 
+# Wait until the PID file is removed by the application.
+WAITING=1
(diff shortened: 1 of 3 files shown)


dist/src/bin/shutdown
```diff
@@ -21,21 +21,56 @@ else
   exit 100
 fi
 
+
```
minwoox: nit: double empty lines?

trustin: Fixed :-)

PR #388, opened 2019-04-09
Fix a dead lock by reverting back to `ReentrantReadWriteLock`

trustin:
Motivation:

Previous fully asynchronous `GitRepository` which uses a `StampedLock`
can cause a dead lock because a pending write lock request can prevent
other read lock requests happened after the write lock request.

Modifications:

- Used `ReentrantReadWriteLock` again.
- Made `GitRepository.findLatestRevision()` delegate its job to
  `blockingFindLatestRevision()`.
  - Changed the caching logic of `CacheableCompareTreesCall` to build a
    blocking version of this operation.
  - Added `CacheableCall.coarseGrainedLock()` so that we can acquire a
    reasonably less-contended lock for a key.
- Reverted `GitRepository.watch()` so that a watch operation uses
  `blockingFindLatestRevision()` instead of its old asynchronous version.
- Removed `ReadLockStamp`

Result:

- No more dead lock.
- Less efficient caching of `compareTrees` operation.


trustin:
Thanks for reviewing :-)

PR #391, opened 2019-04-12
Make client-side watch operations better-documented and less error-prone.

trustin:
Motivation:

It seems like our users sometimes specify way too small timeout value
when they use `watch*()` operations or `Watcher`s. A small timeout value
can lead to a failure to fetch the initial value, preventing the startup
of the user application when it relies entirely on Central Dogma.

Modifications:

- Added `CentralDogma.watchFile/Repository()` variants that do not
  require the `timeoutMillis` parameter.
  - Explicitly recommended using the variants without timeout parameter.
- Added a `Watcher.awaitInitialValue()` variant that falls back to the
  default value when it failed to retrieve the initial value from the
  server.
- Added the 'Preparing for unavailability' subsection.
- Miscellaneous:
  - Cleaned up the example snippets.
  - Improved the indentation of the example snippets.

Result:

- Watch operations are better-documented.
- If a user reads the documentation, he or she will make less mistakes
  and an informed decision about what to do when the server is
  unavailable.

### DIFF
--- a/client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java
+++ b/client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java
@@ -427,7 +427,8 @@ public interface CentralDogma {
      */
     default CompletableFuture<Revision> watchRepository(String projectName, String repositoryName,
                                                         Revision lastKnownRevision, String pathPattern) {
-        return watchRepository(projectName, repositoryName, lastKnownRevision, pathPattern, 60000);
+        return watchRepository(projectName, repositoryName, lastKnownRevision, pathPattern,
+                               WatchConstants.DEFAULT_WATCH_TIMEOUT_MILLIS);
     }
 
     /**
@@ -455,7 +456,8 @@ public interface CentralDogma {
      */
     default <T> CompletableFuture<Entry<T>> watchFile(String projectName, String repositoryName,
                                                       Revision lastKnownRevision, Query<T> query) {
-        return watchFile(projectName, repositoryName, lastKnownRevision, query, 60000);
+        return watchFile(projectName, repositoryName, lastKnownRevision, query,
+                         WatchConstants.DEFAULT_WATCH_TIMEOUT_MILLIS);
     }
 
     /**
--- a/client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java
+++ b/client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java
@@ -419,7 +419,23 @@ public interface CentralDogma {
 
     /**
      * Waits for the files matched by the specified {@code pathPattern} to be changed since the specified
-     * {@code lastKnownRevision}.
+     * {@code lastKnownRevision}. If no changes were made within 1 minute, the returned
+     * {@link CompletableFuture} will be completed with {@code null}.
+     *
+     * @return the latest known {@link Revision} which contains the changes for the matched files.
+     *         {@code null} if the files were not changed for 1 minute since the invocation of this method.
+     */
+    default CompletableFuture<Revision> watchRepository(String projectName, String repositoryName,
+                                                        Revision lastKnownRevision, String pathPattern) {
+        return watchRepository(projectName, repositoryName, lastKnownRevision, pathPattern, 60000);
+    }
+
+    /**
+     * Waits for the files matched by the specified {@code pathPattern} to be changed since the specified
+     * {@code lastKnownRevision}.  If no changes were made within the specified {@code timeoutMillis}, the
+     * returned {@link CompletableFuture} will be completed with {@code null}. It is recommended to specify
+     * the largest {@code timeoutMillis} allowed by the server. If unsure, use
+     * {@link #watchRepository(String, String, Revision, String)}.
      *
      * @return the latest known {@link Revision} which contains the changes for the matched files.
      *         {@code null} if the files were not changed for {@code timeoutMillis} milliseconds
@@ -431,7 +447,23 @@ public interface CentralDogma {
 
     /**
      * Waits for the file matched by the specified {@link Query} to be changed since the specified
-     * {@code lastKnownRevision}.
+     * {@code lastKnownRevision}. If no changes were made within 1 minute, the returned
+     * {@link CompletableFuture} will be completed with {@code null}.
+     *
+     * @return the {@link Entry} which contains the latest known {@link Query} result.
+     *         {@code null} if the file was not changed for 1 minute since the invocation of this method.
+     */
+    default <T> CompletableFuture<Entry<T>> watchFile(String projectName, String repositoryName,
+                                                      Revision lastKnownRevision, Query<T> query) {
+        return watchFile(projectName, repositoryName, lastKnownRevision, query, 60000);
+    }
+
+    /**
+     * Waits for the file matched by the specified {@link Query} to be changed since the specified
+     * {@code lastKnownRevision}. If no changes were made within the specified {@code timeoutMillis}, the
+     * returned {@link CompletableFuture} will be completed with {@code null}. It is recommended to specify
+     * the largest {@code timeoutMillis} allowed by the server. If unsure, use
+     * {@link #watchFile(String, String, Revision, Query)}.
      *
      * @return the {@link Entry} which contains the latest known {@link Query} result.
      *         {@code null} if the file was not changed for {@code timeoutMillis} milliseconds
--- a/client/java/src/main/java/com/linecorp/centraldogma/client/Latest.java
+++ b/client/java/src/main/java/com/linecorp/centraldogma/client/Latest.java
@@ -16,7 +16,11 @@
 
 package com.linecorp.centraldogma.client;
 
+import static java.util.Objects.requireNonNull;
+
 import java.util.Objects;
+
+import javax.annotation.Nullable;
 
 import com.google.common.base.MoreObjects;
 
@@ -30,13 +34,14 @@ import com.linecorp.centraldogma.common.Revision;
 public final class Latest<U> {
 
     private final Revision revision;
+    @Nullable
     private final U value;
 
     /**
      * Creates a new instance with the specified {@link Revision} and value.
      */
-    public Latest(Revision revision, U value) {
-        this.revision = revision;
+    public Latest(Revision revision, @Nullable U value) {
+        this.revision = requireNonNull(revision, "revision");
         this.value = value;
     }
 
@@ -50,6 +55,7 @@ public final class Latest<U> {
     /**
      * Returns the latest known value.
      */
+    @Nullable
     public U value() {
         return value;
     }
@@ -63,12 +69,12 @@ public final class Latest<U> {
             return false;
         }
         final Latest<?> that = (Latest<?>) o;
-        return Objects.equals(revision, that.revision) && Objects.equals(value, that.value);
+        return revision.equals(that.revision) && Objects.equals(value, that.value);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hashCode(revision) * 31 + Objects.hashCode(value);
+        return revision.hashCode() * 31 + Objects.hashCode(value);
     }
 
     @Override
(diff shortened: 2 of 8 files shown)


trustin:
@mauhiz, PTAL

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java
```diff
@@ -419,7 +419,23 @@
...
-     * {@code lastKnownRevision}.
+     * {@code lastKnownRevision}. If no changes were made within 1 minute, the returned
+     * {@link CompletableFuture} will be completed with {@code null}.
+     *
+     * @return the latest known {@link Revision} which contains the changes for the matched files.
+     *         {@code null} if the files were not changed for 1 minute since the invocation of this method.
+     */
+    default CompletableFuture<Revision> watchRepository(String projectName, String repositoryName,
+                                                        Revision lastKnownRevision, String pathPattern) {
+        return watchRepository(projectName, repositoryName, lastKnownRevision, pathPattern, 60000);
```
mauhiz: Let's create a DEFAULT_WATCH_TIMEOUT_MILLIS constant somewhere?

client/java/src/main/java/com/linecorp/centraldogma/client/Latest.java
```diff
@@ -63,12 +69,12 @@ public boolean equals(Object o) {
...
+        return revision.equals(that.revision) && Objects.equals(value, that.value);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hashCode(revision) * 31 + Objects.hashCode(value);
+        return revision.hashCode() * 31 + Objects.hashCode(value);
     }
 
     @Override
```
minwoox: nit: `omitNullValues()`?

site/src/sphinx/client-java.rst
```diff
@@ -318,21 +322,47 @@ the process. The client library provides an easy way to watch a file:
...
+    try {
+        Latest<JsonNode> latest = watcher.awaitInitialValue(1, MINUTES);
+        System.err.printf("Initial: %s at %s%n", latest.value(), latest.revision());
+        initialValue = latest.value();
+    } catch (TimeoutException e) {
+        System.err.printf("Default: %s%n", someDefaultValue);
+        initialValue = someDefaultValue;
+    }
+
+
```
minwoox: nit: Could remove one line?

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
```diff
@@ -58,9 +61,21 @@
...
 
     /**
-     * Waits for the initial value to be available.
+     * Waits for the initial value to be available. Specify the default value with
+     * {@link #awaitInitialValue(long, TimeUnit, Object)} or make sure to handle
+     * a {@link TimeoutException} properly if the initial value must be available
+     * even when the server is unavailable.
+     *
+     * @param timeout the maximum amount of time to wait for the initial value. It is recommended to use a
+     *                value larger than 30 seconds at least so that the client can retry at least a few times
```
mauhiz: Where does this `30s` value come from? CD server or client settings?
Asking because it sounds extremely big for a modern system.
What would you recommend our users if they need to shorten it?

trustin:
image

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
trustin: A user could specify a short timeout and a default value? The recommended timeout is to ensure that there are at least 4-5 retries in case of unavailability, because the client does exponential backoff all the way up from 2 seconds to 1 minute, i.e. t0 -> t2 -> t4 -> t8 -> t16. I'm not sure if we need to expose this level of detail, though.

mauhiz: > the client does exponential backoff all the way up from 2 seconds to 1 minute

Thanks for the info. Just FMI, is this backoff (initial and factor) configurable?

trustin: Updated the Javadoc and client library manual to reflect your feed back. Please check again. :bowing_man: 

> Just FMI, is this backoff (initial and factor) configurable?

The backoff is currently hard-coded. Do you think there's a case a user would want to use non-default backoff? I'm somewhat concerned about cargo-culting too short retry interval.

client/java/src/main/java/com/linecorp/centraldogma/client/Latest.java
trustin: I think we should not omit null in this case because we should tall the caller that the value is null.

site/src/sphinx/client-java.rst
trustin: Fixed

client/java/src/main/java/com/linecorp/centraldogma/client/Latest.java
```diff
@@ -30,13 +34,14 @@
...
+    @Nullable
     private final U value;
 
     /**
      * Creates a new instance with the specified {@link Revision} and value.
      */
-    public Latest(Revision revision, U value) {
-        this.revision = revision;
+    public Latest(Revision revision, @Nullable U value) {
+        this.revision = requireNonNull(revision, "revision");
```
imasahiro: `value` is Nullable but requireNonNull? Could you remove this validation?

trustin:
image

client/java/src/main/java/com/linecorp/centraldogma/client/Latest.java
trustin: This `requireNonNull` is for `revision`. :sweat_smile:

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
mauhiz: > Do you think there's a case a user would want to use non-default backoff?

CD is used for many purposes and I don't dare say I know them all :).

I could see a use case for disabling the retries (failing immediately and letting end-user manage the retry (or not) at his own convenience.
In our flagship server project, we have for instance
- sampling configurations : totally fine with immediate failure/fallback default value, we can peek again when CD goes back online
- server repertory : can't boot without it
If the end-user's use case is all of the first kind, then the time spent sleeping between retries is a waste. WDYT?

trustin: > sampling configurations

In that case, you don't really need to use `awaitInitialValue()` or you can use a very short timeout, and then peek the latest value using `latestValue(defaultValue)` which never blocks. If the server is unavailable at startup, it will start with the default value, but once the server is back up, `latestValue()` will start to return the latest value retrieved from the server.

trustin: Let me add some note about this case as well in `client-java.rst`. Thanks for a good idea.

mauhiz: Thanks! It sounds like our usage of `awaitInitialValue` in both use cases (mostly from copy-pasting, I dare say) was a bit wrong. /cc @imasahiro

trustin:
image

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
trustin: Please check the updated docs :-)

mauhiz: Thanks, LGTM

client/java/src/main/java/com/linecorp/centraldogma/internal/client/AbstractWatcher.java
```diff
@@ -222,8 +221,8 @@ private void doWatch(int numAttemptsSoFar) {
         }
 
         final Revision lastKnownRevision = latest != null ? latest.revision() : Revision.INIT;
-        final CompletableFuture<Latest<T>> f = doWatch(client, projectName, repositoryName, lastKnownRevision,
-                                                       WATCH_TIMEOUT_MILLIS);
+        final CompletableFuture<Latest<T>> f = doWatch(client, projectName, repositoryName, lastKnownRevision
+        );
```
minwoox: This can't be one line? if so, How about splitting before `lastKnownRevision`?

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
```diff
@@ -58,9 +61,27 @@
...
     /**
-     * Waits for the initial value to be available.
+     * Waits for the initial value to be available. Specify the default value with
+     * {@link #awaitInitialValue(long, TimeUnit, Object)} or make sure to handle
+     * a {@link TimeoutException} properly if the initial value must be available
+     * even when the server is unavailable.
+     *
+     * @param timeout the maximum amount of time to wait for the initial value. Note that timeout is basically
+     *                a trade-off. If you specify a smaller timeout, this method will take less time even if
+     *                the server is note responsive, at the risk of {@link TimeoutException}. If you specify
```
minwoox: nit: note -> not?

client/java/src/main/java/com/linecorp/centraldogma/client/WatchConstants.java
```diff
@@ -0,0 +1,26 @@
...
+ * under the License.
+ */
+package com.linecorp.centraldogma.client;
+
+import java.util.concurrent.TimeUnit;
+
+final class WatchConstants {
+
+    static final long DEFAULT_WATCH_TIMEOUT_MILLIS = TimeUnit.MINUTES.toMillis(1);
+    static final int RECOMMENDED_AWAIT_TIMEOUT_SECONDS = 20;
```
jongyeol: How many times can the CD-client retry a request with the default retry policy within 20 seconds? If it's greater than or equal to 3 including the first request, then it seems acceptable.

trustin: - At t0 + 0s: initial try
- At t0 + 2s: 1st retry
- At t0 + 6s: 2nd retry
- At t0 + 14s: 3rd retry
- At t0 + 30s: 4th retry

i.e. 20s timeout will give a user 4 tries.

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
```diff
@@ -73,6 +94,34 @@
         }
     }
 
+    /**
+     * Waits for the initial value to be available and returns the specified default value if failed
+     * to retrieve the initial value from the server.
+     *
+     * @param timeout the maximum amount of time to wait for the initial value. Note that timeout is basically
+     *                a trade-off. If you specify a smaller timeout, this method will take less time even if
+     *                the server is not responsive, at the risk of falling back to the {@code defaultValue}.
```
hyangtack: I'm not sure if it's your intention.
not responsive -> still responsive? (or just 'responsive'?)

client/java/src/main/java/com/linecorp/centraldogma/internal/client/AbstractWatcher.java
trustin: Oops :sweat_smile:

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
trustin: What I meant is: When a user specified a small timeout and the server is not responsive, a user will have a higher chance of getting `defaultValue`.

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
trustin: Fixed :smile:

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
trustin: Rephrased the sentence to make it clearer.

trustin:
Thanks for reviewing!

PR #394, opened 2019-04-22
Make `ArmeriaCentralDogma` handle `2xx` responses better.

trustin:
Motivation:

Central Dogma server sometimes gives a `204 No Content` response when
the result is an empty list. However, `ArmeriaCentralDogma` does not
always handle it properly.

Modifivations:

- Handle `204 No Content` and `200 OK` responses wherever possible, so
  that subtle changes in the response specification do not affect the
  client.

Result:

- `ArmeriaCentralDogma` does not fail for some operations such as
  `listFiles()` even if the server responds with a `204 No Content`.

### DIFF
--- a/client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogma.java
+++ b/client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogma.java
@@ -53,6 +53,7 @@ import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Streams;
 import com.google.common.math.IntMath;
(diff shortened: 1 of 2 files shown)


client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogma.java
```diff
@@ -150,8 +151,10 @@
     }
 
     private static Void createProject(AggregatedHttpMessage res) {
-        if (res.status().code() == 201) { // Created
-            return null;
+        switch (res.status().code()) {
+            case 200:
```
hyangtack: Question: [ProjectServiceV1.createProject()](https://github.com/line/centraldogma/blob/master/server/src/main/java/com/linecorp/centraldogma/server/internal/api/ProjectServiceV1.java#L92) always returns `201` when it succeeded to create a new project. So would it be better to handle this as an error like before? 🤔 
(It's the same question for the removal APIs)

trustin: Just wanted to relax the response status code check so that the client doesn't have to change when we make a server-side change for some reason. Shall I revert this?

hyangtack: No. :-) Thank you for explanation.

PR #396, opened 2019-04-24
Apply `ArmeriaClientConfigurator` after the internal decorators

trustin:
Motivation:

When we introduce an HTTP-based Central Dogma client implementation, we
slightly adjusted how decorators are added. As a result, the
`ArmeriaClientConfigurator` was being invoked before the internal
decorators were added.

Modifications:

- Move the common logic required for configuring an Armeria
  `ClientBuilder` to `AbstractArmeriaCentralDogmaBuilder.newClientBuilder()`.
- The customization of a `ClientBuilder` is always performed in the
  following order:
  - Builder-specific customizers are applied first.
  - `ArmeriaClientConfigurator` is applied next.
  - `ClientBuilder.clientFactory()` is set finally.
- Added a test case that verifies the customization ordering.

Result:

- Consistent customization order.
- Fixes the regression caused by 20583b1e74d583a5d5d12412e9cdb8af232f460f

### DIFF
--- a/client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogmaBuilder.java
+++ b/client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogmaBuilder.java
@@ -34,10 +34,8 @@ public final class ArmeriaCentralDogmaBuilder
         final Endpoint endpoint = endpoint();
         final String scheme = "none+" + (isUseTls() ? "https" : "http") + "://";
         final String uri = scheme + endpoint.authority();
-        final ClientBuilder builder = new ClientBuilder(uri);
-        clientConfigurator().configure(builder);
-        builder.factory(clientFactory());
-        builder.decorator(HttpDecodingClient.newDecorator());
+        final ClientBuilder builder =
+                newClientBuilder(uri, cb -> cb.decorator(HttpDecodingClient.newDecorator()));
         return new ArmeriaCentralDogma(clientFactory().eventLoopGroup(),
                                        builder.build(HttpClient.class), accessToken());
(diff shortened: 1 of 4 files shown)


client/java-armeria/src/test/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogmaBuilderTest.java
```diff
@@ -140,6 +152,42 @@ public void buildingWithMultipleHosts() throws Exception {
         }
     }
 
+    /**
+     * Ensure the customization order of {@link ClientBuilder}.
+     * <ol>
+     *   <li>Builder-specified {@link Consumer}</li>.
```
hyangtack: nit: removing the dot?

trustin: Oops! Thank you.

client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogmaBuilder.java
```diff
@@ -16,6 +16,7 @@
 package com.linecorp.centraldogma.client.armeria;
 
 import java.net.UnknownHostException;
+import java.util.function.Consumer;
```
minwoox: Should remove this?

trustin: Yes!

trustin:
Thanks for reviewing!

PR #399, opened 2019-05-10
Change the Java download links to refer to https://adoptopenjdk.net/

trustin:
.. for the reduced licensing risks of the users.


trustin:
Thanks :smile:

PR #403, opened 2019-05-29
Refer a cross-project task as a path rather than as a reference

trustin:
Related: https://github.com/line/gradle-scripts/pull/61
Motivation:

When configure-on-demand is enabled, inter-project dependencies are
calculated correctly only when task dependency is specified as a task
path string. See:

- https://docs.gradle.org/current/userguide/multi_project_builds.html#sec:configuration_on_demand

> The task dependencies declared via task path are supported and cause
> relevant projects to be configured.
> Example: someTask.dependsOn(":someOtherProject:someOtherTask")

Modifications:

- Use `Task.path` when referring to a task in a different project to
  call `Task.dependsOn()`

Result:

- Fixes a bug where a certain project is not configured when it has to
  be, causing a weird compilation error due to unconfigured dependencies.

### DIFF
--- a/dist/build.gradle
+++ b/dist/build.gradle
@@ -16,10 +16,7 @@ buildscript {
 apply plugin: 'com.boazj.log'
 apply plugin: 'com.bmuschko.docker-remote-api'
 
-evaluationDependsOn(':cli')
-evaluationDependsOn(':server')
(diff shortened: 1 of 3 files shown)


trustin:
Decided to stick to `evaluationDependsOn()` which is less error-prone than `Task.path`.

PR #408, opened 2019-06-18
Update requestTimeout when using Thrift watch API

minwoox:
Motivation:
We update the request timeout when a client sends a watch request using HTTP API, but we don't in Thrift.

Modification:
- Update `requestTimeoutMillis` when using Thrift watch API.

Result:
- No more `RequestAlreadyTimedOutException`.


trustin:
Thanks, @minwoox !

PR #409, opened 2019-06-18
Move metadata package to the public API

trustin:
Motivation:

We need to expose a `MetadataService` to a `Plugin` via a
`PluginContext`, but to no avail because `MetadataService` is not part
of the public API.

Modifications:

- Rename `server.internal.metadata` to `server.metadata`.
- Move `User` and `UserWithToken` from `server.internal.admin.auth` to
  `server.metadata`.

Result:

- It is now possible to expose `MetadataService` in `PluginContext`.
  - Note that we do not expose a `MetadataService` to a `Plugin` just
    yet. It will be done in a follow-up PR.


trustin:
Thanks!

PR #416, opened 2019-07-05
Automatic purging of removed repositories and projects

### LINKED ISSUE #47: Automatic purging of removed repositories and projects
e.g. Permanently delete repositories/projects which were removed 7+ days ago.

.. as well as on-demand purging of course.

ikhoon:
Motivation:
The repositories and projects that were removed by users just renamed with having the suffix of `.removed`.
If the number of deleted files increases, it takes up a lot of disk space.
We can reduce wasted disk space by permanently deleting files which are older than configured.

Modifications:
* Add `PurgeProjectCommand` and `PurgeRepositoryCommand` and their thrift, rest api.
* Add `markForPurge()` and `purgeMarked()` method to mark and permanently delete a project or a repository directory.
* Add `PurgeSchedulingService` and `PurgeSchedulingServicePlugin` to purge automitically.
  * When the scheduler get started, it cleans up purged projects and repositories
    which were not deleted before.
* Add a new configuration `maxRemovedRepositoryAgeMillis` and update its documentation.
  * Setting 0 to `maxRemovedRepositoryAgeMillis` means disable the automatic purging.
* Record removal time into `removal.timestamp` file when a project/repository is removed.
* Change return type of `Set<String> listRemove()` into
  `Map<String, Instant> listRemoved()` to get a removal timestamp.

Result:
* A user who has permission can purge a repository or a project on demand.
* Removed repositories and projects are pursed automatically.

Fixes #47

### DIFF
--- a/client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogma.java
+++ b/client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogma.java
@@ -189,19 +189,10 @@ final class ArmeriaCentralDogma extends AbstractCentralDogma {
             return client.execute(
                     headers(HttpMethod.DELETE, pathBuilder(projectName).append(REMOVED).toString()))
                          .aggregate()
-                         .thenApply(ArmeriaCentralDogma::purgeProject);
+                         .thenApply(ArmeriaCentralDogma::handlePurgeResult);
         } catch (Exception e) {
             return exceptionallyCompletedFuture(e);
         }
-    }
-
-    private static Void purgeProject(AggregatedHttpResponse res) {
-        switch (res.status().code()) {
-            case 200:
-            case 204:
-                return null;
-        }
-        return handleErrorResponse(res);
     }
 
     @Override
@@ -293,13 +284,13 @@ final class ArmeriaCentralDogma extends AbstractCentralDogma {
             return client.execute(headers(HttpMethod.DELETE,
                                           pathBuilder(projectName, repositoryName).append(REMOVED).toString()))
                          .aggregate()
-                         .thenApply(ArmeriaCentralDogma::purgeRepository);
+                         .thenApply(ArmeriaCentralDogma::handlePurgeResult);
         } catch (Exception e) {
             return exceptionallyCompletedFuture(e);
         }
     }
 
-    private static Void purgeRepository(AggregatedHttpResponse res) {
+    private static Void handlePurgeResult(AggregatedHttpResponse res) {
         switch (res.status().code()) {
             case 200:
             case 204:
--- a/client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogma.java
+++ b/client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogma.java
@@ -24,6 +24,7 @@ import static com.google.common.collect.ImmutableSet.toImmutableSet;
 import static com.linecorp.centraldogma.internal.Util.unsafeCast;
 import static com.linecorp.centraldogma.internal.Util.validatePathPattern;
 import static com.linecorp.centraldogma.internal.api.v1.HttpApiV1Constants.PROJECTS_PREFIX;
+import static com.linecorp.centraldogma.internal.api.v1.HttpApiV1Constants.REMOVED;
 import static com.linecorp.centraldogma.internal.api.v1.HttpApiV1Constants.REPOS;
 import static com.spotify.futures.CompletableFutures.exceptionallyCompletedFuture;
 import static java.util.Objects.requireNonNull;
@@ -182,6 +183,28 @@ final class ArmeriaCentralDogma extends AbstractCentralDogma {
     }
 
     @Override
+    public CompletableFuture<Void> purgeProject(String projectName) {
+        try {
+            validateProjectName(projectName);
+            return client.execute(
+                    headers(HttpMethod.DELETE, pathBuilder(projectName).append(REMOVED).toString()))
+                         .aggregate()
+                         .thenApply(ArmeriaCentralDogma::purgeProject);
+        } catch (Exception e) {
+            return exceptionallyCompletedFuture(e);
+        }
+    }
+
+    private static Void purgeProject(AggregatedHttpResponse res) {
+        switch (res.status().code()) {
+            case 200:
+            case 204:
+                return null;
+        }
+        return handleErrorResponse(res);
+    }
+
+    @Override
     public CompletableFuture<Void> unremoveProject(String projectName) {
         try {
             validateProjectName(projectName);
@@ -255,6 +278,28 @@ final class ArmeriaCentralDogma extends AbstractCentralDogma {
     }
 
     private static Void removeRepository(AggregatedHttpResponse res) {
+        switch (res.status().code()) {
+            case 200:
+            case 204:
+                return null;
+        }
+        return handleErrorResponse(res);
+    }
+
+    @Override
+    public CompletableFuture<Void> purgeRepository(String projectName, String repositoryName) {
+        try {
+            validateProjectAndRepositoryName(projectName, repositoryName);
+            return client.execute(headers(HttpMethod.DELETE,
+                                          pathBuilder(projectName, repositoryName).append(REMOVED).toString()))
+                         .aggregate()
+                         .thenApply(ArmeriaCentralDogma::purgeRepository);
+        } catch (Exception e) {
+            return exceptionallyCompletedFuture(e);
+        }
+    }
+
+    private static Void purgeRepository(AggregatedHttpResponse res) {
         switch (res.status().code()) {
             case 200:
             case 204:
--- a/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
@@ -42,12 +42,15 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.LinkedTransferQueue;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.BiFunction;
 import java.util.function.Consumer;
 import java.util.function.Function;
 
@@ -199,6 +202,8 @@ public class CentralDogma implements AutoCloseable {
     @Nullable
     private ExecutorService repositoryWorker;
     @Nullable
+    private ScheduledExecutorService purgeWorker;
+    @Nullable
     private CommandExecutor executor;
     @Nullable
     private PrometheusMeterRegistry meterRegistry;
@@ -311,6 +316,7 @@ public class CentralDogma implements AutoCloseable {
     private void doStart() throws Exception {
         boolean success = false;
         ExecutorService repositoryWorker = null;
+        ScheduledExecutorService purgeWorker = null;
         ProjectManager pm = null;
         CommandExecutor executor = null;
         PrometheusMeterRegistry meterRegistry = null;
@@ -330,8 +336,12 @@ public class CentralDogma implements AutoCloseable {
 
             logger.info("Starting the project manager: {}", cfg.dataDir());
 
-            pm = new DefaultProjectManager(cfg.dataDir(), repositoryWorker, meterRegistry,
-                                           cfg.repositoryCacheSpec());
+            purgeWorker = Executors.newSingleThreadScheduledExecutor(
+                    new DefaultThreadFactory("purge-worker", true));
+
+            pm = new DefaultProjectManager(cfg.dataDir(), repositoryWorker, purgeWorker,
+                                           meterRegistry, cfg.repositoryCacheSpec());
+
             logger.info("Started the project manager: {}", pm);
 
             logger.info("Current settings:\n{}", cfg);
@@ -339,8 +349,8 @@ public class CentralDogma implements AutoCloseable {
             sessionManager = initializeSessionManager();
 
             logger.info("Starting the command executor ..");
-
-            executor = startCommandExecutor(pm, repositoryWorker, meterRegistry, sessionManager);
+            executor = startCommandExecutor(pm, repositoryWorker, purgeWorker,
+                                            meterRegistry, sessionManager);
             if (executor.isWritable()) {
                 logger.info("Started the command executor.");
 
@@ -358,24 +368,28 @@ public class CentralDogma implements AutoCloseable {
         } finally {
             if (success) {
                 this.repositoryWorker = repositoryWorker;
+                this.purgeWorker = purgeWorker;
                 this.pm = pm;
                 this.executor = executor;
                 this.meterRegistry = meterRegistry;
                 this.server = server;
                 this.sessionManager = sessionManager;
             } else {
-                doStop(server, executor, pm, repositoryWorker, sessionManager);
+                doStop(server, executor, pm, repositoryWorker, purgeWorker, sessionManager);
             }
         }
     }
 
     private CommandExecutor startCommandExecutor(
-            ProjectManager pm, Executor repositoryWorker, MeterRegistry meterRegistry,
+            ProjectManager pm, Executor repositoryWorker,
+            ScheduledExecutorService purgeWorker, MeterRegistry meterRegistry,
             @Nullable SessionManager sessionManager) {
+
         final Consumer<CommandExecutor> onTakeLeadership = exec -> {
             if (pluginsForLeaderOnly != null) {
                 logger.info("Starting plugins on the leader replica ..");
-                pluginsForLeaderOnly.start(cfg, pm, exec, meterRegistry).handle((unused, cause) -> {
+                pluginsForLeaderOnly
+                        .start(cfg, pm, exec, meterRegistry, purgeWorker).handle((unused, cause) -> {
                     if (cause == null) {
                         logger.info("Started plugins on the leader replica.");
                     } else {
@@ -389,7 +403,7 @@ public class CentralDogma implements AutoCloseable {
         final Consumer<CommandExecutor> onReleaseLeadership = exec -> {
             if (pluginsForLeaderOnly != null) {
                 logger.info("Stopping plugins on the leader replica ..");
-                pluginsForLeaderOnly.stop(cfg, pm, exec, meterRegistry).handle((unused, cause) -> {
+                pluginsForLeaderOnly.stop(cfg, pm, exec, meterRegistry, purgeWorker).handle((unused, cause) -> {
                     if (cause == null) {
                         logger.info("Stopped plugins on the leader replica.");
                     } else {
@@ -788,6 +802,7 @@ public class CentralDogma implements AutoCloseable {
         final CommandExecutor executor = this.executor;
         final ProjectManager pm = this.pm;
         final ExecutorService repositoryWorker = this.repositoryWorker;
+        final ExecutorService purgeWorker = this.purgeWorker;
         final SessionManager sessionManager = this.sessionManager;
 
         this.server = null;
@@ -801,7 +816,7 @@ public class CentralDogma implements AutoCloseable {
         }
 
         logger.info("Stopping the Central Dogma ..");
-        if (!doStop(server, executor, pm, repositoryWorker, sessionManager)) {
+        if (!doStop(server, executor, pm, repositoryWorker, purgeWorker, sessionManager)) {
             logger.warn("Stopped the Central Dogma with failure.");
         } else {
             logger.info("Stopped the Central Dogma successfully.");
@@ -810,7 +825,8 @@ public class CentralDogma implements AutoCloseable {
 
     private static boolean doStop(
             @Nullable Server server, @Nullable CommandExecutor executor,
-            @Nullable ProjectManager pm, @Nullable ExecutorService repositoryWorker,
+            @Nullable ProjectManager pm,
+            @Nullable ExecutorService repositoryWorker, @Nullable ExecutorService purgeWorker,
             @Nullable SessionManager sessionManager) {
 
         boolean success = true;
@@ -847,28 +863,37 @@ public class CentralDogma implements AutoCloseable {
             logger.warn("Failed to stop the command executor:", t);
         }
 
-        try {
-            if (repositoryWorker != null && !repositoryWorker.isTerminated()) {
-                logger.info("Stopping the repository worker ..");
-                boolean interruptLater = false;
-                while (!repositoryWorker.isTerminated()) {
-                    repositoryWorker.shutdownNow();
-                    try {
-                        repositoryWorker.awaitTermination(1, TimeUnit.SECONDS);
-                    } catch (InterruptedException e) {
-                        // Interrupt later.
-                        interruptLater = true;
+        final BiFunction<ExecutorService, String, Boolean> stopWorker = (worker, name) -> {
+            try {
+                if (worker != null && !worker.isTerminated()) {
+                    logger.info("Stopping the {} worker ..", name);
+                    boolean interruptLater = false;
+                    while (!worker.isTerminated()) {
+                        worker.shutdownNow();
+                        try {
+                            worker.awaitTermination(1, TimeUnit.SECONDS);
+                        } catch (InterruptedException e) {
+                            // Interrupt later.
+                            interruptLater = true;
+                        }
+                    }
+                    logger.info("Stopped the {} worker.", name);
+
+                    if (interruptLater) {
+                        Thread.currentThread().interrupt();
                     }
                 }
-                logger.info("Stopped the repository worker.");
-
-                if (interruptLater) {
-                    Thread.currentThread().interrupt();
-                }
+                return true;
+            } catch (Throwable t) {
+                logger.warn("Failed to stop the " + name + " worker:", t);
+                return false;
             }
-        } catch (Throwable t) {
+        };
+        if (!stopWorker.apply(repositoryWorker, "repository")) {
             success = false;
-            logger.warn("Failed to stop the repository worker:", t);
+        }
+        if (!stopWorker.apply(purgeWorker, "purge")) {
+            success = false;
         }
 
         try {
@@ -905,7 +930,7 @@ public class CentralDogma implements AutoCloseable {
                         final CommandExecutor executor = CentralDogma.this.executor;
                         final MeterRegistry meterRegistry = CentralDogma.this.meterRegistry;
                         if (pm != null && executor != null && meterRegistry != null) {
-                            pluginsForAllReplicas.start(cfg, pm, executor, meterRegistry).join();
+                            pluginsForAllReplicas.start(cfg, pm, executor, meterRegistry, purgeWorker).join();
                         }
                     }
                 } catch (Exception e) {
@@ -922,7 +947,7 @@ public class CentralDogma implements AutoCloseable {
                     final CommandExecutor executor = CentralDogma.this.executor;
                     final MeterRegistry meterRegistry = CentralDogma.this.meterRegistry;
                     if (pm != null && executor != null && meterRegistry != null) {
-                        pluginsForAllReplicas.stop(cfg, pm, executor, meterRegistry).join();
+                        pluginsForAllReplicas.stop(cfg, pm, executor, meterRegistry, purgeWorker).join();
                     }
                 }
                 CentralDogma.this.doStop();
--- a/server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaBuilder.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaBuilder.java
@@ -300,7 +300,7 @@ public final class CentralDogmaBuilder {
 
     /**
      * Sets the maximum allowed age of removed projects and repositories before they are purged.
-     * Set 0 to disable automatic purge.
+     * Set {@code 0} to disable automatic purge.
      * If unspecified, the default of {@value #DEFAULT_MAX_REMOVED_REPOSITORY_AGE_MILLIS} milliseconds is used.
      */
     public CentralDogmaBuilder maxRemovedRepositoryAge(Duration maxRemovedRepositoryAge) {
@@ -312,7 +312,7 @@ public final class CentralDogmaBuilder {
     /**
      * Sets the maximum allowed age, in milliseconds of removed projects and repositories
      * before they are purged.
-     * Set 0 to disable automatic purge.
+     * Set {@code 0} to disable automatic purge.
      * If unspecified, the default of {@value #DEFAULT_MAX_REMOVED_REPOSITORY_AGE_MILLIS} milliseconds is used.
      */
     public CentralDogmaBuilder maxRemovedRepositoryAgeMillis(long maxRemovedRepositoryAgeMillis) {
--- a/server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaBuilder.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaBuilder.java
@@ -70,6 +70,7 @@ public final class CentralDogmaBuilder {
     static final int DEFAULT_NUM_MIRRORING_THREADS = 16;
     static final int DEFAULT_MAX_NUM_FILES_PER_MIRROR = 8192;
     static final long DEFAULT_MAX_NUM_BYTES_PER_MIRROR = 32 * 1048576; // 32 MiB
+    static final long DEFAULT_MAX_REMOVED_REPOSITORY_AGE_MILLIS = 604_800_000;  // 7 days
 
     static final String DEFAULT_REPOSITORY_CACHE_SPEC =
             "maximumWeight=134217728," + // Cache up to apx. 128-megachars.
@@ -91,6 +92,8 @@ public final class CentralDogmaBuilder {
     // Central Dogma properties
     private final File dataDir;
     private int numRepositoryWorkers = DEFAULT_NUM_REPOSITORY_WORKERS;
+    private long maxRemovedRepositoryAgeMillis = DEFAULT_MAX_REMOVED_REPOSITORY_AGE_MILLIS;
+
     @Nullable
     private String repositoryCacheSpec = DEFAULT_REPOSITORY_CACHE_SPEC;
     private boolean webAppEnabled = true;
@@ -292,6 +295,28 @@ public final class CentralDogmaBuilder {
      */
     public CentralDogmaBuilder numRepositoryWorkers(int numRepositoryWorkers) {
         this.numRepositoryWorkers = numRepositoryWorkers;
+        return this;
+    }
+
+    /**
+     * Sets the maximum allowed age of removed projects and repositories before they are purged.
+     * Set 0 to disable automatic purge.
+     * If unspecified, the default of {@value #DEFAULT_MAX_REMOVED_REPOSITORY_AGE_MILLIS} milliseconds is used.
+     */
+    public CentralDogmaBuilder maxRemovedRepositoryAge(Duration maxRemovedRepositoryAge) {
+        maxRemovedRepositoryAgeMillis(
+                requireNonNull(maxRemovedRepositoryAge, "maxRemovedRepositoryAge").toMillis());
+        return this;
+    }
+
+    /**
+     * Sets the maximum allowed age, in milliseconds of removed projects and repositories
+     * before they are purged.
+     * Set 0 to disable automatic purge.
+     * If unspecified, the default of {@value #DEFAULT_MAX_REMOVED_REPOSITORY_AGE_MILLIS} milliseconds is used.
+     */
+    public CentralDogmaBuilder maxRemovedRepositoryAgeMillis(long maxRemovedRepositoryAgeMillis) {
+        this.maxRemovedRepositoryAgeMillis = maxRemovedRepositoryAgeMillis;
         return this;
     }
 
@@ -508,7 +533,8 @@ public final class CentralDogmaBuilder {
         return new CentralDogmaConfig(dataDir, ports, tls, trustedProxyAddresses, clientAddressSources,
                                       numWorkers, maxNumConnections,
                                       requestTimeoutMillis, idleTimeoutMillis, maxFrameLength,
-                                      numRepositoryWorkers, repositoryCacheSpec, gracefulShutdownTimeout,
+                                      numRepositoryWorkers, repositoryCacheSpec,
+                                      maxRemovedRepositoryAgeMillis, gracefulShutdownTimeout,
                                       webAppEnabled, webAppTitle, mirroringEnabled, numMirroringThreads,
                                       maxNumFilesPerMirror, maxNumBytesPerMirror, replicationConfig,
                                       null, accessLogFormat, authCfg);
--- a/server/src/main/java/com/linecorp/centraldogma/server/command/Command.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/command/Command.java
@@ -42,9 +42,11 @@ import com.linecorp.centraldogma.server.auth.Session;
 @JsonSubTypes({
         @Type(value = CreateProjectCommand.class, name = "CREATE_PROJECT"),
         @Type(value = RemoveProjectCommand.class, name = "REMOVE_PROJECT"),
+        @Type(value = PurgeProjectCommand.class, name = "PURGE_PROJECT"),
         @Type(value = UnremoveProjectCommand.class, name = "UNREMOVE_PROJECT"),
         @Type(value = CreateRepositoryCommand.class, name = "CREATE_REPOSITORY"),
         @Type(value = RemoveRepositoryCommand.class, name = "REMOVE_REPOSITORY"),
+        @Type(value = PurgeRepositoryCommand.class, name = "PURGE_REPOSITORY"),
         @Type(value = UnremoveRepositoryCommand.class, name = "UNREMOVE_REPOSITORY"),
         @Type(value = PushCommand.class, name = "PUSH"),
         @Type(value = CreateSessionCommand.class, name = "CREATE_SESSIONS"),
@@ -116,6 +118,29 @@ public interface Command<T> {
     static Command<Void> unremoveProject(@Nullable Long timestamp, Author author, String name) {
         requireNonNull(author, "author");
         return new UnremoveProjectCommand(timestamp, author, name);
+    }
+
+    /**
+     * Returns a new {@link Command} which is used to purge a project that was removed before.
+     *
+     * @param author the author who is restoring the project
+     * @param name the name of the project which is supposed to be restored
+     */
+    static Command<Void> purgeProject(Author author, String name) {
+        requireNonNull(author, "author");
+        return new PurgeProjectCommand(null, author, name);
+    }
+
+    /**
+     * Returns a new {@link Command} which is used to purge a project that was removed before.
+     *
+     * @param timestamp the purging time of the project, in milliseconds
+     * @param author the author who is restoring the project
+     * @param name the name of the project which is supposed to be restored
+     */
+    static Command<Void> purgeProject(@Nullable Long timestamp, Author author, String name) {
+        requireNonNull(author, "author");
+        return new PurgeProjectCommand(timestamp, author, name);
     }
 
     /**
@@ -191,6 +216,33 @@ public interface Command<T> {
                                             String projectName, String repositoryName) {
         requireNonNull(author, "author");
         return new UnremoveRepositoryCommand(timestamp, author, projectName, repositoryName);
+    }
+
+    /**
+     * Returns a new {@link Command} which is used to purge a repository.
+     *
+     * @param author the author who is removing the repository
+     * @param projectName the name of the project
+     * @param repositoryName the name of the repository which is supposed to be purge
+     */
+    static Command<Void> purgeRepository(Author author,
+                                         String projectName, String repositoryName) {
+        requireNonNull(author, "author");
+        return new PurgeRepositoryCommand(null, author, projectName, repositoryName);
+    }
+
+    /**
+     * Returns a new {@link Command} which is used to purge a repository.
+     *
+     * @param timestamp the purging time of the repository, in milliseconds
+     * @param author the author who is removing the repository
+     * @param projectName the name of the project
+     * @param repositoryName the name of the repository which is supposed to be purge
+     */
+    static Command<Void> purgeRepository(@Nullable Long timestamp, Author author,
+                                          String projectName, String repositoryName) {
+        requireNonNull(author, "author");
+        return new PurgeRepositoryCommand(timestamp, author, projectName, repositoryName);
     }
 
     /**
--- a/server/src/main/java/com/linecorp/centraldogma/server/command/Command.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/command/Command.java
@@ -223,7 +223,7 @@ public interface Command<T> {
      *
      * @param author the author who is removing the repository
      * @param projectName the name of the project
-     * @param repositoryName the name of the repository which is supposed to be purge
+     * @param repositoryName the name of the repository which is supposed to be purged
      */
     static Command<Void> purgeRepository(Author author,
                                          String projectName, String repositoryName) {
@@ -237,7 +237,7 @@ public interface Command<T> {
      * @param timestamp the purging time of the repository, in milliseconds
      * @param author the author who is removing the repository
      * @param projectName the name of the project
-     * @param repositoryName the name of the repository which is supposed to be purge
+     * @param repositoryName the name of the repository which is supposed to be purged
      */
     static Command<Void> purgeRepository(@Nullable Long timestamp, Author author,
                                           String projectName, String repositoryName) {
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/DirectoryBasedStorageManager.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/DirectoryBasedStorageManager.java
@@ -21,14 +21,19 @@ import static java.util.Objects.requireNonNull;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.FileAlreadyExistsException;
+import java.nio.file.Files;
+import java.time.Instant;
+import java.time.format.DateTimeFormatter;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.LinkedHashMap;
-import java.util.LinkedHashSet;
 import java.util.Map;
-import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.Executor;
+import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Supplier;
@@ -38,6 +43,8 @@ import javax.annotation.Nullable;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import com.google.common.collect.ImmutableMap;
 
 import com.linecorp.centraldogma.common.Author;
 import com.linecorp.centraldogma.common.CentralDogmaException;
@@ -57,15 +64,20 @@ public abstract class DirectoryBasedStorageManager<T> implements StorageManager<
     private static final Pattern CHILD_NAME =
             Pattern.compile("^[0-9A-Za-z](?:[-+_0-9A-Za-z.]*[0-9A-Za-z])?$");
     private static final String SUFFIX_REMOVED = ".removed";
+    private static final String SUFFIX_PURGED = ".purged";
 
     private final String childTypeName;
     private final File rootDir;
+    private final StorageRemovalManager storageRemovalManager = new StorageRemovalManager();
     private final ConcurrentMap<String, T> children = new ConcurrentHashMap<>();
     private final AtomicReference<Supplier<CentralDogmaException>> closed = new AtomicReference<>();
+    private final Executor purgeWorker;
     private boolean initialized;
 
-    protected DirectoryBasedStorageManager(File rootDir, Class<? extends T> childType) {
+    protected DirectoryBasedStorageManager(File rootDir, Class<? extends T> childType,
+                                           Executor purgeWorker) {
         requireNonNull(rootDir, "rootDir");
+        this.purgeWorker = requireNonNull(purgeWorker, "purgeWorker");
 
         if (!rootDir.exists()) {
             if (!rootDir.mkdirs()) {
@@ -85,6 +97,10 @@ public abstract class DirectoryBasedStorageManager<T> implements StorageManager<
 
         this.rootDir = rootDir;
         childTypeName = Util.simpleTypeName(requireNonNull(childType, "childTypeName"), true);
+    }
+
+    protected Executor purgeWorker() {
+        return purgeWorker;
     }
 
     /**
@@ -228,7 +244,7 @@ public abstract class DirectoryBasedStorageManager<T> implements StorageManager<
                 try {
                     Util.deleteFileTree(f);
                 } catch (IOException e) {
-                    logger.warn("Failed to delete a partially created project: {}", f);
+                    logger.warn("Failed to delete a partially created project: {}", f, e);
                 }
             }
         }
@@ -254,15 +270,15 @@ public abstract class DirectoryBasedStorageManager<T> implements StorageManager<
     }
 
     @Override
-    public Set<String> listRemoved() {
+    public Map<String, Instant> listRemoved() {
         ensureOpen();
-        final Set<String> removed = new LinkedHashSet<>();
         final File[] files = rootDir.listFiles();
         if (files == null) {
-            return Collections.emptySet();
+            return ImmutableMap.of();
         }
 
         Arrays.sort(files);
+        final ImmutableMap.Builder<String, Instant> builder = ImmutableMap.builder();
 
         for (File f : files) {
             if (!f.isDirectory()) {
@@ -279,10 +295,10 @@ public abstract class DirectoryBasedStorageManager<T> implements StorageManager<
                 continue;
             }
 
-            removed.add(name);
+            builder.put(name, storageRemovalManager.readRemoval(f));
         }
 
-        return Collections.unmodifiableSet(removed);
+        return builder.build();
     }
 
     @Override
@@ -295,7 +311,9 @@ public abstract class DirectoryBasedStorageManager<T> implements StorageManager<
 
         closeChild(name, child, () -> newStorageNotFoundException(name));
 
-        if (!new File(rootDir, name).renameTo(new File(rootDir, name + SUFFIX_REMOVED))) {
+        final File file = new File(rootDir, name);
+        storageRemovalManager.mark(file);
+        if (!file.renameTo(new File(rootDir, name + SUFFIX_REMOVED))) {
             throw new StorageException("failed to mark " + childTypeName + " as removed: " + name);
         }
     }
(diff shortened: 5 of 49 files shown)


server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -371,7 +375,8 @@ private void doStart() throws Exception {
 
     private CommandExecutor startCommandExecutor(
             ProjectManager pm, Executor repositoryWorker, MeterRegistry meterRegistry,
-            @Nullable SessionManager sessionManager) {
+            @Nullable SessionManager sessionManager, StoragePurgeService storagePurgeService) {
```
trustin: How about sort the parameters in the order of importance, with an optional one placed at last, i.e.

- `ProjectManager`
- `Executor`
- `StoragePurgeService`
- `MeterRegistry`
- `SessionManager`

?

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -384,6 +389,8 @@ private CommandExecutor startCommandExecutor(
                     return null;
                 });
             }
+            storagePurgeService.start(exec);
+            logger.info("Started the storage purge service on the leader replica ..");
```
trustin: `..` means something is in progress. In this case, you finished starting the purge service already, so there's no need to append `..`.

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -398,6 +405,8 @@ private CommandExecutor startCommandExecutor(
                     return null;
                 });
             }
+            storagePurgeService.stop();
+            logger.info("Stopped the storage purge service on the leader replica ..");
```
trustin: `..` means something is in progress. In this case, you finished starting the purge service already, so there's no need to append `..`.

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -398,6 +405,8 @@ private CommandExecutor startCommandExecutor(
                     return null;
                 });
             }
+            storagePurgeService.stop();
```
trustin: I wonder if stopping the purge service will take longer time than expected if a purge operation is currently in progress. How about adding an additional log message like `Stopping the storage purge service ..`?

server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaBuilder.java
```diff
@@ -295,6 +298,15 @@ public CentralDogmaBuilder numRepositoryWorkers(int numRepositoryWorkers) {
         return this;
     }
 
+    /**
+     * Sets the maximum allowed age of removed projects and repositories before they are purged.
+     * If unspecified, the default of {@value #DEFAULT_MAX_REMOVED_REPOSITORY_AGE_MILLIS} is used.
+     */
+    public CentralDogmaBuilder maxRemovedRepositoryAgeMillis(long maxRemovedRepositoryAgeMillis) {
```
trustin: Could you also add a method that accepts a `Duration`?

server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaBuilder.java
```diff
@@ -295,6 +298,15 @@ public CentralDogmaBuilder numRepositoryWorkers(int numRepositoryWorkers) {
         return this;
     }
 
+    /**
+     * Sets the maximum allowed age of removed projects and repositories before they are purged.
+     * If unspecified, the default of {@value #DEFAULT_MAX_REMOVED_REPOSITORY_AGE_MILLIS} is used.
```
trustin: Could you allow a user to disable automatic purge, by specifying zero?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/DirectoryBasedStorageManager.java
```diff
@@ -323,6 +331,21 @@ public T unremove(String name) {
...
+    @Override
+    public void purge(String name) {
+        ensureOpen();
+        validateChildName(name);
+        final File removed = new File(rootDir, name + SUFFIX_REMOVED);
+        if (!removed.isDirectory()) {
+            throw newStorageNotFoundException(name);
+        }
+        try {
+            Util.deleteFileTree(removed);
```
trustin: Deleting a file tree can be a time-consuming task. A client who issues this request might observe a timeout. How about renaming `*.removed` directory to `*.purged` and scheduling deletion in the background? This change will require some additional changes to your `StoragePurgeScheduler` such as moving the actual file deletion there.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/StoragePurgeService.java
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    public void stop() {
+        storagePurgeScheduler.stop();
+    }
+
+    @VisibleForTesting
+    void schedulePurgeData(CommandExecutor commandExecutor) {
+        final long minAllowedTimestamp = System.currentTimeMillis() - maxRemovedRepositoryAgeMillis;
+        final Predicate<File> olderThanMinAllowed = removed -> removed.lastModified() < minAllowedTimestamp;
```
trustin: Could you use `ProjectMetadata.removal()` and `RepositoryMetadata.removal()` rather than `File.lastModified()`?

server/src/main/java/com/linecorp/centraldogma/server/storage/StorageManager.java
```diff
@@ -81,6 +83,14 @@ default T create(String name, Author author) {
      */
     Set<String> listRemoved();
 
+    /**
+     * Returns a set of names for the elements which have been removed.
+     * The specified @{code filter} evaluates whether an {@link File} is collected or not
+     *
+     * @param filter the filter evaluates whether the {@link File} is collected or not
+     */
+    Set<String> listRemoved(Predicate<File> filter);
```
trustin: This leaks the implementation detail of `DirectoryBasedStorageManager`. Not all storage may be filesystem-based. How about just retrieving the complete list using `listRemoved()` given that the list will not be very large?

ikhoon:
Thanks for fast feedback! @trustin 😊
I will fix it soon.

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
ikhoon: I tended to preserve the order of the existed parameters.
But this method is a private and "an optional one placed at last" is reasonable. Thanks

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
ikhoon: Ah, thanks!

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
ikhoon: Thank you, I added `stopping ...` log message before stopping.

server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaBuilder.java
ikhoon: Oops, I missed it!

ikhoon:
Had chat with @trustin about the strategy of purging scheduler.

* If scheduler receives a directory, delete the directory in order in a single thread. 
* When scheduler started, scan `.purged` directories and rescheduling them
* Automatically purge old projects and repository which were removed

server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaBuilder.java
ikhoon: That is a nice idea!, It was added [here](https://github.com/line/centraldogma/pull/416/files#diff-e3bb9ea78f820637a7452d39d75bb583R166)

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/StoragePurgeService.java
ikhoon: fixed :-)

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/DirectoryBasedStorageManager.java
ikhoon: fixed it :-)

server/src/main/java/com/linecorp/centraldogma/server/storage/StorageManager.java
ikhoon: Oops! This API has been removed and added `listPurged()` to scan files which were `.purged`.

ikhoon:
I will improve test coverage. 🤨

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/ScheduledPurgeService.java
```diff
@@ -36,64 +39,131 @@
...
 
-    public void start(CommandExecutor commandExecutor) {
-       storagePurgeScheduler.start(() -> schedulePurgeData(commandExecutor));
+    public void start(CommandExecutor commandExecutor,
+                      MetadataService metadataService) {
+        if (!isDisabled()) {
+            requireNonNull(commandExecutor, "commandExecutor");
+            requireNonNull(metadataService, "metadataService");
+            cleanPurgedFiles(commandExecutor);
+            storagePurgeScheduler.start(() -> schedulePurgeData(commandExecutor, metadataService).join());
```
trustin: - Could we start the scheduler task after the commands issued by `cleanPurgedFiles()` are all done? It's because otherwise two threads may try to issue `Purge*Command`s at the same time.
- Catch and log an exception.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/DirectoryBasedStorageManager.java
```diff
@@ -323,6 +306,68 @@ public T unremove(String name) {
...
     }
 
+    @Override
+    public void purge(String name) {
+        ensureOpen();
+        validateChildName(name);
+        final File removed = new File(rootDir, name + SUFFIX_REMOVED);
+        final File purged = new File(rootDir, name + SUFFIX_PURGED);
+
+        if (removed.exists() && !purged.exists()) { // try to rename `.removed` into `.purged`
```
trustin: - Could reduce the number of system calls:
  ```java
  if (!purged.exists()) {
     if (removed.exists()) {
         ...
     } else {
         ...
     }
  }
  ```
- Shouldn't we also raise `newStorageNotFoundException(...)` when only `.purged` exists?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/ScheduledPurgeService.java
```diff
@@ -0,0 +1,234 @@
...
+        projectManager.list().values()
+                      .forEach(project -> project
+                              .repos()
+                              .listPurged()
+                              .forEach(repoName -> commandExecutor.execute(
+                                      Command.purgeRepository(Author.SYSTEM, project.name(), repoName))));
+    }
+
+    @VisibleForTesting
+    CompletableFuture<Void> schedulePurgeData(CommandExecutor commandExecutor,
```
trustin: Is it necessary to make `schedulePurgeData`, `purgeProject` and `purgeRepository` asynchronously and concurrently at all? How about simplifying it to run synchronously and sequentially, for example just using `purgeWorker` we created when starting `CentralDogma`.

trustin: .. and then we do not need to return `CompletableFuture` at all.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/DirectoryBasedStorageManager.java
```diff
@@ -336,14 +311,61 @@ public void purge(String name) {
...
         }
-        try {
-            Util.deleteFileTree(removed);
-        } catch (IOException e) {
-            throw new StorageException("failed to purge " + childTypeName + ": " + name);
+        purgeWorker.execute(() -> {
+            try {
+                Util.deleteFileTree(purged);
+            } catch (IOException e) {
+                Exceptions.throwUnsafely(e);
```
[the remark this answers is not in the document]
ikhoon: Thanks for kind advice. I was considering how to handle error here. Logging at WARN level looks good.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/DirectoryBasedStorageManager.java
```diff
@@ -336,14 +311,61 @@ public void purge(String name) {
...
+    }
+
+    @Override
+    public Set<String> listPurged() {
+        return listWithSuffix(SUFFIX_PURGED);
+    }
+
+    private Set<String> listWithSuffix(String suffix) {
+        ensureOpen();
+        final Set<String> removed = new LinkedHashSet<>();
```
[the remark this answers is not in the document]
ikhoon: Good, `ImmutableSet.Builder` would be better than `LinkedHashSet`!

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/ScheduledPurgeService.java
```diff
@@ -36,64 +39,131 @@
...
+import com.linecorp.centraldogma.server.metadata.UserAndTimestamp;
 import com.linecorp.centraldogma.server.storage.project.ProjectManager;
 
 import io.netty.util.concurrent.DefaultThreadFactory;
 
 /**
  * A service class for pursing projects and repositories that were removed before.
  */
-public class StoragePurgeService {
+public class ScheduledPurgeService {
```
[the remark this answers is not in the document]
ikhoon: I was also thinking about `Plugin` too. Because it has a nice start/stop code as you said. 
I hesitated to implement this code as `Plugin` because I don't have a strong conviction that this feature is suitable as `Plugin`.
Thanks for nice advice and I try to make this a `Plugin`. 😊

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/DirectoryBasedStorageManager.java
ikhoon: Ahh, reduce system calls!! I didn't think about that. It's a nice idea. Thanks 🙏

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/DirectoryBasedStorageManager.java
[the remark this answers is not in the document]
ikhoon: fixed.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/DirectoryBasedStorageManager.java
```diff
@@ -336,14 +311,61 @@ public void purge(String name) {
...
+        final File[] files = rootDir.listFiles();
+        if (files == null) {
+            return Collections.emptySet();
```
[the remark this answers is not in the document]
ikhoon: fixed.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/ScheduledPurgeService.java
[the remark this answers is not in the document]
ikhoon: Added [ScheduledPurgingServicePlugin](https://github.com/line/centraldogma/pull/416/files#diff-1522f6e7ffae1368385b728f4d695d0a)

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/ScheduledPurgeService.java
```diff
@@ -36,64 +39,131 @@
...
 
-    private static final Logger logger = LoggerFactory.getLogger(StoragePurgeService.class);
+    private static final Logger logger = LoggerFactory.getLogger(ScheduledPurgeService.class);
 
     /**
      * How often to run the storage purge schedules. i.e. every minute.
      */
-    private static final Duration TICK = Duration.ofSeconds(1);
+    private static final Duration TICK = Duration.ofMinutes(1);
+    private static final Duration INITIAL_DELAY = Duration.ofSeconds(1);
```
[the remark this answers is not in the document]
ikhoon: Thanks. fixed it.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/ScheduledPurgeService.java
```diff
@@ -36,64 +39,131 @@
...
+    @VisibleForTesting
+    void cleanPurgedFiles(CommandExecutor commandExecutor) {
+        projectManager.listPurged()
+                      .forEach(name -> commandExecutor.execute(Command.purgeProject(Author.SYSTEM, name)));
+        projectManager.list().values()
+                      .forEach(project -> project
+                              .repos()
+                              .listPurged()
+                              .forEach(repoName -> commandExecutor.execute(
+                                      Command.purgeRepository(Author.SYSTEM, project.name(), repoName))));
```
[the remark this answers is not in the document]
ikhoon: Without issuing `CommandExecutor`, directly [clear purged files](https://github.com/line/centraldogma/pull/416/files#diff-5bbabea077125f79a5bf3fc645623307R117)

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/ScheduledPurgeService.java
```diff
@@ -36,64 +39,131 @@
...
-        storagePurgeScheduler.stop();
+        if (!isDisabled()) {
+            storagePurgeScheduler.stop();
+        }
+    }
+
+    @VisibleForTesting
+    void cleanPurgedFiles(CommandExecutor commandExecutor) {
+        projectManager.listPurged()
+                      .forEach(name -> commandExecutor.execute(Command.purgeProject(Author.SYSTEM, name)));
```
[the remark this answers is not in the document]
ikhoon: Without using `CommandExecutor`, directly [clear purged files]( https://github.com/line/centraldogma/pull/416/files#diff-5bbabea077125f79a5bf3fc645623307R117].

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/ScheduledPurgeService.java
ikhoon: I changed `schedulePurgeData` into synchronous and sequential logic. It doesn't return CompletableFuture anymore. :-)

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/ScheduledPurgingService.java
```diff
@@ -0,0 +1,252 @@
...
+import com.linecorp.centraldogma.common.Author;
+import com.linecorp.centraldogma.internal.Util;
+import com.linecorp.centraldogma.server.command.Command;
+import com.linecorp.centraldogma.server.command.CommandExecutor;
+import com.linecorp.centraldogma.server.metadata.MetadataService;
+import com.linecorp.centraldogma.server.metadata.UserAndTimestamp;
+import com.linecorp.centraldogma.server.storage.project.ProjectManager;
+
+/**
+ * A service class for pursing projects and repositories that were removed before.
```
trustin: purging

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/ScheduledPurgingService.java
```diff
@@ -0,0 +1,252 @@
...
+ */
+public class ScheduledPurgingService {
```
trustin: `PurgeSchedulingService`?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/ScheduledPurgingService.java
```diff
@@ -0,0 +1,252 @@
...
+                   .forEach(repoName -> cleanPurgedFile(repoName, "repository"));
+        });
+    }
+
+    @VisibleForTesting
+    void cleanPurgedFile(String name, String typeName) {
+        final File purged = new File(rootDir, name + SUFFIX_PURGED);
+        CompletableFuture.supplyAsync(() -> {
+            try {
+                Util.deleteFileTree(purged);
```
trustin: How about:

- Adding some log message around `deleteFileTree()` at INFO just in case it takes long time?
- de-duplicating the same logic in `DirectoryBasedStorageManager`?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/ScheduledPurgingService.java
```diff
@@ -0,0 +1,252 @@
...
+                .forEach((unused, project) -> project
+                        .repos()
+                        .listRemoved()
+                        .stream()
+                        .map(repoName -> metadataService.getRepo(project.name(), repoName).join())
+                        .forEach(metadata -> {
+                            if (olderThanMinAllowed.test(metadata.removal())) {
+                                commandExecutor.execute(
+                                        Command.purgeRepository(Author.SYSTEM, project.name(), metadata.name()))
+                                               .join();
```
trustin: Could be indented this way for aesthetics:

```java
commandExecutor.execute(Command.purgeRepository(Author.SYSTEM,
                                                project.name(),
                                                metadata.name()))
               .join();
```

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/ScheduledPurgingServicePlugin.java
```diff
@@ -0,0 +1,87 @@
...
+
+import com.google.common.base.MoreObjects;
+
+import com.linecorp.centraldogma.server.CentralDogmaConfig;
+import com.linecorp.centraldogma.server.metadata.MetadataService;
+import com.linecorp.centraldogma.server.plugin.Plugin;
+import com.linecorp.centraldogma.server.plugin.PluginContext;
+import com.linecorp.centraldogma.server.plugin.PluginTarget;
+
+public final class ScheduledPurgingServicePlugin implements Plugin {
```
trustin: `PurgeSchedulingServicePlugin`?

server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaBuilder.java
```diff
@@ -295,6 +298,28 @@ public CentralDogmaBuilder numRepositoryWorkers(int numRepositoryWorkers) {
         return this;
     }
 
+    /**
+     * Sets the maximum allowed age of removed projects and repositories before they are purged.
+     * Sets to 0 to disable automatic purge.
+     * If unspecified, the default of {@value #DEFAULT_MAX_REMOVED_REPOSITORY_AGE_MILLIS} is used.
```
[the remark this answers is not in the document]
ikhoon: Thanks!

server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaBuilder.java
```diff
@@ -295,6 +298,28 @@ public CentralDogmaBuilder numRepositoryWorkers(int numRepositoryWorkers) {
...
+    public CentralDogmaBuilder maxRemovedRepositoryAge(Duration maxRemovedRepositoryAge) {
+        maxRemovedRepositoryAgeMillis(
+                requireNonNull(maxRemovedRepositoryAge, "maxRemovedRepositoryAge").toMillis());
+        return this;
+    }
+
+    /**
+     * Sets the maximum allowed age, in milliseconds of removed projects and repositories
+     * before they are purged.
+     * Sets to 0 to disable automatic purge.
```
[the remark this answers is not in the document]
ikhoon: I updated it.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/ScheduledPurgingService.java
ikhoon: I also have concerned about duplication logic. 
I made a new interface `void clear(String name)` to permanently delete a specified file.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/ScheduledPurgingService.java
ikhoon: I like aesthetics, too.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/ScheduledPurgingServicePlugin.java
ikhoon: fixed

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/DirectoryBasedStorageManager.java
```diff
@@ -326,37 +331,39 @@ public void purge(String name) {
...
+    private void deletePurgedFile(File file) {
         try {
-            logger.info("Deleting {} file: {} ..", childTypeName, file);
+            logger.info("Deleting a purged {}: {} ..", childTypeName, file);
             Util.deleteFileTree(file);
-            logger.info("Deleted {} file: {}.", childTypeName, file);
+            logger.info("Deleted a purged {}: {}.", childTypeName, file);
         } catch (IOException e) {
-            logger.warn("Failed to delete " + childTypeName + ": {}", file);
+            logger.warn("Failed to delete a purged {}: {}.", childTypeName, file);
```
trustin: Append `e` to log the stack trace?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/DirectoryBasedStorageManager.java
```diff
@@ -326,37 +331,39 @@ public void purge(String name) {
...
     }
 
-    @Override
-    public Set<String> listPurged() {
-        return listWithSuffix(SUFFIX_PURGED);
-    }
-
-    private Set<String> listWithSuffix(String suffix) {
+    @VisibleForTesting
+    Set<String> listWithSuffix(String suffix) {
```
trustin: This could be merged back into `listRemoved()` now?

server/src/main/java/com/linecorp/centraldogma/server/storage/StorageManager.java
```diff
@@ -96,23 +96,16 @@ default T create(String name, Author author) {
...
     /**
-     * Returns a set of names for the elements which have been purged.
+     * Purges a set of names for the elements which have been marked for purging.
      */
-    Set<String> listPurged();
+    void purgeMarked();
 
     /**
-     * Purges an element with the specified {@code name}.
+     * Marks for purging an element with the specified {@code name}.
```
trustin: Marks an element with the specified name for purge.

server/src/main/java/com/linecorp/centraldogma/server/storage/StorageManager.java
```diff
@@ -96,23 +96,16 @@ default T create(String name, Author author) {
...
      */
-    void purge(String name);
-
-    /**
-     * Clears an element permanently with the specified {@code name}.
-     *
-     * @param name the name of an element which is supposed to be cleared
-     */
-    void clear(String name);
+    void markForPurge(String name);
```
trustin: I think we need to be consistent with the documentation: `markForPurging` or `markForPurge`.

ikhoon: I chose `markForPurge` because it is shorter and simple. 😁
For synchronizing with that, the documentation was updated.

server/src/main/java/com/linecorp/centraldogma/server/storage/StorageManager.java
ikhoon: `purge` looks better. Thanks. :-)

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/DirectoryBasedStorageManager.java
```diff
@@ -267,7 +267,34 @@ private T create0(Author author, String name, long creationTimeMillis) {
...
     public Set<String> listRemoved() {
-        return listWithSuffix(SUFFIX_REMOVED);
+        ensureOpen();
+        final File[] files = rootDir.listFiles();
+        if (files == null) {
+            return ImmutableSet.of();
+        }
+
+        Arrays.sort(files);
+        final Builder<String> builder = ImmutableSet.builder();
```
trustin: nit: How about using `ImmutableSet.Builder`. Importing `Builder` may cause pain when we start to use other types like `ImmutableList` later.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/DirectoryBasedStorageManager.java
```diff
@@ -358,40 +385,8 @@ private void deletePurgedFile(File file) {
...
-        }
-    }
-
-    @VisibleForTesting
-    Set<String> listWithSuffix(String suffix) {
-        ensureOpen();
-        final File[] files = rootDir.listFiles();
-        if (files == null) {
-            return ImmutableSet.of();
+            logger.warn("Failed to delete a purged " + childTypeName + ": " + file, e);
```
trustin:
```java
logger.warn("Failed to delete a purged {}: {}.", childTypeName, file);
```

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/project/DefaultProjectManager.java
```diff
@@ -41,12 +41,13 @@
...
-    public DefaultProjectManager(File rootDir, Executor repositoryWorker, MeterRegistry meterRegistry,
-                                 @Nullable String cacheSpec) {
-        super(rootDir, Project.class);
+    public DefaultProjectManager(File rootDir, Executor repositoryWorker, Executor purgeWorker,
+                                 MeterRegistry meterRegistry, @Nullable String cacheSpec) {
+        super(rootDir, Project.class, purgeWorker);
 
         requireNonNull(meterRegistry, "meterRegistry");
         requireNonNull(repositoryWorker, "repositoryWorker");
+        requireNonNull(purgeWorker, "purgeWorker");
```
trustin: This is redundant because the super constructor will check nullness.

server/src/test/java/com/linecorp/centraldogma/server/internal/storage/PurgeSchedulingServiceTest.java
```diff
@@ -0,0 +1,127 @@
...
+import com.linecorp.centraldogma.server.command.PurgeRepositoryCommand;
+import com.linecorp.centraldogma.server.metadata.MetadataService;
+import com.linecorp.centraldogma.server.metadata.ProjectMetadata;
+import com.linecorp.centraldogma.server.metadata.RepositoryMetadata;
+import com.linecorp.centraldogma.server.metadata.UserAndTimestamp;
+import com.linecorp.centraldogma.server.storage.project.Project;
+import com.linecorp.centraldogma.server.storage.project.ProjectManager;
+import com.linecorp.centraldogma.server.storage.repository.RepositoryManager;
+
+public class PurgeSchedulingServiceTest {
```
trustin: I think we are using mocks excessively here. How about using `ProjectManagerRule`, which will give a fully working `ProjectManager` implementation? See `MissingRepositoryMetadataTest` for an example.

server/src/test/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepositoryManagerTest.java
```diff
@@ -138,6 +139,6 @@ public void testHas() throws IOException {
 
     private GitRepositoryManager newRepositoryManager() {
         return new GitRepositoryManager(mock(Project.class), rootDir(), ForkJoinPool.commonPool(),
-                                        mock(RepositoryCache.class));
+                                        Executors.newSingleThreadExecutor(), mock(RepositoryCache.class));
```
trustin: If the `purgeWorker` is used really, it can lead to a dangling thread. How about using `MoreExecutors.directExecutor()`?

server/src/test/java/com/linecorp/centraldogma/server/internal/storage/repository/DefaultMetaRepositoryTest.java
```diff
@@ -89,7 +90,7 @@
     @BeforeClass
     public static void init() throws Exception {
         pm = new DefaultProjectManager(rootDir.getRoot(), ForkJoinPool.commonPool(),
-                                       NoopMeterRegistry.get(), null);
+                                       Executors.newSingleThreadExecutor(), NoopMeterRegistry.get(), null);
```
trustin: How about using `MoreExecutors.directExecutor()` to avoid a dangling thread?

server/src/test/java/com/linecorp/centraldogma/server/internal/storage/PurgeSchedulingServiceTest.java
```diff
@@ -0,0 +1,127 @@
...
+
+    @Test
+    public void testScheduledPurgingService() throws Exception {
+        final CommandExecutor commandExecutor = mock(CommandExecutor.class);
+        final ProjectManager projectManager = mock(ProjectManager.class);
+        final MetadataService metadataService = mock(MetadataService.class);
+        final ProjectMetadata projectMetadata = mock(ProjectMetadata.class);
+        final UserAndTimestamp userAndTimestamp = mock(UserAndTimestamp.class);
+        final ScheduledExecutorService purgeWorker = Executors.newSingleThreadScheduledExecutor();
```
trustin: How about using `MoreExecutors.directExecutor()`?

testing-internal/src/main/java/com/linecorp/centraldogma/testing/internal/ProjectManagerRule.java
```diff
@@ -73,9 +74,10 @@ public CommandExecutor executor() {
     protected final void before() throws Throwable {
         super.before();
 
-        final Executor worker = newWorker();
-        projectManager = newProjectManager(worker);
-        executor = newCommandExecutor(projectManager, worker);
+        final Executor repositoryWorker = newWorker();
+        final Executor purgeWorker = Executors.newSingleThreadExecutor();
```
trustin: - You need to stop it in `after()`.
- Please use a daemon thread.

ikhoon:
We have discussed that asynchronous purging logic could crash replication just in case:
```
1. createProject(foo)
2. removeProject(foo)
3. purgeProject(foo)
3.1 purgeWoker.execute(() -> {
    deleteFile(foo)  // this runs asynchronously and `foo.purge` don't deleted yet
})
4. createProject(foo) -> success
5. removeProject(foo) -> success
6. purgeProject(foo)  -> fail because `foo.purge` is aleady existed. 
```

Thanks. @trustin 😀

testing-internal/src/main/java/com/linecorp/centraldogma/testing/internal/ProjectManagerRule.java
ikhoon: fix it :-)

server/src/test/java/com/linecorp/centraldogma/server/internal/storage/PurgeSchedulingServiceTest.java
ikhoon: fix it :-)

server/src/test/java/com/linecorp/centraldogma/server/internal/storage/repository/DefaultMetaRepositoryTest.java
ikhoon: fix it :-)

server/src/test/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepositoryManagerTest.java
ikhoon: fix it :-)

server/src/test/java/com/linecorp/centraldogma/server/internal/storage/PurgeSchedulingServiceTest.java
ikhoon: Thanks for guiding me! The test codes have been fully rewritten using `ProjectManagerRule`. 😄

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/project/DefaultProjectManager.java
ikhoon: Oops, thanks. You have an eagle eye!  👀

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/DirectoryBasedStorageManager.java
ikhoon: That is a nice idea. The implementation has been changed a little so I chose `ImmutableMap.Builder`.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/DirectoryBasedStorageManager.java
```diff
@@ -228,7 +243,7 @@ private T create0(Author author, String name, long creationTimeMillis) {
                 try {
                     Util.deleteFileTree(f);
                 } catch (IOException e) {
-                    logger.warn("Failed to delete a partially created project: {}", f);
+                    logger.warn("Failed to delete a partially created project: " + f, e);
```
trustin: `{}", f, e`?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/DirectoryBasedStorageManager.java
```diff
@@ -323,6 +341,60 @@ public T unremove(String name) {
...
+            return new File(rootDir, name + interfix + SUFFIX_PURGED);
+        };
+
+        synchronized (this) {
+            if (!removed.exists() || !removed.isDirectory()) {
+                throw newStorageNotFoundException(name + SUFFIX_REMOVED);
+            }
+
+            marked = newMarkedFile.get();
+            while (!removed.renameTo(marked)) {
```
trustin: I think it's possible to stuck here infinitely. How about using [`Files.move()`](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#move-java.nio.file.Path-java.nio.file.Path-java.nio.file.CopyOption...-) and retrying only on `FileAlreadyExistsException`?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/DirectoryBasedStorageManager.java
```diff
@@ -323,6 +341,60 @@ public T unremove(String name) {
...
+        }
+    }
+
+    private void deletePurgedFile(File file) {
+        try {
+            logger.info("Deleting a purged {}: {} ..", childTypeName, file);
+            Util.deleteFileTree(file);
+            logger.info("Deleted a purged {}: {}.", childTypeName, file);
+        } catch (IOException e) {
+            logger.warn("Failed to delete a purged " + childTypeName + ": " + file, e);
```
trustin: `"Failed to delete a purged {}: {}", childTypeName, file, e`

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/DirectoryBasedStorageManager.java
```diff
@@ -347,11 +419,51 @@ private static boolean isValidChildName(String name) {
...
     }
 
     @Override
     public String toString() {
         return Util.simpleTypeName(getClass()) + '(' + rootDir + ')';
     }
+
+    private final class StorageRemovalManager {
+
+        private static final String REMOVAL_TIMESTAMP_NAME = ".removal.timestamp";
```
trustin: I think we do not need to make this a hidden file. How about just `removal.timestamp`?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/DirectoryBasedStorageManager.java
```diff
@@ -347,11 +419,51 @@ private static boolean isValidChildName(String name) {
...
+                throw new StorageException(
+                        "failed to write a removal timestamp " + childTypeName + ": " + removal);
+            }
+        }
+
+        void unmark(File file) {
+            final File removal = new File(file, REMOVAL_TIMESTAMP_NAME);
+            if (removal.exists()) {
+                if (!removal.delete()) {
+                    logger.warn("Failed to delete a removal timestamp {}: {}", childTypeName, removal);
```
trustin: `a removal timestamp for {}: {}` ?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/DirectoryBasedStorageManager.java
```diff
@@ -347,11 +419,51 @@ private static boolean isValidChildName(String name) {
...
+        Instant readRemoval(File file) {
+            final File removal = new File(file, REMOVAL_TIMESTAMP_NAME);
+            if (!removal.exists()) {
+                return Instant.ofEpochMilli(file.lastModified());
+            }
+            try {
+                final String timestamp = new String(Files.readAllBytes(removal.toPath()),
+                                                    StandardCharsets.UTF_8);
+                return Instant.from(DateTimeFormatter.ISO_INSTANT.parse(timestamp));
+            } catch (IOException e) {
```
trustin: `Exception`? (We should catch a `DateTimeParseException`.)

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/DirectoryBasedStorageManager.java
```diff
@@ -347,11 +419,51 @@ private static boolean isValidChildName(String name) {
...
+                logger.warn("Failed to read a removal timestamp " + childTypeName + ": " + removal, e);
```
trustin: `a removal timestamp for {}: {}`

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/PurgeSchedulingService.java
```diff
@@ -0,0 +1,213 @@
...
+        this.projectManager = requireNonNull(projectManager, "projectManager");
+        this.purgeWorker = requireNonNull(purgeWorker, "purgeWorker");
+        checkArgument(maxRemovedRepositoryAgeMillis >= 0,
+                      "maxRemovedRepositoryAgeMillis: %s (expected: >= 0)", maxRemovedRepositoryAgeMillis);
+        this.maxRemovedRepositoryAgeMillis = maxRemovedRepositoryAgeMillis;
+    }
+
+    public void start(CommandExecutor commandExecutor,
+                      MetadataService metadataService) {
+        if (!isDisabled()) {
```
trustin: nit: How about doing an early return when `isDisabled()` for less indentation?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/PurgeSchedulingService.java
```diff
@@ -0,0 +1,213 @@
...
+                    task,
+                    TICK.getSeconds(), TICK.getSeconds(), TimeUnit.SECONDS);
+
+            Futures.addCallback(future, new FutureCallback<Object>() {
+                @Override
+                public void onSuccess(@Nullable Object result) {}
+
+                @Override
+                public void onFailure(Throwable cause) {
+                    logger.error("storage purge scheduler stopped due to an unexpected exception:", cause);
```
trustin: storage -> Storage

server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java
```diff
@@ -322,6 +322,22 @@ public MetadataService(ProjectManager projectManager, CommandExecutor executor)
...
+     */
+    public CompletableFuture<Revision> purgeRepo(Author author, String projectName, String repoName) {
+        requireNonNull(author, "author");
+        requireNonNull(projectName, "projectName");
+        requireNonNull(repoName, "repoName");
+
+        final JsonPointer path = JsonPointer.compile("/repos" + encodeSegment(repoName));
+        final Change<JsonNode> change = Change.ofJsonPatch(METADATA_JSON,
+                                                           new RemoveOperation(path).toJsonNode());
+        final String commitSummary = "Purge the repo '" + repoName + "' from the project " + projectName;
```
trustin: nit: Could we add quotes for project names as well throughout this class?

server/src/main/resources/META-INF/services/com.linecorp.centraldogma.server.plugin.Plugin
```diff
@@ -1 +1,2 @@
 com.linecorp.centraldogma.server.internal.mirror.DefaultMirroringServicePlugin
+com.linecorp.centraldogma.server.internal.storage.PurgeSchedulingServicePlugin
```
trustin: Missing EOL?

server/src/test/java/com/linecorp/centraldogma/server/internal/api/auth/PermissionTest.java
```diff
@@ -98,7 +99,9 @@ public static void cleanup() throws Exception {
         @Override
         protected void configure(ServerBuilder sb) throws Exception {
             final ProjectManager pm = new DefaultProjectManager(
-                    rootDir.newFolder(), ForkJoinPool.commonPool(), NoopMeterRegistry.get(), null);
+                    rootDir.newFolder(), ForkJoinPool.commonPool(), Executors.newSingleThreadExecutor(),
```
trustin: Probably OK to use `MoreExecutors.directExecutor()`?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/DirectoryBasedStorageManager.java
ikhoon: IIUC, slf4j does support `logger.warn(format, arg, throwable)` api, so I chose [`warn​(String msg, Throwable t)`](https://www.slf4j.org/api/org/slf4j/Logger.html#warn(java.lang.String,java.lang.Throwable)).

https://www.slf4j.org/api/org/slf4j/Logger.html
FYI

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/DirectoryBasedStorageManager.java
ikhoon: Thanks, it is a nice idea.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/DirectoryBasedStorageManager.java
ikhoon: same reason with https://github.com/line/centraldogma/pull/416#discussion_r305659075

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/DirectoryBasedStorageManager.java
trustin: If you try, you'll realize it will just work. If the last argument in `Object...` is a `Throwable` and it has no matching `{}`, its stack trace will be logged.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/DirectoryBasedStorageManager.java
ikhoon: Thanks, I will change this into a regular file. :-)

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/DirectoryBasedStorageManager.java
ikhoon: Thanks, I will keep in mind.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/PurgeSchedulingService.java
ikhoon: It looks good!

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/PurgeSchedulingService.java
ikhoon: Thanks!

server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java
ikhoon: Sure! 😀

server/src/test/java/com/linecorp/centraldogma/server/internal/api/auth/PermissionTest.java
ikhoon: Sorry, I missed it. 😟

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/DirectoryBasedStorageManager.java
ikhoon: Ahh............ I didn't know that. I found the code which extracts throwable from the arguments.

```java
final Object lastEntry = argArray[argArray.length - 1];
if (lastEntry instanceof Throwable) {
    return (Throwable) lastEntry;
}
```
https://github.com/qos-ch/logback/blob/master/logback-classic/src/main/java/ch/qos/logback/classic/spi/EventArgUtil.java#L23-L25

Thanks for kind advice. 🙏

ikhoon: While seeing the `logback` codes, I thought it seems to always extract the last argument exception without matching `{}`.

I tried some case:
```java
logger.warn("hello {}", new Exception());
// [WARN] hello {}                    <- ignore matching and print stack trace
// java.lang.Exception: null
logger.warn("hello {}", "world", new Exception());
// [WARN] hello world                 
// java.lang.Exception: null
logger.warn("hello {}", new Exception(), "world");
// [WARN] hello java.lang.Exception    <- {} matchs with Exception and no stack trace
```

trustin:
Thanks for your patience and nice work, @ikhoon :+1:

ikhoon:
the PR description has updated.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/PurgeSchedulingService.java
```diff
@@ -0,0 +1,214 @@
...
+            return scheduler != null;
+        }
+
+        public synchronized void start(Runnable task) {
+            if (isStarted()) {
+                return;
+            }
+            requireNonNull(task, "task");
+            scheduler = MoreExecutors.listeningDecorator(purgeWorker);
+            final ListenableScheduledFuture<?> future = scheduler.scheduleWithFixedDelay(
```
minwoox: How about:
```
final ListeningScheduledExecutorService scheduler = MoreExecutors.listeningDecorator(purgeWorker);
this.scheduler = scheduler;
scheduler.scheduleWithFixedDelay( ...)
```
not to access the volatile field twice?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/PurgeSchedulingService.java
```diff
@@ -0,0 +1,214 @@
...
+                final boolean interrupted = terminate(scheduler);
+                if (interrupted) {
+                    Thread.currentThread().interrupt();
+                }
+            } finally {
+                this.scheduler = null;
+            }
+        }
+
+        private boolean terminate(ExecutorService executor) {
```
minwoox: `@nullable`?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/PurgeSchedulingService.java
```diff
@@ -0,0 +1,214 @@
...
+
+        public synchronized void start(Runnable task) {
+            if (isStarted()) {
+                return;
+            }
+            requireNonNull(task, "task");
+            scheduler = MoreExecutors.listeningDecorator(purgeWorker);
+            final ListenableScheduledFuture<?> future = scheduler.scheduleWithFixedDelay(
+                    task,
+                    TICK.getSeconds(), TICK.getSeconds(), TimeUnit.SECONDS);
```
minwoox: Do we need the initial delay for this?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/PurgeSchedulingService.java
```diff
@@ -0,0 +1,214 @@
...
+        }
+    }
+
+    private void cleanPurgedFiles() {
+        projectManager.purgeMarked();
+        projectManager.list().forEach((unused, project) -> project.repos().purgeMarked());
+    }
+
+    @VisibleForTesting
+    void schedulePurgeData(CommandExecutor commandExecutor,
```
minwoox: I think this is not scheduling but removing. How about renaming to just `purgeData` or `purgeProjectAndRepository`?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/PurgeSchedulingServicePlugin.java
```diff
@@ -0,0 +1,86 @@
...
+
+    @Nullable
+    public PurgeSchedulingService scheduledPurgingService() {
+        return purgeSchedulingService;
+    }
+
+    @Override
+    public String toString() {
+        return MoreObjects.toStringHelper(this)
+                          .add("target", target())
```
minwoox: Don't need to add `purgeSchedulingService`?
`toStringHelper(this).omitNullValues().add("purgeSchedulingService"...`

server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java
```diff
@@ -318,7 +320,25 @@ public MetadataService(ProjectManager projectManager, CommandExecutor executor)
...
                                                new AddOperation(path, Jackson.valueToTree(
                                                        UserAndTimestamp.of(author)))));
-        final String commitSummary = "Remove the repo '" + repoName + "' from the project " + projectName;
+        final String commitSummary =
+                "Remove the repo '" + repoName + "' from the project '" + projectName + '\'';
+        return metadataRepo.push(projectName, Project.REPO_DOGMA, author, commitSummary, change);
+    }
+
+    /**
+     * Purge a {@link RepositoryMetadata} of the specified {@code repoName} from the specified
```
minwoox: nit: Purges

server/src/main/java/com/linecorp/centraldogma/server/plugin/PluginContext.java
```diff
@@ -42,15 +45,18 @@
...
                          ProjectManager projectManager,
                          CommandExecutor commandExecutor,
-                         MeterRegistry meterRegistry) {
+                         MeterRegistry meterRegistry,
+                         ScheduledExecutorService purgeWorker) {
         this.config = requireNonNull(config, "config");
         this.projectManager = requireNonNull(projectManager, "projectManager");
         this.commandExecutor = requireNonNull(commandExecutor, "commandExecutor");
         this.meterRegistry = requireNonNull(meterRegistry, "meterRegistry");
+        this.purgeWorker = purgeWorker;
```
minwoox: `requireNonNull` for consistency?

server/src/main/java/com/linecorp/centraldogma/server/storage/StorageManager.java
```diff
@@ -95,6 +95,18 @@ default T create(String name, Author author) {
      */
     T unremove(String name);
 
+    /**
+     * Purges a set of names for the elements which have been marked for purging.
+     */
+    void purgeMarked();
+
+    /**
+     * Marks an element with the specified {@code name} for purge.
```
minwoox: Marks the specified {@code name} element for purging?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/DirectoryBasedStorageManager.java
```diff
@@ -347,11 +428,51 @@ private static boolean isValidChildName(String name) {
...
+                if (!removal.delete()) {
+                    logger.warn("Failed to delete a removal timestamp for {}: {}", childTypeName, removal);
+                }
+            }
+        }
+
+        Instant readRemoval(File file) {
+            final File removal = new File(file, REMOVAL_TIMESTAMP_NAME);
+            if (!removal.exists()) {
+                return Instant.ofEpochMilli(file.lastModified());
```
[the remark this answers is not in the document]
ikhoon: If a file has `.removed` postfix but no `REMOVAL_TIMESTAMP`, we determine the removal time by reading `lastModified()` to keep backward compatibility because old removed files don't have `REMOVAL_TIMESTAMP` before deploying this new purge logic.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/PurgeSchedulingService.java
ikhoon: I agree, `purgeProjectAndRepository` looks good for me. Thanks.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/PurgeSchedulingServicePlugin.java
ikhoon: I thought `purgeScheduingService` does not need to be added because `purgeSchedulingService` don't override `toString()` method.
But whether `purgeSchedulingService` is null or not seems to make sense.

server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java
ikhoon: Thanks

server/src/main/java/com/linecorp/centraldogma/server/storage/StorageManager.java
ikhoon: We were supposed to use `markForPurge`. https://github.com/line/centraldogma/pull/416#discussion_r303713795
I'm going to change `element for purge`. :-)

client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogma.java
```diff
@@ -263,6 +286,28 @@ private static Void removeRepository(AggregatedHttpResponse res) {
...
+            return client.execute(headers(HttpMethod.DELETE,
+                                          pathBuilder(projectName, repositoryName).append(REMOVED).toString()))
+                         .aggregate()
+                         .thenApply(ArmeriaCentralDogma::purgeRepository);
+        } catch (Exception e) {
+            return exceptionallyCompletedFuture(e);
+        }
+    }
+
+    private static Void purgeRepository(AggregatedHttpResponse res) {
```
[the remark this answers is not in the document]
ikhoon: Ok, It's good.

server/src/main/java/com/linecorp/centraldogma/server/command/Command.java
```diff
@@ -193,6 +218,33 @@
         return new UnremoveRepositoryCommand(timestamp, author, projectName, repositoryName);
     }
 
+    /**
+     * Returns a new {@link Command} which is used to purge a repository.
+     *
+     * @param author the author who is removing the repository
+     * @param projectName the name of the project
+     * @param repositoryName the name of the repository which is supposed to be purge
```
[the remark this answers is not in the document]
ikhoon: Thanks

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/PurgeSchedulingService.java
ikhoon: It is a good idea.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/PurgeSchedulingService.java
ikhoon: I don't have a strong opinion here. I just followed how other schedulers work. 
This service is not time-critical. But I think slow starting take advantage of avoiding the busy time of server startup.

server/src/main/java/com/linecorp/centraldogma/server/plugin/PluginContext.java
ikhoon: Thanks.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/PurgeSchedulingService.java
minwoox: That makes sense. Thanks for the explanation!

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirroringServicePlugin.java
```diff
@@ -80,6 +80,7 @@ public DefaultMirroringService mirroringService() {
     @Override
     public String toString() {
         return MoreObjects.toStringHelper(this)
+                          .omitNullValues()
```
minwoox: `target()` doesn't return `null` so we don't have to add `omitNullValues`.

client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogma.java
```diff
@@ -181,6 +182,28 @@ private static Void removeProject(AggregatedHttpResponse res) {
...
+            return client.execute(
+                    headers(HttpMethod.DELETE, pathBuilder(projectName).append(REMOVED).toString()))
+                         .aggregate()
+                         .thenApply(ArmeriaCentralDogma::purgeProject);
+        } catch (Exception e) {
+            return exceptionallyCompletedFuture(e);
+        }
+    }
+
+    private static Void purgeProject(AggregatedHttpResponse res) {
```
minwoox: so we can remove this method?

server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaBuilder.java
```diff
@@ -295,6 +298,28 @@ public CentralDogmaBuilder numRepositoryWorkers(int numRepositoryWorkers) {
...
+    public CentralDogmaBuilder maxRemovedRepositoryAge(Duration maxRemovedRepositoryAge) {
+        maxRemovedRepositoryAgeMillis(
+                requireNonNull(maxRemovedRepositoryAge, "maxRemovedRepositoryAge").toMillis());
+        return this;
+    }
+
+    /**
+     * Sets the maximum allowed age, in milliseconds of removed projects and repositories
+     * before they are purged.
+     * Set 0 to disable automatic purge.
```
minwoox: nit: `{@code 0}`?

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirroringServicePlugin.java
ikhoon: Oops, it is added accidentally. Thanks for catching it.

server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaBuilder.java
ikhoon: Oops, sorry for missing it. :-)

client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogma.java
ikhoon: Ah, ok!

trustin:
:tada: Great job!

PR #417, opened 2019-07-05
Do not fail `GitMirrorAuthTest` when failed to load an SSH key

trustin:
Motivation:

- A contributor might see `GitMirrorAuthTest` fails if he or she has
  an SSH key which cannot be loaded by jSch.
- Even if a contributor specified correct System properties,
  `GitMirrorAuthTest` will not test anything because `main` repository,
  the mirror destination, does not exist.

Modifications:

- Do not fail `GitMirrorAuthTest` if jSch failed to load an SSH key.
- Create the `main` repository before running a mirroring test to make
  sure `GitMirrorAuthTest` attempts to mirror.

Result:

- Build stability
- A contributor can test Git authentication (if configured as documented)


trustin:
Thanks for reviewing :smile:

PR #418, opened 2019-07-10
Fix API links in the web site

trustin:
Motivation:

At some point in the past, `:api:` and `:apiplural:` roles stopped to
work.

Modification:

- Do not monkey-patch `HTMLTranslator`.
- Define a node properly.

Result:

- API links are back.


trustin:
Basically same fix with Armeria's

PR #422, opened 2019-07-23
Fix XSS on a search feature

ikhoon:
Motivation:
XSS issue was reported by Line Security Team

Modifications:
* Encode html on error notification message.
* Invalid search query returns `Bad Request` instead of `Internal Server Error`.
* Replace `jGitRepository` into `parent.name() + '/' + name`
  to prevent exposure of a full file system path on the error message.

Result:
Enhanced security

### DIFF
--- a/site/src/sphinx/known-issues.rst
+++ b/site/src/sphinx/known-issues.rst
@@ -17,3 +17,18 @@ Known issues
   - Consider enforcing network-level access control over Thrift calls.
   - Note that the Thrift RPC layer is left only for backward compatibility, and will be removed in the future,
     in favor of the REST API.
+
+Previously known issues
+-----------------------
+- DOM based, Cross-site Scripting (XSS) was found in Central Dogma.
+  An attacker could exploit this by convincing an authenticated user to visit a specifically crafted URL on a CentralDogma server,
+  allowing for the execution of arbitrary scripts on the client-side browser, resulting to perform unauthorized actions.
+
+  - This issue affects: Central Dogma artifacts from 0.17.0 to 0.40.1.
+  - The impact: Attacker is able to have victim execute arbitrary JavaScript code in the browser.
+  - The component: Notification feature
+  - The attack vector: Victim must open a specifically crafted URL.
+  - The fixed version: 0.41.0 and later.
+  - Please check `CVE-2019-6002 <https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2019-6002>`_ to get more information.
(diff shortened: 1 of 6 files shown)


stypr:
I think `ngSanitize` is good enough as a quick fix.
But as I mentioned earlier, It would be better off to encode the text from the notification message itself.

This patch can still load image tags, but seems to filter out harmful attributes well. Also, it blocks `/`, therefore not able to load external domain

For example: https://centraldogma/#/projects/a'%22%3E%3Cimg%20src=1%3E

ikhoon:
@stypr [The notification UI library](https://github.com/alexcrack/angular-ui-notification#usage) we are using is able to embed HTML. I just tried to remove harm tag from HTML on error display to keep the flexibility of HTML. 
But I agree that sanitized HTML is still dangerous. I will fix it to encode all the notification text.

stypr:
Code seems good to go. Built and tested it too.
Test Code seems to work fine now. malicious tag is not rendered anymore.

ikhoon:
@stypr Thank for fast reviewing and feedback. 
As we talked in slack chat, unfortunately, we don’t have test codes and suits for frontend yet. If you need unit test code for this, it takes time to set up that.

The frontend is made of angular 1 which is under LTS, not actively developed anymore. 
We have a plan to migrate front end into a modern framework stack. With the migration, we will set up the front end tests.

stypr:
> As we have talked in slack chat, unfortunately, we don’t have test codes and suits for frontend yet. Thanks for the manual test.

For now, I think it should be ok to approach with the current process. we can decide later about making test cases for frontend side.

site/src/sphinx/known-issues.rst
```diff
@@ -17,3 +19,11 @@ Known issues
...
   - Note that the Thrift RPC layer is left only for backward compatibility, and will be removed in the future,
     in favor of the REST API.
+
+Previously known issues
+=======================
+- A Reflected Cross-site Scripting (XSS) vulnerability
+
+  - A Search feature of Central Dogma allows remote attacker to perform unauthorized actions.
+    This issue affects: Central Dogma 0.17.0 version 0.40.1 and prior versions.
+    Please check `CVE-2009-6002`_ to get more information.
```
trustin: Because the link will be used only once, you could just do an inline reference:

    `CVE-2009-6002 <https:....>`_

ikhoon: Thanks!

stypr: I think there is a typo. 

`CVE-2009-`  -> `CVE-2019`

ikhoon: fixed

PR #427, opened 2019-08-02
Update Git-to-CD mirroring documentation about private key

trustin:
Motivation:

In its recent version, OpenSSH does not generate an RSA key in PEM
format by default. The key pair generated with the default settings are
not recognized by JSch, causing a mysterious `invalid privatekey` error.

Modifications:

- Update the documentation to state that the private key must be an RSA
  key in PEM format and to show some useful commands for conversion and
  generation.

Result:

- A user does not get confused by the non-RSA PEM keypair he or she
  generated.

### DIFF
--- a/site/src/sphinx/mirroring.rst
+++ b/site/src/sphinx/mirroring.rst
@@ -162,24 +162,40 @@ repositories defined in ``/mirrors.json``:
 
 - ``publicKey`` (string)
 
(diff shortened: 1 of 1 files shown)


trustin:
Preview:

image

trustin:
Thanks for reviewing.

PR #431, opened 2019-08-21
Update dependencies

trustin:
- Armeria 0.90.2 -> 0.90.3
- JCommander 1.72 -> 1.75
- Build
  - Gradle 5.5.1 -> 5.6
  - gradle-docker-plugin 4.10.0 -> 5.0.0

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.5.1-all.zip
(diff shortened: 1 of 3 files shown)


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.5.1-all.zip
-distributionSha256Sum=6c4276f97c9059ef4dfb06cc7325c43ed9b933ab04f032e9d9372e8589cb6faf
```
minwoox: Dont need this?

trustin: Nice to have, but I got tired of adding it manually on every version up. Perhaps it's enough to use HTTPS?

minwoox: Yes I think so :)

PR #434, opened 2019-08-28
Implement client-side JSON filtering

mauhiz:
Hi, this is a contribution from our internal project.

The use case is :
- a JSON file contains several, mostly unrelated configuration blocks used by different components
- we don't want each component to maintain a separate watcher
- we don't want each component to be notified for the changes not related to it

This `FilteredJsonWatcher` enables to share a single actual watcher, while restricting the notifications to the concerned components only.


trustin:
Could you explain why `Query.ofJsonPath("/some.json", "$.b")` doesn't work?

mauhiz:
- A watcher can only watch a single query
- If each component has its own watcher, it increases the number of active connections to CD
- If components share a watcher with root query, they are notified for parts they don't care about

trustin:
If I understood correctly, the main purpose of `FilteredJsonWatcher` is to use a single watcher to watch different parts of a document and to reduce the number of connections. Correct?

Then, why not just reducing the number of event loops of Central Dogma client to 1 and watch as many times as each component wants? It would keep the number of connections as small as the number of replicas at worst case, unless you are falling back to HTTP/1.

trustin:
> we don't want each component to maintain a separate watcher

Or.. is it because you don't want to make each component manage the life cycle of each watcher? Then I guess it makes sense. :-)

client/java/src/main/java/com/linecorp/centraldogma/client/FilteredJsonWatcher.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    @Nonnull
+    private Latest<JsonNode> applyFilter(Latest<JsonNode> latestParentNode) {
+        return new Latest<>(latestParentNode.revision(), orMissing(latestParentNode.value()).at(jsonPath));
+    }
+
+    @Override
+    public void close() {
+        // do nothing. parent will take care of itself
```
trustin: Shouldn't we cancel the listeners we added in `watch()` below from `parent`?

client/java/src/main/java/com/linecorp/centraldogma/client/FilteredJsonWatcher.java
```diff
@@ -0,0 +1,89 @@
...
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.MissingNode;
+import com.linecorp.centraldogma.common.Revision;
+
+/**
+ * A watcher for a specific part of a JSON file, that piggy-rides on a watcher of the whole JSON file.
+ * It does not make extra requests, and it will not notify listeners of irrelevant changes.
+ */
+public class FilteredJsonWatcher implements Watcher<JsonNode> {
```
trustin: How about generalizing this a little bit like this:

```java
public interface Watcher<T> {
    ...
    // Maybe throw UnsupportedOperationException when the current Query is 
    // neither Query.ofJsonPath() nor Query.ofJson()?
    Watcher<T> newChild(String... additionalExpressions);
    Watcher<T> newChild(Iterable<String> additionalExpressions);
    <U> Watcher<U> newChild(Function<? super T, ? extends U> function,
                            String... additionalExpressions);
    <U> Watcher<U> newChild(Function<? super T, ? extends U> function,
                            Iterable<String> additionalExpressions);
}

Watcher<JsonNode> parent = dogma.fileWatcher(... Query.ofJson("/foo.json"));
Watcher<JsonNode> child1 = parent.newChild("$.a");
Watcher<Foo> child2 = parent.newChild(node -> new Foo(node), "$.b");
Watcher<JsonNode> grandChild = child1.newChild("$.c");
```

trustin: If `additionalExpressions` is empty, `newChild()` could be used for any `Watcher`s, even `Watcher<Revision>` from `CentralDogma.repositoryWatcher()`.

client/java/src/main/java/com/linecorp/centraldogma/client/FilteredJsonWatcher.java
mauhiz: I see. Yes, it would be cleaner

client/java/src/main/java/com/linecorp/centraldogma/client/FilteredJsonWatcher.java
mauhiz: I considered it and while the fluent API looks nice, I didn't like the fact that it doesn't make sense to refine the query for non-JSON (QueryType only supports IDENTITY and JSON_PATH only).

However, it makes sense to add query transformers (T -> U) that branch from the parent watcher.

Let me consider some refinement here

mauhiz: updated. Thanks for the idea

client/java/src/main/java/com/linecorp/centraldogma/client/TransformingWatcher.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.function.BiConsumer;
+import java.util.function.Function;
+
+import javax.annotation.Nullable;
+
+import com.linecorp.centraldogma.common.Revision;
+
+public class TransformingWatcher<T, U> implements Watcher<U> {
+
+    protected final Watcher<T> parent;
```
ikhoon: How about renaming `parent` into `delegate` or `origin`?
I don't think we have to put `parent` child relationship here. 
This `TransformingWatcher` just map or transform original watcher. 😀
What do you think @trustin @minwoox ?

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
```diff
@@ -178,4 +183,33 @@ default void watch(Consumer<? super T> listener) {
         requireNonNull(listener, "listener");
         watch((revision, value) -> listener.accept(value));
     }
+
+    /**
+     * @return A new watcher, that reuses the supplied watcher and applies transformation.
+     * Note: the returned watcher is effectively filtering in a sense that,
+     * its listeners are *not* notified when a change has no effect on the transformed value.
+     */
+    default <U> Watcher<U> newChild(@Nonnull Function<T, U> transformer) {
```
ikhoon: I prefer `map` or `transform` to `newChild`. I guess, at least this point, we don't need to put parent/child relationship here. 😀

client/java/src/main/java/com/linecorp/centraldogma/client/TransformingWatcher.java
trustin: How about:

- Adding `Watcher.mappedWatcher()` which is a static factory method.
- Renaming `newChild()` into `mappedWatcher()` and making it an overloaded static factory method (so that a user can always look for `mappedWatcher()` when he or she needs to transform.)

...?

client/java/src/main/java/com/linecorp/centraldogma/client/TransformingWatcher.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.function.BiConsumer;
+import java.util.function.Function;
+
+import javax.annotation.Nullable;
+
+import com.linecorp.centraldogma.common.Revision;
+
+public class TransformingWatcher<T, U> implements Watcher<U> {
```
trustin: This class could be package-local and final?

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
```diff
@@ -24,9 +24,14 @@
 import java.util.concurrent.TimeoutException;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
+import java.util.function.Function;
 
+import javax.annotation.Nonnull;
```
trustin: You don't need to use this annotation because everything is non-null by default here.

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
```diff
@@ -178,4 +183,33 @@ default void watch(Consumer<? super T> listener) {
...
+     * Note: the returned watcher is effectively filtering in a sense that,
+     * its listeners are *not* notified when a change has no effect on the transformed value.
+     */
+    default <U> Watcher<U> newChild(@Nonnull Function<T, U> transformer) {
+        return new TransformingWatcher<>(this, transformer);
+    }
+    /**
+     * @return A new child, which transformation is a JSON path query.
+     */
+    static Watcher<JsonNode> atJsonPath(@Nonnull Watcher<JsonNode> watcher, String jsonPath) {
```
trustin: Could accept `Iterable<String>` and `String...`

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
```diff
@@ -178,4 +183,33 @@ default void watch(Consumer<? super T> listener) {
...
+        return watcher.newChild(new Function<JsonNode, JsonNode>() {
+            @Override
+            public JsonNode apply(JsonNode node) {
+                if (node == null) {
+                    return MissingNode.getInstance();
+                } else {
+                    return node.at(jsonPath);
```
trustin: I guess this is not a JSON path expression?

client/java/src/main/java/com/linecorp/centraldogma/client/TransformingWatcher.java
ikhoon: > Renaming newChild() into mappedWatcher()

I like the name, thanks for fast feedback. 😊

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
mauhiz: I'm not sure what would such a filter return. An aggregate of several Json paths in a single Json object?

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
mauhiz: Right, this is a JSON pointer. I got them mixed up.
It looks like JSON path is more powerful, I will look into it

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
mauhiz: 👍

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
trustin: How about sticking to JSON pointer? (It's because JSON path implementation is relatively big and we don't want to introduce too many dependencies.)

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
trustin: Ah, I was walking about JSON path, but yeah, accepting a single JSON pointer is easier and has less dependencies. Could you just rename `jsonPath` to `jsonPointer`?

trustin:
Gentle ping, @mauhiz :wink:

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
mauhiz: 👍

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
mauhiz: 👍

client/java/src/main/java/com/linecorp/centraldogma/client/TransformingWatcher.java
mauhiz: I think the relationship is meaningful because the returned object is "forked" from the current watcher.
- Closing a forked watcher does nothing
- if transformer == identity, we still return a forked watcher that does not close its parent

WDYT?

trustin: > if transformer == identity, we still return a forked watcher that does not close its parent

SGTM.

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
```diff
@@ -178,4 +182,34 @@ default void watch(Consumer<? super T> listener) {
         requireNonNull(listener, "listener");
         watch((revision, value) -> listener.accept(value));
     }
+
+    /**
+     * @return A new watcher, that reuses the supplied watcher and applies transformation.
```
trustin: `@return A new watcher, that ...` -> `Returns a new {@link Watcher} that ...`

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
```diff
@@ -178,4 +182,34 @@ default void watch(Consumer<? super T> listener) {
...
+     * Note: the returned watcher is effectively filtering in a sense that,
+     * its listeners are *not* notified when a change has no effect on the transformed value.
+     */
+    default <U> Watcher<U> newChild(Function<T, U> transformer) {
+        return new TransformingWatcher<>(this, transformer);
+    }
+    /**
+     * @return A new child, which transformation is a
```
trustin: Ditto - We want every Javadoc to have a summary.

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
mauhiz: 👍

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
mauhiz: 👍

client/java/src/main/java/com/linecorp/centraldogma/client/TransformingWatcher.java
mauhiz: 👍sorry I has missed this comment. I've reduced the visibility of this class and made it final.

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
mauhiz: Or did you mean that it should not have a `@return` keyword? This was vague https://github.com/line/centraldogma/blob/master/CONTRIBUTING.md#add-javadoc

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
```diff
@@ -178,4 +182,35 @@ default void watch(Consumer<? super T> listener) {
...
+     * The returned watcher does not need to be closed after use.
+     */
+    default <U> Watcher<U> newChild(Function<T, U> transformer) {
+        return new TransformingWatcher<>(this, transformer);
+    }
+    /**
+     * @return A new child {@link Watcher}, which transformation is a
+     * <a href="http://tools.ietf.org/html/draft-ietf-appsawg-json-pointer-03">JSON pointer</a> query.
+     */
+    static Watcher<JsonNode> atJsonPointer(Watcher<JsonNode> watcher, String jsonPointer) {
```
minwoox: If you want to keep this method we should tell that the pointer should be [encoded](https://tools.ietf.org/html/rfc6901#section-3) in the Javadoc because we don't do that. 😉

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
```diff
@@ -178,4 +182,35 @@ default void watch(Consumer<? super T> listener) {
...
     }
+
+    /**
+     * @return A new {@link Watcher}, that reuses the supplied watcher and applies transformation.
+     * Note: the returned watcher is effectively filtering in a sense that,
+     * its listeners are *not* notified when a change has no effect on the transformed value.
+     * The returned watcher does not need to be closed after use.
+     */
+    default <U> Watcher<U> newChild(Function<T, U> transformer) {
+        return new TransformingWatcher<>(this, transformer);
```
minwoox: nit: `requireNonNull(transformer, "transformer");`

client/java/src/main/java/com/linecorp/centraldogma/client/TransformingWatcher.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    @Override
+    public void close() {
+        closed = true;
+        // do nothing else. We don't own the parent's lifecycle
+    }
+
+    @Override
+    public void watch(BiConsumer<? super Revision, ? super U> listener) {
```
minwoox: nit: `requireNonNull(listener, "listener");`

client/java/src/main/java/com/linecorp/centraldogma/client/TransformingWatcher.java
```diff
@@ -0,0 +1,87 @@
...
+                listener.accept(rev, transformedValue);
+            } // else, transformed value has not changed
+        });
+    }
+
+    @Override
+    public String toString() {
+        String me = getClass().getSimpleName() + " filtering " + parent + " with " + transformer;
+        if (closed) {
+            return me + " (closed)";
```
minwoox: How about using `StringBuilder`?

mauhiz: I don't think anyone will use this toString besides debuggers. Let me keep it simple.

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
trustin: Not really. I just mean that the summary must exist at least. `@return` is a bonus. :-)

client/java/src/main/java/com/linecorp/centraldogma/client/TransformingWatcher.java
trustin: How about just using Guava's `toStringHelper()` for consistency?

client/java/src/main/java/com/linecorp/centraldogma/client/TransformingWatcher.java
mauhiz: 👍

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
mauhiz: 👍

client/java/src/main/java/com/linecorp/centraldogma/client/TransformingWatcher.java
mauhiz: okay. I don't think the output is prettier but at least the code is consistent with other classes.

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
mauhiz: Good point. Added a comment

client/java/src/main/java/com/linecorp/centraldogma/client/TransformingWatcher.java
trustin: I guess this is not fixed yet?

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
```diff
@@ -178,4 +182,43 @@ default void watch(Consumer<? super T> listener) {
...
+     * @return A {@link Watcher} that is effectively filtering in a sense that,
+     * its listeners are *not* notified when a change has no effect on the transformed value.
+     * Furthermore, it does not need to be closed after use.
+     */
+    default <U> Watcher<U> newChild(Function<T, U> transformer) {
+        requireNonNull(transformer, "transformer");
+        return new TransformingWatcher<>(this, transformer);
+    }
+    /**
+     * Creates a forked watcher based on an existing {@link JsonNode}-watching {@link Watcher}.
```
trustin: watcher -> `{@link Watcher}`

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
```diff
@@ -178,4 +182,43 @@ default void watch(Consumer<? super T> listener) {
...
+     *
+     * @param jsonPointer should already be <a href="https://tools.ietf.org/html/rfc6901#section-3">encoded</a>
+     * if needed.
```
trustin: `@param jsonPointer a <a href="...">JSON pointer</a> that is encoded` ?

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
```diff
@@ -178,4 +182,43 @@ default void watch(Consumer<? super T> listener) {
...
+     *
+     * @return A new child {@link Watcher}, which transformation is a
+     * <a href="http://tools.ietf.org/html/draft-ietf-appsawg-json-pointer-03">JSON pointer</a> query.
```
trustin: which -> whose ?

trustin:
There are some Checkstyle violations:

```
[ant:checkstyle] [ERROR] /centraldogma/client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java:190: Line continuation have incorrect indentation level, expected level should be 4. [JavadocTagContinuationIndentation]

[ant:checkstyle] [ERROR] /centraldogma/client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java:191: Line continuation have incorrect indentation level, expected level should be 4. [JavadocTagContinuationIndentation]

[ant:checkstyle] [ERROR] /centraldogma/client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java:201: Line continuation have incorrect indentation level, expected level should be 4. [JavadocTagContinuationIndentation]

[ant:checkstyle] [ERROR] /centraldogma/client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java:206: 'METHOD_DEF' should be separated from previous statement. [EmptyLineSeparator]
```

minwoox:
Thanks @mauhiz!

mauhiz:
Thanks for following up on this PR!

PR #453, opened 2019-09-25
Allow colon for json path

jrhee17:
This is a PR for accepting colons for path names. (specifically, our use case is to create json files like `localhost:8080.json`)

Let me know if this is something you be open to accepting, and I will make this a serious PR w/ test cases, etc :)


minwoox:
I found this very old questions and answers that the colon is invalid in Window OS. Not sure it's still invalid. https://stackoverflow.com/questions/10386344/how-to-get-a-file-in-windows-with-a-colon-in-the-filename

ikhoon:
There is a similar issue on OSX

> A colon was the path delimiter in Classic Mac OS. Apple has kept it as a prohibited character for file names.

https://discussions.apple.com/thread/7540284

jrhee17:
didn't know that:D .      Let me close this PR.  I don't think it would be worth the effort to allow based on platform. Thanks for your comments

PR #458, opened 2019-10-11
Add more debug logging to `ReplicationLagTolerantCentralDogma`

trustin:
Motivation:

We need to add more logging to `ReplicationLagTolerantCentralDogma` to
figure out why some users are still getting `RevisionNotFoundException`.

Modifications:

- Log the call result or cause.
- Add logging to `updateLastestKnownRevision()`

Result:

- Maybe we can find out why some users are still getting `RevisionNotFoundException`.


minwoox:
Thanks, @trustin!

PR #463, opened 2019-11-28
Unwrap CompletionException to check actual thrown exception

imasahiro:
Motivation:
- An exception which is thrown at `CompletableFuture` callback handler will be wrapped with `CompletionException`.
- `ReplicationLagTolerantCentralDogma` does not retry when `CompletionException`-wrapped `RevisionNotFoundException` is thrown

Changes:
- Unwrap `CompletionException`

Results:
- Handle replication lag correctly

### DIFF
--- a/client/java/src/main/java/com/linecorp/centraldogma/internal/client/ReplicationLagTolerantCentralDogma.java
+++ b/client/java/src/main/java/com/linecorp/centraldogma/internal/client/ReplicationLagTolerantCentralDogma.java
@@ -26,6 +26,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.function.BiFunction;
@@ -39,6 +40,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableList;
 import com.spotify.futures.CompletableFutures;
 
@@ -681,6 +683,9 @@ public final class ReplicationLagTolerantCentralDogma extends AbstractCentralDog
             String projectName, String repositoryName, Revision revision, Throwable cause) {
 
         requireNonNull(cause, "cause");
+        if (cause instanceof CompletionException) {
+            cause = Throwables.getRootCause(cause);
+        }
 
         if (!(cause instanceof RevisionNotFoundException)) {
             return false;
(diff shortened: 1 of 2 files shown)


client/java/src/main/java/com/linecorp/centraldogma/internal/client/ReplicationLagTolerantCentralDogma.java
```diff
@@ -681,6 +683,9 @@ private boolean handleRevisionNotFound(
             String projectName, String repositoryName, Revision revision, Throwable cause) {
 
         requireNonNull(cause, "cause");
+        if (cause instanceof CompletionException) {
```
minwoox: We can use `Exceptions.peel(cause)`

imasahiro: Hm, we can use it only when under `client-java-armeria` or `server`.
This class is under `client-java` and it does not have armeria dependency.

ikhoon:
Build error with `Checkstyle` 😉
```
> Task :client:java:checkstyleTest FAILED
[ant:checkstyle] [ERROR] /home/travis/build/line/centraldogma/client/java/src/test/java/com/linecorp/centraldogma/internal/client/ReplicationLagTolerantCentralDogmaTest.java:33:8: Unused import - java.util.concurrent.CountDownLatch. [UnusedImports]
[ant:checkstyle] [ERROR] /home/travis/build/line/centraldogma/client/java/src/test/java/com/linecorp/centraldogma/internal/client/ReplicationLagTolerantCentralDogmaTest.java:34:8: Unused import - java.util.concurrent.ExecutorService. [UnusedImports]
[ant:checkstyle] [ERROR] /home/travis/build/line/centraldogma/client/java/src/test/java/com/linecorp/centraldogma/internal/client/ReplicationLagTolerantCentralDogmaTest.java:37:8: Unused import - java.util.concurrent.TimeUnit. [UnusedImports]
```

client/java/src/main/java/com/linecorp/centraldogma/internal/client/ReplicationLagTolerantCentralDogma.java
minwoox: Ah yes, forgot about it. 😅 
So isn't there any chance that the exception is wrapped by other exceptions such as `ExecutionException`?

client/java/src/main/java/com/linecorp/centraldogma/internal/client/ReplicationLagTolerantCentralDogma.java
```diff
@@ -681,6 +683,9 @@ private boolean handleRevisionNotFound(
...
+            cause = Throwables.getRootCause(cause);
```
mauhiz: root cause might be irrelevant, e.g. if someone throws a `RevisionNotFoundException(cause=IOException)`

PR #469, opened 2020-01-14
Migrate integration tests to JUnit 5

KarboniteKream:
This PR migrates integration tests in `it` to JUnit 5.

This excludes `GitMirrorTest`, which I will migrate after JUnit 5.6 is released (after a small issue with `@TempDir` is resolved).

#### Changes:
- Migrate integration tests
- Add dependency on `junit-jupiter-params`
- Replace `CentralDogmaRuleWithScaffolding`

#### Notes:
- `@DisableOnDebug` for `@Timeout` is not supported by JUnit 5, but version 5.6 will add a global configuration option
- Projects in `CacheTest` depend on the test name (method + parameters), so I had to set it explicitly
- Test parameters cannot be passed to constructors, so I had to do some initialization in the tests themselves (`WatchTest`)
- JUnit 5 does not allow no parameters (`GitMirrorAuthTest`), so I used `Assumptions.assumeFalse()` to skip the tests when there are none


build.gradle
```diff
@@ -159,6 +160,6 @@ allprojects {
 // Require to use JDK 10 when releasing.
 tasks.release.doFirst {
     if (JavaVersion.current() != JavaVersion.VERSION_11) {
-        throw new IllegalStateException("You must release using JDK 11.");
+        throw new IllegalStateException("You must release using JDK 11.")
```
ikhoon: 👍

it/src/test/java/com/linecorp/centraldogma/it/mirror/git/GitMirrorAuthTest.java
```diff
@@ -189,13 +214,14 @@ protected void configure(CentralDogmaBuilder builder) {
...
-                                    '}')
-                    });
+                                    '}')));
                 }
             }
         }
 
-        return builder.build();
+        final Collection<Arguments> parameters = builder.build();
+        assumeFalse(parameters.isEmpty(), "Nothing to test.");
```
ikhoon: Could you use? 
```java
assertThat(parameters).isEmpty();
```

testing-internal/src/main/java/com/linecorp/centraldogma/testing/internal/TestUtil.java
```diff
@@ -35,11 +40,15 @@
...
         assertThatJson(json).isEqualTo(value);
         try {
-            assertEquals(value, Jackson.readValue(json, valueType));
+            assertThat(Jackson.readValue(json, valueType)).isEqualTo(value);
         } catch (IOException e) {
             throw new IOError(e);
         }
     }
 
+    public static String getMethodName(TestInfo testInfo) {
```
ikhoon: nit: How about renaming this to `normalizedDisplayName` for clarifying the method behavior? 😀

minwoox:
Thanks @KarboniteKream! 🙇

it/src/test/java/com/linecorp/centraldogma/it/mirror/git/GitMirrorAuthTest.java
ikhoon: How about?
```java
import static org.assertj.core.api.Assumptions.assumeThat;
assumeThat(parameters).isEmpty();
```

KarboniteKream: Missed this, thank you!

minwoox:
Thanks @KarboniteKream!

PR #470, opened 2020-01-19
Migrate ProjectManagerRule to JUnit 5

KarboniteKream:
This PR migrates internal `ProjectManagerRule` to JUnit 5.

#### Changes:
- Replace `ProjectManagerRule` with `ProjectManagerExtension`
- Migrate tests using the extension
- Extract common extension logic to `TemporaryFolder`

### DIFF
--- a/server/src/test/java/com/linecorp/centraldogma/server/internal/api/TokenServiceTest.java
+++ b/server/src/test/java/com/linecorp/centraldogma/server/internal/api/TokenServiceTest.java
@@ -13,7 +13,6 @@
  * License for the specific language governing permissions and limitations
  * under the License.
  */
-
 package com.linecorp.centraldogma.server.internal.api;
 
 import static org.assertj.core.api.Assertions.assertThat;
--- a/server/src/test/java/com/linecorp/centraldogma/server/internal/api/TokenServiceTest.java
+++ b/server/src/test/java/com/linecorp/centraldogma/server/internal/api/TokenServiceTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 LINE Corporation
+ * Copyright 2020 LINE Corporation
  *
  * LINE Corporation licenses this file to you under the Apache License,
  * version 2.0 (the "License"); you may not use this file except in compliance
@@ -13,6 +13,7 @@
  * License for the specific language governing permissions and limitations
  * under the License.
  */
+
 package com.linecorp.centraldogma.server.internal.api;
 
 import static org.assertj.core.api.Assertions.assertThat;
@@ -21,9 +22,9 @@ import static org.mockito.Mockito.mock;
 
 import java.util.Collection;
 
-import org.junit.BeforeClass;
-import org.junit.ClassRule;
-import org.junit.Test;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
 
 import com.linecorp.armeria.server.HttpResponseException;
 import com.linecorp.armeria.server.ServiceRequestContext;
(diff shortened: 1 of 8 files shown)


server/src/test/java/com/linecorp/centraldogma/server/internal/api/TokenServiceTest.java
```diff
@@ -13,6 +13,7 @@
  * License for the specific language governing permissions and limitations
  * under the License.
  */
+
```
ikhoon: nit: Revert? 😃

testing/common/src/main/java/com/linecorp/centraldogma/testing/internal/TemporaryFolder.java
```diff
@@ -0,0 +1,71 @@
...
+
+    public Path newFolder() throws IOException {
+        return Files.createTempDirectory(getRoot(), "");
+    }
+
+    public Path newFile() throws IOException {
+        return Files.createTempFile(getRoot(), "", "");
+    }
+
+    public void delete() throws IOException {
```
ikhoon: How about implementing `AutoCloseable` to use with try-with-resource by any chance?

server/src/test/java/com/linecorp/centraldogma/server/internal/api/TokenServiceTest.java
KarboniteKream: I wasn't sure which is the preferred way, since I've seen both... Let me revert!

testing/common/src/main/java/com/linecorp/centraldogma/testing/internal/TemporaryFolder.java
KarboniteKream: Let me take a look, but I'm not sure if this will be possible, since the folder is created in `@Before` and cleaned up in `@After`

ikhoon: I didn't mean to refactor the existing logic. 😅 Just thought we can use it as a local variable with try-with-resouces too if needed.

ikhoon: By the way, not strong here. If you think it is weird, feel free to ignore this. 😁

KarboniteKream: Ah, I understand! Let me take a look.

minwoox: I think it's okay as it is.
Shouldn't we rather do the try-with-resource when calling `walk`?
```java
try (Stream<Path> walk = Files.walk(root)) {
    walk.map(...)
}
```

testing/common/src/main/java/com/linecorp/centraldogma/testing/internal/TemporaryFolder.java
```diff
@@ -0,0 +1,71 @@
...
+        if (root == null) {
+            return;
+        }
+
+        Files.walk(root)
+             .sorted(Comparator.reverseOrder())
+             .map(Path::toFile)
+             .forEach(File::delete);
+
+        root = null;
```
minwoox: Does the root directory is deleted as well?

KarboniteKream: Yes, `Files.walk(root)` includes `root`

KarboniteKream:
Interesting, `MirrorTest` starts reporting unused stubs when `MockitoExtension` is applied. Let me fix this

trustin:
Excellent work, @KarboniteKream !

PR #473, opened 2020-01-23
Migrate unit tests to JUnit 5 (Part 2)

KarboniteKream:
This PR continues the migration of unit tests to JUnit 5.

#### Changes:
- Migrate some tests to JUnit 5
- Upgrade to JUnit 5.6.0 and 4.13
  - Enable `disabled_on_debug` for `@Timeout`
- Migrate to Armeria's `ServerExtension`
- Introduce more nested tests, similar to `ProjectServiceV1Test`
  - Extract common logic to `TestUtil.getClient()`

### DIFF
--- a/dependencies.yml
+++ b/dependencies.yml
@@ -6,7 +6,7 @@ boms:
 - com.fasterxml.jackson:jackson-bom:2.10.1
 - com.linecorp.armeria:armeria-bom:0.97.0
 - io.micrometer:micrometer-bom:1.3.2
-
+- org.junit:junit-bom:5.6.0
 
 ch.qos.logback:
   logback-classic:
@@ -152,20 +152,9 @@ junit:
 
 org.junit.jupiter:
   junit-jupiter-api:
-    version: &JUNIT_JUPITER_VERSION '5.6.0'
     # ':site:javadoc' fails when we use a newer version of Javadoc.
     javadocs:
       - https://junit.org/junit5/docs/5.5.2/api/
-  junit-jupiter-engine:
-    version: *JUNIT_JUPITER_VERSION
-  junit-jupiter-params:
-    version: *JUNIT_JUPITER_VERSION
-org.junit.platform:
-  junit-platform-launcher:
-    version: '1.6.0'
-org.junit.vintage:
-  junit-vintage-engine:
-    version: *JUNIT_JUPITER_VERSION
 
 kr.motd.gradle:
   sphinx-gradle-plugin: { version: '2.6.1' }
--- a/dependencies.yml
+++ b/dependencies.yml
@@ -146,13 +146,14 @@ javax.validation:
 
 junit:
   junit:
-    version: '4.12'
+    version: '4.13'
     javadocs:
-    - https://junit.org/junit4/javadoc/4.12/
+      - https://junit.org/junit4/javadoc/4.13/
 
 org.junit.jupiter:
   junit-jupiter-api:
-    version: &JUNIT_JUPITER_VERSION '5.5.2'
+    version: &JUNIT_JUPITER_VERSION '5.6.0'
+    # ':site:javadoc' fails when we use a newer version of Javadoc.
     javadocs:
       - https://junit.org/junit5/docs/5.5.2/api/
   junit-jupiter-engine:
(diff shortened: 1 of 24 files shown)


dependencies.yml
```diff
@@ -146,13 +146,14 @@ javax.validation:
...
+    version: '4.13'
     javadocs:
-    - https://junit.org/junit4/javadoc/4.12/
+      - https://junit.org/junit4/javadoc/4.13/
 
 org.junit.jupiter:
   junit-jupiter-api:
-    version: &JUNIT_JUPITER_VERSION '5.5.2'
+    version: &JUNIT_JUPITER_VERSION '5.6.0'
+    # ':site:javadoc' fails when we use a newer version of Javadoc.
```
KarboniteKream: The API documentation now documents modules, which causes `site:javadoc` to fail.
Similar issue as https://github.com/line/armeria/blob/f9b38ea6c164ccf62b76f72aea74756a7a842226/dependencies.yml#L245

KarboniteKream:
JUnit 5.6 fixed `@TempDir` when removing read-only files, but some tests on Windows are still flaky. This is caused by file locks not being released in time, preventing the extension from deleting the files.

To resolve this problem, I've implemented `TemporaryFolderExtension` to use in these tests. `TemporaryFolder` doesn't throw an exception when files cannot be deleted, and keeps them intact. The extension mimics the behaviour of JUnit 4's `TemporaryFolder` rule.

This issue should only occur on Windows.

dependencies.yml
```diff
@@ -146,13 +146,14 @@ javax.validation:
...
-    version: '4.12'
+    version: '4.13'
     javadocs:
-    - https://junit.org/junit4/javadoc/4.12/
+      - https://junit.org/junit4/javadoc/4.13/
 
 org.junit.jupiter:
   junit-jupiter-api:
-    version: &JUNIT_JUPITER_VERSION '5.5.2'
+    version: &JUNIT_JUPITER_VERSION '5.6.0'
```
ikhoon: How about using [junit-bom](https://search.maven.org/artifact/org.junit/junit-bom/5.6.0/pom)?

server/src/test/java/com/linecorp/centraldogma/server/internal/api/AdministrativeServiceTest.java
```diff
@@ -45,28 +43,19 @@ protected boolean runForEachTest() {
...
-        final String serverUri = "http://127.0.0.1:" + serverAddress.getPort();
-        webClient = WebClient.builder(serverUri)
-                             .addHttpHeader(HttpHeaderNames.AUTHORIZATION, "Bearer anonymous")
-                             .build();
-    }
-
     @Test
     void status() {
-        final AggregatedHttpResponse res = webClient.get(API_V1_PATH_PREFIX + "status").aggregate().join();
+        final WebClient client = getClient(dogma);
```
ikhoon: Don't we use `dogma.httpClient()` here?
I think we can add `configureHttpClient` to `CentralDogmaExtension`.
```java
protected void configureHttpClient(WebClientBuilder builder) {}
```

testing-internal/src/main/resources/junit-platform.properties
```diff
@@ -1 +1,2 @@
+junit.jupiter.execution.timeout.mode = disabled_on_debug
```
ikhoon: Nice 👍

server/src/test/java/com/linecorp/centraldogma/server/internal/api/AdministrativeServiceTest.java
KarboniteKream: I'll add this, thanks for the suggestion. But this PR is already pretty big, so if you don't mind, I'd prefer to do this in a follow-up PR.

testing-internal/src/main/java/com/linecorp/centraldogma/testing/internal/TemporaryFolderExtension.java
```diff
@@ -0,0 +1,84 @@
...
+
+import com.linecorp.centraldogma.testing.junit.AbstractAllOrEachExtension;
+
+/**
+ * A JUnit {@link Extension} that creates a {@link TemporaryFolder}.
+ *
+ * <pre>{@code
+ * > class MyTest {
+ * >     @RegisterExtension
+ * >     static final TemporaryFolder folder = new TemporaryFolder();
```
ikhoon: nit: `TemporaryFolder` -> `TemporaryFolderExtension`

KarboniteKream: Thanks 🙇

testing-internal/src/main/resources/junit-platform.properties
trustin: How about setting the default timeout for all testables and life cycle methods?

```
junit.jupiter.execution.timeout.default = 60s
```

Then we could remove all `@Timeout` annotations.

KarboniteKream: Added, thanks for the suggestion!

PR #483, opened 2020-02-10
Use modern Gradle dependency API

ikhoon:
Motivation:
Gradle introduced a new API for dependencies.

Modifications:
* Change `compile` to `api` or `implmentation`
* Change 'testCompile` to `testImplementation`
* Change optional compile dependency to `optionalImplementation`

Result:
* Maybe faster incremental compile time
* Easy to control the exposed API


trustin:
Thanks, @ikhoon !

PR #490, opened 2020-02-13
Fix version constraints in Spring Boot 1 modules

trustin:
Motivation:

We forgot to migrate the build scripts of `:client:java-spring-boot1-*`
when we migrate from Spring's dependency management plugin to Gradle's
version constraint API.

Modifications:

- Fix dependencies.
- Add `sbrannen/spring-test-junit5` for testing with Spring Boot 1.x.
  - Work around the bug which pulls in a bean from other tests.

Result:

- `armeria-client-spring-boot1-*` modules now have correct dependencies.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3,7 +3,6 @@ plugins {
     id 'me.champeau.gradle.jmh' version '0.5.0' apply false
 }
 
-
 allprojects {
     repositories {
         mavenCentral()
--- a/build.gradle
+++ b/build.gradle
@@ -1,14 +1,8 @@
-buildscript {
-    repositories {
-        gradlePluginPortal()
-        mavenCentral()
-    }
-    dependencies {
-        classpath 'com.google.gradle:osdetector-gradle-plugin:1.6.2'
-        classpath 'io.spring.gradle:dependency-management-plugin:1.0.9.RELEASE'
-        classpath 'me.champeau.gradle:jmh-gradle-plugin:0.5.0'
-    }
+plugins {
+    id 'com.google.osdetector' version '1.6.2' apply false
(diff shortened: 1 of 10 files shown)


trustin:
Need to fix the test failure.

trustin:
@KarboniteKream It seems like it's injecting the `ClientFactory` defined in `CentralDogmaClientAutoConfigurationWithClientFactoryTest. Probably a bug in the test extension?

KarboniteKream:
From what I’ve observed, it seems like it didn’t recognize the `@ForCentralDogma` annotation and used `ClientFactory.ofDefault()` instead.

trustin:
The build should pass now that the workaround is added.

trustin:
@KarboniteKream wrote:

> From what I’ve observed, it seems like it didn’t recognize the `@ForCentralDogma` annotation and used `ClientFactory.ofDefault()` instead.

I did `println` on all `ClientFactory` instances involved and it was indeed from other test (`CentralDogmaClientAutoConfigurationWithClientFactoryTest`). Please check the updated test case with a workaround.

build.gradle
```diff
@@ -1,15 +1,9 @@
...
-        classpath 'com.google.gradle:osdetector-gradle-plugin:1.6.2'
-        classpath 'io.spring.gradle:dependency-management-plugin:1.0.9.RELEASE'
-        classpath 'me.champeau.gradle:jmh-gradle-plugin:0.5.0'
-    }
+plugins {
+    id 'com.google.osdetector' version '1.6.2' apply false
+    id 'me.champeau.gradle.jmh' version '0.5.0' apply false
 }
 
+
```
minwoox: nit: revert?

trustin: Oops :sweat_smile:

PR #491, opened 2020-02-17
Do not fail `watch` when `ClientFactory` is closed

trustin:
Motivation:

Although not recommended, a user can close the `ClientFactory` which is
running a `CentralDogma` client while `watch*()` operations are in
progress. A `ClosedSessionException` will be triggered in such a case,
which can mislead the users to think the server has closed the
connection unexpectedly.

Modifications:

- Do not propagate `ClosedSessionException` if `ClientFactory` is closing.

Result:

- Better user experience


trustin:
Thanks for reviewing!

PR #495, opened 2020-02-19
Do not log warnings about closed connections and streams

trustin:
Motivation:

The disconnection between Central Dogma client and server can happen
very often due to various reasons, such as application shutdown. It's of
dubious value to log them.

Modifications:

- Do not log the exceptions related with unexpected disconnections and
  stream resets at WARN level.
  - Note that they will still be recorded in `RequestLog`s.

Result:

- Less noisy log file

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/api/HttpApiUtil.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/api/HttpApiUtil.java
@@ -50,6 +50,8 @@ import com.linecorp.centraldogma.common.RedundantChangeException;
 import com.linecorp.centraldogma.common.ShuttingDownException;
 import com.linecorp.centraldogma.internal.Jackson;
 
+import io.netty.handler.codec.http2.Http2Exception;
+
 /**
  * A utility class which provides common functions for HTTP API.
  */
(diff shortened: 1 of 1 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/api/HttpApiUtil.java
```diff
@@ -178,8 +182,11 @@ private static HttpResponse newResponse0(RequestContext ctx, HttpStatus status,
         //                 the stack trace of the cause to the trusted client.
         if (status == HttpStatus.INTERNAL_SERVER_ERROR) {
             if (cause != null) {
-                if (!(cause instanceof ShuttingDownException ||
-                      cause instanceof AbortedStreamException)) {
+                if (!(cause instanceof AbortedStreamException ||
+                      cause instanceof ClosedStreamException ||
+                      cause instanceof Http2Exception ||
+                      cause instanceof ClosedSessionException ||
```
ikhoon: Can't we use `Exceptions.isExpected(Throwable)` or `Exceptions.isStreamCancelling(Throwable)` for Armeria exceptions? https://github.com/line/armeria/blob/ea2dbc05089191a4f457d03147440e3a7c27d27f/core/src/main/java/com/linecorp/armeria/common/util/Exceptions.java#L136
https://github.com/line/armeria/blob/ea2dbc05089191a4f457d03147440e3a7c27d27f/core/src/main/java/com/linecorp/armeria/common/util/Exceptions.java#L173

trustin: That's a good idea. :-)

PR #500, opened 2020-02-26
Clean up overall logging configuration

trustin:
Motivation:

- We already log an exception at our side when async loading fails.
- Some WARN-level log messages are triggered by a bad request, such as
  an invalid query expression or a commit with conflicts.

Modifications:

- Disable Caffeine's async loading failure logging to avoid double
  logging/notification.
- Stop using `RequestContextExportingAppender` because we do not use MDC.
- Reduce the default amount of retained log files.

Result:

- No more double logging/notification
- No more WARN logs triggered by bad requests
- Reduced default disk usage

### DIFF
--- a/dist/src/conf/logback.xml
+++ b/dist/src/conf/logback.xml
@@ -1,57 +1,40 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <configuration>
+  <!-- Up to 1GB per file, 7 days and 8GB total by default -->
+  <variable name="LOG_MAX_FILE_SIZE" value="${ACCESS_LOG_MAX_FILE_SIZE:-1GB}" />
+  <variable name="LOG_MAX_HISTORY" value="${ACCESS_LOG_MAX_HISTORY:-7}" />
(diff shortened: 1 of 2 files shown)


dist/src/conf/logback.xml
```diff
@@ -1,57 +1,40 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <configuration>
+  <!-- Up to 1GB per file, 7 days and 8GB total by default -->
+  <variable name="LOG_MAX_FILE_SIZE" value="${ACCESS_LOG_MAX_FILE_SIZE:-1GB}" />
+  <variable name="LOG_MAX_HISTORY" value="${ACCESS_LOG_MAX_HISTORY:-7}" />
+  <variable name="LOG_TOTAL_SIZE_CAP" value="${ACCESS_LOG_TOTAL_SIZE_CAP:-8GB}" />
```
ikhoon: Question: does `LOG_MAX_FILE_SIZE` use `ACCESS_LOG_MAX_FILE_SIZE` evironment variable?

trustin: No. Let me fix them.

PR #507, opened 2020-04-01
Do not halt even if failed to schedule a purge task

trustin:
Motivation:

We observed a replica enters read-only mode when it failed to schedule a
purge task. It's not a critical problem and thus it can be fixed
manually later, rather than halting the system.

Modification:

Do not propagate an exception when:

- Failed to schedule a purge task.
- A repository to purge does not exist.

Result:

Robustness


minwoox:
Thanks @trustin!

PR #511, opened 2020-04-06
Fix a bug where a JSON file cannot be watched and fetched as String

minwoox:
Motivation:
A user might want to watch and fetch a JSON file as String. Currently, it's not possible.

Modifications:
- Add `IDENTITY_TEXT` and `IDENTITY_JSON` to `QueryType`.
- Remove JSON related annotations in `Query` and its subclasses.
  - `Query` is not directly (de)serialized into JSON.
  - Remove `QueryDto` and `NamedQueryDto` which are not used.

Result:
- You can now fetch and watch a JSON as String.


common/src/main/java/com/linecorp/centraldogma/common/QueryType.java
```diff
@@ -28,6 +28,16 @@
      * Retrieves the content as it is.
      */
     IDENTITY(EnumSet.of(EntryType.TEXT, EntryType.JSON)),
+
+    /**
+     * Retrieves the String content as it is.
```
trustin, line 33: String -> textual?

minwoox: Fixed. 😄

client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogma.java
```diff
@@ -894,7 +897,7 @@ private static Revision watchRepository(AggregatedHttpResponse res) {
             return client.execute(builder.build()).aggregate()
                          .handle((res, cause) -> {
                              if (cause == null) {
-                                 return func.apply(res);
+                                 return func.apply(res, queryType);
                              }
 
                              if ((cause instanceof ClosedSessionException ||
```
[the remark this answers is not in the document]
minwoox: Ah fixed. Thanks! 😄

PR #517, opened 2020-06-18
Update broken link for armeria parameters javac option

jrhee17:
As an fyi, I also found the same broken link in armeria project (under `examples/README.md`)


jrhee17:
(force pushed due to wrong git username 😅 )

minwoox:
Thanks!

PR #518, opened 2020-06-18
Fix broken links

ikhoon:
As #517's follow-up work, fix broken links checked by https://github.com/stevenvachon/broken-link-checker.
I ran with:
```
blc https://line.github.io/centraldogma -r
```


ikhoon:
> How about using JDK 11 to build Javadoc? (Not sure it's possible. 🤔 )

Let me try it another PR. 😉 (Not sure it's possible. 🤔 )

PR #535, opened 2020-09-18
Provide a way to configure hierarchical quorums

### LINKED ISSUE #534: Provide a way to configure hierarchical quorums
If users want to deploy Central Dogma to multi-data centers, [hierarchical quorums](https://zookeeper.apache.org/doc/r3.5.8/zookeeperHierarchicalQuorums.html) could reduce write transaction time.

Let's say 15 Central Dogma replicas are distributed to 3 data centers, five each. The responses from 8 instances are required to complete a write command. However, if each data center is grouped using hierarchical quorums, only 6 votes(3 votes from 2 groups) are enough.

I think we can add `groupId` and `weight` properties to `servers` of replication config.
```json
"replication" : {
  "method": "ZOOKEEPER",
  "serverId": 1,
  "servers": {
    "1": {
      "host": "1.1.1.1",
      "quorumPort": 36463,
      "electionPort": 36464,
      "groupId": 1,
      "weight": 1
    },
    "2": {
      "host": "1.1.1.2",
      "quorumPort": 36463,
      "electionPort": 36464,
      "groupId": 1,
      "weight": 1
    },
    "3": {
      "host": "1.1.1.3",
      "quorumPort": 36463,
      "electionPort": 36464,
      "groupId": 2,
      "weight": 1
    },
    ...
}
```

ikhoon:
Motivation:

For multi-data center replication, hierarchical quorums could reduce total
write transaction time than a single quorum.

See #534

Modifications:

- Add `groupId` and `weight` to Zookeeper replication config

Result:

- You can now use hierarchical quorums for the embedded Zookeeper of Central Dogma
- Closes #534

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/ZooKeeperServerConfig.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/ZooKeeperServerConfig.java
@@ -65,7 +65,11 @@ public final class ZooKeeperServerConfig {
                       "clientPort: %s (expected: 0-65535)", clientPort);
         this.clientPort = clientPort;
         this.groupId = groupId;
-        this.weight = weight == null ? 1 : weight;
+        if (weight == null) {
+            weight = 1;
+        }
+        checkArgument(weight >= 0, "weight: %s (expected: >= 0)", weight);
+        this.weight = weight;
     }
 
     private static int validatePort(int port, String name) {
--- a/server/src/main/java/com/linecorp/centraldogma/server/ZooKeeperServerConfig.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/ZooKeeperServerConfig.java
@@ -15,6 +15,7 @@
  */
 package com.linecorp.centraldogma.server;
 
+import static com.google.common.base.MoreObjects.firstNonNull;
 import static com.google.common.base.Preconditions.checkArgument;
 import static java.util.Objects.requireNonNull;
 
@@ -53,7 +54,7 @@ public final class ZooKeeperServerConfig {
     public ZooKeeperServerConfig(@JsonProperty(value = "host", required = true) String host,
                                  @JsonProperty(value = "quorumPort", required = true) int quorumPort,
                                  @JsonProperty(value = "electionPort", required = true) int electionPort,
-                                 @JsonProperty(value = "clientPort", defaultValue = "0") int clientPort,
+                                 @JsonProperty(value = "clientPort", defaultValue = "0") Integer clientPort,
                                  @JsonProperty("groupId") @Nullable Integer groupId,
                                  @JsonProperty(value = "weight", defaultValue = "1") Integer weight) {
 
@@ -61,6 +62,9 @@ public final class ZooKeeperServerConfig {
         this.quorumPort = validatePort(quorumPort, "quorumPort");
         this.electionPort = validatePort(electionPort, "electionPort");
 
+        if (clientPort == null) {
+            clientPort = 0;
+        }
         checkArgument(clientPort >= 0 && clientPort <= 65535,
                       "clientPort: %s (expected: 0-65535)", clientPort);
         this.clientPort = clientPort;
--- a/server/src/main/java/com/linecorp/centraldogma/server/ZooKeeperServerConfig.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/ZooKeeperServerConfig.java
@@ -55,7 +55,7 @@ public final class ZooKeeperServerConfig {
                                  @JsonProperty(value = "electionPort", required = true) int electionPort,
                                  @JsonProperty(value = "clientPort", defaultValue = "0") int clientPort,
                                  @JsonProperty("groupId") @Nullable Integer groupId,
-                                 @JsonProperty(value = "weight", defaultValue = "1") int weight) {
+                                 @JsonProperty(value = "weight", defaultValue = "1") Integer weight) {
 
         this.host = requireNonNull(host, "host");
         this.quorumPort = validatePort(quorumPort, "quorumPort");
@@ -65,7 +65,7 @@ public final class ZooKeeperServerConfig {
                       "clientPort: %s (expected: 0-65535)", clientPort);
         this.clientPort = clientPort;
         this.groupId = groupId;
-        this.weight = weight;
+        this.weight = weight == null ? 1 : weight;
     }
 
     private static int validatePort(int port, String name) {
--- a/server/src/main/java/com/linecorp/centraldogma/server/ZooKeeperServerConfig.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/ZooKeeperServerConfig.java
@@ -15,7 +15,6 @@
  */
 package com.linecorp.centraldogma.server;
 
-import static com.google.common.base.MoreObjects.firstNonNull;
 import static com.google.common.base.Preconditions.checkArgument;
 import static java.util.Objects.requireNonNull;
 
@@ -47,16 +46,17 @@ public final class ZooKeeperServerConfig {
      * @param quorumPort the quorum port number
      * @param electionPort the election port number
      * @param clientPort the client port number (0-65535)
-     * @param groupId the group ID to use hierarchical quorums
+     * @param groupId the group ID to use for hierarchical quorums
      * @param weight the weight of the Zookeeper server
      */
     @JsonCreator
-    public ZooKeeperServerConfig(@JsonProperty(value = "host", required = true) String host,
-                                 @JsonProperty(value = "quorumPort", required = true) int quorumPort,
-                                 @JsonProperty(value = "electionPort", required = true) int electionPort,
-                                 @JsonProperty(value = "clientPort", defaultValue = "0") Integer clientPort,
-                                 @JsonProperty("groupId") @Nullable Integer groupId,
-                                 @JsonProperty(value = "weight", defaultValue = "1") Integer weight) {
+    public ZooKeeperServerConfig(
+            @JsonProperty(value = "host", required = true) String host,
+            @JsonProperty(value = "quorumPort", required = true) int quorumPort,
+            @JsonProperty(value = "electionPort", required = true) int electionPort,
+            @JsonProperty(value = "clientPort", defaultValue = "0") @Nullable Integer clientPort,
+            @JsonProperty("groupId") @Nullable Integer groupId,
+            @JsonProperty(value = "weight", defaultValue = "1") @Nullable Integer weight) {
 
         this.host = requireNonNull(host, "host");
         this.quorumPort = validatePort(quorumPort, "quorumPort");
--- a/server/src/main/java/com/linecorp/centraldogma/server/ZooKeeperServerConfig.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/ZooKeeperServerConfig.java
@@ -20,6 +20,8 @@ import static java.util.Objects.requireNonNull;
 
 import java.util.Objects;
 
+import javax.annotation.Nullable;
+
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.base.MoreObjects;
@@ -27,12 +29,15 @@ import com.google.common.base.MoreObjects;
 /**
  * Represents the address and port numbers of a ZooKeeper node.
  */
-public final class ZooKeeperAddress {
+public final class ZooKeeperServerConfig {
 
     private final String host;
     private final int quorumPort;
     private final int electionPort;
     private final int clientPort;
+    @Nullable
+    private final Integer groupId;
+    private final int weight;
 
     /**
      * Creates a new instance.
@@ -41,12 +46,16 @@ public final class ZooKeeperAddress {
      * @param quorumPort the quorum port number
      * @param electionPort the election port number
      * @param clientPort the client port number (0-65535)
+     * @param groupId the group ID to use hierarchical quorums
+     * @param weight the weight of the Zookeeper server
      */
     @JsonCreator
-    public ZooKeeperAddress(@JsonProperty(value = "host", required = true) String host,
-                            @JsonProperty(value = "quorumPort", required = true) int quorumPort,
-                            @JsonProperty(value = "electionPort", required = true) int electionPort,
-                            @JsonProperty(value = "clientPort", defaultValue = "0") int clientPort) {
+    public ZooKeeperServerConfig(@JsonProperty(value = "host", required = true) String host,
+                                 @JsonProperty(value = "quorumPort", required = true) int quorumPort,
+                                 @JsonProperty(value = "electionPort", required = true) int electionPort,
+                                 @JsonProperty(value = "clientPort", defaultValue = "0") int clientPort,
+                                 @JsonProperty("groupId") @Nullable Integer groupId,
+                                 @JsonProperty(value = "weight", defaultValue = "1") int weight) {
 
         this.host = requireNonNull(host, "host");
         this.quorumPort = validatePort(quorumPort, "quorumPort");
@@ -55,6 +64,8 @@ public final class ZooKeeperAddress {
         checkArgument(clientPort >= 0 && clientPort <= 65535,
                       "clientPort: %s (expected: 0-65535)", clientPort);
         this.clientPort = clientPort;
+        this.groupId = groupId;
+        this.weight = weight;
     }
 
     private static int validatePort(int port, String name) {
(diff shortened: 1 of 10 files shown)


server/src/test/java/com/linecorp/centraldogma/server/CentralDogmaConfigTest.java
```diff
@@ -141,11 +141,8 @@ void trustedProxyAddress_withDefaultClientAddressSources_withoutProxyProtocol()
 
         final List<ClientAddressSource> sources = cfg.clientAddressSourceList();
         assertThat(sources).hasSize(2);
-        // TODO(hyangtack) Need to add equals/hashCode to ClientAddressSource class.
-        //                 https://github.com/line/armeria/pull/1804
-        assertThat(sources.get(0).toString()).isEqualTo(ClientAddressSource.ofHeader("forwarded").toString());
-        assertThat(sources.get(1).toString()).isEqualTo(
-                ClientAddressSource.ofHeader("x-forwarded-for").toString());
+        assertThat(sources.get(0)).isEqualTo(ClientAddressSource.ofHeader("forwarded"));
```
minwoox, line 144: Thank you!

server/src/main/java/com/linecorp/centraldogma/server/ZooKeeperServerConfig.java
```diff
@@ -55,6 +64,8 @@ public ZooKeeperAddress(@JsonProperty(value = "host", required = true) String ho
         checkArgument(clientPort >= 0 && clientPort <= 65535,
                       "clientPort: %s (expected: 0-65535)", clientPort);
         this.clientPort = clientPort;
+        this.groupId = groupId;
+        this.weight = weight == null ? 1 : weight;
```
minwoox, line 68: Just wondering this could be a negative value. If so we should validate. 🤔

server/src/test/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutorTest.java
```diff
@@ -254,13 +255,55 @@ void stopWhileWaitingForInitialQuorum() throws Exception {
                                .hasMessageContaining("before joining");
     }
 
+    @Test
+    void hierarchicalQuorums() throws Throwable {
```
minwoox, line 259: Question:
Can we check how it works when the weight is different? for example:
```
(x) means the weight of the replica

group.1=1(2):2(1):3(1) (and what happens if 1 is down?)
or
group.1=1(1):2(0):3(0)
or 
group.1=1(0):2(0):3(0)

with
group.2=4:5:6
group.3=7:8:9
```

minwoox (commented):
Left some questions. 😄

server/src/main/java/com/linecorp/centraldogma/server/ZooKeeperServerConfig.java
ikhoon: There is no specification or documentation for negative weights.
Base on the implementation of QuorumHierarchical: 
- The negative server weight is also added to group weight.
- If a group weight is zero, the group falls out of the hierarchical quorums.
https://github.com/apache/zookeeper/blob/4a2d58219b7435c3b8cdf8f7ab04b158c1900223/zookeeper-server/src/main/java/org/apache/zookeeper/server/quorum/flexible/QuorumHierarchical.java#L312-L321

So I think the weight of servers does not need to be validated and it looks good to follow the ZooKeeper policy.

ikhoon: In the following example, Group 2 is moved out from hierarchical quorums.
```
group.1=1:2:3
group.2=4:5:6
group.3=7:8:9

weight.1=1
weight.2=1
weight.3=1
weight.4=-2
weight.5=1
weight.6=1
weight.7=1
weight.8=1
weight.9=1
```

minwoox: What if `weight.4=-3`?
It seems like they don't do `weight <= 0` but `weight == 0`.
Does it mean that weight cannot be negative?

ikhoon: > What if weight.4=-3?

Good question. I don't think about that. Let me test it how it works.

ikhoon: From the documentation, the negative weight seems to be excluded from quorums.
https://zookeeper.apache.org/doc/r3.5.8/zookeeperHierarchicalQuorums.html

> When running the system, we are able to form a quorum once we have a majority of votes from a majority of non-zero-weight groups. Groups that have zero weight are discarded and not considered when forming quorums.

But it is worth to add tests for verification. Let me add more tests for various weights of servers.

server/src/test/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutorTest.java
ikhoon: > Can we check how it works when the weight is different?

I will add test cases for that.

> group.1=1(2):2(1):3(1) (and what happens if 1 is down?)

(Maybe) In this case, Group 1 is excluded from quorums, however, the majority is still satisfied because the remaining two groups can vote.

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -269,6 +276,10 @@ public ZooKeeperCommandExecutor(ZooKeeperReplicationConfig cfg,
 
         // Register the metrics which are accessible even before started.
         Gauge.builder("replica.id", this, self -> replicaId()).register(meterRegistry);
+        Gauge.builder("replica.groupId", this, self -> {
+            final Integer groupId = self.cfg.serverConfig().groupId();
+            return groupId != null ? groupId : -1;
```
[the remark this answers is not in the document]
ikhoon: > How about not registering groupId if it's null

That sounds good.

minwoox:
Could you also add tests that:
- Writing a file to the replica which is participating in the quorum in hierarchical mode. 
- Writing a file to the replica which is not participating in the quorum in hierarchical mode.

server/src/main/java/com/linecorp/centraldogma/server/ZooKeeperServerConfig.java
ikhoon: There is undefined behavior on negative weight in ZooKeeper hierarchical quorums. It seems to be better not to accept negative weights. Now, an `IllegalArgumentException` is raised on negative weights.

ikhoon:
> Writing a file to the replica which is not participating in the quorum in hierarchical mode.

I think writing a file is not strongly related to whether a replica has participated or not. Because
- A replica which received a command execute the command locally (which is not related to ZooKeeper).
- After the job is completed, the command is saved to the ZooKeeper cluster via Curator.
- Other replicas, which listen to ReplicationLog from a Curator client, will execute the command when the event is relayed.

The ZooKeeper server status and mode are transparent from Centra Dogma replicas because the replicas only commute with others via the Curator client.

ikhoon:
Anyway, I will add some tests for it. :-)

minwoox:
> Anyway, I will add some tests for it. :-)

Yeah, let's see how it goes. 😄

server/src/main/java/com/linecorp/centraldogma/server/ZooKeeperServerConfig.java
```diff
@@ -41,12 +46,16 @@
...
-    public ZooKeeperAddress(@JsonProperty(value = "host", required = true) String host,
-                            @JsonProperty(value = "quorumPort", required = true) int quorumPort,
-                            @JsonProperty(value = "electionPort", required = true) int electionPort,
-                            @JsonProperty(value = "clientPort", defaultValue = "0") int clientPort) {
+    public ZooKeeperServerConfig(@JsonProperty(value = "host", required = true) String host,
+                                 @JsonProperty(value = "quorumPort", required = true) int quorumPort,
+                                 @JsonProperty(value = "electionPort", required = true) int electionPort,
+                                 @JsonProperty(value = "clientPort", defaultValue = "0") int clientPort,
+                                 @JsonProperty("groupId") @Nullable Integer groupId,
+                                 @JsonProperty(value = "weight", defaultValue = "1") Integer weight) {
```
trustin, line 58: Could we use `int` here given that we specified `defaultValue`?

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -417,14 +428,53 @@ private EmbeddedZooKeeper startZooKeeper() throws Exception {
...
+                boolean isHierarchical = true;
+                for (Entry<Integer, ZooKeeperServerConfig> entry : servers.entrySet()) {
+                    final Integer groupId = entry.getValue().groupId();
+                    if (groupId == null) {
+                        isHierarchical = false;
+                        final List<ZooKeeperServerConfig> noGroupIds =
+                                servers.values().stream()
+                                       .filter(serverConfig -> serverConfig.groupId() == null)
+                                       .collect(toImmutableList());
+                        logger.warn("Hierarchical quorums are disabled. 'groupId' are missed in {}",
```
trustin, line 460: missed -> missing

trustin (commented):
Looking nice!

server/src/main/java/com/linecorp/centraldogma/server/ZooKeeperServerConfig.java
ikhoon: Initially, I tried to use `int`. However, the `defaultValue` did not work.
From the [Javadoc](https://fasterxml.github.io/jackson-annotations/javadoc/2.8/com/fasterxml/jackson/annotation/JsonProperty.html#defaultValue()), it is only used for documentation. 🤪

trustin: Ah, that sounds unintuitive 😓

trustin: Then should we also update other primitive parameters with `defaultValue`, e.g. `clientPort`?

ikhoon: The default value of `clientPort` is 0. I thought we don't need to update it because the initialized value is 0. Anyway, let me check our codebase and update where `defaultValue` is used. 😀

ikhoon: Fixed :-)

server/src/test/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutorTest.java
```diff
@@ -254,6 +258,186 @@ void stopWhileWaitingForInitialQuorum() throws Exception {
...
+
+            final Command<Void> command3 = Command.createRepository(Author.SYSTEM, "project", "repo3");
+            assertThatThrownBy(() -> replica1.rm.execute(command3).get(10, TimeUnit.SECONDS))
+                    .isInstanceOf(TimeoutException.class);
+
+            // Restart two instances in Group 3, so the hierarchical quorums should be working again.
+            final CompletableFuture<Void> replica7Start = cluster.get(7).rm.start();
+            final CompletableFuture<Void> replica8Start = cluster.get(8).rm.start();
+            replica7Start.join();
+            replica8Start.join();
```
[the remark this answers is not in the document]
ikhoon: I think so. Let me add more assertion for that.

server/src/main/java/com/linecorp/centraldogma/server/ZooKeeperServerConfig.java
```diff
@@ -41,20 +47,33 @@
      * @param quorumPort the quorum port number
      * @param electionPort the election port number
      * @param clientPort the client port number (0-65535)
+     * @param groupId the group ID to use hierarchical quorums
```
trustin, line 50: use -> use for?

server/src/main/java/com/linecorp/centraldogma/server/ZooKeeperServerConfig.java
```diff
@@ -41,20 +47,33 @@
...
-    public ZooKeeperAddress(@JsonProperty(value = "host", required = true) String host,
-                            @JsonProperty(value = "quorumPort", required = true) int quorumPort,
-                            @JsonProperty(value = "electionPort", required = true) int electionPort,
-                            @JsonProperty(value = "clientPort", defaultValue = "0") int clientPort) {
+    public ZooKeeperServerConfig(@JsonProperty(value = "host", required = true) String host,
+                                 @JsonProperty(value = "quorumPort", required = true) int quorumPort,
+                                 @JsonProperty(value = "electionPort", required = true) int electionPort,
+                                 @JsonProperty(value = "clientPort", defaultValue = "0") Integer clientPort,
+                                 @JsonProperty("groupId") @Nullable Integer groupId,
+                                 @JsonProperty(value = "weight", defaultValue = "1") Integer weight) {
```
trustin, line 59: `@Nullable`?

server/src/main/java/com/linecorp/centraldogma/server/ZooKeeperServerConfig.java
```diff
@@ -41,20 +47,33 @@
...
      */
     @JsonCreator
-    public ZooKeeperAddress(@JsonProperty(value = "host", required = true) String host,
-                            @JsonProperty(value = "quorumPort", required = true) int quorumPort,
-                            @JsonProperty(value = "electionPort", required = true) int electionPort,
-                            @JsonProperty(value = "clientPort", defaultValue = "0") int clientPort) {
+    public ZooKeeperServerConfig(@JsonProperty(value = "host", required = true) String host,
+                                 @JsonProperty(value = "quorumPort", required = true) int quorumPort,
+                                 @JsonProperty(value = "electionPort", required = true) int electionPort,
+                                 @JsonProperty(value = "clientPort", defaultValue = "0") Integer clientPort,
```
trustin, line 57: `@Nullable`?

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -417,14 +428,53 @@ private EmbeddedZooKeeper startZooKeeper() throws Exception {
...
+                        groupBuilder.put(groupId, entry.getKey());
+                    }
+                }
+                if (isHierarchical) {
+                    groupBuilder.build().asMap().forEach((groupId, serverIds) -> {
+                        // group.1=1:2:3
+                        zkProps.setProperty("group." + groupId, colonJoiner.join(serverIds));
+                    });
+                    servers.forEach((serverId, serverConfig) -> {
+                        // weight.1=1
```
trustin, line 473: How about prepending `e.g.` to the examples in the comments?

server/src/test/java/com/linecorp/centraldogma/server/CentralDogmaConfigTest.java
```diff
@@ -141,11 +141,8 @@ void trustedProxyAddress_withDefaultClientAddressSources_withoutProxyProtocol()
 
         final List<ClientAddressSource> sources = cfg.clientAddressSourceList();
         assertThat(sources).hasSize(2);
-        // TODO(hyangtack) Need to add equals/hashCode to ClientAddressSource class.
```
trustin, line 144: Thanks for cleaning up. :bow:

trustin (approved):

server/src/main/java/com/linecorp/centraldogma/server/ZooKeeperServerConfig.java
```diff
@@ -41,20 +47,33 @@
...
+                                 @JsonProperty(value = "weight", defaultValue = "1") Integer weight) {
 
         this.host = requireNonNull(host, "host");
         this.quorumPort = validatePort(quorumPort, "quorumPort");
         this.electionPort = validatePort(electionPort, "electionPort");
 
+        if (clientPort == null) {
+            clientPort = 0;
+        }
         checkArgument(clientPort >= 0 && clientPort <= 65535,
```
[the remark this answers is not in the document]
ikhoon: Those values are checked by `validatePort(...)`. 😉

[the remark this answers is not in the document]
minwoox: I missed that. 😅

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
ikhoon: Nice suggestion!

minwoox (approved):
Great job, @ikhoon!

trustin:
Awesome. One step closer to multi-datacenter support :+1:

ikhoon:
Thanks for reviewing 😀

PR #542, opened 2020-10-23
introduce client-side micrometer, start with exposing watcher revision metrics

jrhee17:
Let me know what you guys think of the following approach

If all looks good, will add tests/make the PR more serious


client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogma.java
```diff
@@ -201,4 +205,9 @@ protected final ScheduledExecutorService executor() {
             return CompletableFuture.completedFuture(revision);
         }
     }
+
+    @Override
+    public MeterRegistry meterRegistry() {
```
ikhoon, line 210: Add `final`?

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogma.java
```diff
@@ -36,21 +36,25 @@
...
+    private final MeterRegistry meterRegistry;
 
     /**
      * Creates a new instance.
      *
      * @param executor the {@link ScheduledExecutorService} which will be used for scheduling the tasks
      *                 related with automatic retries.
      */
-    protected AbstractCentralDogma(ScheduledExecutorService executor) {
+    protected AbstractCentralDogma(ScheduledExecutorService executor, MeterRegistry meterRegistry) {
```
ikhoon, line 55: This could cause breaking changes.
Could you add a new one?

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java
```diff
@@ -546,4 +548,6 @@
      */
     <T> Watcher<T> repositoryWatcher(String projectName, String repositoryName, String pathPattern,
                                      Function<Revision, ? extends T> function);
+
+    MeterRegistry meterRegistry();
```
ikhoon, line 552: Javadoc?

ikhoon (commented):
Oops... I missed this PR. Mostly LGTM. 🙇‍♂️

minwoox:
This looks good!

minwoox:
> I guess later if needed *CentralDogma can use ClientFactory.meterRegistry by default if not specified.

How about implementing it in this PR?

> *CentralDogma.meterRegistry and ClientFactory.meterRegistry might be different,

Shouldn't we warn in this case?

jrhee17:
> *CentralDogma.meterRegistry and ClientFactory.meterRegistry might be different,
> Shouldn't we warn in this case?

I tried this out, but this means for the user either:
1) Provide the same `meterRegistry` should be provided for both `ClientFactory`, `CentralDogma`
2) Or set `meterRegistry` for `ClientFactory` only relying on the fallback logic

I'm thinking whether simply removing the option to set `meterRegistry` in `CentralDogma` and relying on `ClientFactory` only may provide a more intuitive API

Let me know what you think

jrhee17:
rebased once - gentle ping, as I still think this PR has value

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogmaBuilder.java
```diff
@@ -415,4 +419,16 @@ public final B retryIntervalOnReplicationLagMillis(long retryIntervalOnReplicati
...
+
+    /**
+     * Sets the {@link MeterRegistry} used to collect metrics.
+     */
+    public final B meterRegistry(MeterRegistry meterRegistry) {
+        this.meterRegistry = Optional.of(requireNonNull(meterRegistry, "meterRegistry"));
+        return self();
+    }
+
+    protected Optional<MeterRegistry> meterRegistry() {
```
ikhoon, line 431: Could we use `@Nullable` instead of `Optional` for better interop with other JVM languages such as Kotlin that translate `@Nullable` as `?` type?
https://kotlinlang.org/docs/java-interop.html#nullability-annotations

client/java/src/main/java/com/linecorp/centraldogma/internal/client/AbstractWatcher.java
```diff
@@ -296,18 +309,28 @@ private void notifyListeners() {
...
-            });
+            }, executor);
+            futures.add(future);
         }
+        CompletableFuture.allOf(futures.toArray(new CompletableFuture[0])).thenAccept(ignored -> {
+            final boolean result = futures.stream().allMatch(CompletableFuture::join);
+            if (result) {
+                latestNotifiedRevision.set(latest.revision().major());
+            }
+        });
```
ikhoon, line 333: Could use `CompletableFutures.allAsList(...)`?
```java
CompletableFutures.allAsList(futures).thenApply(results -> {
   final boolean result = results.stream().allMatch(x -> x);
   ...
})
```

ikhoon (commented):
Sorry for the late review. Mostly looking good to me!

trustin:
> I'm thinking whether simply removing the option to set meterRegistry in CentralDogma and relying on ClientFactory only may provide a more intuitive API
>
> Let me know what you think

+1 for always using `ClientFactory.meterRegistry()`. How about adding `disableMetrics()` instead for those who want to opt out of metrics? (although not very likely)

jrhee17:
Ready for the next round of reviews 🙏

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java
```diff
@@ -591,4 +595,13 @@
      */
     <T> Watcher<T> repositoryWatcher(String projectName, String repositoryName, String pathPattern,
                                      Function<Revision, ? extends T> function, Executor executor);
+
+    /**
+     * Returns the {@link MeterRegistry} which collects {@link CentralDogma} specific metrics.
+     * If this value is null, metric collection is disabled.
```
minwoox, line 601: nit:
```suggestion
     * If this value is {@code null}, metric collection is disabled.
```

client/java/src/main/java/com/linecorp/centraldogma/internal/client/AbstractWatcher.java
```diff
@@ -127,6 +134,16 @@ protected AbstractWatcher(CentralDogma client, ScheduledExecutorService watchSch
         this.repositoryName = requireNonNull(repositoryName, "repositoryName");
         this.pathPattern = requireNonNull(pathPattern, "pathPattern");
 
+        final MeterRegistry meterRegistry = client.meterRegistry();
+        if (meterRegistry != null) {
+            Gauge.builder("centraldogma.client.watcher.revision",
+                          this, watcher -> watcher.latestNotifiedRevision.get())
```
minwoox, line 140: Question: Can't we use a counter here?
Can't we use an `AtomicInteger`? 🤔

client/java/src/main/java/com/linecorp/centraldogma/internal/client/AbstractWatcher.java
```diff
@@ -296,18 +313,29 @@ private void notifyListeners() {
...
                                 projectName, repositoryName, pathPattern, latest.revision(), e);
+                    return false;
                 }
-            });
+            }, executor);
+            futures.add(future);
         }
+
+        CompletableFutures.allAsList(futures).thenAccept(results -> {
+            final boolean result = results.stream().allMatch(x -> x);
```
minwoox, line 334: Shouldn't we update the `latestNotifiedRevision` regardless of a listener throws an exception?

minwoox (commented):
Thanks! left some questions. 😄

client/java/src/main/java/com/linecorp/centraldogma/internal/client/AbstractWatcher.java
jrhee17: I've added another metric which updates the revision irregardless of whether the notify succeeds.

I think the two metrics combined can give users a good idea of whether properties have been propagated correctly.

trustin:
What's the status of this PR since then? 😅

jrhee17:
superseded by https://github.com/line/centraldogma/pull/1094

PR #548, opened 2021-01-13
Make client DNS query timeout configurable

mauhiz:
Motivation:
Central Dogma client (based on Armeria) currently uses a non-configurable default of 5s.
We want to be able to configure it based on our network situation.

Modifications:
- Add `DnsAddressEndpointGroupConfigurator` to configure `DnsAddressEndpointGroup`.

Result:
- You can now customize `DnsAddressEndpointGroup` in `ArmeriaCentralDogmaBuilder`.

### DIFF
--- a/client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/AbstractArmeriaCentralDogmaBuilder.java
+++ b/client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/AbstractArmeriaCentralDogmaBuilder.java
@@ -144,10 +144,9 @@ public class AbstractArmeriaCentralDogmaBuilder<B extends AbstractArmeriaCentral
             if (addr.isUnresolved()) {
                 final DnsAddressEndpointGroupBuilder dnsAddressEndpointGroup = DnsAddressEndpointGroup
                         .builder(addr.getHostString())
-                        .eventLoop(clientFactory.eventLoopGroup().next())
-                        .port(addr.getPort());
+                        .eventLoop(clientFactory.eventLoopGroup().next());
                 dnsAddressEndpointGroupConfigurator.configure(dnsAddressEndpointGroup);
-                groups.add(dnsAddressEndpointGroup.build());
+                groups.add(dnsAddressEndpointGroup.port(addr.getPort()).build());
             } else {
                 staticEndpoints.add(toResolvedHostEndpoint(addr));
             }
--- a/client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/AbstractArmeriaCentralDogmaBuilder.java
+++ b/client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/AbstractArmeriaCentralDogmaBuilder.java
@@ -56,7 +56,7 @@ public class AbstractArmeriaCentralDogmaBuilder<B extends AbstractArmeriaCentral
     private ClientFactory clientFactory = ClientFactory.ofDefault();
     private ArmeriaClientConfigurator clientConfigurator = cb -> {};
     private Duration healthCheckInterval = Duration.ofMillis(DEFAULT_HEALTH_CHECK_INTERVAL_MILLIS);
-    private Consumer<DnsAddressEndpointGroupBuilder> dnsAddressEndpointGroupCustomizer = b -> {};
+    private DnsAddressEndpointGroupConfigurator dnsAddressEndpointGroupConfigurator = b -> {};
 
     /**
      * Returns the {@link ClientFactory} that will create an underlying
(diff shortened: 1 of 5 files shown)


mauhiz:
tests pass locally. Is this a known issue?

minwoox:
> tests pass locally. Is this a known issue?

Yeah, it seems flaky. Sorry about it. 😅

client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/AbstractArmeriaCentralDogmaBuilder.java
```diff
@@ -82,6 +87,15 @@ public final B clientConfigurator(ArmeriaClientConfigurator clientConfigurator)
         return self();
     }
 
+    /**
+     * Sets the DNS query timeout for looking up Central Dogma server.
+     * If not set, it will use <a href="https://line.github.io/armeria/">Armeria</a> client built-in defaults.
+     */
+    public final B dnsQueryTimeout(@Nullable Duration dnsQueryTimeout) {
```
minwoox, line 94: Could you add `dnsQueryTimeoutMillis` as well because we always have the pair of setters for the time?
Also, could you remove `@Nullable` and delegate to the `dnsQueryTimeoutMillis` method, please?

mauhiz: I don't get the need to use all available setters

mauhiz: Can you please elaborate on the reason why having 2 setters here too is better?

minwoox: It's sort of our convention to provide the same API experience. 😄

mauhiz: okay, understood. Let me add it.

minwoox (commented):
Sorry for making you do this twice but
How about removing the setters and adding
`dnsAddressEndpointGroupCustomizer(Consumer<DnsAddressEndpointGroupBuilder>)`?
Because a user may want to set different properties for the `DnsAddressEndpointGroup` but we cannot add all the setters to this `ArmeriaCentralDogmaBuilder`.

mauhiz:
okay, is it fine to keep the springboot-based injection for the value? Or should users of this feature use a custom bean declaration

minwoox:
> okay, is it fine to keep the springboot-based injection for the value? Or should users of this feature use a custom bean declaration

That's a good question. How about using the same approach as we do for `ArmeriaClientConfigurator`?

mauhiz:
force-pushed as I redid it from scratch

client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/AbstractArmeriaCentralDogmaBuilder.java
```diff
@@ -82,6 +84,13 @@ public final B clientConfigurator(ArmeriaClientConfigurator clientConfigurator)
         return self();
     }
 
+    public final B dnsAddressEndpointGroupCustomizer(
```
ikhoon, line 87: Javadoc?

client/java-spring-boot-autoconfigure/src/main/java/com/linecorp/centraldogma/client/spring/CentralDogmaClientAutoConfiguration.java
```diff
@@ -87,13 +89,15 @@ public CentralDogma dogmaClient(
             Environment env,
             CentralDogmaSettings settings,
             @ForCentralDogma ClientFactory clientFactory,
-            Optional<ArmeriaClientConfigurator> armeriaClientConfigurator) throws UnknownHostException {
+            Optional<ArmeriaClientConfigurator> armeriaClientConfigurator,
+            Optional<Consumer<DnsAddressEndpointGroupBuilder>> dnsAddressEndpointGroupCustomizer) throws UnknownHostException {
```
ikhoon, line 93: Don't we need a dedicated interface if we choose to approach in the same way as `ArmeriaClientConfigurator` such as:
```java
public interface DnsAddressEndpointGroupConfigurator {
    void configure(DnsAddressEndpointGroupBuilder builder);
}

Comment (mauhiz) [reply client/java-spring-boot-autoconfigure/src/main/java/com/linecorp/centraldogma/client/spring/CentralDogmaClientAutoConfiguration.java:93]: okay, it works well without but it could be useful. Let me add it.

**Review by ikhoon — APPROVED** [review]
Thanks! @mauhiz

File: client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/AbstractArmeriaCentralDogmaBuilder.java [hunk 4]
```diff
@@ -129,10 +142,12 @@ protected final EndpointGroup endpointGroup() throws UnknownHostException {
...
         for (final InetSocketAddress addr : hosts) {
             if (addr.isUnresolved()) {
-                groups.add(DnsAddressEndpointGroup.builder(addr.getHostString())
-                                                  .eventLoop(clientFactory.eventLoopGroup().next())
-                                                  .port(addr.getPort())
-                                                  .build());
+                final DnsAddressEndpointGroupBuilder dnsAddressEndpointGroup = DnsAddressEndpointGroup
+                        .builder(addr.getHostString())
+                        .eventLoop(clientFactory.eventLoopGroup().next())
+                        .port(addr.getPort());
```
minwoox, line 148: Could you call this `.port(addr.getPort())` after
`dnsAddressEndpointGroupConfigurator.configure(dnsAddressEndpointGroup);` is called?
Because there's a chance that a user sets the port by mistake.

minwoox (commented):
Just one nit. Thanks! 😄

client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/AbstractArmeriaCentralDogmaBuilder.java
mauhiz: Changed. Meanwhile, as a "user", let me protest against babysitting : when I make mistakes I can fix them, but when the library designer prevents me from doing what he thinks is a mistake I can only weep...

minwoox (approved):
Thanks a ton!

minwoox:
🎉 🎉 🎉

PR #553, opened 2021-02-01
Add CentralDogmaEndpointGroupBuilder

minwoox:
Motivation:
A user might want to use a different `EndpointSelectionStrategy` which is not possible currently.

Modification:
- Add `CentralDogmaEndpointGroupBuilder` so a user can specify an `EndpointSelectionStrategy`.

Result:
- You can now use a different `EndpointSelectionStrategy` when using `CentralDogmaEndpointGroup`.


trustin (approved):

ikhoon (approved):
Thanks! @trustin

minwoox:
Thanks, reviewers!

PR #554, opened 2021-02-02
Fix that `mergeFiles` API modifies the content of the merge sources

minwoox:
Motivation:
`mergeFiles` API should not modify the content of the merge sources

Modifications:
- Fix a bug where the `mergeFiles` API modifies the content of the first file of the merge sources

Result:
- You no longer see that the `mergeFiles` API does not modify the content of the merge sources.

### DIFF
--- a/common/src/main/java/com/linecorp/centraldogma/common/MergeQuery.java
+++ b/common/src/main/java/com/linecorp/centraldogma/common/MergeQuery.java
@@ -78,7 +78,7 @@ public interface MergeQuery<T> {
      */
     static MergeQuery<JsonNode> ofJsonPath(Iterable<MergeSource> mergeSources,
                                            Iterable<String> jsonPaths) {
-        if (Iterables.isEmpty(jsonPaths)) {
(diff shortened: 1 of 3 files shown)


common/src/main/java/com/linecorp/centraldogma/common/MergeQuery.java
```diff
@@ -77,6 +78,9 @@
      */
     static MergeQuery<JsonNode> ofJsonPath(Iterable<MergeSource> mergeSources,
                                            Iterable<String> jsonPaths) {
+        if (Iterables.isEmpty(jsonPaths)) {
```
ikhoon, line 81: `requireNonNull`?

ikhoon (approved):
Thanks!

common/src/main/java/com/linecorp/centraldogma/common/MergeQuery.java
minwoox: Thanks fixed. 😉

minwoox:
Thanks for reviewing!

PR #567, opened 2021-03-02
Add metadata only once for a repository

minwoox:
Motivation:
CentralDogma currently adds metadata information of a repository into metadata.json if it doesn't exist.
However, if `MetadataService.getProject(...) is called multiple times before the metadata is added, there can be multiple commits of adding operation.
(We have watched over 1000 times for the same repository.)

Modifications:
- Fix to commit only once for the metadata of a repository.

Result:
- You no longer see multiple `Adding missing repository metadata` for the same repository.

Todo:
- Change to add metadata after `CreateRepository` is called.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java
@@ -26,9 +26,11 @@ import static com.linecorp.centraldogma.server.metadata.Tokens.validateSecret;
 import static java.util.Objects.requireNonNull;
 
 import java.util.Collection;
+import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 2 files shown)


trustin (approved):
Thanks, @minwoox and @ikhoon :smile:

server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java
```diff
@@ -119,8 +123,26 @@ public MetadataService(ProjectManager projectManager, CommandExecutor executor)
                     continue;
                 }
 
+                final CompletableFuture<Revision> future = new CompletableFuture<>();
+                final CompletableFuture<Revision> futureInMap =
+                        reposInAddingMetadata.computeIfAbsent(repo, key -> future);
```
ikhoon, line 128: Don't we need to include `projectName` in the cache key?

minwoox: Oops thanks!

minwoox: Fixed. 😉

ikhoon (approved):
Thanks! @minwoox 🙇‍♂️

minwoox:
Thanks @trustin and @ikhoon!

PR #568, opened 2021-03-07
Normalize chagnes and commit in the same lock when pushing

minwoox:
Motivation:
Central Dogma server does not directly apply the changes in a push command but applies
the changes produced by the preview diff call. It reduces the size of changes so we can store
smaller amount of data for the ZooKeeper replication log.
However, it has two drawbacks (or bugs):
- After the preview diff call, the `TestAbsenseOperation` or `Test` operation in a JSON patch call is gone.
  - Because we do not operate two calls in the same write lock if there's a change of the target file between the preview diff call and the commit call, it's not working as expected.
- All replicas have to do two operations (preview diff and commit) even though only one replica can handle it.

Modifications:
- Rename `PushAsIsCommand` to `PushCommand`.
- Introduce `NormalizingPushCommand`.
  - `NormalizingPushCommand` is used for the replica who does the commit first.
    - The replica normalizes changes and commits in the same lock.
    - The replica creates `PushAsIsCommand` with the normalized changes and sends it to other replicas.
    - Other replicas do not have to normalize changes again.

Result:
- You no longer see the multiple commits that have the same contents.


server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -1072,7 +1070,19 @@ private static String pathFromRevision(long revision) {
...
             final T result = delegate.execute(command).get();
-            final ReplicationLog<T> log = new ReplicationLog<>(replicaId(), command, result);
+            final ReplicationLog<?> log;
+            if (command.type() == CommandType.APPLYING_DIFF_PUSH) {
+                final ApplyingDiffPushCommand pushCommand = (ApplyingDiffPushCommand) command;
+                assert result instanceof CommitResult : result;
+                final CommitResult commitResult = (CommitResult) result;
+                final Command<Revision> replicationPushCommand = Command.replicationPush(
+                        pushCommand.timestamp(), pushCommand.author(), pushCommand.projectName(),
+                        pushCommand.repositoryName(), pushCommand.baseRevision(), pushCommand.summary(),
```
trustin, line 1080: - `pushCommand.baseRevision()` -> `commitResult.revision() - 1`
- `pushCommand.changes()` -> `commitResult.changes()`

server/src/main/java/com/linecorp/centraldogma/server/command/Command.java
```diff
@@ -48,6 +48,9 @@
         @Type(value = RemoveRepositoryCommand.class, name = "REMOVE_REPOSITORY"),
         @Type(value = PurgeRepositoryCommand.class, name = "PURGE_REPOSITORY"),
         @Type(value = UnremoveRepositoryCommand.class, name = "UNREMOVE_REPOSITORY"),
+        @Type(value = ApplyingDiffPushCommand.class, name = "APPLYING_DIFF_PUSH"),
+        @Type(value = ReplicationPushCommand.class, name = "REPLICATION_PUSH"),
+        // Deprecated. Should use OriginalPushCommand or ReplicationPushCommand.
         @Type(value = PushCommand.class, name = "PUSH"),
```
trustin, line 54: How about this:

- Rename `PushCommand` to `PushAsIsCommand` or `JustPushCommand`.
- Remove `ReplicationPushCommand` and use `PushAsIsCommand`.
- Rename `ApplyingDiffPushCommand` to `NormalizingPushCommand`.

minwoox: Those are good suggestions. Applied. 😄

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
minwoox: 😱

server/src/main/java/com/linecorp/centraldogma/server/command/AbstractPushCommand.java
```diff
@@ -38,7 +35,7 @@
 /**
  * A {@link Command} which is used for pushing changes to the repository.
  */
-public final class PushCommand extends RepositoryCommand<Revision> {
+class AbstractPushCommand<T> extends RepositoryCommand<T> {
```
ikhoon, line 38: Is it better to add `abstract`?

server/src/main/java/com/linecorp/centraldogma/server/command/StandaloneCommandExecutor.java
```diff
@@ -249,33 +254,35 @@ protected void doStop(@Nullable Runnable onReleaseLeadership) {
         }, repositoryWorker);
     }
 
-    private CompletableFuture<Revision> push(PushCommand c) {
+    private CompletableFuture<CommitResult> push(AbstractPushCommand<?> c, boolean applyPreviewDiff) {
```
ikhoon, line 257: Not sure this is better. 😅
```suggestion
    private CompletableFuture<CommitResult> push(AbstractPushCommand<?> c, boolean normalizing) {
```

server/src/main/java/com/linecorp/centraldogma/server/command/StandaloneCommandExecutor.java
```diff
@@ -171,8 +171,13 @@ protected void doStop(@Nullable Runnable onReleaseLeadership) {
...
         }
 
-        if (command instanceof PushCommand) {
-            return (CompletableFuture<T>) push((PushCommand) command);
+        if (command instanceof NormalizingPushCommand) {
+            return (CompletableFuture<T>) push((NormalizingPushCommand) command, true);
+        }
+
+        if (command instanceof PushAsIsCommand) {
+            return (CompletableFuture<T>) push((AbstractPushCommand<Revision>) command, false)
```
ikhoon, line 179:
```suggestion
            return (CompletableFuture<T>) push((PushAsIsCommand) command, false)
```

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -1072,7 +1070,23 @@ private static String pathFromRevision(long revision) {
...
+                        normalizingPushCommand.timestamp(), normalizingPushCommand.author(),
+                        normalizingPushCommand.projectName(), normalizingPushCommand.repositoryName(),
+                        commitResultRevision.backward(1),
+                        normalizingPushCommand.summary(), normalizingPushCommand.detail(),
+                        normalizingPushCommand.markup(),
+                        commitResult.changes());
+                log = new ReplicationLog<>(replicaId(), pushAsIsCommand, commitResultRevision);
+            } else {
+                log = new ReplicationLog<>(replicaId(), command, result);
+            }
```
ikhoon, line 1089: - Could it be convenient if we add `Command<Revision> pushAsIsCommand = normalizingPushCommand.asIs(commitResult)`?
- Add `assert !(command instanceof NormalizingPushCommand)` to the constructor of `ReplicationLog`?

trustin (approved):
LGTM once the comments from @ikhoon are addressed.

ikhoon (approved):
Thanks! @minwoox

minwoox:
Thanks, @trustin and @ikhoon!

PR #570, opened 2021-03-18
Remove a write lock from git gc

ikhoon:
Motivation:

A git gc can run while commiting new revisions.
We don't need to use a write lock for git gc.

Futhermore, a gc takes about 15 seconds for 10k commits.
It is a huge block. So the long locking time could make `repositoryWorker` exhausted

Modifications:

- Remove a write lock from `GitRepository.gc()`
- Call `start()` if a stopwatch is reset in `RepositoryGarbageCollectionPlugin`

Result:

- A git gc no longer blocks git commits.


ikhoon:
@minwoox pointed out that it is risky to run git gc with other processes concurrently.
https://git-scm.com/docs/git-gc#_notes
Let me close this PR. 😅

minwoox:
Thanks a lot for your work. 🙇‍♂️

PR #571, opened 2021-03-18
Add a lock timeout for ZooKeeperCommandExecutor

ikhoon:
Motivation:

A command submitted to ZooKeeperCommandExecutor could be waiting for a lock for a long time
if a command acquired the lock executes blocked by another lock.

Modifications:

- Add a timeout for ZooKeeper global locking

Result:

- A pending task in a `ZooKeeperCommandExecutor` could be fail fast when failed to get a lock.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
@@ -46,9 +46,11 @@ import java.util.Objects;
 import java.util.StringJoiner;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executor;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.function.BiConsumer;
 import java.util.function.Supplier;
@@ -164,6 +166,8 @@ class GitRepository implements Repository {
     }
 
     private final ReadWriteLock rwLock = new ReentrantReadWriteLock();
+    @VisibleForTesting
+    final ReentrantLock gcLock = new ReentrantLock();
     private final Project parent;
     private final Executor repositoryWorker;
     @VisibleForTesting
@@ -886,7 +890,7 @@ class GitRepository implements Repository {
     @Override
     public CompletableFuture<CommitResult> commit(
             Revision baseRevision, long commitTimeMillis, Author author, String summary,
-            String detail, Markup markup, Iterable<Change<?>> changes, boolean normalizing) {
+            String detail, Markup markup, Iterable<Change<?>> changes, boolean directExecution) {
         requireNonNull(baseRevision, "baseRevision");
         requireNonNull(author, "author");
         requireNonNull(summary, "summary");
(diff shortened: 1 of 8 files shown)


minwoox (commented):
Thanks!
How about adding a lock timeout in `GitRepository` as well?
I think there's a such scenario:
- A commit acquires the ZooKeeper lock
- Before the commit acquires a write lock in `GitRepository`, gc executor acquires the write lock.
- The ZooKeeper command executor waits to acquire the lock.
- Other commands are queued.

ikhoon:
> Before the commit acquires a write lock in GitRepository, gc executor acquires the write lock.

That's a good point. I am working on acquiring a Zookeeper lock before executing a gc task. If it is applied, the problem could be fixed together.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
```diff
@@ -1497,14 +1505,15 @@ private Revision blockingFindLatestRevision(Revision lastKnownRevision, String p
 
     @Override
     public Revision gc() throws Exception {
-        rwLock.writeLock().lock();
+        // Should not acquire `writeLock()` which prevents reading commits from this repository.
```
minwoox, line 1508: 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
```diff
@@ -1590,6 +1599,44 @@ private void readUnlock() {
...
 
+    @VisibleForTesting
+    boolean writeLock(boolean directExecution) {
+        // Should not commit changes if gc is running
+        boolean hasGcLock;
+        if (directExecution) {
+            try {
+                // Waits only 10 seconds the gc lock to prevent a command worker queue from filing up
+                // due to a long gc, if the command was directly submitted to this server.
+                hasGcLock = gcLock.tryLock() || gcLock.tryLock(10, TimeUnit.SECONDS);
```
minwoox, line 1610: `gcLock.tryLock()` does not honor the fairness.
Shouldn't we honor the fairness because there can be multiple commits waiting?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
```diff
@@ -1590,6 +1599,44 @@ private void readUnlock() {
...
+            } catch (InterruptedException e) {
+                hasGcLock = false;
+            }
+        } else {
+            // Should not fail replaying a log due to a long gc.
+            // Because if a log is failed to replay, this replica will enter read-only mode.
+            gcLock.lock();
```
minwoox, line 1617: Shoudn't we acquire this lock when this repository is closed?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
ikhoon: I don't think of that case. It would be better to remove `tryLock()` for that.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
ikhoon: Ah.. `gcLock` is needed.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
```diff
@@ -1590,6 +1601,44 @@ private void readUnlock() {
...
+    boolean writeLock(boolean directExecution) {
+        // Should not commit changes if gc is running
+        boolean hasGcLock;
+        if (directExecution) {
+            try {
+                // Waits only 10 seconds the gc lock to prevent a command worker queue from filing up
+                // due to a long gc, if the command was directly submitted to this server.
+                hasGcLock = gcLock.tryLock(10, TimeUnit.SECONDS);
+            } catch (InterruptedException e) {
+                hasGcLock = false;
```
minwoox, line 1614: Let's do
`Thread.currentThread().interrupt();`

minwoox (approved):
One nit. Thanks! 😄

minwoox:
@ikhoon 👍 👍 👍

PR #578, opened 2021-04-05
Add CentralDogmaClientFactoryConfigurator

### LINKED ISSUE #577: Autoconfigured client creates an armeria client factory bean that can interfere with the user's spring context
Example: https://github.com/m50d/centraldogma/commit/fc2f5703a4c64fd99e7a074a1ed19e78c5881cbf - this test fails, but I think it should pass. `dogmaClientFactory` is an internal implementation detail of the CentralDogma client, it shouldn't interfere with the user's existing use of Armeria.

I can't immediately see how to fix this though - setting `autowireCandidate=false` breaks the way `dogmaClientFactory` is currently used. And of course any change to this would be potentially breaking from a user's point of view.

minwoox:
Motivation:
Related #577
We use `@ForCentralDogma` to return the `ClientFactory` which is used to build `ArmeriaCentralDogma` client.
However, if there's a user who uses another bean that returns the `ClientFactory` to use for Armeria client,
there might be a conflict.

Modifications:
- Add `CentralDogmaClientFactoryConfigurator` to configure the `ClientFactory` which is used to build the `ArmeriaCentralDogma` client.

Result:
- Fix #577
- You can now use `CentralDogmaClientFactoryConfigurator` to configure the `ClientFactory` for `ArmeriaCentralDogma` client.
  ```java
  @Bean
  CentralDogmaClientFactoryConfigurator configurator() {
      return builder -> builder.connectTimeoutMillis(3000);
  }
  ```
- (Breaking) `@ForCentralDomga` is now gone. Use `CentralDogmaClientFactoryConfigurator` to configure `ClientFactory`.


minwoox:
I realized that as long as `@ForCentralDogma` exists, we cannot solve the problem illustrated in #577.
So let me remove the `@ForCentralDogma` completely.

ikhoon (approved):
Thanks! @minwoox

minwoox:
Thanks, @ikhoon for reviewing!

Issue #575, opened 2021-04-01
Provide a way to remove the old commits


minwoox:
Central Dogma uses jGit to store data. Due to the nature of Git that stores unlimited history,
Central Dogma will eventually get in trouble managing disk usage.
We can handle this by removing old commits that are not used anymore. In order to do that, we cannot just
use Git squash because that changes the hash of the commits.

So what we are going to do is (that is suggested by @trustin):
- Creating a new Git repository whenever the commits of the previously created repository
  reaches a certain number(`N`).
  - So we are going to have `R`, `R + 1`, `R + 2`, `R + 3`, ...
  - When we create the repository `R + 2`, we can remove the repository created two times earlier that is `R` in this case.
    (We can also use a batch to remove that)
- Limiting the number of history that a user can get to `N`
  - That means that `Revision.INIT` is not `1` anymore. It will be `max(1, Revision.HEAD - N + 1)`

PR #586, opened 2021-04-15
Add DiffFormatter for the diff of revisions in two repositories

minwoox:
Motivation:
To implement #575, we should be able to make diffs between two different repositories.

Modifications:
- Add `TwoRepositoriesDiffFormatter` and `TwoRepositoriesTreeWalk` to make diffs between two different repositories.
- Misc
  - Use `@TempDir` in `GitRepositoryManagerTest` so that the temporarily created directories for the test are not left anymore.

Result:
- Ready to implement #575


KarboniteKream:
Looks like `@TempDir` still has trouble deleting the directory on Windows 😭 

It seems to be due to having read-only files in the directory, but AFAIK the extension should try to make them readable before deletion. I wonder if the issue is related to JUnit or JGit...

minwoox:
> When does it leave the files? I've tried replicating it (on MacOS), but TemporaryFolderExtension always deletes the files after itself 🤔

I think https://github.com/line/centraldogma/blob/master/server/src/test/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepositoryManagerTest.java#L124
this line creates a directory and does not remove it. 😅 

> Looks like @TempDir still has trouble deleting the directory on Windows 😭

Ah, I didn't know about it. 😅 

> Perhaps GitRepositoryManager is still holding some file locks?

I guess so, let me see what happened. Thanks a lot for your investigation. 😄

KarboniteKream:
> Ah, I didn't know about it. 😅

It's possible it's due to the test implementation. When I was implementing JUnit 5 support, I wasn't too familiar with how Central Dogma works, so I wasn't able to figure it out. If your latest commit works, I'll check other tests and cleanup `TemporaryFolderExtension`.

minwoox:
> It's possible it's due to the test implementation. When I was implementing JUnit 5 support, I wasn't too familiar with how Central Dogma works, so I wasn't able to figure it out. If your latest commit works, I'll check other tests and cleanup TemporaryFolderExtension.

I guess it worked at this time. 😄  Oh, could you? Thanks a lot for taking care of this. 🙇‍♂️

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/TwoRepositoriesTreeWalk.java
```diff
@@ -0,0 +1,545 @@
...
+            }
+            if (trace[1] == -1) {
+                throw new ArrayIndexOutOfBoundsException(delta);
+            }
+            prevPtr = trace[0];
+            currPtr = trace[1];
+            parseEntry();
+        }
+
+        private void parseEntry() {
```
ikhoon, line 440: Could you leave an entry format of a tree as a comment?
```
[mode] [file/folder name]\0[SHA-1 of referencing blob or tree]
```

ikhoon (commented):
Mostly looking good. I've learned a lot of Git internals.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/TwoRepositoriesDiffFormatter.java
```diff
@@ -0,0 +1,168 @@
...
+import org.eclipse.jgit.diff.DiffEntry;
+import org.eclipse.jgit.diff.DiffEntry.ChangeType;
+import org.eclipse.jgit.lib.AbbreviatedObjectId;
+import org.eclipse.jgit.lib.FileMode;
+import org.eclipse.jgit.lib.MutableObjectId;
+import org.eclipse.jgit.lib.ObjectId;
+
+import com.google.common.collect.ImmutableList;
+
+final class TwoRepositoriesDiffFormatter {
```
trustin, line 41: `InterRepositoryDiffFormatter` or `CrossRepositoryDiffFormatter` ?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/TwoRepositoriesTreeWalk.java
```diff
@@ -0,0 +1,545 @@
...
+import org.eclipse.jgit.lib.ObjectId;
+import org.eclipse.jgit.lib.ObjectReader;
+import org.eclipse.jgit.lib.Repository;
+import org.eclipse.jgit.revwalk.RevTree;
+import org.eclipse.jgit.revwalk.RevWalk;
+import org.eclipse.jgit.treewalk.AbstractTreeIterator;
+import org.eclipse.jgit.treewalk.filter.TreeFilter;
+import org.eclipse.jgit.util.RawParseUtils;
+
+final class TwoRepositoriesTreeWalk implements AutoCloseable {
```
trustin, line 52: `InterRepositoryTreeWalk` or `CrossRepositoryTreeWalk` ?

server/src/test/java/com/linecorp/centraldogma/server/internal/storage/repository/git/DiffFormatterTest.java
```diff
@@ -0,0 +1,197 @@
...
+        //  The revision number - 1 of Single repo is the corresponding revision of New repo.
+
+        //  New repo:
+        //  Revision     1            2                          3              4           5             6
+        //  initial commit  Add /a.json              Modify /a.json                           Rename to /aa.json
+        //  initial commit  Add /a.txt
+        //  initial commit  Add /b/b.json
+        //  initial commit  Add /b/b.txt                           Remove /b/b.txt
+        //  initial commit                                                         Add /c.json
+        //  initial commit                                                         Add /c.txt
```
trustin, line 108: How about using some vertical lines (`|`) to make it clear which change belongs to which commit?

trustin (commented):
I didn't check the implementation detail assuming that jGit folks did a good job already and no big changes were made when forking it. Let me know if you need more detailed review. :smile:

minwoox:
> I didn't check the implementation detail assuming that jGit folks did a good job already and no big changes were made when forking it.

Thanks for the review and no problem at all. The point that I wanted to get reviewed was whether I missed some test cases or not. 😄

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/TwoRepositoriesTreeWalk.java
```diff
@@ -0,0 +1,545 @@
...
+        /**
+         * Offset one past the current entry (first byte of next entry).
+         */
+        private int nextPtr;
+
+        private final ObjectReader reader;
+        @Nullable
+        CanonicalTreeParser matches;
+
+        CanonicalTreeParser(CanonicalTreeParser parent, ObjectReader reader) {
```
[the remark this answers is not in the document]
minwoox: Can I just use as it is? 🤔 
I still have to implement the method `createSubtreeIterator` by myself and also I have to override all methods in `AbstractTreeIterator` to delegate the call which is not necessary when just extending `AbstractTreeIterator`. 😄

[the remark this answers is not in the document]
minwoox: @ikhoon Do you have any concerns about it? 🤔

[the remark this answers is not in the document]
ikhoon: Sorry, I missed your comment. I didn't deeply think about the details. Anyway, I'm also fine with the current implementation.

ikhoon (approved):
Nice work! @minwoox 👍 and sorry for the late reivew.

minwoox:
No problem at all. Thanks a lot for reviewing. 😉

minwoox:
We do not need this diff formatter because we are going to implement the new repository in a way that the newly-created subrepository contains at least N commits that it needs. 
So let me close this. 😄

PR #589, opened 2021-04-22
Use a blocking task executor for the watch callacks of Central Dogma client

### LINKED ISSUE #587: Introduce a blocking task executor for client
It is an essential feature to watch changes in a repository.
A `Revision` can be converted to an object using `Function<Revision, ? extends T>` while wathing.
https://github.com/line/centraldogma/blob/d650417c3f38e2bff68a843bf89f6af55e845d91/client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java#L547-L548

Since the signature of `function` is synchronous, users can not convert a result asynchronously. They usually have to do like the followings:
```java
Watcher<String> watcher = dogma.repositoryWatcher("xyz", "foo", "/*.txt",
                                                  revision -> {
                                                      dogma.getFiles("xyz", "foo",
                                                                     revision,
                                                                     "/*.txt").join();
                                                      return "Whatever";
                                                  });
```
Central Dogma does not have a blocking task executor. So the callback is executed in an event loop which we don't want. It would be a disaster to block an event loop.

An idea from @kawamuray 
> 1. having an executor in CD, using it by default in the existing method signature helps users by: 
    1. avoid pitfall to call subsequent blocking call and 
    2. enabled to use blocking operation in watcher callback w/o needing to introduce manual executor or similar, or alternatively,
> 2. Changing the method signature to require users explicitly tell whether they want the direct execution (by IO thread) or bypassed execution (through executor) helps them to avoid pitfall anyway

We need to create a sentible default executor for Central Dogma client and provide both way to run the callback in an event loop or a blocking task excutor.

ikhoon:
Motivation:

The callacks for watcher are synchronous APIs such as `Function<T, U>`.
Users can not compose an asynchronous API for the callback function.
They have to wait for the result and return it.
```java
dogma.repositoryWatcher("xyz", "foo", "/*.txt",
                        revision -> {
                            dogma.getFiles("xyz", "foo",
                                           revision,
                                           "/*.txt").join();
                            return "Whatever";
                        });
```

CentralDomga internally invokes the callback by an event loop which must not be blocked.
However, most users do not aware that the callback is called by an event loop.
It would be safe to call the callbacks of watcher with a blocking task executor.

Modifications:

- Replace `EventLoop` executor with a blocking task executor.

Result:

- You can now safely execute a block IO for the callbacks of watcher.
- Fixes #587

### DIFF
--- a/client/java-armeria-legacy/src/main/java/com/linecorp/centraldogma/client/armeria/legacy/LegacyCentralDogmaBuilder.java
+++ b/client/java-armeria-legacy/src/main/java/com/linecorp/centraldogma/client/armeria/legacy/LegacyCentralDogmaBuilder.java
@@ -16,6 +16,7 @@
 package com.linecorp.centraldogma.client.armeria.legacy;
 
 import java.net.UnknownHostException;
+import java.util.concurrent.ScheduledExecutorService;
 
 import com.linecorp.armeria.client.ClientBuilder;
 import com.linecorp.armeria.client.ClientRequestContext;
@@ -28,8 +29,6 @@ import com.linecorp.centraldogma.client.armeria.AbstractArmeriaCentralDogmaBuild
 import com.linecorp.centraldogma.client.armeria.ArmeriaCentralDogmaBuilder;
 import com.linecorp.centraldogma.internal.client.ReplicationLagTolerantCentralDogma;
 import com.linecorp.centraldogma.internal.thrift.CentralDogmaService.AsyncIface;
-
-import io.netty.channel.EventLoopGroup;
 
 /**
  * Builds a legacy {@link CentralDogma} client based on Thrift.
@@ -65,14 +64,17 @@ public class LegacyCentralDogmaBuilder extends AbstractArmeriaCentralDogmaBuilde
             return delegate.execute(ctx, req);
         });
 
-        final EventLoopGroup executor = clientFactory().eventLoopGroup();
+        final ScheduledExecutorService blockingTaskExecutor = blockingTaskExecutor();
+
         final int maxRetriesOnReplicationLag = maxNumRetriesOnReplicationLag();
-        final CentralDogma dogma = new LegacyCentralDogma(executor, builder.build(AsyncIface.class));
+        final CentralDogma dogma = new LegacyCentralDogma(blockingTaskExecutor,
+                                                          builder.build(AsyncIface.class));
         if (maxRetriesOnReplicationLag <= 0) {
             return dogma;
         } else {
             return new ReplicationLagTolerantCentralDogma(
-                    executor, dogma, maxRetriesOnReplicationLag, retryIntervalOnReplicationLagMillis(),
+                    blockingTaskExecutor(), dogma, maxRetriesOnReplicationLag,
+                    retryIntervalOnReplicationLagMillis(),
                     () -> {
                         // FIXME(trustin): Note that this will always return `null` due to a known limitation
                         //                 in Armeria: https://github.com/line/armeria/issues/760
--- a/client/java-armeria-legacy/src/main/java/com/linecorp/centraldogma/client/armeria/legacy/LegacyCentralDogmaBuilder.java
+++ b/client/java-armeria-legacy/src/main/java/com/linecorp/centraldogma/client/armeria/legacy/LegacyCentralDogmaBuilder.java
@@ -73,7 +73,7 @@ public class LegacyCentralDogmaBuilder extends AbstractArmeriaCentralDogmaBuilde
             return dogma;
         } else {
             return new ReplicationLagTolerantCentralDogma(
-                    blockingTaskExecutor(), dogma, maxRetriesOnReplicationLag,
+                    blockingTaskExecutor, dogma, maxRetriesOnReplicationLag,
                     retryIntervalOnReplicationLagMillis(),
                     () -> {
                         // FIXME(trustin): Note that this will always return `null` due to a known limitation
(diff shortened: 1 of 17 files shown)


trustin (approved):
Thanks, @ikhoon :+1:

client/java-armeria-legacy/src/main/java/com/linecorp/centraldogma/client/armeria/legacy/LegacyCentralDogmaBuilder.java
```diff
@@ -65,14 +64,17 @@ public CentralDogma build() throws UnknownHostException {
...
         final int maxRetriesOnReplicationLag = maxNumRetriesOnReplicationLag();
-        final CentralDogma dogma = new LegacyCentralDogma(executor, builder.build(AsyncIface.class));
+        final CentralDogma dogma = new LegacyCentralDogma(blockingTaskExecutor,
+                                                          builder.build(AsyncIface.class));
         if (maxRetriesOnReplicationLag <= 0) {
             return dogma;
         } else {
             return new ReplicationLagTolerantCentralDogma(
-                    executor, dogma, maxRetriesOnReplicationLag, retryIntervalOnReplicationLagMillis(),
+                    blockingTaskExecutor(), dogma, maxRetriesOnReplicationLag,
```
minwoox, line 76: nit:
```suggestion
                    blockingTaskExecutor, dogma, maxRetriesOnReplicationLag,
```

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java
```diff
@@ -540,10 +562,32 @@
...
      */
     <T> Watcher<T> repositoryWatcher(String projectName, String repositoryName, String pathPattern,
                                      Function<Revision, ? extends T> function);
+
+    /**
+     * Returns a {@link Watcher} which notifies its listeners when the specified repository has a new commit
+     * that contains the changes for the files matched by the given {@code pathPattern}. e.g:
+     * <pre>{@code
+     * Watcher<Map<String, Entry<?>> watcher = client.repositoryWatcher(
+     *         "foo", "bar", "/*.json", revision -> client.getFiles("foo", "bar", revision, "/*.json").get());
```
minwoox, line 580: nit: 
```suggestion
     *         "foo", "bar", "/*.json", revision -> client.getFiles("foo", "bar", revision, "/*.json").join(), executor);
```
We need to use `join` if we don't want to handle `InterruptedException`.
Could you also fix the other examples, please? 😄

client/java/src/main/java/com/linecorp/centraldogma/internal/client/RepositoryWatcher.java
```diff
@@ -28,27 +29,30 @@
...
     private final String pathPattern;
     private final Function<Revision, ? extends T> function;
+    private final Executor callbackExecutor;
 
     /**
      * Creates a new instance.
      */
-    public RepositoryWatcher(CentralDogma client, ScheduledExecutorService executor,
+    public RepositoryWatcher(CentralDogma client, ScheduledExecutorService watchScheduler,
+                      Executor callbackExecutor,
```
minwoox, line 38: nit: indentation?

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java
```diff
@@ -512,10 +513,31 @@
...
     <T, U> Watcher<U> fileWatcher(String projectName, String repositoryName,
                                   Query<T> query, Function<? super T, ? extends U> function);
 
+    /**
+     * Returns a {@link Watcher} which notifies its listeners after applying the specified
+     * {@link Function} when the result of the given {@link Query} becomes available or changes. e.g:
+     * <pre>{@code
+     * Watcher<MyType> watcher = client.fileWatcher(
+     *         "foo", "bar", Query.ofJson("/baz.json"),
+     *         content -> new ObjectMapper().treeToValue(content, MyType.class));
```
minwoox, line 529:
```suggestion
     *         content -> new ObjectMapper().treeToValue(content, MyType.class), executor);
```

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java
```diff
@@ -512,10 +513,31 @@
...
+     *
+     * watcher.watch((revision, myValue) -> {
+     *     assert myValue instanceof MyType;
+     *     ...
+     * });}</pre>
+     *
+     * @param executor the {@link Executor} that executes the {@link Function}
+     */
+    <T, U> Watcher<U> fileWatcher(String projectName, String repositoryName,
+                                  Query<T> query, Function<? super T, ? extends U> function, Executor executor);
```
minwoox, line 539: I'm wondering if we need to provide this method that takes an additional `executor` because the client is already created with a blocking task executor. 🤔

minwoox (commented):
Thanks! Left some minor comments. 🙇‍♂️

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java
ikhoon: I thought, sometimes, users want to specify different executors depending on the query they use.

minwoox: That makes sense. 👍

ikhoon:
Thanks for reviewing!

PR #591, opened 2021-04-23
Replace some usages of TemporaryFolderExtension

KarboniteKream:
#### Motivation
- Try to use `@TempDir` in some tests to see if file deletion logic on Windows has been improved

#### Modifications
- Replace some usages of `TemporaryFolderExtension` with JUnit's `@TempDir`

### DIFF
--- a/server/src/test/java/com/linecorp/centraldogma/server/internal/api/auth/PermissionTest.java
+++ b/server/src/test/java/com/linecorp/centraldogma/server/internal/api/auth/PermissionTest.java
@@ -27,6 +27,7 @@ import java.util.concurrent.ForkJoinPool;
 import java.util.function.Function;
 import java.util.stream.Stream;
 
+import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.Order;
 import org.junit.jupiter.api.extension.RegisterExtension;
 import org.junit.jupiter.api.io.TempDir;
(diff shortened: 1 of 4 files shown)


KarboniteKream:
Remaining usages of `TemporaryFolderExtension`:

- `PermissionTest` can be made to work locally on Windows, but fails on CI
- `GitMirrorTest` has an issue with `remoteToLocal_submodule()`. Even after closing all repositories, JGit states there's 0 references, but the `.pack` file referencing the submodule cannot be deleted
- `ReplicationWriteQuotaTest` and `ReplicatedLoginAndLogoutTest` have a similar problem. Looks like some file locks remain, even after shutting down replicas.

Looks like the timing of `@TempDir` is different to the extension I implemented. I don't completely understand why it works better. Looks like the tests have much more time to cleanup and release file locks before the extension cleanup runs.

I was hoping to be able to completely remove `TemporaryFolderExtension`, but no such luck 😭

trustin (approved):
Thanks a lot! What stopped you from replacing 'all' usages?

KarboniteKream:
> What stopped you from replacing 'all' usages?

I explained it in [this comment](https://github.com/line/centraldogma/pull/591#issuecomment-826299414). There seems to be some timing/file lock issues on Windows, which I couldn't completely figure out. Even after stopping the services that seems to keep file locks, `@TempDir` seems to be unable to delete the files.

Please let me know if you have any ideas 🙇

ikhoon (approved):
Thanks! @KarboniteKream

minwoox (approved):
>  but no such luck 😭

😭 
Thanks a lot for your effort on this. 😄  I think we can keep using `TemporaryFolderExtension` for the rest of them.

ikhoon:
Thanks again @KarboniteKream

Issue #601, opened 2021-05-27
Consider renaming the watch API


minwoox:
Currently, get file(s) API and watch API have the same endpoint.
https://github.com/line/centraldogma/blob/master/server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java#L231
So access logs and metrics have the same naming.
That really makes it hard to debug when something happens because the watch API usually has high latency.

PR #602, opened 2021-05-27
Change the service name of watch API calls

minwoox:
Motivation:
Related #601
The Watch API and the get files API are served by the same method in `ContentServiceV1` which means they have the same metric name.
That really makes it hard to debug when something happens because the watch API usually has high latency.

Modifications:
- Change the service name of watch API calls to `WatchContentServiceV1` so that we can easily distinguish.

Result:
- You can now see the different metrics for the watch API and get files API.


ikhoon (approved):
Cool!

minwoox:
Thanks for reviewing. 😉

PR #598, opened 2021-05-19
Update dependencies

ikhoon:
- Armeria 1.7.2 -> 1.8.0
- Caffeine 2.9.0 -> 2.9.1
- jcommander 1.78 -> 1.81
- JGit 5.11.0 -> 5.11.1
- Micrometer 1.6.6 -> 1.7.0
- Mocktio 3.9.0 -> 3.10.0
- Build
  - awaitility 4.0.3 -> 4.1.0
  - gradle-docker-plugin 6.7.0 -> 7.0.0
  - jmh-gradle-plugin 0.5.3 -> 0.6.4
  - jmh 1.29 -> 1.31
  - JUnit 5.7.1 -> 5.7.2
  - NodeJS 10.19.0 -> 16.1.0
  - Yarn 1.17.3 -> 1.22.10

PR #604, opened 2021-06-01
Downgrade JMH to 0.5.3

minwoox:
Motivation:
me.champeau.jmh 0.6.X is compatible with Gradle 6.8+
However, we have an issue with the Gradle 6.8+
https://discuss.gradle.org/t/unable-to-exclude-task-after-upgrading-from-6-6-1-to-6-8-or-later/40098
So let me just revert it temporarily before I solve the problem.

Modification:
- Downgrade JMH to 0.5.3

Result:
- N/A

### DIFF
--- a/dependencies.yml
+++ b/dependencies.yml
@@ -161,8 +161,8 @@ org.junit.jupiter:
 kr.motd.gradle:
   sphinx-gradle-plugin: { version: '2.10.1' }
 
-me.champeau.jmh:
-  jmh-gradle-plugin: { version: '0.6.4' }
+me.champeau.gradle:
+  jmh-gradle-plugin: { version: '0.5.3' }
 
 net.javacrumbs.json-unit:
   json-unit: { version: &JSON_UNIT_VERSION '2.25.0' }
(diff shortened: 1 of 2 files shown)


dependencies.yml
```diff
@@ -161,8 +161,8 @@ org.junit.jupiter:
 kr.motd.gradle:
   sphinx-gradle-plugin: { version: '2.10.1' }
 
-me.champeau.jmh:
-  jmh-gradle-plugin: { version: '0.6.4' }
+me.champeau.gradle:
+  jmh-gradle-plugin: { version: '0.5.3' }
```
ikhoon, line 165: How about rollbacking to `0.5.0`?
https://github.com/line/armeria/pull/3561/files#diff-1d8ca348d900d958b9aa55ebd96110af6deb206336d2f7c9778b7d8bcd60f99bL334

ikhoon (approved):
Oops... 😱

dependencies.yml
minwoox: https://github.com/line/centraldogma/pull/598/files#diff-1d8ca348d900d958b9aa55ebd96110af6deb206336d2f7c9778b7d8bcd60f99bL164
It was 0.5.3 and it works well so let me just use the version. 😉

ikhoon: OK, go ahead.

PR #607, opened 2021-06-04
Add a path from a project page to a metadata page on UI

hokkun-dayo:
It seems nice in practice to me if there is a path from a project page (`#/projects/xxx`) to a metadata page (`#/metadata/xxx`) on the UI.

Here is a screenshot:
Screen Shot 2021-06-04 at 15 45 49

### DIFF
--- a/server/src/main/resources/webapp/i18n/en.main.json
+++ b/server/src/main/resources/webapp/i18n/en.main.json
@@ -40,6 +40,7 @@
     "button_delete": "Delete",
     "button_edit": "Edit",
     "button_fetch_again": "Fetch again",
+    "button_go_to_metadata": "Go to metadata",
     "button_history": "History",
     "button_query": "Query",
     "button_restore": "Restore",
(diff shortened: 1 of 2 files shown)


ikhoon (approved):
Looking good! Thanks, @hokkun-dayo.

trustin:
Should we move the button down so all buttons are grouped together?

minwoox:
Can we also remove the `meta` repository from the list? 🤔

hokkun-dayo:
@trustin 
> Should we move the button down so all buttons are grouped together? Or perhaps we could move the Create a repository button up. 😄

Sure, I'll try it.

hokkun-dayo:
@minwoox 
IIUC, contents in `meta` directory and what we can see in `#/metadata/xxx` are different, so it seems that we don't have to remove that. Any thoughts?

hokkun-dayo:
Modified one
Screen Shot 2021-06-10 at 18 20 14

minwoox:
> @minwoox
IIUC, contents in meta directory and what we can see in #/metadata/xxx are different, so it seems that we don't have to remove that. Any thoughts?

Yes, exactly. Sorry about the comment. 😅

minwoox (approved):

minwoox:
Thanks @hokkun-dayo!

Issue #532, opened 2020-09-07
FileWatcher should throw NoSuchFileException?


minwoox:
Currently, `FileWatcher.awaitInitialValue(...);` throws `TimeoutException` rather than `NoSuchFileException` if the file, which the watcher tries to watch, does not exist.
It makes the user hard to find the cause.

trustin:
I think there should be an option for this behavior, because some users might want to wait until the file is created. I agree that the default behavior should be to throw an exception.

din-seo:
I try to handle this issue.

minwoox:
Closed via #653

PR #610, opened 2021-06-11
Provide a way to get notified if entries are not found when watching

minwoox:
Motivation:
Related #532
Currently, when we try to watch a file that doesn't exist,
the watch API waits until the file is added or timeout happens.
It sometimes makes user difficult to debug because the client cannot distinguish such cases.

Modifications:
- Add `notify-entry-not-found` preference to the watch `Prefer` header.
  - https://datatracker.ietf.org/doc/html/rfc7240#section-4

Result:
- You can now get notified if the entry doesn't exist when watching.

To-do:
- Add the option to client-side


ikhoon (approved):
Thanks! @minwoox

minwoox:
Thanks for reviewing. 😉

PR #606, opened 2021-06-02
Add missing package-info.java

minwoox:
Motivation:
So that IDE does not complain about nullabilty anymore.

Modifications:
- Add missing package-info.java files
- Add `ElementType.TYPE_USE` to `NonNullByDefault`
  - Related https://github.com/line/armeria/pull/3558

Result:
- Cleanup

PR #611, opened 2021-06-16
Fails when package-info.java is missing in internal packages

minwoox:
Motivation:
Related https://github.com/line/centraldogma/pull/606#issuecomment-855197216
We suppress the check style error when we miss package-info.java in internal packages.
However, we need the package-info.java files even in internal packages because we add `@NonNullByDefault` using it.

Modifications:
- Remove the suppression rule for `JavadocPackage`.

Result:
- No more warning from IntelliJ about nullability.


trustin (approved):
Thanks, @minwoox !

Issue #151, opened 2018-03-12
Better YAML support


trustin:
We should provide support for YAML that's as good as that of JSON:

- Retrieving YAML as JSON
- Querying YAML
- Including YAML from YAML

.. so that we can say 'Central Dogma has first-class support for JSON *and* YAML.'

Dogacel:
It seems like there is already a PR #616 but it is very old. I wonder if there is any initiative at the moment for this?

PR #616, opened 2021-07-06
Add better YAML supports

chacham:
### Motivation

#151 Need better YAML supports.
This is my first time trying to contribute to an open source project.
Please let me know if I'm doing something wrong!

### Modifications (In Progress)

- Upsert YAML
- Implement YAML Entry

### TO-DO

- Client-side supports
- Patching YAML
- Querying YAML
- Retrieving YAML as JSON
- Including YAML from YAML

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -63,6 +63,9 @@ configure(projectsWithFlags('java')) {
             implementation "com.fasterxml.jackson.core:jackson-$it"
         }
 
+        // SnakeYAML
+        implementation 'org.yaml:snakeyaml'
+
         // javax.inject
         api 'javax.inject:javax.inject'
 
--- a/common/src/main/java/com/linecorp/centraldogma/common/Change.java
+++ b/common/src/main/java/com/linecorp/centraldogma/common/Change.java
@@ -33,11 +33,14 @@ import java.util.stream.Stream;
 
 import javax.annotation.Nullable;
 
+import org.yaml.snakeyaml.nodes.Node;
+
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 
 import com.linecorp.centraldogma.internal.Jackson;
+import com.linecorp.centraldogma.internal.SnakeYaml;
 import com.linecorp.centraldogma.internal.Util;
 import com.linecorp.centraldogma.internal.jsonpatch.JsonPatch;
 import com.linecorp.centraldogma.internal.jsonpatch.ReplaceMode;
@@ -102,6 +105,32 @@ public interface Change<T> {
     static Change<JsonNode> ofJsonUpsert(String path, JsonNode jsonNode) {
         requireNonNull(jsonNode, "jsonNode");
         return new DefaultChange<>(path, ChangeType.UPSERT_JSON, jsonNode);
+    }
+
+    /**
+     * Returns a newly-created {@link Change} whose type is {@link ChangeType#UPSERT_YAML}.
+     *
+     * @param path the path of the file
+     * @param yamlText the content of the file
+     */
+    static Change<Node> ofYamlUpsert(String path, String yamlText) {
+        requireNonNull(yamlText, "yamlText");
+
+        final Node yamlNode;
+        yamlNode = SnakeYaml.readTree(yamlText);
+
+        return new DefaultChange<>(path, ChangeType.UPSERT_YAML, yamlNode);
+    }
+
+    /**
+     * Returns a newly-created {@link Change} whose type is {@link ChangeType#UPSERT_YAML}.
+     *
+     * @param path the path of the file
+     * @param yamlNode the content of the file
+     */
+    static Change<Node> ofYamlUpsert(String path, Node yamlNode) {
+        requireNonNull(yamlNode, "yamlNode");
+        return new DefaultChange<>(path, ChangeType.UPSERT_YAML, yamlNode);
     }
 
     /**
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
@@ -97,6 +97,7 @@ import org.eclipse.jgit.treewalk.filter.AndTreeFilter;
 import org.eclipse.jgit.treewalk.filter.TreeFilter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.yaml.snakeyaml.nodes.Node;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.JsonNodeFactory;
@@ -119,6 +120,7 @@ import com.linecorp.centraldogma.common.Revision;
 import com.linecorp.centraldogma.common.RevisionNotFoundException;
 import com.linecorp.centraldogma.common.RevisionRange;
 import com.linecorp.centraldogma.internal.Jackson;
+import com.linecorp.centraldogma.internal.SnakeYaml;
 import com.linecorp.centraldogma.internal.Util;
 import com.linecorp.centraldogma.internal.jsonpatch.JsonPatch;
 import com.linecorp.centraldogma.internal.jsonpatch.ReplaceMode;
@@ -565,6 +567,10 @@ class GitRepository implements Repository {
                             final JsonNode jsonNode = Jackson.readTree(content);
                             entry = Entry.ofJson(normRevision, path, jsonNode);
                             break;
+                        case YAML:
+                            final Node yamlNode = SnakeYaml.readTree(content);
+                            entry = Entry.ofYaml(normRevision, path, yamlNode);
+                            break;
                         case TEXT:
                             final String strVal = sanitizeText(new String(content, UTF_8));
                             entry = Entry.ofText(normRevision, path, strVal);
@@ -576,6 +582,9 @@ class GitRepository implements Repository {
                     switch (entryType) {
                         case JSON:
                             entry = Entry.ofJson(normRevision, path, Jackson.nullNode);
+                            break;
+                        case YAML:
+                            entry = Entry.ofYaml(normRevision, path, SnakeYaml.nullNode);
                             break;
                         case TEXT:
                             entry = Entry.ofText(normRevision, path, "");
@@ -867,6 +876,13 @@ class GitRepository implements Repository {
                                 putChange(changeMap, newPath, Change.ofJsonUpsert(newPath, jsonNode));
                                 break;
                             }
+                            case YAML: {
+                                final Node yamlNode = SnakeYaml.readTree(
+                                        reader.open(diffEntry.getNewId().toObjectId()).getBytes());
+
+                                putChange(changeMap, newPath, Change.ofYamlUpsert(newPath, yamlNode));
+                                break;
+                            }
                             case TEXT: {
                                 final String text = sanitizeText(new String(
                                         reader.open(diffEntry.getNewId().toObjectId()).getBytes(), UTF_8));
@@ -1081,6 +1097,16 @@ class GitRepository implements Repository {
                         // Upsert only when the contents are really different.
                         if (!Objects.equals(newJsonNode, oldJsonNode)) {
                             applyPathEdit(dirCache, new InsertJson(changePath, inserter, newJsonNode));
+                            numEdits++;
+                        }
+                        break;
+                    }
+                    case UPSERT_YAML: {
+                        final Node oldYamlNode = oldContent != null ? SnakeYaml.readTree(oldContent) : null;
+                        final Node newYamlNode = firstNonNull((Node) change.content(), null);
+
+                        if (!Objects.equals(oldYamlNode, newYamlNode)) {
+                            applyPathEdit(dirCache, new InsertYaml(changePath, inserter, newYamlNode));
                             numEdits++;
                         }
                         break;
@@ -1805,6 +1831,27 @@ class GitRepository implements Repository {
                 ent.setFileMode(FileMode.REGULAR_FILE);
             } catch (IOException e) {
                 throw new StorageException("failed to create a new JSON blob", e);
+            }
+        }
+    }
+
+    private static final class InsertYaml extends PathEdit {
+        private final ObjectInserter inserter;
+        private final Node yamlNode;
+
+        InsertYaml(String entryPath, ObjectInserter inserter, Node yamlNode) {
+            super(entryPath);
+            this.inserter = inserter;
+            this.yamlNode = yamlNode;
+        }
+
+        @Override
+        public void apply(DirCacheEntry ent) {
+            try {
+                ent.setObjectId(inserter.insert(Constants.OBJ_BLOB, SnakeYaml.serialize(yamlNode).getBytes()));
+                ent.setFileMode(FileMode.REGULAR_FILE);
+            } catch (IOException e) {
+                throw new StorageException("failed to create a new YAML blob", e);
             }
         }
     }
--- a/client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogma.java
+++ b/client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogma.java
@@ -1072,6 +1072,8 @@ final class ArmeriaCentralDogma extends AbstractCentralDogma {
                 switch (receivedEntryType) {
                     case JSON:
                         return entryAsJson(revision, node, entryPath);
+                    case YAML:
+                        return entryAsYaml(revision, node, entryPath);
                     case TEXT:
                         return entryAsText(revision, node, entryPath);
                     case DIRECTORY:
@@ -1094,6 +1096,17 @@ final class ArmeriaCentralDogma extends AbstractCentralDogma {
 
     private static <T> Entry<T> entryAsJson(Revision revision, JsonNode node, String entryPath) {
         return unsafeCast(Entry.ofJson(revision, entryPath, getField(node, "content")));
+    }
+
+    private static <T> Entry<T> entryAsYaml(Revision revision, JsonNode node, String entryPath) {
+        final JsonNode content = getField(node, "content");
+        final String content0;
+        if (content.isContainerNode()) {
+            content0 = content.toString();
+        } else {
+            content0 = content.asText();
+        }
+        return unsafeCast(Entry.ofYaml(revision, entryPath, content0));
     }
 
     private static Commit toCommit(JsonNode node) {
--- a/common/src/main/java/com/linecorp/centraldogma/common/ChangeType.java
+++ b/common/src/main/java/com/linecorp/centraldogma/common/ChangeType.java
@@ -19,6 +19,8 @@ package com.linecorp.centraldogma.common;
 import static com.google.common.base.Preconditions.checkArgument;
 import static java.util.Objects.requireNonNull;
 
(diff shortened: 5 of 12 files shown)


minwoox:
This looks awesome! Thanks for creating this PR.
Let us take a look at it soon. 😉

build.gradle
```diff
@@ -63,6 +63,9 @@ configure(projectsWithFlags('java')) {
             implementation "com.fasterxml.jackson.core:jackson-$it"
         }
 
+        // SnakeYAML
+        implementation 'org.yaml:snakeyaml'
```
minwoox, line 67: Have you ever thought about using Jackson instead of SnakeYAML?
https://www.baeldung.com/jackson-yaml
https://stackabuse.com/reading-and-writing-yaml-files-in-java-with-snakeyaml
https://stackabuse.com/reading-and-writing-yaml-files-in-java-with-jackson

I think we need to investigate the advantages and disadvantages before we decide to use SnakeYAML. 🤔

chacham: Oh, I didn't know about jackson-yaml. I'll see. Thank you!!

chacham: It seems that jackson-dataformat-yaml is also using SnakeYAML internally.
It is a bit confusing that `JsonNode` is returned even when parsing YAML, but it seems convenient to develop by referring to `Jackson.java` by using the `ObjectMapper` interface as it is.
For now, I'll proceed with jackson-dataformat-yaml.

minwoox: @chacham Sorry for the late reply.
I didn't know that Jackson uses SnakeYAML internally. 😅 

> It is a little bit confusing that JsonNode

If so and Jackson uses SnakeYAML anyway, I think it's okay to use SnakeYAML directly unless Jackson provides better usability. 😅 
@ikhoon Do you have any idea about this?

client/java-armeria-legacy/src/main/java/com/linecorp/centraldogma/client/armeria/legacy/LegacyCentralDogma.java
```diff
@@ -258,6 +258,12 @@
                                                     " (expected: " + queryType + ')');
                 }
                 return entryAsJson(query, normRev, content);
+            case IDENTITY_YAML:
+                if (receivedEntryType != com.linecorp.centraldogma.internal.thrift.EntryType.YAML) {
+                    throw new CentralDogmaException("invalid entry type. entry type: " + receivedEntryType +
```
minwoox, line 263: How about extracting this duplicate logic?

minwoox (commented):
Looks good so far. 😉 
Let us know when this is ready. 😄

chacham:
@minwoox, I read your previous comment and continued with using SnakeYAML directly.
> If so and Jackson uses SnakeYAML anyway, I think it's okay to use SnakeYAML directly unless Jackson provides better usability. 😅

But, as I continue to work, using SnakeYAML's `Node` seems to have several limitations.
- Unlike Jackson's `JsonNode`, even if a SnakeYAML `Node` has the same value (content), it becomes equal only when it is the same `Node` -> It is difficult to compare YAML content, and need to change implementations a lot such as `Entry`.
- Although it is not a problem specific to SnakeYAML, there is no standardized patch/query syntax. So, I think using jsonpatch/jsonqurey is good for now. (And it would be easy with jackson-dataformat-yaml)

So, I'll modify it to use Jackson instead of using SnakeYAML directly.
If you have any opinion, please tell me. 😅

minwoox:
> So, I'll modify it to use Jackson instead of using SnakeYAML directly.
If you have any opinion, please tell me.

I haven't known about the limitations. 😅
Yeah, jackson-dataformat-yaml seems a good choice for now. 😄

chacham:
Since jackson-dataformat-yaml also uses the Jackson `JsonNode` class, there is quite a lot of duplicate code. For now, I'll focus on implementation, and I'll organize it later. 😅
In `ContentHolder.contentAsText()`, if `content instanceof JsonNode`, format content as JSON. (Even if the content is YAML actually) I am thinking about how to proceed.

build.gradle
ikhoon: I also prefer using `jackson-dataformat-yaml`. If we directly use SnakeYAML, we eventually convert the parsed node from SnakeYAML into JsonNode for reusing other code.

minwoox:
> In ContentHolder.contentAsText(), if content instanceof JsonNode, format content as JSON. (Even if the content is YAML actually) I am thinking about how to handle it.

I think we can check the `EntryType` in `contentAsText()` so that we can produce the YAML text properly.

chacham:
> I think we can check if the `EntryType` is YAML in `contentAsText()` so that we can produce the YAML text properly. (We probably need to refactor `Entry` a bit.)

The problem is `Query.ofText`. With `Query.ofText`, the `EntryType` will set as `TEXT`, so it can't be checked now :(

minwoox:
> The problem is Query.ofText. With Query.ofText, the EntryType will set as TEXT, so it can't be checked now :(

IIRC, the `EntryType` is decided by the file extension. https://github.com/line/centraldogma/blob/master/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java#L494
Then, the fetched `Entry` (with the `EntryType`) is transferred to the client and client converts the content using the query type. https://github.com/line/centraldogma/blob/master/client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogma.java#L463-L464
However, at this point, we know if the `Entry` is pure text or YAML.
Did I miss something? 🤔

chacham:
> IIRC, the `EntryType` is decided by the file extension. https://github.com/line/centraldogma/blob/master/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java#L494
> Then, the fetched `Entry` (with the `EntryType`) is transferred to the client and client converts the content using the query type. https://github.com/line/centraldogma/blob/master/client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogma.java#L463-L464
> At this point, we know if the `Entry` is pure text or YAML so that we can convert the content as text properly.
> Did I miss something? 🤔

Oh, I'll check your comment and see it again. Thank you!
I saw the assertion fails with JSON format content at https://github.com/line/centraldogma/pull/616/files#diff-32bfb9d1fc9f4d4388019340cc31985f5ac3e41dd32aa11283cd78805398ba57R74-R76, so I thought as I wrote.
I'll have to check again after work! 😄

minwoox:
> I saw the assertion fails with JSON format content at #616 (files), so I thought as I wrote. Maybe the code I wrote is doing something wrong.

That's probably because we didn't modify the client? 😄 
https://github.com/line/centraldogma/blob/master/client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogma.java#L463-L464

ikhoon:
Do we need to add new features for this legacy(Thrift)? 🤔
How about only support `ArmeriaCentralDogma`?

common/src/main/java/com/linecorp/centraldogma/common/Change.java
```diff
@@ -256,6 +287,63 @@
...
+     *                               not a valid YAML
+     */
+    static Change<JsonNode> ofYamlPatch(String path, @Nullable String oldYamlText, String newYamlText) {
+        requireNonNull(newYamlText, "newYamlText");
+
+        final JsonNode oldYamlNode;
+        final JsonNode newYamlNode;
+        try {
+            oldYamlNode = oldYamlText == null ? JacksonYaml.nullNode
+                                              : JacksonYaml.readTree(oldYamlText);
```
ikhoon, line 307: How about making `JacksonYaml.readTree()` return `JacksonYaml.nullNode` if `oldYamlText` is null?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
```diff
@@ -817,6 +825,7 @@ private static TreeFilter pathPatternFilterOrTreeFilter(@Nullable String pathPat
                 switch (diffEntry.getChangeType()) {
                     case MODIFY:
                         final EntryType oldEntryType = EntryType.guessFromPath(oldPath);
+                        final JsonPatch patch;
```
ikhoon, line 828: If `patch` is not used outside of `case`, isn't it better to revert and minimize its scope?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
```diff
@@ -829,14 +838,31 @@ private static TreeFilter pathPatternFilterOrTreeFilter(@Nullable String pathPat
...
+                                if (!oldPath.equals(newPath)) {
+                                    putChange(changeMap, oldPath, Change.ofRename(oldPath, newPath));
+                                }
+
+                                final JsonNode oldYamlNode =
+                                        JacksonYaml.readTree(
+                                                reader.open(diffEntry.getOldId().toObjectId()).getBytes());
+                                final JsonNode newYamlNode =
+                                        JacksonYaml.readTree(
+                                                reader.open(diffEntry.getNewId().toObjectId()).getBytes());
```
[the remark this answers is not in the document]
ikhoon: On second thought, it would be better to make `Jackson` interface and use like:
```java
Jackson.ofJson().readTree(...)
Jackson.ofYaml().readTree(...)
// dynamically get the implementation 
Jackson.of(YAML).readTree(...)
```

common/src/main/java/com/linecorp/centraldogma/common/Change.java
chacham: Oh, I copied it from `ofJsonPatch`. I'll apply it after modify `Jackson.readTree()` receive `EntryType`.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
chacham: I'll revert it when when making `Jackson.readTree()` receive `EntryType`.

chacham:
Because the `Jackson` is used for both `Entry` content and metadata/request/response, some methods are duplicated and confusing.
The methods for the content of `Entry` has been modified to receive `EntryType` as a parameter, but there are still methods without `EntryType`(for metadata/request/response), so it looks easy to make mistake. (maybe I did already)
Would it better to separate Jackson for Content and Jackson for Metadata? Could somebody help me? 🤔

minwoox:
> but there are still methods without EntryType(for metadata/request/response), so it looks easy to make mistake. (maybe I did already)

Could you give us a specific example of the case, please?

chacham:
> Could you give us a specific example of the case, please?

- `Jackson.readTree()` for parsing git revision from full commit messgae
  - https://github.com/line/centraldogma/blob/master/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/CommitUtil.java#L60
- `Jackson.writeValueAsBytes()` for write session for file based management
  - https://github.com/line/centraldogma/blob/master/server/src/main/java/com/linecorp/centraldogma/server/internal/admin/auth/FileBasedSessionManager.java#L185
- `Jackson.writeValueAsPrettyString()` for data objects `toString()` (not related to Entry)
  - `AuthConfig`, `CentralDogmaConfig`, `ClientProfile`, `SamlAuthProperties`

Especially, `Jackson.writeValueAsBytes()` is used frequently for HTTP/File/ZooKeeper (where entries are not related)
There are two implementations, one for with `EntryType`, and one for without `EntryType`

ikhoon:
> Would it better to separate Jackson for Content and Jackson for Metadata?

`Jackson` should be not strongly connected to any domain. As you know it is a thin wrapper of `ObjectMapper`.
What do you think about https://github.com/line/centraldogma/pull/616#discussion_r679253140 style?
You don't want to implement some methods that are not used in YAML, we can introduce an abstract class.
Additionally, some methods using `compactFactory`, `prettyFactory` move to `JacksonUtil`?

- An interface for factory methods and APIs.
```java
public interface Jackson {
   static ofJson() {
      return JacksonJson.INSTANCE;
   }

   static ofYaml() {
      return JacksonYaml.INSTANCE;
   }
   
   static of(EntryType type) {
       if (type == JSON) {
           return JacksonJson.INSTANCE;
       } else if (type == YAML) {
           ...
       }
   }

   T readValue(String data, Class<T> type);
   ...
} 
``` 
- An abstract class for marshaling and unmarshaling objects
```
abstract class AbstractJackson {
    AbstractJackson(ObjectMapper compactMapper, ObjectMapper prettyMapper) {
       ...
    }
    
    @Override
    public <T> T readValue(String data, Class<T> type) throws JsonParseException, JsonMappingException {
        try {
            return compactMapper.readValue(data, type);
        } catch (JsonParseException | JsonMappingException e) {
            throw e;
        } catch (IOException e) {
            throw new IOError(e);
        }
    }

    // Copy all other methods from the old Jackson.java
    ...
}
```
- Create `JacksonJson` and `JacksonYaml`
```java
// Could an enum class
public final class JacksonJson extends AbstractJackson {

    private static final ObjectMapper compactMapper = new ObjectMapper();
    private static final ObjectMapper prettyMapper = new ObjectMapper();
    private static JacksonJson INSTANCE = new JacksonJson();
    static {
       ...
    }

    private JacksonJson() {
        super(compactMapper, prettyMapper);
    }
}

public final class JacksonYaml extends AbstractJackson {

    private static final YAMLMapper yamlMapper = new YAMLMapper();

    private static JacksonYaml INSTANCE = new JacksonYaml();
    static {
       ...
    }

    private JacksonJson() {
        // Perhaps, we don't support compack and pretty for YAML
        super(yamlMapper, yamlMapper);
    }
}
```

ikhoon:
> Do we need to add new features for the legacy Thrift client and service? 🤔
> How about only supporting `ArmeriaCentralDogma` and REST API?

Any though on this? @minwoox

chacham:
> > Do we need to add new features for the legacy Thrift client and service? 🤔
> > How about only supporting `ArmeriaCentralDogma` and REST API?
> 
> Any though on this? @minwoox

It seems there are several `ParameterizedTest`. Should the tests be separated if the new features do not support legacy client?

ikhoon:
> > > Do we need to add new features for the legacy Thrift client and service? 🤔
> > > How about only supporting `ArmeriaCentralDogma` and REST API?
> > 
> > 
> > Any though on this? @minwoox
> 
> It seems there are several `@ParameterizedTest`. Should the tests be separated if the new features do not support legacy client?

If it is more complex not to support legacy client, we can ignore the suggestion. I thought we don't put in a lot of work to legacy APIs.

chacham:
> > > > Do we need to add new features for the legacy Thrift client and service? 🤔
> > > > How about only supporting `ArmeriaCentralDogma` and REST API?
> > > 
> > > 
> > > Any though on this? @minwoox
> > 
> > 
> > It seems there are several `@ParameterizedTest`. Should the tests be separated if the new features do not support legacy client?
> 
> If it is more complex not to support legacy client, we can ignore the suggestion. I thought we don't put in a lot of work to legacy APIs.

I added tests for YAML (similar to the existing JSON/TEXT), and there doesn't seem to be big problems for now.
I'll ask for help if there's any difficulties. 😁

chacham:
I changed Jackson to interface and separated JSON/YAML implementation using abstract class. But because of the json path config and some methods, it looks a bit weird.
Now, there are many `Jackson.ofJson()`. I'm confused if I'm going in the right direction. 😵‍💫

minwoox:
> Now, there are many Jackson.ofJson(). I'm confused if I'm going in the right direction. 

I think it's okay! Please, go ahead! 😄

common/src/main/java/com/linecorp/centraldogma/common/MergeQuery.java
```diff
@@ -38,21 +38,47 @@
...
         return new JsonMergeQuery(IDENTITY, mergeSources, ImmutableList.of());
     }
 
+    /**
+     * Returns a newly-created {@link MergeQuery} that merges the JSON/YAML contents as specified in
+     * the {@code mergeSources}.
+     *
+     * @param mergeSources the paths of JSON files being merged and indicates whether it is optional
+     */
+    static MergeQuery<JsonNode> ofJsonOrYaml(MergeSource... mergeSources) {
```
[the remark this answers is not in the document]
chacham: At first, I tried to implement it as you said, but since there is a part that uses `ofJson` as default merge query in the client, I thought that a common MergeQuery would be better. (https://github.com/line/centraldogma/pull/616/files#diff-f08760fa027ab0e795025c0410578002d9b1b5313714f83c79940f2d1b174807L215)
Now, there is a problem that the `EntryType` of `MergedEntry` is all JSON, but merging JSON and YAML files also works. However, if you prefer `ofJson`/`ofYaml` over `ofJsonOrYaml`, could you help how to handle the default merge query in `mergeFiles()`?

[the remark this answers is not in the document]
ikhoon: If we support another file format such as TOML or INI, this style of public API might not be extensible.
I guess we need some breaking changes to type-safely merge multiple formats by taking an additional type parameter for `MergeSource`.
For example:
```java
class MergeSource<T> {
   static MergeSource<JsonNode> ofJson(String path) {
       ...
   }

   static MergeSource<JsonNode> ofJson(String path, boolean required) {
       ...
   }

   static MergeSource<JsonNode> ofYaml(String path) {
       ...
   }
}

interface CentralDogma {

   default CompletableFuture<MergedEntry<T>> mergeFiles(
            String projectName, String repositoryName,
            Revision revision, Iterable<MergeSource<T>> mergeSources) {
       ...
   }
}
```

It might be out of the scope of PR. How about only supporting `mergeFiles` for JSON at the moment so that we can consider YAML in the following PR as a different topic?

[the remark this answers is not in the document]
minwoox: > It might be out of the scope of PR. How about excluding YAML from supporting mergeFiles at the moment so that we can consider YAML in the following PR as a different topic?

+ 1 for this.

minwoox:
I guess this is ready so let me change the status this PR. 😉

minwoox:
Let me do the final review after  https://github.com/line/centraldogma/pull/616#discussion_r752182517 is addressed. 🙇‍♂️

chacham:
Old clients crashes when `EntryType.YAML` goes down to the client.
I tested `dogma.listFiles()` after making yaml file, and below is the log.
```
java.lang.IllegalArgumentException: No enum constant com.linecorp.centraldogma.common.EntryType.YAML
java.util.concurrent.ExecutionException: java.lang.IllegalArgumentException: No enum constant com.linecorp.centraldogma.common.EntryType.YAML
	at java.base/java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:395)
	at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:1999)
	at com.example.demo.CentralDogmaClientTest.listFiles(CentralDogmaClientTest.java:37)
```

minwoox:
> Old clients crashes when EntryType.YAML goes down to the client.
I tested dogma.listFiles() after making yaml file, and below is the log.

Had a chat with @ikhoon and I think it's time to go for `/api/v2/` https://github.com/line/centraldogma/blob/master/common/src/main/java/com/linecorp/centraldogma/internal/api/v1/HttpApiV1Constants.java#L27
which supports YAML. 
Also we can add JSON5 support with `/api/v2/`. /cc @ks-yim

I think there's a lot to do to use `/api/v2` so let's finish this PR by just implementing YAML.
I might implement `/api/v2/` based on your work. 😄

ikhoon:
@chacham Thanks for your hard work. 🙇‍♂️ YAML support is superseded by #1229 and I added you as a co-author as your work was referenced when implementing #1229.

PR #617, opened 2021-07-07
Use Git repository format version 1

minwoox:
Motivation:
Related: #83
We use Git repository format version 1 from 0.19.0.
I think we can force version 1 and throw an exception when CD tries to open a Git repository whose format version is 0.

Modifications:
- Throw an exception if CD tries to open a Git repository whose format version is 0.
- Remove the migration code that helps migrate from format version 0 to 1.

Result:
- Simplicity


ikhoon (approved):
👋👋👋

minwoox:
Thanks for reviewing. 😉

PR #615, opened 2021-07-06
Revert "Run git gc periodically (#564)"

minwoox:
Motivation:
We no longer need Git gc because we are going to provide a way to remove old commits.
https://github.com/line/centraldogma/issues/575

Modifications:
- Revert Git gc.

Result:
- This reverts commit e6caf26a91d6fea1fa2ae36e4e0472488d804013.

PR #619, opened 2021-07-16
Provide a way to remove old commits

### LINKED ISSUE #575: Provide a way to remove the old commits

minwoox:
Motivation:
Central Dogma uses jGit to store data. Due to the nature of Git that stores unlimited history,
Central Dogma will eventually get in trouble managing disk usage.
We can handle this by maintaining the primary and secondary Git repositories internally.
This works in this way:
- Commits are pushed to the primary Git repository.
- If the number of commits exceeds the threshold (`minRetentionCommits`), then the secondary Git repository is created.
- Commits are pushed to both primary and secondary Git repositories.
- If the secondary Git repository has the number of commits more than the threshold;
  - The secondary Git repository is promoted to the primary Git repository.
  - The primary Git repository is removed completely.
  - Another secondary Git repository is created.
- Back to the third.

Modifications:
- Add `CreateRollingRepositoryCommand` that creates the rolling repository by the `CommitRetentionManagementPlugin`.
- Add `GitRepositoryV2` that manages the rolling jGit repositories.
  - The name of internal repositories will be: `foo_0000000000`, `foo_0000000001`, `foo_0000000002` and so on
    - So that we can only store the suffix of the primary repo to maintain the repository. (i.e. We don't have to maintain the directory information of the secondary repo.)
    - `RepositoryMetadataDatabase` has the suffix in its file database.
  - `GitRepository` is not removed for the migration test. 
- Add `InternalRepository` that has jGit repository and `CommitIdDatabase`.
- What happens if the revision of an operation(such as `diff`, `watch`, `history`, etc.) is lower than the `firstRevision` of the current primary repo?
  - If `Revision.INIT(1)` is used, the `firstRevision` is used instead.
    - e.g. `diff(INIT, new Revision(100) ...)` is equals to `diff(new Revision(firstRevisionNumber), new Revision(100) ...)`
  - If the `Revision` between `Revision.INIT(1)` and the `firstRevision` is used, a `RevisionNotFoundException` is raised.
    - except `watch` and `findLatestRevision`.

Result:
- Close #575
- A repository now removes the commits that are older than `minRetentionCommits` while keeping the commits made in the recent `minRetentionDay`.

Todo:
- Provide a way to set `minRetentionCommits` and `minRetentionDay` for each repository.
- Unused internal repositories are removed by purging service.
- Support mirroring from CD to external Git.


minwoox:
I think this is ready. Please, review this PR when you have time. 😄

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepositoryV2.java
```diff
@@ -0,0 +1,850 @@
...
+                    }
+                    laggedCommitsForSecondary.clear();
+                }
+                isCreatingSecondaryRepo = false;
+                this.secondaryRepo = secondaryRepo;
+            } finally {
+                writeUnLock();
+            }
+            logger.info("The secondary repository {} is created in {}/{}.",
+                        secondaryRepoDir.getName(), parent.name(), originalRepoName, headRevision);
```
ikhoon, line 801:
```suggestion
            logger.info("The secondary repository {} is created in {}/{}. head revision: {}",
                        secondaryRepoDir.getName(), parent.name(), originalRepoName, headRevision);
```

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepositoryV2.java
```diff
@@ -0,0 +1,850 @@
...
+        } catch (Throwable t) {
+            logger.warn("Failed to create the secondary repository", t);
```
ikhoon, line 803: Don't we need to throw an exception and stop replication rather than catching it if an exception is raised while creating a secondary repository?
Additionally, we might need to delete the incomplete secondary repository just in case only half of `LaggedCommit`s is saved.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepositoryV2.java
```diff
@@ -0,0 +1,850 @@
...
+        if (secondaryRepo != null) {
+            if (exceedsMinRetention(secondaryRepo, minRetentionCommits, minRetentionDays)) {
+                promoteSecondaryRepo(secondaryRepo);
+            }
+        } else if (exceedsMinRetention(primaryRepo, minRetentionCommits, minRetentionDays)) {
+            createSecondaryRepo();
+        }
+    }
+
+    private boolean exceedsMinRetention(InternalRepository repo,
```
ikhoon, line 709: `static`?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepositoryV2.java
```diff
@@ -0,0 +1,850 @@
...
+            final InternalRepository primaryRepo = this.primaryRepo;
+            this.primaryRepo = secondaryRepo;
+            this.secondaryRepo = null;
+            repositoryWorker.execute(() -> {
+                closeInternalRepository(primaryRepo);
+                final File repoDir = primaryRepo.repoDir();
+                try {
+                    if (!deleteDirectory(repoDir)) {
+                        logger.warn("Failed to delete the old primary repository: {}", repoDir);
+                    }
```
ikhoon, line 742: How about 1) renaming the old repo with `repo.removed` and 2) letting the old repos be removed by `PurgeSchedulingService` after 7 days.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepositoryManager.java
```diff
@@ -69,7 +69,7 @@ public Project parent() {
 
     @Override
     protected Repository openChild(File childDir) throws Exception {
-        return new GitRepository(parent, childDir, repositoryWorker, cache);
+        return GitRepositoryV2.open(parent, childDir, repositoryWorker, cache);
     }
 
     private static void deleteCruft(File dir) throws IOException {
```
ikhoon, line 75: Could remove this and `MigrationProgressLogger`?

server/src/main/java/com/linecorp/centraldogma/server/CommitRetentionConfig.java
```diff
@@ -0,0 +1,98 @@
...
+
+import com.linecorp.centraldogma.server.storage.repository.Repository;
+
+/**
+ * A configuration for retaining commits in a {@link Repository}.
+ */
+public final class CommitRetentionConfig {
+
+    // The minimum of minRetentionCommits
+    private static final int MINIMUM_MINIMUM_RETENTION_COMMITS = 1000;
```
ikhoon, line 39: 1,000 seems too small as the default min value. How about 5,000? Some mega repos in LINE are larger than 250,000 commits but most repo are less than 2000 commits. I guess 5,000 would keep all histories of normal repos and will not trigger creating a secondary repo.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepositoryV2.java
```diff
@@ -0,0 +1,850 @@
...
+        if (minRetentionCommits != 0) {
+            if (headRevision.major() - firstRevision.major() <= minRetentionCommits) {
+                return false;
+            }
+        }
+
+        if (minRetentionDays != 0) {
+            final Instant secondCommitCreationTime = repo.secondCommitCreationTimeInstant();
+            return secondCommitCreationTime != null &&
+                   Instant.now().minus(Duration.ofDays(minRetentionDays)).isBefore(secondCommitCreationTime);
```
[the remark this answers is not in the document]
minwoox: OOps. 😓

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepositoryV2.java
minwoox: > Additionally, we might need to delete the incomplete secondary repository just in case only half of LaggedCommits is saved.

👍 

> Don't we need to throw an exception and stop replication rather than catching it if an exception is raised while creating a secondary repository?

I thought we can continue to serve because we can serve the request without a second repository.
The secondary repository does not have to be created in a timely manner and the plugin tries that in the next schedule. 🤔

minwoox (commented):
@ikhoon All fixed. PTAL. 😉

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepositoryV2.java
```diff
@@ -0,0 +1,857 @@
...
+        InternalRepository secondaryRepo = null;
+        try {
+            writeLock();
+            final Revision headRevision = this.headRevision;
+            isCreatingSecondaryRepo = true;
+            writeUnLock();
+
+            logger.info("Creating the secondary repository in {}/{}. head revision: {}.",
+                        parent.name(), originalRepoName, headRevision);
+            final Map<String, Entry<?>> entries = find(headRevision, ALL_PATH, ImmutableMap.of()).join();
```
ikhoon, line 764: Should we use multiple of `minRetentionCommits` so that all replias have the same histories?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepositoryV2.java
ikhoon: Should we need to set `isCreatingSecondaryRepo` to false and clear `laggedCommitsForSecondary` if there is an error while committing data to the secondary repository?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepositoryV2.java
minwoox: Thanks for the valuable comment. 😄 
I introduced `CreateRollingRepositoryCommand` and made the leader use its own head revision to create the rolling repository. PTAL. 🙇

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepositoryV2.java
```diff
@@ -0,0 +1,874 @@
...
+        }
+        return null;
+    }
+
+    private static boolean exceedsMinRetention(InternalRepository repo, Revision headRevision,
+                                               int minRetentionCommits, int minRetentionDays) {
+        final Revision firstRevision = repo.firstRevision();
+        if (minRetentionCommits != 0) {
+            if (headRevision.major() - firstRevision.major() <= minRetentionCommits) {
+                return false;
```
jrhee17, line 735: Q) My initial understanding was that both conditions (days and commits) have to be satisfied to start the commit retention process.
Don't we also have to look at `minRetentionDays` before deciding?

jrhee17 (commented):
I haven't looked through everything, but looks awesome! 👍 Left a question on retention for starters 🙇

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepositoryV2.java
```diff
@@ -0,0 +1,874 @@
...
+            writeLock();
+            try {
+                final Revision headRevision = this.headRevision;
+                if (!rollingRepositoryInitialRevision.equals(headRevision)) {
+                    assert rollingRepositoryInitialRevision.major() < headRevision.major();
+                    // There were commits after the createRollingRepositoryCommand is created
+                    // so we should catch up.
+                    final RevisionRange revisionRange = new RevisionRange(
+                            rollingRepositoryInitialRevision.forward(1), headRevision);
+                    final List<Commit> commits = primaryRepo.listCommits(ALL_PATH, MAX_MAX_COMMITS,
```
jrhee17, line 811: Q) I'm not sure how realistic/unrealistic the magic number 1000 is, but is it safe to say this won't be exceeded?
I'm asking because it seems like this implementation can break if `secondaryRepo.HEAD != primaryRepo.HEAD` (no problem if this isn't realistic)

I guess an alternative may be to allow partial `secondaryRepo` states.
This might also allow shorter write lock durations at the cost of additional complexity.

(or maybe we can just set an arbitrarily high value 🤔 )

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepositoryV2.java
```diff
@@ -0,0 +1,874 @@
...
+        } else {
+            createSecondaryRepo(rollingRepositoryInitialRevision);
+        }
+    }
+
+    private void promoteSecondaryRepo(InternalRepository secondaryRepo,
+                                      Revision rollingRepositoryInitialRevision) {
+        writeLock();
+        try {
+            logger.info("Promoting the secondary repository in {}/{}.", parent.name(), originalRepoName);
```
jrhee17, line 767: Q) What do you think of validating `secondaryRepo.HEAD == primaryRepo.HEAD` here? (inside the write lock)

server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaBuilder.java
```diff
@@ -520,6 +523,22 @@ public CentralDogmaBuilder writeQuotaPerRepository(int writeQuota, int timeWindo
...
 
+    /**
+     * Sets the configuration for retaining commits in a {@link Repository}.
+     * The {@link Repository} retains at least the number of {@code minRetentionCommits} when more than
+     * {@code minRetentionCommits} are made.
+     * The {@link Repository} also retains commits at least {@code minRetentionDays}. If both conditions are
+     * not satisfied, the commits are not removed.
+     * For example, when {@code minRetentionCommits} is set to 2000 and {@code minRetentionDays} is set to 14,
+     * the commits that are created more than 2000 are not removed until 14 days have passed. Set 0 to retain
+     * all commits.
```
ikhoon, line 534: I'm a bit confused with the policy when both `minRetentionCommits` and `minRetentionDays` are set.

> the commits that are created more than 2000 are not removed until 14 days have passed.

IIUC, if commits seem not to be removed in 14 days even if the number of commits exceeds 2000.
2000 `minRetentionCommits` with 14 `minRetentionDays` sounds like Central Dogma allows 2000 commits for 14 days.
If commits are created more than 2000 in 14 days, some old commits need to be removed. Although the old commits were less than 14 days old.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepositoryV2.java
jrhee17: Or maybe we can also make `MAX_MAX_COMMITS` a variable and only log if `HEAD - INIT > MAX_MAX_COMMITS` 🤔

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepositoryV2.java
```diff
@@ -0,0 +1,874 @@
...
+        checkState(shouldCreateRollingRepository(rollingRepositoryInitialRevision,
+                                                 minRetentionCommits, minRetentionDays) ==
+                   rollingRepositoryInitialRevision, "aaa");
+
+        final InternalRepository secondaryRepo = this.secondaryRepo;
+        if (secondaryRepo != null) {
+            promoteSecondaryRepo(secondaryRepo, rollingRepositoryInitialRevision);
+        } else {
+            createSecondaryRepo(rollingRepositoryInitialRevision);
+        }
```
ikhoon, line 760: As per SPR, how about removing `createSecondaryRepo()` in `promoteSecondaryRepo()`?
```suggestion
        if (secondaryRepo != null) {
            promoteSecondaryRepo(secondaryRepo, rollingRepositoryInitialRevision);
        }
        createSecondaryRepo(rollingRepositoryInitialRevision);
```

ks-yim:
I am just asking out of curiosity 😅.

I thought of a very naive approach of doing something jGit equivalent for `git rebase -i HEAD~N` to rewrite commit history via `fixup` or `squash` when I first looked at the issue.

Rolling repository is awesome but at the same time looks relatively complicated compared to the rebase option.
What would be the benefit of rolling repo strategy assuming that we can do the same thing with rebase?

Maybe rewriting thousands of commit history could be too heavy or rebase doesn't give us a desired result at all?

jrhee17:
@minwoox is away for a while, so leaving my guess/analysis 😅  Feel free to correct me

> I thought of a very naive approach of doing something jGit equivalent for git rebase -i HEAD~N to rewrite commit history via fixup or squash when I first looked at the issue.

From what I understand, this is what we're trying to do. We're trying to squash commits between `INIT~HEAD^N`.

> Rolling repository is awesome but at the same time looks relatively complicated compared to the rebase option.
What would be the benefit of rolling repo strategy assuming that we can do the same thing with rebase?

My guess is so that we can make the switch atomically. If any of the intermediate steps fails while squashing (which I think would involve multiple IOs to disk) the repository may result in a corrupt state.

ks-yim:
@jrhee17 Thanks for the reply!!

I thought we could just do something like..:
```java
Git.wrap(jGitRepository).rebase().setUpstream("HEAD~2000").runInteractively(new InteractiveHandler() {
    @Override
    public void prepareSteps(List<RebaseTodoLine> steps) {
        if (steps.size() <= 2000) {
            return;
        }
        for (int i = 0; i < 1000; i++) {
            try {
                step.setAction(Action.SQUASH);
            } catch (IllegalTodoFileModification e) {
                // exception handling..
            }
        }
    }

    @Override
    public String modifyCommitMessage(String commit) {
        return commit;
    }
})
```

> My guess is so that we can make the switch atomically. If any of the intermediate steps fails while squashing (which I think would involve multiple IOs to disk) the repository may result in a corrupt state.

I am not sure if I am getting it correctly, but agree that there seems to be no easy way to handle IO errors or merge conflict with the rebase option (though I doubt squashing will cause conflict).

ikhoon:
> Git.wrap(jGitRepository).rebase().setUpstream("HEAD~2000")

If a repository has a long history, it would not release a write lock for a long time. The long lock causes to block threads waiting for the lock.
So we decided to avoid manipulating the working repository.
We tried to apply Git GC periodically, but it was reverted. #615
A mega repo in Central Dogma in LINE has over 380k commits. IIRC, it took almost 10 minutes to run Git GC on the repository.

minwoox:
That's a good question, @ks-yim and Thanks @jrhee17 and @ikhoon for the answer. 😄 

There's another reason that I didn't use squash because it modifies the Git Hashes that we are internally using for revision mapping.
Let's look at the following example:
```
1th commit(hash: foo) -> 2nd commit(hash: bar) -> 3rd commit(hash: baz)
```
If we squash 1th and 2nd then the commits will be:
```
1th and 2nd commit(hash: fooX) -> 3rd commit(hash: bazX)
                                             This one is also changed.
```
The hashes of all commits after the squash point are changed so I can't just do the squash. 😄

ks-yim:
Thanks for the detailed comments!!
There's a lot of heavy lifting done for users under the hood in CD 👍!

minwoox:
> I'm a bit confused with the policy when both minRetentionCommits and minRetentionDays are set.

Had a chat with @ikhoon and we decided to keep the current behavior. 😄

minwoox (commented):
Address all comments, PTAL. 🙇

minwoox:
https://github.com/line/centraldogma/runs/5300517235?check_suite_focus=true#step:5:659

It seems like the cache of the build exists and isn't cleaned somehow. 😓 
Let me close this PR and send it again.

PR #644, opened 2021-11-23
Fix flaky test in CentralDogmaEndpointGroupTest

minwoox:
https://github.com/line/centraldogma/blob/centraldogma-0.52.5/it/src/test/java/com/linecorp/centraldogma/it/CentralDogmaEndpointGroupTest.java#L105
`latch` is decremented before the `Endpoint` is set
https://github.com/line/centraldogma/blob/centraldogma-0.52.5/it/src/test/java/com/linecorp/centraldogma/it/CentralDogmaEndpointGroupTest.java#L118
because the listener that decrements the `latch` is executed before the `endpointListDecoder` is executed.


ikhoon (approved):
Thanks, @minwoox!

Issue #38, opened 2017-10-11
Advanced file filtering for mirrors


trustin:
Some users want to apply more complicated inclusion/exclusion rules for a mirror. We could support rsync-like exclusion settings.

minwoox:
The files that matched to the `localPath` are pushed in this method.
https://github.com/line/centraldogma/blob/centraldogma-0.52.5/server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/GitMirror.java#L88

yyuunn0044:
Let me try this one

PR #645, opened 2021-11-26
Add file exclusion rule for mirroring

yyuunn0044:
**Motivation**
Related to https://github.com/line/centraldogma/issues/38

**Modifications**

- Add `remoteExclude` property to `SingleMirrorConfig`.
  - It takes string of `gitignore` format

### DIFF
--- a/it/src/test/java/com/linecorp/centraldogma/it/mirror/git/GitMirrorTest.java
+++ b/it/src/test/java/com/linecorp/centraldogma/it/mirror/git/GitMirrorTest.java
@@ -213,8 +213,8 @@ class GitMirrorTest {
     }
 
     @Test
-    void remoteToLocal_remoteExcludePath() throws Exception {
-        pushMirrorSettings(null, null, "second");
+    void remoteToLocal_remoteExclude() throws Exception {
+        pushMirrorSettings(null, null, "/first/ex*" + "\\n" + "second");
 
         final Revision rev0 = client.normalizeRevision(projName, REPO_FOO, Revision.HEAD).join();
 
@@ -243,6 +243,7 @@ class GitMirrorTest {
         //// This file should not be mirrored because it does not conform to CD's file naming rule.
         addToGitIndex(".gitkeep", "");
         addToGitIndex("first/light.txt", "26-Aug-2014");
+        addToGitIndex("first/exclude.txt", "26-Aug-2014");
         addToGitIndex("second/son.json", "{\"release_date\": \"21-Mar-2014\"}");
         git.commit().setMessage("Add the release dates of the 'Infamous' series").call();
 
@@ -452,12 +453,12 @@ class GitMirrorTest {
     }
 
     private void pushMirrorSettings(@Nullable String localPath, @Nullable String remotePath,
-                                    @Nullable String remoteExcludePath) {
-        pushMirrorSettings(REPO_FOO, localPath, remotePath, remoteExcludePath);
+                                    @Nullable String remoteExclude) {
+        pushMirrorSettings(REPO_FOO, localPath, remotePath, remoteExclude);
     }
 
     private void pushMirrorSettings(String localRepo, @Nullable String localPath, @Nullable String remotePath,
-                                    @Nullable String remoteExcludePath) {
+                                    @Nullable String remoteExclude) {
         client.push(projName, Project.REPO_META, Revision.HEAD, "Add /mirrors.json",
                     Change.ofJsonUpsert("/mirrors.json",
                                         "[{" +
@@ -466,9 +467,7 @@ class GitMirrorTest {
                                         "  \"localRepo\": \"" + localRepo + "\"," +
                                         (localPath != null ? "\"localPath\": \"" + localPath + "\"," : "") +
                                         "  \"remoteUri\": \"" + gitUri + firstNonNull(remotePath, "") + '"' +
-                                        (remoteExcludePath != null ?
-                                         ",\"remoteExcludePath\": \"" + firstNonNull(remoteExcludePath, "") +
-                                         '"' : "") +
+                                        ",\"remoteExclude\": \"" + firstNonNull(remoteExclude, "") + "\"" +
                                         "}]")).join();
     }
 
--- a/it/src/test/java/com/linecorp/centraldogma/it/mirror/git/GitMirrorTest.java
+++ b/it/src/test/java/com/linecorp/centraldogma/it/mirror/git/GitMirrorTest.java
@@ -467,8 +467,8 @@ class GitMirrorTest {
                                         (localPath != null ? "\"localPath\": \"" + localPath + "\"," : "") +
                                         "  \"remoteUri\": \"" + gitUri + firstNonNull(remotePath, "") + '"' +
                                         (remoteExcludePath != null ?
-                                         ",\"remoteExcludePath\": \"" + firstNonNull(remoteExcludePath, "")
-                                         + '"' : "") +
+                                         ",\"remoteExcludePath\": \"" + firstNonNull(remoteExcludePath, "") +
+                                         '"' : "") +
                                         "}]")).join();
     }
 
--- a/it/src/test/java/com/linecorp/centraldogma/it/mirror/git/GitMirrorTest.java
+++ b/it/src/test/java/com/linecorp/centraldogma/it/mirror/git/GitMirrorTest.java
@@ -146,7 +146,7 @@ class GitMirrorTest {
 
     @Test
     void remoteToLocal() throws Exception {
-        pushMirrorSettings(null, null);
+        pushMirrorSettings(null, null, null);
 
         final Revision rev0 = client.normalizeRevision(projName, REPO_FOO, Revision.HEAD).join();
 
@@ -213,8 +213,56 @@ class GitMirrorTest {
     }
 
     @Test
+    void remoteToLocal_remoteExcludePath() throws Exception {
+        pushMirrorSettings(null, null, "second");
+
+        final Revision rev0 = client.normalizeRevision(projName, REPO_FOO, Revision.HEAD).join();
+
+        // Mirror an empty git repository, which will;
+        // - Create /mirror_state.json
+        // - Remove the sample files created by createProject().
+        mirroringService.mirror().join();
+
+        //// Make sure a new commit is added.
+        final Revision rev1 = client.normalizeRevision(projName, REPO_FOO, Revision.HEAD).join();
+        assertThat(rev1).isEqualTo(rev0.forward(1));
+
+        //// Make sure /mirror_state.json exists (and nothing else.)
+        final Entry<JsonNode> expectedInitialMirrorState = expectedMirrorState(rev1, "/");
+        assertThat(client.getFiles(projName, REPO_FOO, rev1, "/**").join().values())
+                .containsExactly(expectedInitialMirrorState);
+
+        // Try to mirror again with no changes in the git repository.
+        mirroringService.mirror().join();
+
+        //// Make sure it does not try to produce an empty commit.
+        final Revision rev2 = client.normalizeRevision(projName, REPO_FOO, Revision.HEAD).join();
+        assertThat(rev2).isEqualTo(rev1);
+
+        // Now, add some files to the git repository and mirror.
+        //// This file should not be mirrored because it does not conform to CD's file naming rule.
+        addToGitIndex(".gitkeep", "");
+        addToGitIndex("first/light.txt", "26-Aug-2014");
+        addToGitIndex("second/son.json", "{\"release_date\": \"21-Mar-2014\"}");
+        git.commit().setMessage("Add the release dates of the 'Infamous' series").call();
+
+        mirroringService.mirror().join();
+
+        //// Make sure a new commit is added.
+        final Revision rev3 = client.normalizeRevision(projName, REPO_FOO, Revision.HEAD).join();
+        assertThat(rev3).isEqualTo(rev2.forward(1));
+
+        //// Make sure only the file under '/first' is there, because '/second' was excluded.
+        final Entry<JsonNode> expectedSecondMirrorState = expectedMirrorState(rev3, "/");
+        assertThat(client.getFiles(projName, REPO_FOO, rev3, "/**").join().values())
+                .containsExactlyInAnyOrder(expectedSecondMirrorState,
+                                           Entry.ofDirectory(rev3, "/first"),
+                                           Entry.ofText(rev3, "/first/light.txt", "26-Aug-2014\n"));
+    }
+
+    @Test
     void remoteToLocal_subdirectory() throws Exception {
-        pushMirrorSettings("/target", "/source/main");
+        pushMirrorSettings("/target", "/source/main", null);
 
         client.push(projName, REPO_FOO, Revision.HEAD, "Add a file that's not part of mirror",
                     Change.ofTextUpsert("/not_mirrored.txt", "")).join();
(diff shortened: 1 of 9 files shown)


it/src/test/java/com/linecorp/centraldogma/it/mirror/git/GitMirrorTest.java
```diff
@@ -416,6 +467,7 @@ private void pushMirrorSettings(String localRepo, @Nullable String localPath, @N
                                         "  \"localRepo\": \"" + localRepo + "\"," +
                                         (localPath != null ? "\"localPath\": \"" + localPath + "\"," : "") +
                                         "  \"remoteUri\": \"" + gitUri + firstNonNull(remotePath, "") + '"' +
+                                        ",\"remoteExclude\": \"" + firstNonNull(remoteExclude, "") + "\"" +
```
minwoox, line 470: How about using `gitIgnore` instead of `remoteExclude` so that users easily notice what it is?
Also we don't have to change the field name when we implement CD_TO_GIT mirroring. 😉

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/AbstractMirror.java
```diff
@@ -54,12 +54,14 @@
...
     private final String remoteBranch;
+    private final String gitIgnore;
     private final ExecutionTime executionTime;
     private final long jitterMillis;
 
     protected AbstractMirror(Cron schedule, MirrorDirection direction, MirrorCredential credential,
                              Repository localRepo, String localPath,
-                             URI remoteRepoUri, String remotePath, @Nullable String remoteBranch) {
+                             URI remoteRepoUri, String remotePath, @Nullable String remoteBranch,
+                             @Nullable String gitIgnore) {
```
minwoox, line 64: How about accepting an array of String which is `Iterable<String> gitIgnore` instead of just a String?
People usually use multiple lines to exclude the files. >_<

minwoox: Had a chat with @yyuunn0044 and it's fine as it is. 😄 
We are eventually going to implement UI for mirroring setup so it's better to take a String instead of an array of Strings. 😄

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/GitMirror.java
```diff
@@ -148,6 +166,10 @@ protected void mirrorRemoteToLocal(File workDir, CommandExecutor executor,
                     final FileMode fileMode = treeWalk.getFileMode();
                     final String path = '/' + treeWalk.getPathString();
 
+                    if (ignoreNode.isIgnored(path, fileMode == FileMode.TREE) == MatchResult.IGNORED) {
```
minwoox, line 169: Food for thought: Should we strip the `remotePath` from the `path` to check if it's ignored or not?

minwoox: Would it be better to check only when `gitIgnore` is specified?
```java
@Nullable
private final IgnoreNode ignoreNode;

...
if (gitIgnore != null) {
    ignoreNode = new IgnoreNode();
    ignoreNode.parse(...);
}

...
if (ignoreNode != null && ...) {
    ...
}
```

minwoox (commented):
Basically looks good to me. 😉 
Left some comments.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/GitMirror.java
yyuunn0044: > Should we strip the remotePath from the path to check if it's ignored or not?

I think it should be.
And gitignore should be written in relative to `remotePath`
ex. remotePath is `/aaa` and excluded file is `/aaa/ex.txt`, gitignore will be `/ex.txt`
Let me try this way

it/src/test/java/com/linecorp/centraldogma/it/mirror/git/GitMirrorTest.java
```diff
@@ -212,9 +212,65 @@ void remoteToLocal() throws Exception {
                                            Entry.ofText(rev4, "/final_fantasy_xv.txt", "29-Nov-2016\n"));
     }
 
+    @Test
+    void remoteToLocal_gitIgnore() throws Exception {
+        pushMirrorSettings(null, "/first", "/exclude_if_root.txt" + "\\n" + "exclude_dir");
```
minwoox, line 217: Could you add more test cases that the `localPath` and `remotePath` have sub-directories?
e.g.
- `localPath == /a` and `remotePath == /b/c`
- `localPath == /a/b` and `remotePath == /c`

minwoox (commented):
Thanks! I guess it's almost done. 😆

it/src/test/java/com/linecorp/centraldogma/it/mirror/git/GitMirrorTest.java
yyuunn0044: Q: By the tests, what cases will be covered additionally?
I'm a little confused because they're being covered by `remoteToLocal_subdirectory`

minwoox: Ah, exactly we don't need the case. 😅 
How about also adding a case that the remote repository has a new commit that only contains ignored files?

minwoox (commented):
Just one last nit. Thanks a lot. 😆

minwoox (approved):
Congratulations on your first PR, @yyuunn0044! 🎉 🎉 🎉 
Great work! I'm really looking forward to your next PR. 😆

yyuunn0044:
😆  Thank you for your help~~

minwoox:
Could you also update the documentation as the follow-up? 😆 
https://github.com/line/centraldogma/blob/master/site/src/sphinx/mirroring.rst

it/src/test/java/com/linecorp/centraldogma/it/mirror/git/GitMirrorTest.java
```diff
@@ -416,6 +491,7 @@ private void pushMirrorSettings(String localRepo, @Nullable String localPath, @N
                                         "  \"localRepo\": \"" + localRepo + "\"," +
                                         (localPath != null ? "\"localPath\": \"" + localPath + "\"," : "") +
                                         "  \"remoteUri\": \"" + gitUri + firstNonNull(remotePath, "") + '"' +
+                                        ",\"gitIgnore\": \"" + firstNonNull(gitIgnore, "") + "\"" +
```
ikhoon, line 494: nit:
```suggestion
                                        ",\"gitIgnore\": \"" + firstNonNull(gitIgnore, "") + '"' +
```

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/DefaultMetaRepository.java
```diff
@@ -219,6 +219,7 @@ static MirrorCredential findCredential(Iterable<MirrorCredential> credentials, U
         final String localRepo;
         final String localPath;
         final URI remoteUri;
+        final String gitIgnore;
```
ikhoon, line 222: nit:
```suggestion
        @Nullable
        final String gitIgnore;
```

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/DefaultMetaRepository.java
```diff
@@ -230,6 +231,7 @@ static MirrorCredential findCredential(Iterable<MirrorCredential> credentials, U
                            @JsonProperty(value = "localRepo", required = true) String localRepo,
                            @JsonProperty("localPath") @Nullable String localPath,
                            @JsonProperty(value = "remoteUri", required = true) URI remoteUri,
+                           @JsonProperty("gitIgnore") @Nullable String gitIgnore,
```
ikhoon, line 234: I heard that we decided to use `String` because we will introduce a UI for mirroring in the future.
However, I think even though we provide web UI, a list of ignored files is more human-readable and it would be good for debugging.
What do you think that we allow both `String` and `List<String>`?
```java
{
    "gitIgnore": "/foo\n/bar/quz\n/qux.txt",
    "gitIgnore": [
                   "/foo", 
                   "/bar", 
                   "/quz",
                   "qux.txt" 
                  ]
}
```
We can just change the type of `gitIgnore` into `Object`. For example:
```
@JsonProperty("gitIgnore") @Nullable Object gitIgnore, ...) {
    if (gitIgnore != null) {
	if (gitIgnore instanceof Iterable) {
	    gitIgnore = ...; // join the collection with "\n"
	} else if (gitIgnore instanceof String) {
	     gitIgnore = gitIgnore;
	} else {
              throw new IllegalAguementExcepiton("gitIgnore: %s (expected: either string or array)", gitIgnore);
        }
    }

    }
}
```

yyuunn0044: LGTM!!

minwoox: SGTM. 😄

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/DefaultMetaRepository.java
```diff
@@ -238,6 +241,18 @@ static MirrorCredential findCredential(Iterable<MirrorCredential> credentials, U
             this.localRepo = requireNonNull(localRepo, "localRepo");
             this.localPath = firstNonNull(localPath, "/");
             this.remoteUri = requireNonNull(remoteUri, "remoteUri");
+            if (gitIgnore != null) {
+                if (gitIgnore instanceof Iterable) {
+                    this.gitIgnore = String.join("\n", (Iterable<String>) gitIgnore);
```
ikhoon, line 246: There was one thing I missed.
Could you check that all elements in the `Iterable` are `String` before casting it to `Iterable<String>`?

site/src/sphinx/mirroring.rst
```diff
@@ -104,6 +108,13 @@ You need to put two files into the ``meta`` repository of your Central Dogma pro
     the credential whose ``hostnamePattern`` is matched by the host name part of the ``remoteUri`` value will
     be selected automatically.
 
+- ``gitIgnore`` (string or array of string, optional)
+
+  - a `gitIgnore <https://git-scm.com/docs/gitignore>`_ pattern for the files excluded from mirroring.
+    The type of gitIgnore can be a full string of gitIgnore (e.g. ``/filename.txt\ndirectory``) or an array of
+    strings which consists of each line of gitIgnore. The file path expressed in gitIgnore is relative to the
+    path of ``remoteUri``.
```
ikhoon, line 116: By the way, which is better `gitIgnore` vs `gitignore`.
The official expression is `gitignore`. I think we can regard `gitignore` as a single word rather than a composition of `git` + `ignore`.
I'm leaning toward `gitignore`. Any thoughts? @minwoox @jrhee17

minwoox: Thanks! Let's use `gitignore`. 😄

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/CentralDogmaMirror.java
```diff
@@ -35,9 +35,11 @@
 
     public CentralDogmaMirror(Cron schedule, MirrorDirection direction, MirrorCredential credential,
                               Repository localRepo, String localPath,
-                              URI remoteRepoUri, String remoteProject, String remoteRepo, String remotePath) {
+                              URI remoteRepoUri, String remoteProject, String remoteRepo, String remotePath,
+                              String gitignore) {
```
ikhoon, line 39: `@Nullable`?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/DefaultMetaRepository.java
```diff
@@ -238,6 +242,20 @@ static MirrorCredential findCredential(Iterable<MirrorCredential> credentials, U
             this.localRepo = requireNonNull(localRepo, "localRepo");
             this.localPath = firstNonNull(localPath, "/");
             this.remoteUri = requireNonNull(remoteUri, "remoteUri");
+            if (gitignore != null) {
+                if (gitignore instanceof Iterable &&
+                    StreamSupport.stream(((Iterable<?>) gitignore).spliterator(), false)
```
ikhoon, line 247: nit: Could use Guava’s `Streams.stream()`.

ikhoon (approved):
Thanks a lot, @yyuunn0044! 🚀
Central Dogma will love this feature.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/DefaultMetaRepository.java
minwoox: Would it be possible to add a test case that takes an array of Strings? 😆

yyuunn0044:
retest this, please

jrhee17 (approved):
💯  Only left nit comments on documentation! Looks really nice 👍  Thank you! 🎉 🥳 🙇

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/CentralDogmaMirror.java
```diff
@@ -35,9 +37,11 @@
 
     public CentralDogmaMirror(Cron schedule, MirrorDirection direction, MirrorCredential credential,
                               Repository localRepo, String localPath,
-                              URI remoteRepoUri, String remoteProject, String remoteRepo, String remotePath) {
+                              URI remoteRepoUri, String remoteProject, String remoteRepo, String remotePath,
+                              @Nullable String gitignore) {
```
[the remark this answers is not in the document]
minwoox: It's not used at the moment. 😆 
There's no harm adding it so it should be fine. 😉

it/src/test/java/com/linecorp/centraldogma/it/mirror/git/GitMirrorTest.java
```diff
@@ -212,9 +212,138 @@ void remoteToLocal() throws Exception {
...
+                                           Entry.ofText(rev4, "/light.txt", "26-Aug-2014\n"),
+                                           Entry.ofDirectory(rev4, "/subdir"),
+                                           Entry.ofText(rev4, "/subdir/exclude_if_root.txt", "26-Aug-2014\n"));
+    }
+
+    @Test
+    void remoteToLocal_gitignore_with_array() throws Exception {
+        pushMirrorSettings(null, "/first", "[\"/exclude_if_root.txt\", \"exclude_dir\"]");
+
+        final Revision rev0 = client.normalizeRevision(projName, REPO_FOO, Revision.HEAD).join();
```
minwoox, line 292: It seems like there're lots of duplicate logic here and `remoteToLocal_gitignore`.
Could we dedup them by extracting the logic? 🤔 
```java
void remoteToLocal_gitignore() throws Exception {
    pushMirrorSettings(null, "/first", "\"/exclude_if_root.txt\\nexclude_dir\"");
    checkGitignore();
}

void remoteToLocal_gitignore_with_array() throws Exception {
    pushMirrorSettings(null, "/first", "[\"/exclude_if_root.txt\", \"exclude_dir\"]");
    checkGitignore();
}
```

minwoox:
Excellent work!
Congratulations, @yyuunn0044! 🎉 🎉 🎉 
I'm really looking forward to your next PR. 😄

PR #649, opened 2021-11-29
Disallow to create non-reserved files in internal repositories

### LINKED ISSUE #633: Disallow to create non-reserved files in meta repository
A meta repository is a special repository to manage a project.
So it has restricted permission. 
However, some users use a meta repository as a normal repository that we do not expect.
It would be nice to prohibit users from creating personal files.

ghkim3221:
### Motivation

 - #633

### Modifications

 - Checks a commit not to create non-reserved files in internal repositories (`dogma` and `meta`).
 - Create `RepositoryNotAllowedException` which is raised in this case.

### Results

 - Close #633
 - Users can't create a file except for `mirrors.json` in the internal repositories.
 - The HTTP API will respond `403 Forbidden` with the `RepositoryNotAllowedException` in this case.
 - The Thrift API will respond `BAD_REQUEST` in this case due to backward compatibility.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
@@ -24,6 +24,7 @@ import static com.linecorp.centraldogma.internal.Util.isValidDirPath;
 import static com.linecorp.centraldogma.internal.Util.isValidFilePath;
 import static com.linecorp.centraldogma.server.internal.api.DtoConverter.convert;
 import static com.linecorp.centraldogma.server.internal.api.HttpApiUtil.returnOrThrow;
+import static com.linecorp.centraldogma.server.internal.storage.repository.DefaultMetaRepository.metaRepoFiles;
 import static java.util.Objects.requireNonNull;
 
 import java.util.Collection;
@@ -34,7 +35,6 @@ import java.util.Optional;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Function;
-import java.util.function.Predicate;
 
 import javax.annotation.Nullable;
 
@@ -99,9 +99,6 @@ import com.linecorp.centraldogma.server.storage.repository.Repository;
 public class ContentServiceV1 extends AbstractService {
 
     private static final String MIRROR_LOCAL_REPO = "localRepo";
-    private static final Predicate<Change<?>> NOT_INTERNAL_FILES =
-            change -> !DefaultMetaRepository.PATH_CREDENTIALS.equals(change.path()) &&
-                      !DefaultMetaRepository.PATH_MIRRORS.equals(change.path());
 
     private final WatchService watchService;
 
@@ -416,15 +413,16 @@ public class ContentServiceV1 extends AbstractService {
      */
     public static void checkPush(String repoName, Iterable<Change<?>> changes) {
         if (Project.REPO_META.equals(repoName)) {
-            final boolean hasChangesOtherThanInternalFiles = Streams.stream(changes)
-                                                                    .anyMatch(NOT_INTERNAL_FILES);
-            if (hasChangesOtherThanInternalFiles) {
+            final boolean hasChangesOtherThanMetaRepoFiles =
+                    Streams.stream(changes).anyMatch(change -> !metaRepoFiles.contains(change.path()));
+            if (hasChangesOtherThanMetaRepoFiles) {
                 throw new InvalidPushException(
                         "The " + Project.REPO_META + " repository is reserved for internal usage.");
             }
 
             final Optional<String> notAllowedLocalRepo =
                     Streams.stream(changes)
+                           .filter(change -> DefaultMetaRepository.PATH_MIRRORS.equals(change.path()))
                            .filter(change -> change.content() != null)
                            .map(change -> {
                                final Object content = change.content();
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
@@ -61,6 +61,7 @@ import com.linecorp.centraldogma.common.Entry;
 import com.linecorp.centraldogma.common.Markup;
 import com.linecorp.centraldogma.common.MergeQuery;
 import com.linecorp.centraldogma.common.Query;
+import com.linecorp.centraldogma.common.RepositoryNotAllowedException;
 import com.linecorp.centraldogma.common.Revision;
 import com.linecorp.centraldogma.common.RevisionRange;
 import com.linecorp.centraldogma.internal.api.v1.ChangeDto;
@@ -186,6 +187,7 @@ public class ContentServiceV1 extends AbstractService {
             CommitMessageDto commitMessage,
             @RequestConverter(ChangesRequestConverter.class) Iterable<Change<?>> changes) {
         checkMirrorLocalRepo(repository.name(), changes);
+        checkPushLocalRepo(repository.name(), changes);
 
         final long commitTimeMillis = System.currentTimeMillis();
         return push(commitTimeMillis, author, repository, new Revision(revision), commitMessage, changes)
@@ -438,6 +440,22 @@ public class ContentServiceV1 extends AbstractService {
             if (notAllowedLocalRepo.isPresent()) {
                 throw new IllegalArgumentException("invalid " + MIRROR_LOCAL_REPO + ": " +
                                                    notAllowedLocalRepo.get());
+            }
+        }
+    }
+
+    /**
+     * Checks if the commit is for creating a file and raises a {@link RepositoryNotAllowedException} if the
+     * given {@code repoName} field is one of {@code meta} and {@code dogma} which are internal repositories.
+     */
+    public static void checkPushLocalRepo(String repoName, Iterable<Change<?>> changes) {
+        if (Project.isReservedRepoName(repoName)) {
+            final boolean hasChangesWithoutMirroring =
+                    Streams.stream(changes)
+                           .anyMatch(change -> !DefaultMetaRepository.PATH_MIRRORS.equals(change.path()));
+            if (hasChangesWithoutMirroring) {
+                throw new RepositoryNotAllowedException(
+                        "The " + repoName + " repository is reserved for internal usage.");
             }
         }
     }
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
@@ -58,10 +58,10 @@ import com.linecorp.armeria.server.annotation.RequestConverter;
 import com.linecorp.centraldogma.common.Author;
 import com.linecorp.centraldogma.common.Change;
 import com.linecorp.centraldogma.common.Entry;
+import com.linecorp.centraldogma.common.InvalidPushException;
 import com.linecorp.centraldogma.common.Markup;
 import com.linecorp.centraldogma.common.MergeQuery;
 import com.linecorp.centraldogma.common.Query;
-import com.linecorp.centraldogma.common.RepositoryNotAllowedException;
 import com.linecorp.centraldogma.common.Revision;
 import com.linecorp.centraldogma.common.RevisionRange;
 import com.linecorp.centraldogma.internal.api.v1.ChangeDto;
@@ -408,8 +408,8 @@ public class ContentServiceV1 extends AbstractService {
     }
 
     /**
-     * Checks if the commit is for mirroring setting and raises an exception if the {@code localRepo} field
-     * is one of {@code meta} and {@code dogma} which are internal repositories.
+     * Checks if the commit is for mirroring setting and raises a {@link InvalidPushException} if the given
+     * {@code localRepo} field is one of {@code meta} and {@code dogma} which are internal repositories.
      */
     public static void checkMirrorLocalRepo(String repoName, Iterable<Change<?>> changes) {
         // TODO(minwoox): Provide an internal API for mirroring setup with a better UI(?) and check this there.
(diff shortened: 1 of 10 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
```diff
@@ -441,4 +443,20 @@ public static void checkMirrorLocalRepo(String repoName, Iterable<Change<?>> cha
...
+     * Checks if the commit is for creating a file and raises a {@link RepositoryNotAllowedException} if the
+     * given {@code repoName} field is one of {@code meta} and {@code dogma} which are internal repositories.
+     */
+    public static void checkPushLocalRepo(String repoName, Iterable<Change<?>> changes) {
+        if (Project.isReservedRepoName(repoName)) {
+            final boolean hasChangesWithoutMirroring =
+                    Streams.stream(changes)
+                           .anyMatch(change -> !DefaultMetaRepository.PATH_MIRRORS.equals(change.path()));
+            if (hasChangesWithoutMirroring) {
+                throw new RepositoryNotAllowedException(
```
minwoox, line 457: Maybe `InvalidPushException`?

ghkim3221: Nice idea.
If so, which HTTP status do you expect when `InvalidPushException` is raised? `400 Bad Request` of `403 Forbidden`?

minwoox: I would go for 403. 😄

minwoox: I think we can also throw `InvalidPushException` in `checkMirrorLocalRepo`. 😉

server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
```diff
@@ -436,8 +438,24 @@ public static void checkMirrorLocalRepo(String repoName, Iterable<Change<?>> cha
...
+            }
+        }
+    }
+
+    /**
+     * Checks if the commit is for creating a file and raises a {@link InvalidPushException} if the
+     * given {@code repoName} field is one of {@code meta} and {@code dogma} which are internal repositories.
+     */
+    public static void checkPushLocalRepo(String repoName, Iterable<Change<?>> changes) {
+        if (Project.isReservedRepoName(repoName)) {
```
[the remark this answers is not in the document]
ghkim3221: Right, it would be better to refuse all changes to the `dogma` repository.

While applying, I merged `checkMirrorLocalRepo` and `checkPushLocalRepo` to `checkPush`. Is it fine for you? Would it be breaking changes?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
```diff
@@ -406,12 +407,24 @@ private static Object handleWatchFailure(Throwable thrown) {
...
+                    Streams.stream(changes)
+                           .anyMatch(change -> !DefaultMetaRepository.PATH_MIRRORS.equals(change.path()));
+            if (hasChangesWithoutMirroring) {
+                throw new InvalidPushException(
+                        "The " + Project.REPO_META + " repository is reserved for internal usage.");
+            }
+
             final Optional<String> notAllowedLocalRepo =
                     Streams.stream(changes)
                            .filter(change -> DefaultMetaRepository.PATH_MIRRORS.equals(change.path()))
```
minwoox, line 430: I think we can remove this line because we have already checked it above. 😄

server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
```diff
@@ -406,12 +407,24 @@ private static Object handleWatchFailure(Throwable thrown) {
...
-    public static void checkMirrorLocalRepo(String repoName, Iterable<Change<?>> changes) {
-        // TODO(minwoox): Provide an internal API for mirroring setup with a better UI(?) and check this there.
+    public static void checkPush(String repoName, Iterable<Change<?>> changes) {
+        if (Project.REPO_DOGMA.equals(repoName)) {
+            throw new InvalidPushException(
+                    "The " + Project.REPO_DOGMA + " repository is reserved for internal usage.");
+        }
+
         if (Project.REPO_META.equals(repoName)) {
+            final boolean hasChangesWithoutMirroring =
```
minwoox, line 420: `hasChangesOtherThanMirroring` or
`hasInvalidChanges`?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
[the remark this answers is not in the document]
minwoox: > While applying, I merged checkMirrorLocalRepo and checkPushLocalRepo to checkPush. Is it fine for you? Would it be breaking changes?

I think it's okay because the client will get a CentralDogmaException anyway.
So, let's merge them. 😉

server/src/main/java/com/linecorp/centraldogma/server/internal/thrift/Converter.java
```diff
@@ -168,7 +169,7 @@ static Project convert(
                 com.linecorp.centraldogma.common.Revision.HEAD,
                 com.linecorp.centraldogma.common.Revision.HEAD,
                 com.linecorp.centraldogma.server.storage.repository.Repository.ALL_PATH, 1).thenApply(
-                        history -> new Repository(name).setHead(convert(history.get(0))));
+                history -> new Repository(name).setHead(convert(history.get(0))));
```
ikhoon, line 172: revert?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
```diff
@@ -406,12 +407,24 @@ private static Object handleWatchFailure(Throwable thrown) {
...
+                    Streams.stream(changes)
+                           .anyMatch(change -> !DefaultMetaRepository.PATH_MIRRORS.equals(change.path()));
```
ikhoon, line 422: Could we check the permission of the current user? We should allow only `OWNER` permission.
https://github.com/line/centraldogma/blob/082d87bd172fc0705579f2774458466e0dbfc1c6/server/src/main/java/com/linecorp/centraldogma/server/internal/api/MetadataApiService.java#L63

ghkim3221: Hmm, good suggestion. Let me find a way.

ghkim3221: How about calling `findRole` while checking and returning `CompletableFutre<Void>`?

https://github.com/line/centraldogma/blob/81d0be9597ad409ebbae90f9884eb85e129931a9/server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java#L786

minwoox: I think we don't have to because `RequiresPermissionDecorator` handles it.
https://github.com/line/centraldogma/blob/master/server/src/main/java/com/linecorp/centraldogma/server/internal/api/auth/RequiresPermissionDecorator.java#L90

minwoox (approved):
Great job! 🎉 🎉 🎉 
Looking forward to your next PR. 😆

server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
ikhoon: Ah, you are right.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
```diff
@@ -406,15 +407,26 @@ private static Object handleWatchFailure(Throwable thrown) {
...
+     * Checks if the commit is for creating a file and raises a {@link InvalidPushException} if the
+     * given {@code repoName} field is one of {@code meta} and {@code dogma} which are internal repositories.
      */
-    public static void checkMirrorLocalRepo(String repoName, Iterable<Change<?>> changes) {
-        // TODO(minwoox): Provide an internal API for mirroring setup with a better UI(?) and check this there.
+    public static void checkPush(String repoName, Iterable<Change<?>> changes) {
+        if (Project.REPO_DOGMA.equals(repoName)) {
+            throw new InvalidPushException(
+                    "The " + Project.REPO_DOGMA + " repository is reserved for internal usage.");
+        }
```
ikhoon, line 417: According to the `RequiresPermissionDecorator`, requests sent by non-admin users never reach here.
Could we remove this?

minwoox: That's correct. We can remove 414~ 417 lines. 😉

ghkim3221: Thanks 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
```diff
@@ -406,15 +407,21 @@ private static Object handleWatchFailure(Throwable thrown) {
...
+     * Checks if the commit is for creating a file and raises a {@link InvalidPushException} if the
+     * given {@code repoName} field is one of {@code meta} and {@code dogma} which are internal repositories.
      */
-    public static void checkMirrorLocalRepo(String repoName, Iterable<Change<?>> changes) {
-        // TODO(minwoox): Provide an internal API for mirroring setup with a better UI(?) and check this there.
+    public static void checkPush(String repoName, Iterable<Change<?>> changes) {
         if (Project.REPO_META.equals(repoName)) {
+            final boolean hasChangesOtherThanMirroring =
+                    Streams.stream(changes)
+                           .anyMatch(change -> !DefaultMetaRepository.PATH_MIRRORS.equals(change.path()));
```
minwoox, line 417: `DefaultMetaRepository.PATH_CREDENTIALS` is valid also. 😱

ikhoon (approved):
Thanks, @ghkim3221! 🎉
Please address minwoox's comment.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
```diff
@@ -98,6 +99,9 @@
 public class ContentServiceV1 extends AbstractService {
 
     private static final String MIRROR_LOCAL_REPO = "localRepo";
+    private static final Predicate<Change<?>> NOT_INTERNAL_FILES =
```
minwoox, line 102: Instead of adding it here, how about adding it to `DefaultMetaRepository` and using it here?
```java
public class DefaultMetaRepository ... {

    public static final String PATH_CREDENTIALS = "/credentials.json";

    public static final String PATH_MIRRORS = "/mirrors.json";

    public static final List<String> metaRepoFiles = ImmutableList.of(PATH_CREDENTIALS, PATH_MIRRORS);

    ...
}

final boolean hasChangesOtherThanMetaRepoFiles =
                    Streams.stream(changes).anyMatch(change -> metaRepoFiles.contains(change.path()));
```

jrhee17 (approved):
💯  Left a minor comment. Thank you! 🙇 🥳 🎉

server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
ikhoon: * `hasChangesOtherThanMetaRepoFiles` -> `hasInternalFiles`
* No strong reason to use `static final Predicate<Change<?>>`. `(change) -> { ... }` is converted into a singleton lambda object that is created only once by JVM.
  Inline lambda expression might be better for readability.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
```diff
@@ -406,15 +411,20 @@ private static Object handleWatchFailure(Throwable thrown) {
...
+            final boolean hasChangesOtherThanInternalFiles = Streams.stream(changes)
+                                                                    .anyMatch(NOT_INTERNAL_FILES);
+            if (hasChangesOtherThanInternalFiles) {
+                throw new InvalidPushException(
+                        "The " + Project.REPO_META + " repository is reserved for internal usage.");
+            }
+
             final Optional<String> notAllowedLocalRepo =
                     Streams.stream(changes)
-                           .filter(change -> DefaultMetaRepository.PATH_MIRRORS.equals(change.path()))
```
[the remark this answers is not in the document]
minwoox: Yeah, exactly. We should revert this.
Thanks for catching it. @jrhee17 😄

server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
```diff
@@ -436,8 +446,8 @@ public static void checkMirrorLocalRepo(String repoName, Iterable<Change<?>> cha
                                return null;
                            }).filter(Objects::nonNull).findFirst();
             if (notAllowedLocalRepo.isPresent()) {
-                throw new IllegalArgumentException("invalid " + MIRROR_LOCAL_REPO + ": " +
-                                                   notAllowedLocalRepo.get());
+                throw new InvalidPushException("invalid " + MIRROR_LOCAL_REPO + ": " +
```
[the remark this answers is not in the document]
minwoox: >  (I think it doesn't matter though)

I think so. It's no big deal. 😄

server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
ghkim3221: Thanks !!

> hasChangesOtherThanMetaRepoFiles -> hasInternalFiles?

I used `hasChangesOtherThanMetaRepoFiles` because we should check files that are not internal.

> public static final List<String> metaRepoFiles = ImmutableList.of(PATH_CREDENTIALS, PATH_MIRRORS);

Like `Project#internalRepos()`, how about making the field private and exposing a method? Maybe it's ok since the `DefaultMetaRepository` is in the internal package?

https://github.com/line/centraldogma/blob/88ce886c8c6bf336839ae36a7c292cdc97137565/server/src/main/java/com/linecorp/centraldogma/server/storage/project/Project.java#L78-L80

server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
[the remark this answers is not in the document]
ghkim3221: Nice catch 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
minwoox: Either way is fine to me. 😄

minwoox:
Great work, @ghkim3221!
Glad that you are also becoming familiar with this project. 😆 
Thanks, reviewers also. 🙇‍♂️

PR #650, opened 2021-12-01
Throw error when entries are not found on watching if want

di-seo:
### Motivation
- https://github.com/line/centraldogma/issues/532
- To-Do of https://github.com/line/centraldogma/pull/610

### Modifications
- Add `WatchOptions` object to put additional options on watching file/repository
- Put value of `notify-entry-not-found` on `Prefer` request header using `WatchOptions#errorOnEntryNotFound`
- Propagate error that entries are not found to `Watcher#initialValueFuture` if `WatchOptions#errorOnEntryNotFound` is true

### Result
- You can get error on watching file/repository when the entry doesn't exist.

### DIFF
--- a/client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogma.java
+++ b/client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogma.java
@@ -77,6 +77,7 @@ import com.linecorp.armeria.common.util.SafeCloseable;
 import com.linecorp.armeria.common.util.TimeoutMode;
 import com.linecorp.centraldogma.client.AbstractCentralDogma;
 import com.linecorp.centraldogma.client.RepositoryInfo;
+import com.linecorp.centraldogma.client.WatchOptions;
 import com.linecorp.centraldogma.common.Author;
 import com.linecorp.centraldogma.common.AuthorizationException;
 import com.linecorp.centraldogma.common.CentralDogmaException;
@@ -801,12 +802,13 @@ final class ArmeriaCentralDogma extends AbstractCentralDogma {
     @Override
     public CompletableFuture<Revision> watchRepository(String projectName, String repositoryName,
                                                        Revision lastKnownRevision, String pathPattern,
-                                                       long timeoutMillis) {
+                                                       WatchOptions watchOptions) {
         try {
             validateProjectAndRepositoryName(projectName, repositoryName);
             requireNonNull(lastKnownRevision, "lastKnownRevision");
             validatePathPattern(pathPattern, "pathPattern");
-            checkArgument(timeoutMillis > 0, "timeoutMillis: %s (expected: > 0)", timeoutMillis);
+            checkArgument(watchOptions.getTimeoutMillis() > 0,
+                          "timeoutMillis: %s (expected: > 0)", watchOptions.getTimeoutMillis());
 
             final StringBuilder path = pathBuilder(projectName, repositoryName);
             path.append("/contents");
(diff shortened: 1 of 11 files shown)


client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogma.java
```diff
@@ -801,12 +802,13 @@ private static PushResult push(AggregatedHttpResponse res) {
     @Override
     public CompletableFuture<Revision> watchRepository(String projectName, String repositoryName,
                                                        Revision lastKnownRevision, String pathPattern,
-                                                       long timeoutMillis) {
+                                                       WatchOptions watchOptions) {
```
minwoox, line 805: Sorry for making you do this twice but let's remove `WatchOptions`. 😅 
What I'm trying to add in [another PR](https://github.com/line/centraldogma/pull/651) is that a fluent style for `CentralDogma`:
```java
dogma.forRepo("foo", "bar")
     .watchingFiles("/**")
     .timeoutMillis(...)
     .errorOnEntryNotFound(...)
     .watch(revision);
```
If we do this, I think we no longer need `WatchOptions` because we can just add a method that takes the whole parameters and deprecate others:
```java
// We are going to use this method.
CompletableFuture<Revision> watchRepository(String projectName, String repositoryName,
                                            Revision lastKnownRevision, String pathPattern,
                                            long timeoutMillis, boolean errorOnEntryNotFound);

// Other methods are deprecated because users can use fluent style API.
// dogma.forRepo("foo", "bar").watchingFiles(pathPattern).watch(lastKnownRevision);
@Deprecated
CompletableFuture<Revision> watchRepository(String projectName, String repositoryName,
                                            Revision lastKnownRevision, String pathPattern);

```

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java
```diff
@@ -517,8 +566,10 @@
      * <p>Note that {@link Function} by default is executed by a blocking task executor so that you can
      * safely call a blocking operation.
      */
-    <T, U> Watcher<U> fileWatcher(String projectName, String repositoryName,
-                                  Query<T> query, Function<? super T, ? extends U> function);
+    default <T, U> Watcher<U> fileWatcher(String projectName, String repositoryName,
+                                  Query<T> query, Function<? super T, ? extends U> function) {
```
minwoox, line 570: It seems like the indentation is wrong. Could you check your settings? 😉 
https://armeria.dev/community/developer-guide#setting-up-your-ide

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java
```diff
@@ -535,8 +586,41 @@
...
+     *        {@link WatchOptions#timeoutMillis} is the waiting time about the matched files have a change.
+     *        If any no changes during {@link WatchOptions#timeoutMillis}, re-wait the matched files.
+     *        <p>Note: Legacy client only support {@link WatchOptions#timeoutMillis}</p>
+     */
+
     <T, U> Watcher<U> fileWatcher(String projectName, String repositoryName,
-                                  Query<T> query, Function<? super T, ? extends U> function, Executor executor);
+                                  Query<T> query,
+                                  Function<? super T, ? extends U> function,
+                                  @Nullable Executor executor,
```
minwoox, line 622: Conventionally, we usually don't take a nullable parameter in our public API. 
Could you revert the change, please? which means that we should revert this change:
https://github.com/line/centraldogma/pull/650/files#diff-f08760fa027ab0e795025c0410578002d9b1b5313714f83c79940f2d1b174807R569

client/java/src/main/java/com/linecorp/centraldogma/client/WatchOptions.java
```diff
@@ -0,0 +1,46 @@
+package com.linecorp.centraldogma.client;
```
minwoox, line 1: missing copyright, but we don't need this class anymore so no big deal. 😅 
You can check if there are any style issues at your local using Gradle:
```bash
./gradlew clean build --parallel --no-daemon --max-workers=4
```
This is what I usually do. 😉

minwoox (commented):
Basically looking good. 😄 
Left some comments that we need to make due to the change of #651. 😉

client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogma.java
di-seo: I'm interested in the fluent style. Also I think it is good idea than my changes.
I try remake codes base on your changes

di-seo:
I will request pull again based on your changes [https://github.com/line/centraldogma/pull/651]

PR #651, opened 2021-12-02
Provide the fluent API for CentralDogma client

minwoox:
Motivation:
Currently, `CentralDogma` client has a bunch of methods that take a bunch of parameters.
It makes it harder to add another optional parameter to the existing methods because we need to add another pair of existing methods. For example:
```java
// Existing methods:
void watch(String a);
void watch(String a, String b);
void watch(String a, String b, String c);

// If we want to add an optional long value then we need another three methods:
void watch(String a, long d);
void watch(String a, String b, long d);
void watch(String a, String b, String c, long d);
```

Modifications:
- Add `PathPattern` interface to pass as the parameter instead of String.
- Add `CentralDogma.forRepo(...)` method that returns `CentralDogmaRepository`.
  - Users can use the instance to send a request:
     ```java
     centralDogma.forRepo("foo", "bar").file("/foo.json").get();
     centralDogma.forRepo("foo", "bar").watch("/foo.json").start();
     centralDogma.forRepo("foo", "bar").watcher(PathPattern.all()).start();
     ```
Result:
- You can now send a request to the Central Dogma server via fluent APIs.
- (Deprecated) Various methods in `CentralDogma` client are deprecated. Use fluent APIs.


minwoox:
This is ready for review. 😉

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaWatchingFileRequest.java
```diff
@@ -0,0 +1,137 @@
...
+
+    /**
+     * Sets the {@link Function} that convert the result of a watch. The {@link Function} is executed by
+     * a the specified {@link Executor}.
+     */
+    public <U> CentralDogmaWatchingFileRequest<U> map(Function<? super T, ? extends U> function,
+                                                      Executor executor) {
+        requireNonNull(function, "function");
+        requireNonNull(executor, "executor");
+        return new CentralDogmaWatchingFileRequest<>(requestPreparation, query, this.function.andThen(function),
```
di-seo, line 92: I'm wondering why you create new object not set variable?

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaRequestPreparation.java
```diff
@@ -0,0 +1,295 @@
...
+     * <ul>
+     *   <li>{@code "/**"} - find all files recursively</li>
+     *   <li>{@code "*.json"} - find all JSON files recursively</li>
+     *   <li>{@code "/foo/*.json"} - find all JSON files under the directory {@code /foo}</li>
+     *   <li><code>"/&#42;/foo.txt"</code> - find all files named {@code foo.txt} at the second depth level</li>
+     *   <li>{@code "*.json,/bar/*.txt"} - use comma to specify more than one pattern. A file will be matched
+     *                                     if <em>any</em> pattern matches.</li>
+     * </ul>
+     */
+    public CentralDogmaDiffsRequest diffs(String pathPattern) {
```
ikhoon, line 208: `diffs` is a bit unnatural to me. How about introducing a new query type that can represent multiple files using a path pattern?
For example:
```java
CentralDogmaDiffsRequest file(MultiQuery query){
}
...
CentralDogmaDiffsRequest diff(MultiQuery query){
}
interface MultiQuery {
    static all() {
        return new DefaultMultiQuery("/**");
    }
    // Could represent *.json
    static pattern(String pattern) {    
        ...
    }
    // Could represent *.json, /bar/*.txt
    static patterns(String patrerns...) {
    
    }
}
```

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaRequestPreparation.java
```diff
@@ -0,0 +1,295 @@
...
+        return new CentralDogmaCommitRequest(this, summary, changes);
+    }
+
+    /**
+     * Returns a new {@link CentralDogmaWatchingFileRequest} that is used to send
+     * {@link CentralDogma#watchFile(String, String, Revision, Query, long)} request to
+     * the Central Dogma repository or create a new
+     * {@link CentralDogma#fileWatcher(String, String, Query, Function, Executor)}.
+     */
+    public CentralDogmaWatchingFileRequest<?> watchingFile(String path) {
```
ikhoon, line 259: I prefer `watch` for the name of these APIs. We might express the watching API with `watch()`. For  example:
```java
repo.watch(path)
    .timeout(timeout)
    .once(Revision); // This can replace the orignal `watch()`

var watcher = 
repo.watch(path)
    .timeout(timeout)
    .start(); // This can replace the orignal `newWatcher()`
```

ikhoon (commented):
Looks fluent!

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaRequestPreparation.java
minwoox: I have to distinguish `watch(path)` (file watch) and `watch(pathPattern)` (repository watch).
Do you have any suggestions for this?

ikhoon: How about introducing `PathPattern` API? https://github.com/line/centraldogma/pull/651#discussion_r762017789

minwoox: So the benefit of introducing the new `PathPattern` interface is
- using the same method name e.g. `watch(String)`, `watch(PathPattern)`
- Specifying multiple pathPattern (even though I'm not sure how we concat the string using a special character which shouldn't be used in the pathPattern)
Is there anything I missed except those?

ikhoon: Actually, I’m not 100% sure that `PathPattern` is a good solution. But I guess we might provide more constant and fluent API with `PathPattern`.
If it is not your favorite, don't need to take the suggestion. 😄

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaRequestPreparation.java
```diff
@@ -0,0 +1,295 @@
...
+        requireNonNull(pathPattern, "pathPattern");
+        return new CentralDogmaDiffsRequest(this, pathPattern);
+    }
+
+    /**
+     * Returns a new {@link CentralDogmaPreviewDiffsRequest} that is used to send
+     * {@link CentralDogma#getPreviewDiffs(String, String, Revision, Iterable)} request to the
+     * Central Dogma repository.
+     */
+    public CentralDogmaPreviewDiffsRequest previewDiffs(Change<?>... changes) {
```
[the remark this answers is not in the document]
minwoox: Currently, we use
- `diff(path)` to see the diff of the specified path from the specified `from` revision to `to` revision.
- `previewDiff(Change)` to see the diff from the `HEAD` and the next revision which will be created if the specified Change is committed. 

But you want to use the same name
- `diff(path)` and `diff(Change)`

Is that correct?

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaRequestPreparation.java
minwoox: I like most of your suggestions but I just wanted to clarify if there's something that I missed so I left questions. 😆

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaRequestPreparation.java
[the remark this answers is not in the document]
ikhoon: Yes, it is. Although the behavior is different, I thought the parameter can explain the feature of the API. Preview might be a kind of diff that show the staging diff.

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaFilesRequest.java
```diff
@@ -0,0 +1,64 @@
...
+        this.requestPreparation = requestPreparation;
+        this.pathPattern = pathPattern;
+    }
+
+    /**
+     * Retrieves the list of the files matched by the given path pattern.
+     *
+     * @return a {@link Map} of file path and type pairs
+     */
+    public CompletableFuture<Map<String, EntryType>> list(Revision revision) {
```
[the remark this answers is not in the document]
minwoox: I used `get()` to align with other API names but I didn't have to. 😅 

How about using `contents()` and `list()`?
`list()` returns the `EntryType` as well so I prefer `list()` to `names()`. 🤔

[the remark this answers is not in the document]
ikhoon: contents() and list() sound good to me. 👍

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaRequestPreparation.java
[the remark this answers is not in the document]
minwoox: That makes sense. 👍 Let me just use `diff`.

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaRequestPreparation.java
minwoox: I still leaning toward using different method names because the returned instances are different depending on it is a file or files.
If `diffs` is vague how about appending files?
```java
repo.file(path);
repo.files(pathPattern);
repo.diff(path);
repo.diffFiles(pathPattern);
repo.diffFiles(Change...);
repo.watch(path);
repo.watchFiles(pathPattern);
```

ikhoon: Sounds good to me except for `repo.diffFiles(Change...)`. Because technically, `Change` is a are not written to a file yet.

minwoox: On second thought, I think it's good to use `PathPattern` interface because we can put the scattered Javadoc description into one place and we can validate the path pattern when users put a comma into a path pattern by mistake.
Thanks!

ikhoon: I guess `under` factory is also useful. 😆
```java
interface Pattern {
     Pattern under(String path) {
        return path + "/**";
     }
}

Pattern.under("/foo/bar");
```

minwoox: Let me create an issue for adding useful APIs to `PathPattern`. 😆

minwoox:
@di-seo Please, let me know if there's anything we can improve. 😄

client/java/src/main/java/com/linecorp/centraldogma/client/MappingWatcher.java
```diff
@@ -0,0 +1,179 @@
...
+        return mappedLatest;
+    }
+
+    @Override
+    public void close() {
+        closed = true;
+        if (!initialValueFuture.isDone()) {
+            initialValueFuture.cancel(false);
+        }
+        if (closeParentWhenClosing) {
```
jrhee17, line 142: On second thought, I'm concerned that users may forget to close the `DefaultWatcher`.

i.e.
```
final Watcher<Boolean> watcher = dogma.client()
                                      .forRepo("foo", "bar")
                                      .watch(Query.ofText("/baz.txt")).forever().newChild(s -> s).map(s -> true);
// default watcher won't be closed
watcher.close();
```

I'm unsure if there's a way we can enforce users to avoid this pattern though. If this becomes too confusing, I think just dropping support for `newChild` may also be an option 😅 (a breaking change may be better than having to investigate potential leaks..)

common/src/main/java/com/linecorp/centraldogma/internal/Jackson.java
```diff
@@ -129,9 +129,9 @@ public static void registerSubtypes(Class<?>... subtypes) {
         prettyMapper.registerSubtypes(subtypes);
     }
 
-    public static <T> T readValue(String data, Class<T> type) throws JsonParseException, JsonMappingException {
```
jrhee17, line 132: nit; revert

client/java/src/main/java/com/linecorp/centraldogma/client/MappingWatcher.java
```diff
@@ -0,0 +1,179 @@
...
+        requireNonNull(mapper, "mapper");
+        requireNonNull(executor, "executor");
+        // TODO(minwoo): extract mapper function and combine it with the new mapper.
+        return new MappingWatcher<>(parent, mapper, executor, closeParentWhenClosing);
+    }
+
+    private final Watcher<T> parent;
+    private final Function<? super T, ? extends U> mapper;
+    private final boolean closeParentWhenClosing;
+    private final CompletableFuture<Latest<U>> initialValueFuture = new CompletableFuture<>();
```
jrhee17, line 56: Q) I wonder if we can't just use `parent.initialValueFuture`.   I think it's possible that users use a pattern like the following where intermediate mapped/child watchers aren't closed.

```
final Watcher<String> watcher = dogma.client()
                                      .forRepo("foo", "bar")
                                      .watch(Query.ofText("/baz.txt")).forever();
// the mapped watcher won't be closed
Consumer<String> consumer = c -> { /* do something */ };
watcher.map(a -> {
    consumer.accept(a);
    return null;
});
watcher.close();
```

I'm wondering if this may be a potential source of memory leak (if the client uses `initialValueFuture` do to something)

jrhee17 (commented):
The overall API looks really nice! 🙇 👍  Left some minor questions on the `Watcher` implementation 🙏

di-seo (commented):
I left some comments.
I look forward to this changes. :)

common/src/main/java/com/linecorp/centraldogma/common/PathPattern.java
```diff
@@ -0,0 +1,73 @@
...
+import static java.util.Objects.requireNonNull;
+
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Streams;
+
+/**
+ * A path pattern that represents a variant of glob. For example:
+ * <ul>
+ *   <li>{@code "/**"} - all files</li>
+ *   <li>{@code "*.json"} - all JSON files</li>
```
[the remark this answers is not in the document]
minwoox: https://github.com/line/centraldogma/pull/651#discussion_r774300289
Left a comment for that. 😉

client/java/src/main/java/com/linecorp/centraldogma/client/WatcherOptionsBuilder.java
```diff
@@ -0,0 +1,96 @@
...
+        return this;
+    }
+
+    /**
+     * Sets the delays and multiplier which is used to calculate the delay
+     * for sending the next watch request when the previous request fails.
+     * Currently, it uses exponential backoff. File a feature request if you need another algorithm.
+     */
+    public WatcherOptionsBuilder backoffOnFailure(long initialDelayMillis, long maxDelayMillis,
+                                                  double multiplier) {
```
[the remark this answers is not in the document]
minwoox: This module does not have Armeria dependency which means we cannot use `Backoff`. 😉 
Let's add it when we really need that.

common/src/main/java/com/linecorp/centraldogma/internal/Jackson.java
minwoox: Oops 😓

client/java/src/main/java/com/linecorp/centraldogma/client/MappingWatcher.java
minwoox: >  I wonder if we can't just use parent.initialValueFuture

The complete value of the future will be different from the parent one so I think we cannot use it.
If the mapper returns `null`, a watcher currently leaves a warning log here.
https://github.com/line/centraldogma/pull/651/files#diff-5b249ef74ec0481e49cfd183e95e9629d59dd78ee63f424b3bdee6ed00a5c7daR79

If that's reasonable to return null in the mapper, I think we can remove the warning log and complete the future with the `null`. Let me do it in that way. 😉

client/java/src/main/java/com/linecorp/centraldogma/client/MappingWatcher.java
minwoox: Let's continue to discuss on there: https://github.com/line/centraldogma/pull/651#discussion_r774466273

client/java/src/main/java/com/linecorp/centraldogma/client/WatchFilesRequest.java
```diff
@@ -0,0 +1,121 @@
...
+        requireNonNull(watcherOptions, "watcherOptions");
+        final String proName = centralDogmaRepo.projectName();
+        final String repoName = centralDogmaRepo.repositoryName();
+        final DefaultWatcher<Revision> watcher = new DefaultWatcher<>(
+                blockingTaskExecutor, proName, repoName, pathPattern.get(),
+                lastKnownRevision ->
+                        centralDogmaRepo.centralDogma()
+                                        .watchRepository(proName, repoName, lastKnownRevision, pathPattern,
+                                                         timeoutMillis(), errorOnEntryNotFound())
+                                        .thenApply(revision -> new Latest<>(revision, revision)),
```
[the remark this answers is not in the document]
minwoox: Oops, thanks! 😄

client/java/src/main/java/com/linecorp/centraldogma/client/DefaultWatcher.java
```diff
@@ -28,108 +28,74 @@
...
     private volatile Latest<T> latest;
+    @Nullable
     private volatile ScheduledFuture<?> currentScheduleFuture;
+    @Nullable
     private volatile CompletableFuture<?> currentWatchFuture;
 
-    protected AbstractWatcher(CentralDogma client, ScheduledExecutorService watchScheduler,
-                              String projectName, String repositoryName, String pathPattern) {
-        this.client = requireNonNull(client, "client");
+    DefaultWatcher(ScheduledExecutorService watchScheduler,
```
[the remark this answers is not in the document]
minwoox: That's a good suggestion. 👍 
So basically the file watcher and files watcher are created in `WatchRequest` and `WatchFilesRequest` respectively, but you want to move the creating logic into the static methods of this `DefaultWatcher`.

Well either way is fine to me. 😄  But I slightly prefer to leave it as it is because `DefaultWatcher` is pretty complex now so I want to keep it as simple as possible to reduce the cognitive load. In contrast, `WatchRequest` class is simple so there's not so much distraction.

Like I said, either way is fine so please let me know if you want it strongly, then I'll move. 😄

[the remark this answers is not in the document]
di-seo: I also think `DefaultWatcher` is pretty complex.
So it is not bad to create new instance in `WatchRequest` and `WatchFilesRequest`.
I was just worried that `watchFunction` looks like key function to `DefaultWatcher`  :)

client/java/src/main/java/com/linecorp/centraldogma/client/WatcherOptionsBuilder.java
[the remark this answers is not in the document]
ikhoon: > have Armeria dependency

Oops... 😅

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaRequestPreparation.java
ikhoon: Agreed.

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
```diff
@@ -231,14 +253,42 @@ default void watch(Consumer<? super T> listener, Executor executor) {
...
-     * @return A {@link Watcher} that is effectively filtering in a sense that,
-     *         its listeners are <b>not</b> notified when a change has no effect on the transformed value.
-     *         Furthermore, it does not need to be closed after use.
+     * Returns a {@link Watcher} that applies the {@link Function} for the {@link Latest#value()}.
+     * Unlike {@link #newChild(Function)}, when the returned {@link Watcher} is closed, this {@link Watcher} is
+     * also closed so you don't have to close this {@link Watcher}.
+     */
+    default <U> Watcher<U> map(Function<? super T, ? extends U> mapper) {
+        return map(mapper, watchScheduler());
+    }
```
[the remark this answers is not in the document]
minwoox: Had a chat with maintainers and it's hard for users to manage the life-cycle of a watcher if we provide the `map` method to `Watcher`.
Instead, we have decided to add `WatcherRequest` for building a watcher and add the `map` method to it so it's convenient to manage the life-cycle.

client/java/src/main/java/com/linecorp/centraldogma/client/DefaultWatcher.java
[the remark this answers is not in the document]
minwoox: Based on the comment, I revived `FileWatcher` and `FilesWatcher` to distinguish that. PTAL. 😄

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
```diff
@@ -32,11 +33,14 @@
...
 
+import com.linecorp.centraldogma.common.EntryNotFoundException;
+import com.linecorp.centraldogma.common.PathPattern;
 import com.linecorp.centraldogma.common.Query;
 import com.linecorp.centraldogma.common.Revision;
 
 /**
- * Watches the changes of a repository or a file.
+ * Watches the changes of a repository or a file. The {@link Watcher} must be closed via
+ * {@link Watcher#close()} after use.
```
ikhoon, line 43: How about using `com.google.errorprone.annotations.MustBeClosed`?
I don't think we need to handle it in this PR. Let me create an issue.

client/java/src/main/java/com/linecorp/centraldogma/client/WatchRequest.java
```diff
@@ -0,0 +1,86 @@
...
+    WatchRequest(CentralDogmaRepository centralDogmaRepo, Query<T> query) {
+        this.centralDogmaRepo = centralDogmaRepo;
+        this.query = query;
+    }
+
+    /**
+     * Sets the last known {@link Revision} to get notified if there's a change since the {@link Revision}.
+     * {@link Revision#HEAD} is used by default.
+     */
+    public WatchRequest from(Revision lastKnownRevision) {
```
ikhoon, line 47: `WatchRequest<T>`

client/java-armeria/src/test/java/com/linecorp/centraldogma/client/armeria/CentralDogmaRepositoryTest.java
```diff
@@ -0,0 +1,125 @@
...
+import com.linecorp.centraldogma.testing.junit.CentralDogmaExtension;
+
+class CentralDogmaRepositoryTest {
+
+    @RegisterExtension
+    static final CentralDogmaExtension centralDogma = new CentralDogmaExtension() {
+        @Override
+        protected void scaffold(CentralDogma client) {
+            client.createProject("foo").join();
+            client.createRepository("foo", "bar").join();
```
ikhoon, line 51: Now, it might be natual that `client.createRepository("foo", "bar")` returns a `CompletableFuture<CentralDogmaRepository>`, although it is a breaking change. 😅

client/java/src/main/java/com/linecorp/centraldogma/client/FileRequest.java
```diff
@@ -0,0 +1,51 @@
...
+        this.query = query;
+        this.centralDogmaRepo = centralDogmaRepo;
+    }
+
+    /**
+     * Queries a file at the specified revision and path with the {@link Query}.
+     *
+     * @return the {@link Entry} that is matched by the given {@link Query}
+     */
+    public CompletableFuture<Entry<T>> get(Revision revision) {
```
ikhoon, line 45: Global comment: Could we provide a no argument function where a `Revision` is taken as the last input so that users use `Revision.HEAD` by default?
```java
public CompletableFuture<Entry<T>> get() {
    return get(Revision.HEAD);
}
```

ikhoon (commented):
Mostly looks good.

jrhee17 (commented):
Overall looks good! Only got through half of the files 😅  Let me continue reviewing next week 🙏

client/java/src/main/java/com/linecorp/centraldogma/client/WatchRequest.java
```diff
@@ -0,0 +1,86 @@
...
+     * Waits for the file matched by the {@link Query} to be changed since the {@code lastKnownRevision}.
+     * If no changes were made within the {@link #timeoutMillis(long)}, the
+     * returned {@link CompletableFuture} will be completed with {@code null}.
+     *
+     * @return the {@link Entry} which contains the latest known {@link Query} result.
+     *         {@code null} if the file was not changed for {@link #timeoutMillis(long)} milliseconds
+     *         since the invocation of this method. {@link EntryNotFoundException} is raised if the
+     *         target does not exist.
+     */
+    public CompletableFuture<Entry<T>> await() {
```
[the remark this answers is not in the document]
minwoox: That makes sense. I'd like `start` most. What do you think? @ikhoon

common/src/main/java/com/linecorp/centraldogma/common/DefaultPathPattern.java
```diff
@@ -0,0 +1,110 @@
...
+    @Override
+    public String encoded() {
+        if (encoded != null) {
+            return encoded;
+        }
+        return encoded = encodePathPattern(patterns);
+    }
+
+    @VisibleForTesting
+    static String encodePathPattern(String pathPattern) {
```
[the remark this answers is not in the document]
minwoox: Oops, forgot to remove it. 😅

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java
```diff
@@ -164,20 +183,25 @@
...
-     *   <li>{@code "/foo/*.json"} - find all JSON files under the directory {@code /foo}</li>
-     *   <li><code>"/&#42;/foo.txt"</code> - find all files named {@code foo.txt} at the second depth level</li>
-     *   <li>{@code "*.json,/bar/*.txt"} - use comma to specify more than one pattern. A file will be matched
-     *                                     if <em>any</em> pattern matches.</li>
-     * </ul>
+     * Retrieves the files matched by the path pattern.
+     *
+     * @return a {@link Map} of file path and {@link Entry} pairs
+     *
+     * @deprecated Use {@link #getFiles(String, String, Revision, PathPattern)}.
```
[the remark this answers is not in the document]
minwoox: Let me just add the usage of fluent APIs. 😄 Thanks!

client/java/src/main/java/com/linecorp/centraldogma/client/WatchRequest.java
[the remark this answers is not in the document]
ikhoon: +1 for `start`. 😄

client/java/src/main/java/com/linecorp/centraldogma/client/FileRequest.java
ikhoon: > I didn't intentionally add methods that take from and to revisions though.

Makes sense.

minwoox: Removed `from(revision)` from `WatchRequest` and added `start(revision)` and `start()` to align with other methods. 😉

minwoox:
@ikhoon and @jrhee17 All addressed, PTAL. 🙇‍♂️

client/java/src/main/java/com/linecorp/centraldogma/client/WatcherRequest.java
```diff
@@ -0,0 +1,188 @@
...
+        this.pathPattern = pathPattern;
+        this.blockingTaskExecutor = blockingTaskExecutor;
+        executor = blockingTaskExecutor;
+    }
+
+    /**
+     * Sets the {@link Function} to apply to the result of a watch request.
+     */
+    @SuppressWarnings("unchecked")
+    public <U> WatcherRequest<U> map(Function<? super T, ? extends U> mapper) {
```
jrhee17, line 83: Is there no need to validate nullness?

```
requireNonNull(mapper, "mapper");
```

client/java/src/main/java/com/linecorp/centraldogma/client/WatcherRequest.java
```diff
@@ -0,0 +1,188 @@
...
+            this.mapper = (Function<Object, ? extends T>) this.mapper.andThen(mapper);
+        }
+        return (WatcherRequest<U>) this;
+    }
+
+    /**
+     * Sets the {@link Executor} to execute the {@link #map(Function)}.
+     */
+    public WatcherRequest<T> mapperExecutor(Executor executor) {
+        this.executor = executor;
```
jrhee17, line 96: Is there no need to validate nullness?
```
requireNonNull(executor, "executor");
```

jrhee17 (approved):
Left a minor question, but overall looks good to me! Thank you @minwoox 🙇 👍 🙇

client/java/src/main/java/com/linecorp/centraldogma/client/MappingWatcher.java
```diff
@@ -0,0 +1,182 @@
...
+            if (oldLatest != null && oldLatest.value() == mappedValue) {
+                return;
+            }
+
+            // mappedValue can be nullable which is fine.
+            final Latest<U> newLatest = new Latest<>(revision, mappedValue);
+            mappedLatest = newLatest;
+            notifyListeners(newLatest);
+            if (!initialValueFuture.isDone()) {
+                initialValueFuture.complete(newLatest);
```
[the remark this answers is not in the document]
minwoox: > If upstream notify fails for any reason, I think it is possible that initialValueFuture isn't completed. (exception while mapping, executor rejection when notifying, etc..)

I think we can use `try catch` for those cases. I forgot to catch the exception from the `mapper` but I intentionally remove the `RejectedExecutionException` cases because we don't do that in other code and I think it's really rare and only happens when a server shuts down. (If the watcher is closed before the server shuts down, that would not happen.)
Should we also catch the `RejectedExecutionException`?

ikhoon (approved):
Good shot!

client/java/src/main/java/com/linecorp/centraldogma/client/MappingWatcher.java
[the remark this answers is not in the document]
jrhee17: >  I think it's really rare and only happens when a server shuts down

I think it's fine to leave unhandled 😅  the previous behavior was to just close the watcher which was unexpected for me anyways.

> If the watcher is closed before the server shuts down, that would not happen.)

I do think it can still happen in other cases since the executor can be user provided. i.e. if the user intentionally passes an executor with queueSize = 1, maxPoolSize = 1 and adds a bunch of watchers... or the executor could be filled due to a different code path.
Having said this, this example is probably unrealistic + we probably don't have to worry about it 😄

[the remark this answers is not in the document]
minwoox: > Having said this, this example is probably unrealistic + we probably don't have to worry about it

That's what I have thought. 😄 
It really makes code dirty and we don't do the same approach on our codebase including Armeria. Let's get back to this later if we have a problem with it. 😄

it/src/test/java/com/linecorp/centraldogma/it/CacheTest.java
```diff
@@ -105,20 +108,22 @@ void history(ClientType clientType, TestInfo testInfo) {
         client.createProject(project).join();
         client.createRepository(project, REPO_FOO).join();
 
-        final PushResult res1 = client.push(project, REPO_FOO, HEAD, "Add a file",
-                                            Change.ofTextUpsert("/foo.txt", "bar")).join();
+        final PushResult res1 = client.forRepo(project, REPO_FOO)
+                                      .commit("Add a file", Change.ofTextUpsert("/foo.txt", "bar"))
+                                      .push(HEAD).join();
```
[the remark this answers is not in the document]
minwoox: Thanks! Found more usages and change them. 😄

minwoox:
Thanks, reviewers for the valuable comments. 🙇‍♂️

PR #655, opened 2021-12-07
Add JSON5 support

### LINKED ISSUE #538: Support JSON5
https://json5.org/

JSON5 is a superset of JSON which adds some nice-to-haves to JSON, such as comments, trailing commas in arrays, etc. Will be useful for those who write configuration in JSON.

Related: #151

ks-yim:
#### Motivation:
* #538 

#### Modifications:
* `Server & ArmeriaCentralDogma` client JSON5 support
  - `getFile()` as `Entry<JsonNode>` but can still get the raw content via `#contentAsText()`
  - `watchFile()` as `Entry<JsonNode>` but can still get the raw content via `#contentAsText()`
  - `getFile()` & `watchFile()` with `JSON_PATH`
  - `mergeFiles()`

* Update Admin UI's ace editor version to `1.4.13`
  - To support JSON5 syntax highlighting

#### Results:
* Closes #538


minwoox:
This looks awesome. 😆

client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogma.java
```diff
@@ -1092,8 +1093,17 @@ private static String toString(AggregatedHttpResponse res) {
         return unsafeCast(Entry.ofText(revision, entryPath, content0));
     }
 
-    private static <T> Entry<T> entryAsJson(Revision revision, JsonNode node, String entryPath) {
-        return unsafeCast(Entry.ofJson(revision, entryPath, getField(node, "content")));
+    private static <T> Entry<T> entryAsJson(Revision revision, JsonNode node, String entryPath,
+                                            EntryType type) {
+        final JsonNode content = getField(node, "content");
+        if (type == EntryType.JSON5) {
```
minwoox, line 1099: This method isn't called with `EntryType.JSON5`?

I guess users want to retrieve the JSON5 entry as a `JsonNode` or text with comments. So it will be
```java
Entry<JsonNode> entry = ...;
JsonNode content = entry.content(); // 주석 제거됨
String contextText = entry.contentAsText(); // 주석 보임
```
So I guess we have to use `EntryType` for JSON5 with `JsonNode` even though the bytes of String are transported.

common/src/main/java/com/linecorp/centraldogma/internal/Jackson.java
```diff
@@ -74,8 +77,16 @@
         compactMapper.disable(SerializationFeature.INDENT_OUTPUT);
         prettyMapper.enable(SerializationFeature.INDENT_OUTPUT);
         // Sort the attributes when serialized via the mapper.
-        compactMapper.enable(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS);
-        prettyMapper.enable(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS);
+        enable(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS);
+
+        // Enable JSON5 features.
+        enable(JsonReadFeature.ALLOW_SINGLE_QUOTES,
```
minwoox, line 83: I think we shouldn't use these options for `JSON`.
We need to add `JacksonJson5` as we do in https://github.com/line/centraldogma/pull/616/files#diff-202d69d7f9ebcf08990eda99e556947155f88acc967601e20a3df8361c4aef6bR28

minwoox (commented):
Sorry for the late review. 😅 
Left some suggestions. 😉

common/src/main/java/com/linecorp/centraldogma/internal/Jackson.java
ks-yim: Indeed, I tried the YAML PR's approach(`Jackson.ofJson()`, `Jackson.ofJson5()`) which I think a bit verbose.
We actually need only a few methods of what `Jackson` currently have to support `YAML` and `JSON5` while `Jackson` class is currently being used as a general utility for json ser/deser.

So what about having separate utility classes for `YAML` and `JSON5` and use it instead?
We can then save lots of meaningless code modifications.

client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogma.java
ks-yim: @minwoox
For `Entry` of `EntryType.JSON5` to give `JsonNode` with `content()` method, it should be defined as `JSON5(JsonNode.class)` which will make it committed as PLAIN JSON.

So more viable option here would be..:
```java
Entry<String> entry = ...;
JsonNode contentAsJson = entry.contentAsJsonNode(); // without comments
String content = entry.content(); // with comments
```

And type parameter of `Entry<T>` should be in line with the one for `Query<T>` so I think we can give following entries for JSON5 files depending on the query type...:
* `Query<JsonNode>` - `Query.ofJson()` or `Query.ofJsonPath()` -> `Entry<JsonNode>`
* `Query<String>` - `Query.ofText()` -> `Entry<String>` - `Entry.ofJson5()`
* `Query<?>` - `Query.identity()` -> `Entry<String>` - `Entry.ofJson5()`

Hope to get some advice for this 😅..

common/src/main/java/com/linecorp/centraldogma/internal/Jackson.java
minwoox: > We actually need only a few methods of what Jackson currently have to support YAML and JSON5 while Jackson class is currently being used as a general utility for json ser/deser.

Hmm, could you tell me more about the "few methods" part, please?
Because IIRC, there are quite a lot of methods that take `String` and `byte[]`, which should exist in the shape of JSON5.
We should use different mappers for JSON/JSON5, otherwise, we are going to loosen the validation.

client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogma.java
minwoox: > JsonNode contentAsJson = entry.contentAsJsonNode(); // without comments

I'm reluctant to add this method to `Entry` because an `Entry` can be a text file.

> JSON5(JsonNode.class) which will make it committed as PLAIN JSON.

Can we specially treat it on the client/server side? I thought we can transport/commit in the form of String and just validate if it's really a JSON5 before transporting/commiting.

> Query<String> - Query.ofText() -> Entry<String> - Entry.ofJson5()

It's the same. Can we just do?:
`Query<JsonNode>` - `Query.ofJson5()` -> `Entry<JsonNode>` - `Entry.ofJson5()`

common/src/main/java/com/linecorp/centraldogma/internal/Jackson.java
ks-yim: > > Because IIRC, there are quite a lot of methods that take String and byte[], which should exist in the shape of JSON5.

Right, there're quite a lot of read methods but `readTree` method is the only method used for `JSON5` validation and `JsonNode` conversion in this PR.

client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogma.java
ks-yim: > I'm reluctant to add this method to Entry because an Entry can be a text file
`Entry` already has this method since it implements `ContentHolder` (actually, the method name is `contentAsJson()`... 😂)

minwoox: So basically, 
- We transmit/store the JSON5 as a plain String.
- But before transmitting/storing, validate the String if it's really in the form of JSON5.
- Use `Query<JsonNode>` and get `Entry<JsonNode>`.
  - In the JSON5 `Entry`, it has the String that contains comments and we can retrieve it via `entry.contentAsText()`.
  - We can retrieve the `JsonNode` via `entry.content()` which cannot have comments.
  
I could probably miss something, so let me know your opinion. 😄

minwoox: > Entry already has this method since it implements ContentHolder (actually, the method name is contentAsJson()... 😂

Oops, I missed that. 😅

ks-yim: > Can we specially treat it on the client/server-side? I thought we can transmit/commit in the form of String and just validate if it's really a JSON5 before transmitting/committing.

Okay, let me try this..!

common/src/main/java/com/linecorp/centraldogma/internal/Jackson.java
minwoox: Hmm, I might misunderstad something. 😅 
Let's see one of the methods:
```java
public static <T> T readValue(String data, Class<T> type) {
    try {
        return compactMapper.readValue(data, type);
    ...
}
```
Let's say that we set JSON5 configuration to the `compactMapper` and use it to read JSON/JSON5 data together.
When users read JSON data, which should throw a `JsonParseException` because it's invalid, it might not throw the exception because it's a valid JSON5 type. I think this is unexpected behavior.

Or did you mean specifying a different `mapper`? 🤔

ks-yim: > Or did you mean specifying a different mapper? 🤔

Yeah, since what I only needed was `readTree(...)` method with JSON5 capabilities, I thought we can leave `Jackson` class as it is and make another utility class for `JSON5`(with JSON5 enabled mapper) whose only method is `readTree(...)` and use it where we need to read `JSON5` string.

minwoox: Had a chat with @ks-yim and we agreed that there's no need to introduce `Jaskson.json5()`.
It's an internal class and there are not so many methods needed for JSON5 so introducing `Json5` will be enough. 😄

client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogma.java
minwoox: By the way, was there any reason that we wanted to introduce `EntryType.JSON5`?
While talking with other maintainers, we can probably retrieve the JSON5 entry without introducing new `EntryType.JSON5` 😅 :
```java
// Use the file extension to distinguish JSON/JSON5.
Entry<JsonNode> json5Entry = Entry.ofJson(..., "/foo.json5", ...);

// The file extension is also used for the query.
Query<JsonNode> json5Query = Query.ofJson("/foo.json5);
```

common/src/main/java/com/linecorp/centraldogma/common/Change.java
```diff
@@ -65,10 +66,20 @@
     static Change<String> ofTextUpsert(String path, String text) {
         requireNonNull(text, "text");
         validateFilePath(path, "path");
-        if (EntryType.guessFromPath(path) == EntryType.JSON) {
```
ikhoon, line 68: It seems not to make sense to use `ofTextUpsert` for not JSON but JSON5.
How about modifying `Change.ofJsonUpsert(String path, String jsonText)` to take JSON5?
```java
static Change<JsonNode> ofJsonUpsert(String path, String jsonText) {
  
    final JsonNode jsonNode = ...;
   
    return new DefaultChange<>(path, ChangeType.UPSERT_JSON, jsonNode, jsonText);
}
```
I think we can give `jsonText` a higher priority to serialize the data.

jrhee17 (commented):
The PR overall looks really nice! Left some random ideas/comments 😄 Looking forward to this feature a lot 👍

client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogma.java
ks-yim: @jrhee17 Thanks for the advice..! I am heading into the same direction.
But, here's one thing that I am not sure of.

Now that we need to send JSON5 string to the client, the shape of `EntryDto<T>` should also be changed.
Maybe we can also introduce `String contentAsText` property in there but it sounds like a bit redundant in most of cases.

Or we can just remove the type parameter `T` of `EntryDto` and just use `Object content` though we would lose type safety by doing so..

I am more inclined to the 2nd option since `EntryDto<?>` is being used all across the place.

server/src/test/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1Test.java
```diff
@@ -633,6 +671,72 @@ void editFileWithTextPatch() throws IOException {
                     .isEqualTo("text in some file.\n");
         }
 
+        @Test
+        void editJson5FileWithTextPatch() throws IOException {
```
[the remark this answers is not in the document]
ks-yim: @jrhee17 I was thinking of supporting both `APPLY_JSON_PATCH` and `APPLY_TEXT_PATCH` for JSON5 file but if updated via `APPLY_JSON_PATCH` it loses all `JSON5` features and degraded to ordinary JSON.
I am not sure if it is the right direction, though...

I will ping you again when I revise the PR and hope to get some advice from there..!

client/java-armeria/src/test/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogmaTest.java
```diff
@@ -57,33 +78,94 @@ void testEncodePathPattern() {
...
+        final JsonNode node = Jackson.readTree("{\"a\": \"bar\"}");
+        when(webClient.execute(ArgumentMatchers.<RequestHeaders>any())).thenReturn(
+                HttpResponse.ofJson(new MockEntryDto("/foo.json5", "JSON", node)));
+
+        final Entry<?> entry = getFile(client, Query.ofJsonPath("/foo.json5", "$.a"));
+        assertThat(entry.content()).isEqualTo(node);
     }
 
     @Test
-    void pushMirrorsJsonFileToMetaRepository() throws UnknownHostException {
```
ks-yim, line 76: Moved to `it/MetadataPushTest.java`.

client/java-armeria/src/test/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogmaTest.java
```diff
@@ -57,33 +78,94 @@ void testEncodePathPattern() {
     }
 
     @Test
-    void pushFileToMetaRepositoryShouldFail() throws UnknownHostException {
```
ks-yim, line 60: Moved to it/MetadataPushTest.java.

ks-yim:
@jrhee17 @ikhoon @minwoox I've re-written the PR as per you guys' advices, PTAL 🙏.

* Use `EntryType.JSON` for JSON5
* Use dedicated mapper for JSON5
* Give higher priority to `contentAsText` property when transmitting JSON5 `content` of `Entry` or `Change` over the wire.

Sorry that it's delayed. 😓

server/src/main/java/com/linecorp/centraldogma/server/internal/api/DtoConverter.java
```diff
@@ -58,29 +60,38 @@ public static RepositoryDto convert(Repository repository) {
...
-                                           String path, EntryType type) {
+    private static <T> EntryDto convert(Repository repository, Revision revision,
+                                        String path, EntryType type) {
         return convert(repository, revision, path, type, null);
     }
 
-    private static <T> EntryDto<T> convert(Repository repository, Revision revision, String path,
-                                           EntryType type, @Nullable T content) {
+    private static <T> EntryDto convert(Repository repository, Revision revision, String path,
+                                        EntryType type, @Nullable T content) {
```
minwoox, line 87: We don't need `<T>` anymore throughout this class :
```suggestion
    private static EntryDto convert(Repository repository, Revision revision, String path,
                                    EntryType type, @Nullable Object content) {
```

server/src/main/java/com/linecorp/centraldogma/server/internal/api/DtoConverter.java
```diff
@@ -58,29 +60,38 @@ public static RepositoryDto convert(Repository repository) {
...
+        return convert(repository, revision, entry, withContent, null);
+    }
+
+    public static <T> EntryDto convert(Repository repository, Revision revision,
+                                       Entry<T> entry, boolean withContent, @Nullable QueryType queryType) {
         requireNonNull(entry, "entry");
         if (withContent && entry.hasContent()) {
-            return convert(repository, revision, entry.path(), entry.type(), entry.content());
+            // TODO(ks-yim): Not sure DtoConverter having business logic and it looks hacky to receive
+            //   queryType to handle JSON_PATH query for JSON5 files.
```
minwoox, line 73: I don't have a better solution. 😅 
Let's leave it as a known issue. 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/GitMirror.java
```diff
@@ -240,8 +239,8 @@ protected void mirrorRemoteToLocal(File workDir, CommandExecutor executor,
                     final byte[] content = reader.open(objectId).getBytes();
                     switch (EntryType.guessFromPath(localPath)) {
                         case JSON:
-                            final JsonNode jsonNode = Jackson.readTree(content);
-                            changes.putIfAbsent(localPath, Change.ofJsonUpsert(localPath, jsonNode));
+                            final String jsonText = new String(content, StandardCharsets.UTF_8);
+                            changes.putIfAbsent(localPath, Change.ofJsonUpsert(localPath, jsonText));
```
minwoox, line 243: How about also adding the variant method that takes `byte[]` so that we can remove String conversion?
```java
static Change<JsonNode> ofJsonUpsert(String path, byte[] jsonBytes) {...}
```

minwoox (commented):
You have beautifully implemented it. 😄 
Left some minor comments.

common/src/main/java/com/linecorp/centraldogma/common/Change.java
```diff
@@ -85,12 +88,16 @@
...
-            jsonNode = Jackson.readTree(jsonText);
-        } catch (IOException e) {
+            if (maybeJson5(path)) {
+                jsonNode = Json5.readTree(jsonText);
+                return new DefaultChange<>(path, ChangeType.UPSERT_JSON, jsonNode, jsonText);
+            } else {
+                jsonNode = Jackson.readTree(jsonText);
+                return new DefaultChange<>(path, ChangeType.UPSERT_JSON, jsonNode);
+            }
+        } catch (JsonParseException e) {
```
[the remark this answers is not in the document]
ks-yim: `IOException` is handled in `Jackson` and `Json` utils. They throw `IOError` upon `IOException`.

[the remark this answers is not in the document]
minwoox: Oops, that's correct. I thought `IOException` is rethrown somehow. 😅

it/src/test/java/com/linecorp/centraldogma/it/WatchTest.java
```diff
@@ -468,4 +477,34 @@ private static void revertTestFiles(ClientType clientType) {
...
+            // Make change to an irrelevant file.
+            client.push(dogma.project(), dogma.repo1(), Revision.HEAD, "Add foo.json",
+                        Change.ofJsonUpsert("/test/foo.json", "{}")).join();
+
+            assertThatThrownBy(() -> future.get(500, TimeUnit.MILLISECONDS))
+                    .isInstanceOf(TimeoutException.class);
+
+            // Make change to a relevant file.
+            final PushResult result = client.push(
+                    dogma.project(), dogma.repo1(), Revision.HEAD, "Edit test1.json5",
```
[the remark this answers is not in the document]
ks-yim: > I think we should specially treat the case not to notify the watchers.

This is a nice idea!!
But there seems to be no easy way to detect if the actual JSON content is changed before notifying watchers.

Maybe we can add the following code at `GitRepository.java:L1497 (GitRepository#notifyWatchers)` but it doesn't seem to be a right choice.
```java
case MODIFY: 
    if (isJson5(entry.getOldPath())) {
        try (ObjectReader reader = jGitRepository.newObjectReader()) {
            final JsonNode oldJsonNode =
                    Json5.readTree(reader.open(entry.getOldId().toObjectId()).getBytes());
            final JsonNode newJsonNode =
                    Json5.readTree(reader.open(entry.getNewId().toObjectId()).getBytes());
            if (oldJsonNode == newJsonNode) {
                break;
            }
        } catch (Exception e) {
            // Do nothing.
        }
    }
    /* fall-thru */
```

common/src/main/java/com/linecorp/centraldogma/common/Entry.java
```diff
@@ -69,7 +72,22 @@
...
+     *
+     * @param revision the revision of the JSON file
+     * @param path the path of the JSON file
+     * @param content the content of the JSON file in byte array.
+     *
+     * @throws JsonParseException if the {@code content} is not a valid JSON
+     */
+    public static Entry<JsonNode> ofJson(Revision revision, String path, byte[] content)
+            throws JsonParseException {
+        return ofJson(revision, path, new String(content, StandardCharsets.UTF_8));
```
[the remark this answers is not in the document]
ks-yim: I found a weird bug with Jackson's `ALLOW_SINGLE_QUOTES` features when it deserializes JSON5 from a byte-array source.

I've summarized the issue at `Json5#readTree(byte[])` method...:
```java
// String conversion is a workaround for Jackson bug with 'ALLOW_SINGLE_QUOTES' feature
// when deserializing from byte array source.
// If double quotes are used within a single quote string, it either raises JsonParseException or
// removes double quotes while deserializing.
// e.g. 'I can use "double quotes" here' deserialized into "I can use double quotes here"
//      'I can use double quotes "here"' raises JsonParseException.
```

So, we have to convert the `byte[]` into `String` anyways to workaround this issue.. 🤔.
I will report the issue to Jackson.

ks-yim:
@minwoox I've resolved most of your comments except the feature of not notifying watcher for any changes that does not affect the actual content of JSON5.

And I added a validation before committing a patch on JSON5 files (since they are updated via text patch..).

Plus, I found a weird bug with Jackson's `ALLOW_SINGLE_QUOTES` features when it deserializes JSON5 from a byte-array source.

I've summarized the issue at `Json5#readTree(byte[])` method...:
```java
// String conversion is a workaround for Jackson bug with 'ALLOW_SINGLE_QUOTES' feature
// when it deserializes JSON5 from byte array source.
// If double quotes are used within a single quoted string, it either raises JsonParseException or
// removes double quotes while deserializing.
// e.g. 'I can use "double quotes" here' deserialized into "I can use double quotes here"
//      'I can use double quotes "here"' raises JsonParseException.
```

I will report the issue to Jackson but I think we should always convert `byte[]` into `String` before deserializing JSON5 content as of now.

minwoox:
> And I added a validation before committing a patch on JSON5 files (since they are updated via text patch..).

👍 

> I will report the issue to Jackson but I think we should always convert byte[] into String before deserializing JSON5 content as of now.

Oops, I didn't notice that. 😅 Yeah, let's report it to the upstream and use String for now.

> except the feature of not notifying watchers for any changes that does not affect the actual content of JSON5.

Could you tell me more about it, please?

it/src/test/java/com/linecorp/centraldogma/it/WatchTest.java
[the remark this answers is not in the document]
minwoox: Oops, I missed this comment. 😄 
I thought we could just manipulate here:
https://github.com/line/centraldogma/blob/master/server/src/main/java/com/linecorp/centraldogma/server/storage/repository/RepositoryUtil.java#L182

If it gets too complicated, let's make a separate PR for that. 😄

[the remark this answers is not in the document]
ks-yim: > RepositoryUtil.java#L182

Seems to be a good place to put this check..!
I didn't know that it eventually gets the old content from the repo..!

Let me work on it in this PR.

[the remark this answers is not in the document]
ks-yim: And it seems we don't even need a change. It already works in a way you wanted it to be.
I've just added a test case to prove it..!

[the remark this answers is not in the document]
minwoox: I didn't know that it's going to work. That's awesome. 😄

server/src/main/java/com/linecorp/centraldogma/server/internal/api/DtoConverter.java
```diff
@@ -58,29 +60,38 @@ public static RepositoryDto convert(Repository repository) {
...
+
+    public static EntryDto convert(Repository repository, Revision revision,
+                                   Entry<?> entry, boolean withContent, @Nullable QueryType queryType) {
         requireNonNull(entry, "entry");
         if (withContent && entry.hasContent()) {
-            return convert(repository, revision, entry.path(), entry.type(), entry.content());
+            // TODO(ks-yim): Not sure of 'DtoConverter' having business logic and it looks hacky to receive
+            //   queryType to handle JSON_PATH query for JSON5 files.
+            return convert(repository, revision, entry.path(), entry.type(),
+                           isJson5(entry) && queryType != QueryType.JSON_PATH ? entry.contentAsText()
```
jrhee17, line 75: Q) Did we agree to move this functionality to api v2?
Since `type == JSON` and `content` is `json5`, I feel like this will break backwards-compatibility

jrhee17 (commented):
Overall looks good! 🙇 Left a couple more comments.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/DtoConverter.java
ks-yim: Haven't talked about V2 part, yet.

> Since type == JSON and content is json5, I feel like this will break backwards-compatibility

I don't think there will be a breaking change with `ArmeriaCentralDogma` client unless users are using deprecated `Query.identity(...)` because JSON5 can now only be queried via `Query.ofText(..)` and this PR still supports it.

The only breaking change that I can think of is that users won't be able to upsert JSON5 text with `Change.ofTextUpsert(...)` after this PR on anymore, though I doubt there will be many users who do this.

But I am not sure if this PR won't break admin UI and its APIs somehow.
Let me check this out.

common/src/main/java/com/linecorp/centraldogma/common/DefaultChange.java
```diff
@@ -106,6 +110,7 @@
 
         this.path = path;
         this.content = content;
+        this.contentAsText = contentAsText;
```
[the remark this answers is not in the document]
ks-yim: Hmm.. thanks for pointing this out. I think we should handle the case that `Change#ofJsonUpsert` is used to add a new JSON5 file(`ChangeType.ADD`) as it relies on `contentAsText()` which isn't serialized into replication log. 

No need to care about `Change#ofJsonUpsert` to modify the file content(`ChangeType.MODIFY`) since it internally converted into `Change.ofTextPatch`.

I think we should have `Change#contentAsText` in the serialized content to make it work (or a simple workaround would be internally converting `Change#ofJsonUpsert` to `Change#ofTextUpsert` like we did for `ChangeType.MODIFY`).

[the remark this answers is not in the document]
ks-yim: > I wonder if it's difficult to tweak behavior so that serde happens in the same way as we've been doing so far

The type of `content` in `Change#ofJsonUpsert` is `JsonNode` because of the decision we've made for JSON5 entry type (use `JsonNode` over `String`).

Maybe we can do the tweak that `Change` to a JSON5 file have `TextNode` as its `content` and do `if-else` in all places where change gets applied, but this also seems to me extremely unmanageable.

I agree that we should avoid unnecessary complexity, but either way(tweaking serde vs tweak `Change) is not very appealing 😅..

The 3rd(and the most simple) option would be converting `Change#ofJsonUpsert` request to `Change#ofTextUpsert` internally like we already did for `ChangeType.MODIFY`.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/DtoConverter.java
ks-yim: > I was thinking of the following scenario:

Oh, I've thought CD client's `getFile` API always accept `Query` but it seems not.
There could be a compatibility issue if a user uses the following API.

The return type of `Entry#content()` retrieved via this API is no longer `String` but `TextNode` for JSON5 files.
```java
    default CompletableFuture<Entry<?>> getFile(String projectName, String repositoryName,
                                                Revision revision, String path) {
        final CompletableFuture<Entry<?>> f = (CompletableFuture<Entry<?>>) (CompletableFuture<?>)
                getFile(projectName, repositoryName, revision, Query.of(QueryType.IDENTITY, path));
        return f;
    }
```

common/src/main/java/com/linecorp/centraldogma/common/DefaultChange.java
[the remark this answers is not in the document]
ks-yim: It turns out that the 3rd option will never work since the serializing and sending replication log happens earlier than we can do conversion.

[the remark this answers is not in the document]
ks-yim: Had a chat with @minwoox and @jrhee17 about this and I was advised to put denormalized JSON5 `Change` to replication logs to preserve its original content.

I tried this approach only to find there's no easy way to pass the original JSON5 content to replication log even with this. `NormalizingPushCommand#changes` gives the original JSON5 content in it, but once it gets JSON-serialized for replication log it loses its JSON5 content since `contentAsText` is ignored in the serialization process.

Meanwhile, it turned out that 3rd option that I mentioned above works just fine.
It works by normalizing JSON5 `Change#ofJsonUpsert` to `Change#ofTextUpsert` and then putting the normalized change to replication log (at ).

server/src/main/java/com/linecorp/centraldogma/server/internal/api/DtoConverter.java
ks-yim: Also had a chat about this and decided to go for `ContentServiceV2` which will be handled in another PR.

minwoox:
I think it's almost done, excellent job, @ks-yim!
https://github.com/line/centraldogma/runs/5271252784?check_suite_focus=true#step:5:1211
Could you check the test failures, please? 😄

ks-yim:
@minwoox It seems Jackson's `ALLOW_BACKSLASH_ESCAPING_ANY_CHARACTER` feature doesn't work well in `Windows`..? I should test it in my Windows laptop.

minwoox:
> @minwoox Thanks for the patience.

I really appreciate your all effort. Please, don't worry about it. 😄 
> It seems Jackson's ALLOW_BACKSLASH_ESCAPING_ANY_CHARACTER feature doesn't work well in Windows..? I should test it in my Windows laptop.

Oops, let me also take a look at it.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/DtoConverter.java
ikhoon: > Not sure of 'DtoConverter' having business logic and it looks hacky to receive queryType to handle JSON_PATH query for JSON5 files.

Agreed. `QueryType` and `isJson5()` don't match with DTO. How about taking a `Function` rather than adding a specific logic here?
``` java
public static EntryDto convert(Repository repository, Revision revision,
                               Entry<?> entry, boolean withContent, Function<Entry, ?> contentConverter) {
    return convert(repository, revision, entry.path(), entry.type(), contentCoverter.apply(entry));
}
```

client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogma.java
```diff
@@ -978,7 +979,9 @@ private static ArrayNode toJson(Iterable<? extends Change<?>> changes) {
             changeNode.put("path", c.path());
             changeNode.put("type", c.type().name());
             final Class<?> contentType = c.type().contentType();
-            if (contentType == JsonNode.class) {
+            if (isJson5(c.path())) {
```
ikhoon, line 982: How about adding the `FileType` mentioned to `Change` so that we check file type more statically?
```java
public interface Change<T> {

    FileType fileType() {
         ...
    }
}
```

jyterencekim:
👀 @ks-yim Thanks for your work. I was just wondering whether this is still in progress.

ks-yim:
@jyterencekim No 😭. Stopped working on this for now. You can take this issue if you want..

Just some notes tracing back an old memory..

This PR somewhat has achieved JSON5 support in CentralDogma, but is littered with hacky workarounds and improvisations. As @ikhoon has pointed out in his suggestion of adding `FileType`, a more structured approach is needed so that other config formats(`.yaml`, `.toml`, etc.) support can be added without pain later in the future, too.

Although JSON5 can be parsed as a normal JSON, parsed data can't be reversed back into the original JSON5 content so the data should be treated as `EntryType.TEXT` when it is patched & stored but should support ALL `EntryType.JSON` features such as..:
  - `getFile()` as `Entry<JsonNode>` but can still get the raw content via `#contentAsText()`
  - `watchFile()` as `Entry<JsonNode>` but can still get the raw content via `#contentAsText()`
  - `getFile()` & `watchFile()` with `JSON_PATH`
  - `mergeFiles()`

minwoox:
Let us continue to work on this after adding API v2. @ks-yim Sorry to burden you. You are free to go. 😆 

@jyterencekim Could you use a plain text file for JSON5 before we implement this? If you add the validation logic before pushing a file, I think you can use JSON5 without problems. 🤔

ks-yim:
@minwoox thanks you for all the valuable feedback you gave while I am working on this 😄.

Closing this PR in favor of API v2.

minwoox:
What I meant was I was going to work on this PR again after adding V2. 😓 Could you reopen this?

ks-yim:
@minwoox Oops, my mistake 🤣. Reopened the PR! Looking forward to the rework!

PR #656, opened 2021-12-10
Fix `updateStatus` doesn't update replication when read-only mode

### LINKED ISSUE #360: Cannot update status of writable/replicating
Changing the status to `false` by the following command:
```
curl -v -X PATCH http://your-centraldogma-domain/api/v1/status -H'authorization: bearer appToken-******' -H'Content-Type:application/json-patch+json' -d'[{"op":"replace","path":"/writable","value":false}, {"op":"replace","path":"/replicating","value":false}]'
```
Then, the following command doesn't work:
```
curl -v -X PATCH http://your-centraldogma-domain/api/v1/status -H'authorization: bearer appToken-******' -H'Content-Type:application/json-patch+json' -d'[{"op":"replace","path":"/writable","value":true}, {"op":"replace","path":"/replicating","value":true}]'
```

Also,`304 Not Modified` is returned when trying to change `/replicating` only.

ghkim3221:
### Motivation

 - #360

### Modification

 - Enable replication when read-only with not-replication mode and given `replicating` is `true`.
 - Disable replication when read-only with replication mode and given `replicating` is `false`.

### Result

 - Fix #360 (?)
 - Users can enable/disable replication when read-only mode.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/api/AdministrativeService.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/api/AdministrativeService.java
@@ -114,7 +114,7 @@ public final class AdministrativeService extends AbstractService {
         if (oldStatus.writable != writable) {
             executor().setWritable(writable);
             if (writable) {
-                logger.warn("Left read-only mode. replication: {}, replicating");
+                logger.warn("Left read-only mode. replication: {}", replicating);
             } else {
                 logger.warn("Entered read-only mode. replication: {}, replicating");
             }
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/api/AdministrativeService.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/api/AdministrativeService.java
@@ -116,7 +116,7 @@ public final class AdministrativeService extends AbstractService {
             if (writable) {
                 logger.warn("Left read-only mode. replication: {}", replicating);
             } else {
-                logger.warn("Entered read-only mode. replication: {}, replicating");
+                logger.warn("Entered read-only mode. replication: {}", replicating);
             }
         }
 
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/api/AdministrativeService.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/api/AdministrativeService.java
@@ -126,7 +126,7 @@ public final class AdministrativeService extends AbstractService {
                     if (cause != null) {
                         logger.warn("Failed to start the command executor:", cause);
                     } else {
-                        logger.info("Enabled replication");
+                        logger.info("Enabled replication. read-only: {}", !writable);
                     }
                     return status();
                 });
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/api/AdministrativeService.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/api/AdministrativeService.java
@@ -114,7 +114,7 @@ public final class AdministrativeService extends AbstractService {
         if (oldStatus.writable != writable) {
             executor().setWritable(writable);
             if (writable) {
-                logger.warn("Left read-only mode. replication: {}", replicating);
+                logger.warn("Left read-only mode.");
             } else {
                 logger.warn("Entered read-only mode. replication: {}", replicating);
             }
@@ -139,7 +139,6 @@ public final class AdministrativeService extends AbstractService {
                 }
                 return status();
             });
-            }
         }
 
         return CompletableFuture.completedFuture(status());
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/api/AdministrativeService.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/api/AdministrativeService.java
@@ -130,15 +130,15 @@ public final class AdministrativeService extends AbstractService {
                     }
                     return status();
                 });
-            } else {
-                return executor().stop().handle((unused, cause) -> {
-                    if (cause != null) {
-                        logger.warn("Failed to stop the command executor:", cause);
-                    } else {
-                        logger.info("Disabled replication");
-                    }
-                    return status();
-                });
+            }
+            return executor().stop().handle((unused, cause) -> {
+                if (cause != null) {
+                    logger.warn("Failed to stop the command executor:", cause);
+                } else {
+                    logger.info("Disabled replication");
+                }
+                return status();
+            });
             }
         }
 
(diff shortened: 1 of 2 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/api/AdministrativeService.java
```diff
@@ -124,6 +124,28 @@ public ServerStatus status() {
...
+        } else if (replicating) { // unreplicating -> replicating
+            logger.warn("Enabling replication with read-only mode");
+            return executor().start().handle((unused, cause) -> {
+                if (cause != null) {
+                    logger.warn("Failed to start the command executor:", cause);
+                } else {
+                    logger.info("Enabled replication");
+                }
+                return status();
+            });
```
ikhoon, line 148: Could we simplify this if-else chains? For example:
```java
if (oldStatus.writable != writable) {
     executor().setWritable(writable);
}
if (oldStatus.replicating != replicating) {
   if (replicating) {
        // log
        executor().start().handle(...);
   } else {
        // log
        executor().stop().handle(...);
   }
}
```

ikhoon: Otherwise, extract common code to template methods?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/AdministrativeService.java
```diff
@@ -124,6 +124,28 @@ public ServerStatus status() {
                 }
                 return status();
             });
+        } else if (oldStatus.replicating) {
```
minwoox, line 127: How about adding a comment that describes the status change?
```
//    writable       replicating      result
//----------------------------------------------------------------------- 
//    true -> true                    not_modified exception
//-----------------------------------------------------------------------
//    true -> false  true -> true     setWritable(false)
//                   true -> false    setWritable(false) & Stop executor.
//-----------------------------------------------------------------------
//    false -> true  true -> true     setWritable(true)
//                   false -> true    setWritable(true) & start executor.
//-----------------------------------------------------------------------
//    false -> false true -> false    Stop executor.
//                   false -> true    Start executor.
```
And I think if we implement the logic in that way we can use the early return style and also we are not going to miss any cases.
Currently, I think we missed the fourth case which is `false -> true |  true -> true  |   setWritable(true)`.
Also, it would be nice if we add:
```
if (oldStatus.writable == writable && oldStatus.replicating == replicating) {
    throw HttpStatusException.of(HttpStatus.NOT_MODIFIED);
}
```

server/src/main/java/com/linecorp/centraldogma/server/internal/api/AdministrativeService.java
ghkim3221: I added tailing else-if because of the more specific logging messages (such as _with replication enabled_ or with _read-only mode_). I agree that we need simplification. If we can give up the messages, let me simplify.

> extract common code to template methods?

May I ask what _template methods_ means?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/AdministrativeService.java
```diff
@@ -88,45 +88,61 @@ public ServerStatus status() {
...
+        if (oldStatus.writable == writable && oldStatus.replicating == replicating) {
+            throw HttpStatusException.of(HttpStatus.NOT_MODIFIED);
+        }
+
+        if (oldStatus.writable != writable) {
+            executor().setWritable(writable);
+            if (writable) {
+                logger.warn("Left read-only mode");
+            } else {
+                logger.warn("Entered read-only mode");
```
ikhoon, line 119:
```suggestion
                logger.warn("Left read-only mode. replication: {}, replicating");
            } else {
                logger.warn("Entered read-only mode. replication: {}, replicating");
```

ikhoon (approved):
Looks good to me! 💯

server/src/main/java/com/linecorp/centraldogma/server/internal/api/AdministrativeService.java
```diff
@@ -88,45 +88,61 @@ public ServerStatus status() {
...
                                              "'replicating' must be 'true' if 'writable' is 'true'.");
         }
+        if (oldStatus.writable == writable && oldStatus.replicating == replicating) {
+            throw HttpStatusException.of(HttpStatus.NOT_MODIFIED);
+        }
+
+        if (oldStatus.writable != writable) {
+            executor().setWritable(writable);
+            if (writable) {
+                logger.warn("Left read-only mode. replication: {}, replicating");
```
minwoox, line 117: nit:
```suggestion
                logger.warn("Left read-only mode. replication: {}", replicating);
```

server/src/main/java/com/linecorp/centraldogma/server/internal/api/AdministrativeService.java
```diff
@@ -88,45 +88,61 @@ public ServerStatus status() {
...
+            } else {
+                logger.warn("Entered read-only mode. replication: {}, replicating");
```
minwoox, line 119: nit:
```suggestion
                logger.warn("Entered read-only mode. replication: {}", replicating);
```

server/src/main/java/com/linecorp/centraldogma/server/internal/api/AdministrativeService.java
```diff
@@ -88,45 +88,61 @@ public ServerStatus status() {
...
+                });
+            } else {
+                return executor().stop().handle((unused, cause) -> {
+                    if (cause != null) {
+                        logger.warn("Failed to stop the command executor:", cause);
+                    } else {
+                        logger.info("Disabled replication");
+                    }
+                    return status();
+                });
```
minwoox, line 141: nit: we can return early.
```suggestion
            }
            return executor().stop().handle((unused, cause) -> {
                if (cause != null) {
                    logger.warn("Failed to stop the command executor:", cause);
                } else {
                    logger.info("Disabled replication");
                }
                return status();
            });
```

minwoox (approved):
This really looks awesome. 👍 
Thanks a lot!

server/src/main/java/com/linecorp/centraldogma/server/internal/api/AdministrativeService.java
ghkim3221: 😱

server/src/main/java/com/linecorp/centraldogma/server/internal/api/AdministrativeService.java
```diff
@@ -88,45 +88,60 @@ public ServerStatus status() {
...
-                        } else {
-                            logger.info("Entered read-only mode with replication disabled");
-                        }
-                        return status();
-                    });
-                }
+        if (oldStatus.writable != writable) {
+            executor().setWritable(writable);
+            if (writable) {
+                logger.warn("Left read-only mode.");
```
ghkim3221, line 117: Removed `replication` because always true in this case.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/AdministrativeService.java
ikhoon: 😱

minwoox:
Awesome! 🎉 🎉 🎉 
I can't wait to see your next PR. 😆

PR #657, opened 2021-12-20
Execute mapper only once when the child watcher gets notified

minwoox:
Motivation:
I found out while I implementing #651.
The child watcher created via `Watcher#newChild(mapper)` executes the specified mapper
every time when a listener is notified. If there are 1000 listeners, mapper is called 1000 times.
This could be a problem if the mapper has heavy logic.

Modifications:
- Execute mapper only once when the entry is changed.
- Add `Watcher.newWatcher(Function, Executor)` that executes the mapper using the executor.

Result:
- The mapper `Function` of a Child watcher is executed only once when the child watcher gets notified.

### DIFF
--- a/client/java/src/main/java/com/linecorp/centraldogma/client/TransformingWatcher.java
+++ b/client/java/src/main/java/com/linecorp/centraldogma/client/TransformingWatcher.java
@@ -18,91 +18,148 @@ package com.linecorp.centraldogma.client;
 import static com.google.common.base.MoreObjects.toStringHelper;
 import static java.util.Objects.requireNonNull;
 
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.Executor;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.function.BiConsumer;
 import java.util.function.Function;
 
 import javax.annotation.Nullable;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.collect.Maps;
+
 import com.linecorp.centraldogma.common.Revision;
 
 final class TransformingWatcher<T, U> implements Watcher<U> {
 
+    private static final Logger logger = LoggerFactory.getLogger(TransformingWatcher.class);
+
     private final Watcher<T> parent;
-    private final Function<T, U> transformer;
+    private final Function<? super T, ? extends U> mapper;
+    private final CompletableFuture<Latest<U>> initialValueFuture = new CompletableFuture<>();
+    private final List<Entry<BiConsumer<? super Revision, ? super U>, Executor>> updateListeners =
+            new CopyOnWriteArrayList<>();
 
     @Nullable
-    private volatile Latest<U> transformedLatest;
+    private volatile Latest<U> mappedLatest;
     private volatile boolean closed;
-    private volatile boolean firstWatch = true;
 
-    TransformingWatcher(Watcher<T> parent, Function<T, U> transformer) {
+    TransformingWatcher(Watcher<T> parent, Function<? super T, ? extends U> mapper, Executor mapperExecutor) {
         this.parent = parent;
-        this.transformer = transformer;
+        this.mapper = mapper;
+        parent.initialValueFuture().exceptionally(cause -> {
+            initialValueFuture.completeExceptionally(cause);
+            return null;
+        });
+        parent.watch((revision, value) -> {
+            if (closed) {
+                return;
+            }
+            mapperExecutor.execute(() -> {
+                final U mappedValue = mapper.apply(value);
+                final Latest<U> oldLatest = mappedLatest;
+                if (oldLatest != null && oldLatest.value() == mappedValue) {
+                    return;
+                }
+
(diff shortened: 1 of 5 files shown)


client/java/src/main/java/com/linecorp/centraldogma/client/TransformingWatcher.java
```diff
@@ -18,91 +18,148 @@
...
+        parent.initialValueFuture().exceptionally(cause -> {
+            initialValueFuture.completeExceptionally(cause);
+            return null;
+        });
+        parent.watch((revision, value) -> {
+            if (closed) {
+                return;
+            }
+            mapperExecutor.execute(() -> {
+                final U mappedValue = mapper.apply(value);
```
ikhoon, line 66: nit: `requireNonNull(mappedValue, "mapper.apply() returned null")`

client/java/src/main/java/com/linecorp/centraldogma/client/TransformingWatcher.java
```diff
@@ -18,91 +18,148 @@
...
+        if (mappedLatest != null) {
+            // There's a chance that listener.accept(...) is called twice for the same value
+            // if this watch method is called:
+            // - after "mappedLatest = newLatest;" is invoked.
+            // - and before notifyListener() is called.
+            // However, it's such a rare case and we usually call `watch` method after creating a Watcher,
+            // which means mappedLatest is probably not set yet, so we don't use a lock to guarantee
+            // the atomicity.
+            executor.execute(() -> {
+                listener.accept(mappedLatest.revision(), mappedLatest.value());
```
ikhoon, line 152: How about reading `mappedLatest` again? The value might be changed before starting the `Runnable` task.
```suggestion
                final Latest<U> mappedLatest0 = this.mappedLatest;
                listener.accept(mappedLatest0.revision(), mappedLatest0.value());
```

client/java/src/main/java/com/linecorp/centraldogma/client/TransformingWatcher.java
```diff
@@ -18,91 +18,148 @@
...
+        this.mapper = mapper;
+        parent.initialValueFuture().exceptionally(cause -> {
+            initialValueFuture.completeExceptionally(cause);
+            return null;
+        });
+        parent.watch((revision, value) -> {
+            if (closed) {
+                return;
+            }
+            mapperExecutor.execute(() -> {
```
ikhoon, line 65: Could we directly execute this if `mapperExecutor == parent.watchScheduler()`?

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
```diff
@@ -246,14 +252,22 @@ default void watch(Consumer<? super T> listener, Executor executor) {
...
+     * This {@link Watcher} must be closed regardless of closing the returned {@link Watcher}
+     * when it's not used anymore.
      */
-    default <U> Watcher<U> newChild(Function<T, U> transformer) {
-        requireNonNull(transformer, "transformer");
-        return new TransformingWatcher<>(this, transformer);
+    default <U> Watcher<U> newChild(Function<? super T, ? extends U> mapper, Executor mapperExecutor) {
+        requireNonNull(mapper, "mapper");
+        requireNonNull(mapperExecutor, "mapperExecutor");
+        return new TransformingWatcher<>(this, mapper, watchScheduler());
```
ikhoon, line 271:
```suggestion
        return new TransformingWatcher<>(this, mapper, mapperExecutor);
```

client/java/src/main/java/com/linecorp/centraldogma/internal/client/AbstractWatcher.java
```diff
@@ -134,7 +136,11 @@ protected AbstractWatcher(CentralDogma client, ScheduledExecutorService watchSch
 
         updateListeners = new CopyOnWriteArrayList<>();
         state = new AtomicReference<>(State.INIT);
-        initialValueFuture = new CompletableFuture<>();
+    }
+
+    @Override
+    public ScheduledExecutorService watchScheduler() {
```
ikhoon, line 142:
```suggestion
    public final ScheduledExecutorService watchScheduler() {
```

client/java/src/main/java/com/linecorp/centraldogma/internal/client/AbstractWatcher.java
```diff
@@ -134,7 +136,11 @@ protected AbstractWatcher(CentralDogma client, ScheduledExecutorService watchSch
 
         updateListeners = new CopyOnWriteArrayList<>();
         state = new AtomicReference<>(State.INIT);
```
ikhoon, line 138: nit: Inline these fields together?

ikhoon (commented):
Nice catch! Left some comments.

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
minwoox: 😱

minwoox (commented):
@ikhoon All addressed. PTAL. 😄

ikhoon (approved):
Thanks, @minwoox!

minwoox:
Thanks, @ikhoon  for reviewing. 😉
Let me merge this and continue to work on #651

PR #662, opened 2022-01-05
Change to return CentralDogmaRepository from client.createRepository

minwoox:
Motivation:
We introduced `CentralDogmaRepository` class in #651 and it would be nice if we return it from
`centralDogma.createRepository(...)` so users can chain the next call.
```java
dogma.createRepository("foo", "bar)
     .join()
     .commit(...)
     .push();
```

Modifications:
- Change to return `CentralDogmaRepository` from `centralDogma.createRepository(...)` and `centralDogma.unremoveRepository(...).
- Fix flaky test in `WatchTest`.`

Result:
- (Breaking) `createRepository` and `unremoveRepository` from `CentralDogma` now returns `CentralDogmaRepository`.


ikhoon (approved):
Thanks!

jrhee17 (approved):
Looks great! Thank you! 🙇 👍 💯

minwoox:
Thanks reviewers! 😄

PR #664, opened 2022-01-06
Add maxCommits to CentralDogma.getHistory()

minwoox:
Motivation:
It would be nice if we can specify the number of maximum commits for history call
as we have in Go client.

Modification:
- Add `maxCommit` parameter to `CentralDogma.getHistory()` and `maxCommit` setter to `HistoryRequest`.

Result:
- You can now limit the number of commits when retrieving the commit history.
  ```java
  centralDogma.forRepo("foo", "bar")
              .history()
              .maxCommits(5)
              .get(Revision.HEAD, Revision.INIT);
  ```

### DIFF
--- a/client/java-armeria-legacy/src/main/java/com/linecorp/centraldogma/client/armeria/legacy/LegacyCentralDogma.java
+++ b/client/java-armeria-legacy/src/main/java/com/linecorp/centraldogma/client/armeria/legacy/LegacyCentralDogma.java
@@ -328,7 +328,7 @@ final class LegacyCentralDogma extends AbstractCentralDogma {
                                                       Revision to,
                                                       PathPattern pathPattern,
                                                       int maxCommits) {
-        // maxCommits is not supported for the legacy client.
+        checkArgument(maxCommits == 0, "maxCommits is not supported in LegacyCentralDogma.");
         validateProjectAndRepositoryName(projectName, repositoryName);
         requireNonNull(from, "from");
         requireNonNull(to, "to");
@@ -456,7 +456,7 @@ final class LegacyCentralDogma extends AbstractCentralDogma {
                                                        PathPattern pathPattern,
                                                        long timeoutMillis,
                                                        boolean errorOnEntryNotFound) {
-        // Legacy client does not support 'errorOnEntryNotFound'
+        checkArgument(!errorOnEntryNotFound, "errorOnEntryNotFound is not supported in LegacyCentralDogma.");
         validateProjectAndRepositoryName(projectName, repositoryName);
         requireNonNull(lastKnownRevision, "lastKnownRevision");
         requireNonNull(pathPattern, "pathPattern");
@@ -479,7 +479,7 @@ final class LegacyCentralDogma extends AbstractCentralDogma {
     public <T> CompletableFuture<Entry<T>> watchFile(String projectName, String repositoryName,
                                                      Revision lastKnownRevision, Query<T> query,
                                                      long timeoutMillis, boolean errorOnEntryNotFound) {
-        // Legacy client does not support 'errorOnEntryNotFound'
+        checkArgument(!errorOnEntryNotFound, "errorOnEntryNotFound is not supported in LegacyCentralDogma.");
         validateProjectAndRepositoryName(projectName, repositoryName);
         requireNonNull(lastKnownRevision, "lastKnownRevision");
         requireNonNull(query, "query");
--- a/client/java-armeria-legacy/src/main/java/com/linecorp/centraldogma/client/armeria/legacy/LegacyCentralDogma.java
+++ b/client/java-armeria-legacy/src/main/java/com/linecorp/centraldogma/client/armeria/legacy/LegacyCentralDogma.java
@@ -326,7 +326,9 @@ final class LegacyCentralDogma extends AbstractCentralDogma {
                                                       String repositoryName,
                                                       Revision from,
                                                       Revision to,
-                                                      PathPattern pathPattern) {
+                                                      PathPattern pathPattern,
+                                                      int maxCommits) {
+        // maxCommits is not supported for the legacy client.
         validateProjectAndRepositoryName(projectName, repositoryName);
         requireNonNull(from, "from");
         requireNonNull(to, "to");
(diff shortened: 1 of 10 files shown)


client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java
```diff
@@ -343,7 +364,8 @@
      *         {@link PathPattern} in the specified repository
      */
     CompletableFuture<List<Commit>> getHistory(
-            String projectName, String repositoryName, Revision from, Revision to, PathPattern pathPattern);
```
jrhee17, line 346: I just realized with the current pattern, we will need to make a breaking change every time a new parameter is added 🤔 
I wonder if we should add `@UnstableApi` to methods of this sort in this class

client/java/src/main/java/com/linecorp/centraldogma/client/HistoryRequest.java
```diff
@@ -23,21 +24,36 @@
...
+    private int maxCommits;
 
     HistoryRequest(CentralDogmaRepository centralDogmaRepo, PathPattern pathPattern) {
         this.centralDogmaRepo = centralDogmaRepo;
         this.pathPattern = pathPattern;
     }
 
+    /**
+     * Sets the maximum number of commits to retrieve. {@code 0} is used by default which means retrieving
+     * all commits. The number of retrieved commits can't be greater than
```
jrhee17, line 46:
```suggestion
     * Sets the maximum number of commits to retrieve. {@code 0} is used by default which means to retrieve
     * all commits. The number of retrieved commits can't be greater than
```

jrhee17 (approved):
Thank you also for the cleanups!  🙇 👍 💯

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java
ikhoon: I am also worried about that. I think we have to design a new API to add new options without deprecation or overloaded methods.
How about designing a new API with command and interpreter pattern. That might be pretty similar to the server-side command pattern.
```java

public interface CentralDogma {
    <T> CompletableFuture<T> execute(CentralDogmaRequest<T> request);
}

public interface CentralDogmaRequest<T> {
}

public class CreateRepository implements CentralDogmaRequest<CentralDogmaRepository> {
    private String projectName;
    private String repositoryName;
    
    // constructor and getters
    ...
}

public class HistoryRequest implements CentralDogmaRequest<List<Commit>> {
    ...

    public CompletableFuture<List<Commit>> get(Revision from, Revision to) {
        this.from = from;
        this.to = to;

        return centralDogmaRepo.centralDogma().execute(this);
    }
}
```

minwoox: That's a good suggestion. 👍 Let me use the command pattern. 😄

minwoox: Had a chat with @ikhoon and @jrhee17, and we decide to keep the current style because `CentralDogma` client plays a role of API spec.
However, we might reconsider thie decision later if we need to add more parameter frequently or we add Armeria dependency to `centraldogma-client` module. #673

client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogma.java
```diff
@@ -143,9 +144,8 @@
 
     @Override
     public CompletableFuture<Void> createProject(String projectName) {
+        validateProjectName(projectName);
         try {
-            validateProjectName(projectName);
```
ikhoon, line 147: I think Trustin intentionally wrapped `validateProjectName(projectName)` with `try...catch` block 
so that users handle an error only in one place, a callback of `CompletableFuture`.
Question: Which one is better for users' experience?

ikhoon (approved):
Thanks, @minwoox!

client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogma.java
minwoox: We throw an exception directly if the input parameters are invalid without wrapping it even in the method that returns a future throughout our APIs including Armeria.
So I thought we have to throw an exception here as well to ensure consistency. 😄

client/java-armeria-legacy/src/main/java/com/linecorp/centraldogma/client/armeria/legacy/LegacyCentralDogma.java
```diff
@@ -326,7 +326,9 @@
                                                       String repositoryName,
                                                       Revision from,
                                                       Revision to,
-                                                      PathPattern pathPattern) {
+                                                      PathPattern pathPattern,
+                                                      int maxCommits) {
```
jrhee17, line 330: nit; what do you think of explicitly throwing an exception if `maxCommits` is specified?

```
checkArgument(maxCommits == 0, "maxCommits is not supported in LegacyCentralDogma");
```

client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogma.java
jrhee17: Maybe just me, but I think validating and throwing as quickly as possible is better for user experience.

jrhee17 (approved):
Left some nits, but still looks good to me

client/java-armeria-legacy/src/main/java/com/linecorp/centraldogma/client/armeria/legacy/LegacyCentralDogma.java
minwoox: Thanks, added. 😄

client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogma.java
minwoox: @ikhoon If it's okay with you, let me merge this PR. 😉

minwoox:
Thanks, reviewers. 😉

PR #667, opened 2022-01-27
Add a workaround for `EmptyEndpointException` with `HealthCheckedEndpoint`

ikhoon:
Motivation:

Even if an `EndopintGroup` resolves the initial endpoints,
EndpointSelector could continue to return a null endpoint.
See https://github.com/line/armeria/pull/3978 for the details.
The bug has been fixed in Armeria but some users who use Armeria 1.13.4
could see `EmptyEndpointException` and difficult to apply a workaround.
Because a `CentralDogma` client creates a `HealthCheckedEndpoint`
internally that users can't access.

Modifications:

- Wait for the initial endpoints before returning a new client.

Result:

- You no longer see an `EmptyEndpointException` with your `CentralDogma` client.


jrhee17 (approved):
Thanks for the quick fix @ikhoon 👍 🙇 🏎️

ikhoon:
There are new features in the master branch. Let me reopen a PR against 0.53.0 branch.

PR #668, opened 2022-01-27
Add a workaround for EmptyEndpointException with `HealthCheckedEndpoint`

ikhoon:
Motivation:

Even if an `EndopintGroup` resolves the initial endpoints,
EndpointSelector could continue to return a null endpoint.
See https://github.com/line/armeria/pull/3978 the details.
The bug has been fixed in Armeria but some users whou use Armeria 1.13.4
could see `EmptyEndpointException` and difficult to apply a workaround.
Because a `HealthCheckedEndpoint` is created a `CentralDogma` client
internally.

Modifications:

- Wait for the initial endpoints before returns a new client.

Result:

- You no longer see `EmptyEndpointException` with `CentralDogma` client.


ikhoon:
The changes was reviewed at https://github.com/line/centraldogma/pull/667

ikhoon:
Let me merge this PR if the CI builds pass.

jrhee17 (approved):
Seems like the test failure isn't related to this PR. Looks good 👍

PR #674, opened 2022-02-15
Let `AbstractArmeriaCentralDogmaBuilder` retry health checks by default

jrhee17:


client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/AbstractArmeriaCentralDogmaBuilder.java
```diff
@@ -52,10 +56,18 @@
         extends AbstractCentralDogmaBuilder<B> {
 
     private static final int DEFAULT_HEALTH_CHECK_INTERVAL_MILLIS = 15000;
+    private static final ClientOptions DEFAULT_HEALTH_CHECK_CLIENT_OPTIONS =
+            ClientOptions.builder()
+                         .decorator(RetryingClient.newDecorator(
+                                 RetryConfig.builder(RetryRule.onUnprocessed())
+                                            .maxTotalAttempts(3)
```
ikhoon, line 63: I'm not sure what is a sensible default for retry. 🤔
How about setting `maxTotalAttempts` to 1 so that a client removes unhealthy endpoints quickly?

jrhee17: Note: after discussion with @minwoox @ikhoon we decided to try using armeria's default `retryInterval`.

We believe at the time of writing the default health check interval (15s) is a bug.
This is because pre-`centraldogma-0.53.1` the constructor waited only 10s, not allowing for any health check retries on failure.

client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/AbstractArmeriaCentralDogmaBuilder.java
```diff
@@ -164,15 +163,18 @@ protected final EndpointGroup endpointGroup() throws UnknownHostException {
...
-            return HealthCheckedEndpointGroup.builder(group, HttpApiV1Constants.HEALTH_CHECK_PATH)
-                                             .clientFactory(clientFactory)
-                                             .protocol(isUseTls() ? SessionProtocol.HTTPS
-                                                                  : SessionProtocol.HTTP)
-                                             .retryInterval(healthCheckInterval)
-                                             .build();
-        } else {
+        if (healthCheckInterval != null && healthCheckInterval.isZero()) {
             return group;
+        } else {
```
minwoox, line 168: nit: we could remove `else` block and return early.

client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/AbstractArmeriaCentralDogmaBuilder.java
```diff
@@ -164,15 +163,18 @@ protected final EndpointGroup endpointGroup() throws UnknownHostException {
...
+            final HealthCheckedEndpointGroupBuilder healthCheckedEndpointGroupBuilder =
```
minwoox, line 169: This is not your fault but how about also applying the `HealthCheckedEndpointGroup` which is created at line 138?

minwoox (commented):
Thanks! Left some minor suggestions. 😄

minwoox:
Thanks, @jrhee17!

client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/AbstractArmeriaCentralDogmaBuilder.java
ikhoon: Makes sense.

ikhoon (approved):
Thanks, @jrhee17!

PR #680, opened 2022-02-23
Fix to set the author of a repository correctly

minwoox:
Motivation:
The author of a repository is not set correctly to metadata when
`MetadataService.fetchMetadata()` is called before the metadata information of a repository is added.

Modification:
- Use the author of the repository to when adding the metadata of the repository.

Result:
- The repository author is now correctly set.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java
@@ -133,7 +133,8 @@ public class MetadataService {
                 }
 
                 logger.warn("Adding missing repository metadata: {}/{}", projectName, repo);
-                final CompletableFuture<Revision> addRepoFuture = addRepo(Author.SYSTEM, projectName, repo);
+                final Author author = projectManager.get(projectName).repos().get(repo).author();
+                final CompletableFuture<Revision> addRepoFuture = addRepo(author, projectName, repo);
                 addRepoFuture.handle((revision, cause) -> {
                     if (cause != null) {
                         future.completeExceptionally(cause);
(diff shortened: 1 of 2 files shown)


ikhoon (approved):
Thanks!

server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java
```diff
@@ -133,7 +133,8 @@ public MetadataService(ProjectManager projectManager, CommandExecutor executor)
                 }
 
                 logger.warn("Adding missing repository metadata: {}/{}", projectName, repo);
-                final CompletableFuture<Revision> addRepoFuture = addRepo(Author.SYSTEM, projectName, repo);
+                final Author author = projectManager.get(projectName).repos().get(repo).author();
```
jrhee17, line 136: Q) Doesn't `author()` call `history(..).join()` which is a blocking call?

jrhee17 (commented):
Left a post-review 😅

server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java
jrhee17: Seems like we use `CachingRepository` which caches the `author()` value. Feel free to ignore the above comment 😅

minwoox: If `CachingRepository` is applied, it doesn't do blocking call because it gets the author from the first commit. 
However, if the cache is not applied, yeah it does the blocking call.
Let me use the same approach in the `GitRepository`. Thanks!

PR #682, opened 2022-02-23
Add Author as a field in Repository

minwoox:
Motivation:
`Repository.author()` and `Repository.creationTimeMillis()` have a blocking call in it.
We can add `author` and `creationTimeMillis` as member variables and assign them in the constructor to avoid the blocking call.

Modifications:
- Add `author` and `creationTimeMillis` as member variables
- Change `Repository.author()` and `Repository.creationTimeMillis()` methods not to have implementation

Result:
- `Repository.author()` and `Repository.creationTimeMillis()` do not block anymore.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
@@ -164,7 +164,10 @@ class GitRepository implements Repository {
     private final ReadWriteLock rwLock = new ReentrantReadWriteLock();
     private final Project parent;
     private final Executor repositoryWorker;
+    private final long creationTimeMillis;
+    private final Author author;
     @VisibleForTesting
+    @Nullable
     final RepositoryCache cache;
     private final String name;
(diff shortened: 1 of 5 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
```diff
@@ -315,6 +318,9 @@
                 commitIdDatabase.rebuild(jGitRepository);
                 assert headRevision.equals(commitIdDatabase.headRevision());
             }
+            final Commit initialCommit = blockingHistory(Revision.INIT, Revision.INIT, ALL_PATH, 1).get(0);
```
ikhoon, line 321: Could we also update `CachingRepository` to delegate `author` and `creationTimeMillis`?

minwoox: Thanks, fixed. 😉

ikhoon (approved):
Thanks, @minwoox!

minwoox:
Thanks, @ikhoon!

PR #692, opened 2022-05-11
Allow configuring the initialization timeout of `CentralDogma` client for Spring integration.

### LINKED ISSUE #691: Wait for the initial endpoints of a `CentralDogma` client for Spring integrations
A `CentralDogma` client asynchronously fetches the initial endpoints using `DynamicEndpointGroup`.
https://github.com/line/centraldogma/blob/a74df4688da85782da150413724b27b572c28c52/client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/AbstractArmeriaCentralDogmaBuilder.java#L140
https://github.com/line/centraldogma/blob/a74df4688da85782da150413724b27b572c28c52/client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/AbstractArmeriaCentralDogmaBuilder.java#L163
The created `CentralDogma` client can't get/watch stored documentation from Central Dogma replicas if the endpoints are not resolved.
If `CentralDogma.whenReadyEndpoint().get(..)` is not called, `CentralDogma` waits 3.2 seconds for endpoints to be completed.
https://github.com/line/armeria/blob/790a2db01a8514b591bde90932ab18f63693829e/core/src/main/java/com/linecorp/armeria/client/DefaultClientRequestContext.java#L320
3.2 seconds is the default connection timeout for an Armeria client.
While a server is starting, it could take longer than 3.2 seconds because the server is busy initializing the application context.

A `CentralDogma` bean is a managed instance provided by `centraldogma-client-spring-boot-autoconfigure`.
So the `CentralDogma` bean is created in the main thread when a Spring application context is initialized. 
We can safely block the current thread with `CentralDogma.whenReadyEnpoint().get(...)`.
For example:
```java
class CentralDogmaSettings {
    private long initializationTimeoutMillis = 15000;
}

class CentralDogmaClientAutoConfiguration {
    @Bean
    public CentralDogma dogmaClient(...) {
        ...

        if (initializationTimeoutMillis > 0) {
            centralDogma.whenReadyEndpoint().get(initializationTimeoutMillis, ...);
        }
        return centralDogma;
    }
}
```

ikhoon:
Motivation:

3.2 seconds is used for the default timeout if an endpoint is not
determined while executing a request.
While a server is starting up, a resolution of endpoints may take
longer than 3.2 seconds because the server could be busy initializing
the application context.
To solve the problem, `CentralDogma.whenEndpointReady()` was introduced
to wait for the initial endpoints before starting a request.

However, users found it easy to forget to call the API and use directly a
`CentralDogma` client without waiting for it. It is difficult to call
`whenEndpointReady().get()` because it is to block the current thread.
But a `CentralDogma` bean is a managed instance created in the main
thread when a Spring application context is initialized. So we can
safely call `CentralDogma.whenEndpointReady().get()` before returning a
client.

Modifications:

- Add `initializationTimeoutMillis` to `CentralDogmaSettings` so as to
  to let users configure the timeout
  - If unspecified, 15 seconds is used by default.
  - 0 disables the waiting for initialization.

Result:

- You can now configure the initialization timeout for a `CentralDogma`
  client when using Spring integration modules.
  ```yml
  centraldogma:
    hosts:
       - ...
    initialization-timeout-millis: 15000
  ```
- Fixes #691

### DIFF
--- a/client/java-spring-boot-autoconfigure/src/main/java/com/linecorp/centraldogma/client/spring/CentralDogmaClientAutoConfiguration.java
+++ b/client/java-spring-boot-autoconfigure/src/main/java/com/linecorp/centraldogma/client/spring/CentralDogmaClientAutoConfiguration.java
@@ -18,6 +18,9 @@ package com.linecorp.centraldogma.client.spring;
 import java.net.UnknownHostException;
 import java.util.List;
 import java.util.Optional;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -45,6 +48,8 @@ import com.linecorp.centraldogma.client.armeria.DnsAddressEndpointGroupConfigura
 public class CentralDogmaClientAutoConfiguration {
 
     private static final Logger logger = LoggerFactory.getLogger(CentralDogmaClientAutoConfiguration.class);
+
+    static final long DEFAULT_INITIALIZATION_TIMEOUT_MILLIS = 15;
 
     /**
      * Returns a newly created {@link CentralDogma} client.
@@ -125,6 +130,18 @@ public class CentralDogmaClientAutoConfiguration {
             builder.retryIntervalOnReplicationLagMillis(retryIntervalOnReplicationLagMillis);
         }
 
-        return builder.build();
+        final CentralDogma centralDogma = builder.build();
+        Long initializationTimeoutMillis = settings.initializationTimeoutMillis();
+        if (initializationTimeoutMillis == null) {
+            initializationTimeoutMillis = DEFAULT_INITIALIZATION_TIMEOUT_MILLIS;
+        }
+        if (initializationTimeoutMillis > 0) {
+            try {
+                centralDogma.whenEndpointReady().get(initializationTimeoutMillis, TimeUnit.MILLISECONDS);
+            } catch (InterruptedException | ExecutionException | TimeoutException e) {
+                throw new IllegalStateException("Failed to initialize the endpoints of " + centralDogma, e);
+            }
+        }
+        return centralDogma;
     }
 }
--- a/client/java-spring-boot-autoconfigure/src/main/java/com/linecorp/centraldogma/client/spring/CentralDogmaClientAutoConfiguration.java
+++ b/client/java-spring-boot-autoconfigure/src/main/java/com/linecorp/centraldogma/client/spring/CentralDogmaClientAutoConfiguration.java
@@ -139,7 +139,9 @@ public class CentralDogmaClientAutoConfiguration {
             try {
                 centralDogma.whenEndpointReady().get(initializationTimeoutMillis, TimeUnit.MILLISECONDS);
             } catch (InterruptedException | ExecutionException | TimeoutException e) {
-                throw new IllegalStateException("Failed to initialize the endpoints of " + centralDogma, e);
+                throw new IllegalStateException(
+                        "Failed to initialize the endpoints of " + centralDogma + " in " +
+                        initializationTimeoutMillis + " milliseconds", e);
             }
         }
         return centralDogma;
--- a/client/java-spring-boot-autoconfigure/src/main/java/com/linecorp/centraldogma/client/spring/CentralDogmaSettings.java
+++ b/client/java-spring-boot-autoconfigure/src/main/java/com/linecorp/centraldogma/client/spring/CentralDogmaSettings.java
@@ -246,7 +246,7 @@ public class CentralDogmaSettings {
 
     /**
      * Returns the number of milliseconds the Central Dogma client waits for initialization to complete.
-     * If {@code 0} is specified, Central Dogma client is immediately returned without waiting.
+     * If {@code 0} is specified, the Central Dogma client is immediately returned without waiting.
      * If unspecified, defaults to
      * {@value CentralDogmaClientAutoConfiguration#DEFAULT_INITIALIZATION_TIMEOUT_MILLIS}.
      */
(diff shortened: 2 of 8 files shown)


client/java-spring-boot-autoconfigure/src/main/java/com/linecorp/centraldogma/client/spring/CentralDogmaClientAutoConfiguration.java
```diff
@@ -46,6 +49,8 @@ public class CentralDogmaClientAutoConfiguration {
 
     private static final Logger logger = LoggerFactory.getLogger(CentralDogmaClientAutoConfiguration.class);
 
+    static final long DEFAULT_INITIALIZATION_TIMEOUT_MILLIS = 15;
```
KarboniteKream, line 52:
```suggestion
    static final long DEFAULT_INITIALIZATION_TIMEOUT_MILLIS = 15000;
```

ikhoon: 😱

client/java-spring-boot-autoconfigure/src/test/java/com/linecorp/centraldogma/client/spring/CentralDogmaInitializationTimeoutTest.java
```diff
@@ -0,0 +1,85 @@
...
+import com.linecorp.centraldogma.internal.api.v1.HttpApiV1Constants;
+
+@ExtendWith(SpringExtension.class)
+@SpringBootTest(classes = TestConfiguration.class)
+@ActiveProfiles({ "local", "initTimeout" })
+class CentralDogmaInitializationTimeoutTest {
+
+    // TODO(ikhoon): Randomize the port number to avoid flakiness.
+    private static final Lock lock = new ReentrantLock();
+    private static final int TEST_SERVER_PORT = 56463;
```
minwoox, line 49: Why do we need this fixed port?

ikhoon: Because `CentralDogma` client bean needs a pre-known hostname (`host:port`) to access a server.
https://github.com/line/centraldogma/pull/692/files#diff-4264d927674e9eb11e891891fce0d0305d917df8904caff14930b7df456b2a7f

client/java-spring-boot-autoconfigure/src/main/java/com/linecorp/centraldogma/client/spring/CentralDogmaSettings.java
```diff
@@ -222,13 +233,35 @@ public void setRetryIntervalOnReplicationLagMillis(@Nullable Long retryIntervalO
...
     }
 
+    /**
+     * Returns the number of milliseconds the Central Dogma client waits for initialization to complete.
+     * If {@code 0} is specified, Central Dogma client is immediately returned without waiting.
+     *
+     * @return {@code null} if not specified.
+     */
+    @Nullable
+    public Long initializationTimeoutMillis() {
```
jrhee17, line 243: Not strong on this, but initialization sounds a little general (vs. `endpointReadyTimeoutMillis` or `endpointInitializationTimeoutMillis`)

client/java-spring-boot-autoconfigure/src/test/java/com/linecorp/centraldogma/client/spring/CentralDogmaInitializationTimeoutTest.java
minwoox: Ah, thanks for the explanation.
I'm a bit worried about the flakiness but we can fix it later if this causes the problem. 😉

minwoox (approved):
Thanks, @ikhoon!

client/java-spring-boot-autoconfigure/src/test/java/com/linecorp/centraldogma/client/spring/CentralDogmaInitializationTimeoutTest.java
```diff
@@ -0,0 +1,85 @@
...
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+
+import com.linecorp.armeria.common.HttpResponse;
+import com.linecorp.armeria.server.Server;
+import com.linecorp.centraldogma.client.CentralDogma;
+import com.linecorp.centraldogma.client.spring.CentralDogmaClientAutoConfigurationWithDnsEndpointGroupConfiguratorTest.TestConfiguration;
+import com.linecorp.centraldogma.internal.api.v1.HttpApiV1Constants;
+
+@ExtendWith(SpringExtension.class)
+@SpringBootTest(classes = TestConfiguration.class)
```
jrhee17, line 43: Why import this `TestConfiguration` as opposed to the one defined in this file? 🤔

jrhee17 (commented):
Overall looks good! Left a minor question 🙇

client/java-spring-boot-autoconfigure/src/main/java/com/linecorp/centraldogma/client/spring/CentralDogmaSettings.java
ikhoon: Because 1) I didn't want to expose the implementation detail and 2) we might add another logic later to warm up `CentralDogma` somehow.

client/java-spring-boot-autoconfigure/src/test/java/com/linecorp/centraldogma/client/spring/CentralDogmaInitializationTimeoutTest.java
ikhoon: Oops... it wasn't meant. 😱

client/java-spring-boot-autoconfigure/src/main/java/com/linecorp/centraldogma/client/spring/CentralDogmaSettings.java
jrhee17: I see 👍  Makes sense

jrhee17 (approved):
Thanks @ikhoon 👍 🙇 👍

minwoox:
Thanks!

PR #701, opened 2022-07-12
Use `Objects.equals()` for the `mappedValue`  in `MappingWatcher`

kcvlex:
Motivation:
We don't have to use reference equality for comparing `mappedValue`. 

Modification:
- Use `Objects.equals()`.


ikhoon (approved):
Thanks, @kcvlex!

minwoox (approved):
Thanks, @kcvlex!

PR #719, opened 2022-08-25
Provide a way to log successful and failed requests

### LINKED ISSUE #718: Support for `LoggingService` in Central Dogma server
To diagnose an exception error on the client-side, request logs are important to check request parameters, response duration and application errors.
We may provide `RequestLogConfig` to let users customize:
- Sampling rate
- Logging level
- Logging target such as `API`, `METRICS`, `HEALTH`, `WEB`, `DOCS`.
- ...

ikhoon:
Motivation:

To diagnose an exception error on the client-side, request logs are important to check request parameters, response duration and application errors.

Modifications:

- Add `RequestLogConfig` which can limit the sampling rate, adjust target groups, and set the logging levels.
- `TransientServiceOption.WITH_SERVICE_LOGGING` is specified to `HealthCheckService` or `PrometheusExpositionService` 
  if `METRICS` or `HEALTH` group is defined in `RequestLogConfig`

Result:

- Logging configuration example:
  ```json
  {
    "targetGroups": [ "API", "HEALTH" ],
    "loggerName": "centraldogma.test",
    "requestLogLevel": "DEBUG",
    "successfulResponseLevel": "DEBUG",
    "failureResponseLevel": "ERROR",
    "successSamplingRate": 0.4,
    "failureSamplingRate": 0.6
  }
  ```
- Fixes #718

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
@@ -21,6 +21,7 @@ import static com.google.common.base.Preconditions.checkState;
 import static com.google.common.base.Strings.isNullOrEmpty;
 import static com.linecorp.centraldogma.internal.api.v1.HttpApiV1Constants.API_V0_PATH_PREFIX;
 import static com.linecorp.centraldogma.internal.api.v1.HttpApiV1Constants.API_V1_PATH_PREFIX;
+import static com.linecorp.centraldogma.internal.api.v1.HttpApiV1Constants.DOCS_PATH;
 import static com.linecorp.centraldogma.internal.api.v1.HttpApiV1Constants.HEALTH_CHECK_PATH;
 import static com.linecorp.centraldogma.internal.api.v1.HttpApiV1Constants.METRICS_PATH;
 import static com.linecorp.centraldogma.server.auth.AuthProvider.BUILTIN_WEB_BASE_PATH;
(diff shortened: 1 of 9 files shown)


server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -521,9 +527,67 @@ private Server startServer(ProjectManager pm, CommandExecutor executor,
...
+                                  .successSamplingRate(requestLogConfig.successSamplingRate())
+                                  .failureSamplingRate(requestLogConfig.failureSamplingRate())
+                                  .newDecorator();
+            final Set<RequestLogGroup> requestLogGroups = requestLogConfig.targetGroups();
+            if (requestLogGroups.contains(RequestLogGroup.ALL)) {
+                sb.decorator(loggingService);
+                logHealthCheck = true;
+                logMetrics = true;
+            } else {
+                for (RequestLogGroup logGroup : requestLogGroups) {
```
jrhee17, line 549: Question: Is it possible that users may want to configure different logging parameters for different paths?

e.g.
```
- METRIC: debug
- API: info
- ...
```

ikhoon: Oh... It is so nice! 👍

ikhoon: We can implement it by adding different `LoggingService`s.

minwoox, line 549: How about also supporting the specified path? e.g. `/api/v1`

minwoox (commented):
Thanks! Left some suggestions. 😄

trustin:
- Could a user specify different logging settings for different target groups? For example, a user might want to lower the log level or rate for WEB but not for API.
- Can we also add `OTHERS`, which is used for all other requests?
- How about categories rather than target groups?

PR #722, opened 2022-09-21
Add basic setup for new webapp using modern web development stacks

ikhoon:
Motivation:

This PR contains a minimal setup for new Central Domga admin page. Typescript, Webpack, React and https://chakra-ui.com/ is experimentally chosen as the build tools and the front-end framework.

Some settings could be while prototyping some pages in the following PRs. Because I still don't know which framework could be suitable for Central Dogma. Please feel free to comment on other frameworks or tools if you want to recommend them.

Modifications:

- Add `webapp` module and nodejs to build.gradle
- Configure `webpack` and add dependencies to package.json
- Implement hello world page using Chakra UI

Result:

Start the first step for the new webapp journey.


settings/eslint/eslintrc.js
```diff
@@ -0,0 +1,96 @@
+const commonParserOptions = {
+  ecmaVersion: '2020',
```
jrhee17, line 2: Question) Is there a reason not to use the most recent version?
https://www.ecma-international.org/publications-and-standards/standards/ecma-262/

(I'm not too familiar with the ecosystem so maybe this is too bleeding edge though.. 😅 )

webapp/tsconfig-webpack.json
```diff
@@ -0,0 +1,8 @@
+{
+  "compilerOptions": {
+    "allowSyntheticDefaultImports": true,
+    "esModuleInterop": true,
+    "module": "commonjs",
+    "target": "es5"
```
jrhee17, line 6: IIRC browsers these days usually support es6. Fine to leave as is though

jrhee17 (approved):
Sorry about the late review 😅   Let's go~~ 🚀 🚀 🚀

settings/eslint/eslintrc.js
ikhoon: I just copied the file from Armeria without deep thoughts. 😆 But I wanted to sync the configuration for the two projects.
https://github.com/line/armeria/blob/master/settings/eslint/eslintrc.js
We may create a separate repo to share it later.

jrhee17: Sounds good to me 👍

webapp/build.gradle
```diff
@@ -0,0 +1,18 @@
+plugins {
+    id 'base'
+    // TODO(ikhoon): Use version catalog once https://github.com/line/centraldogma/pull/716 is merged
```
minwoox, line 3: It's merged. 😉

minwoox (approved):
🎉 🎉 🎉 
Let's get going!

webapp/tsconfig-webpack.json
ikhoon: This is an option to use Typescript for Webpack configurations which are only used in nodejs runtime. I followed https://webpack.js.org/configuration/configuration-languages/ documentation. Anyway, es6 looks good.

PR #726, opened 2022-09-30
Fix the test failure in `CentralDogmaInitializationTimeoutTest`

ikhoon:
Motivation:

Since `CentralDogmaInitializationTimeoutTest` uses a pre-defined port, the test is easily broken.
```
CentralDogmaInitializationTimeoutTest > initializationError FAILED
    java.util.concurrent.CompletionException: java.net.BindException: Address already in use
        at java.base/java.util.concurrent.CompletableFuture.reportJoin(CompletableFuture.java:413)
        at java.base/java.util.concurrent.CompletableFuture.join(CompletableFuture.java:2118)
        at com.linecorp.armeria.common.util.EventLoopCheckingFuture.join(EventLoopCheckingFuture.java:87)
        at com.linecorp.centraldogma.client.spring.CentralDogmaInitializationTimeoutTest.beforeAll(CentralDogmaInitializationTimeoutTest.java:68)
```

Modifications:

- Retry to start the server up to 8 times with the default `Backoff`.

Result:

Hope the test doesn't fail. 🙏🙏🙏

### DIFF
--- a/client/java-spring-boot-autoconfigure/src/test/java/com/linecorp/centraldogma/client/spring/CentralDogmaInitializationTimeoutTest.java
+++ b/client/java-spring-boot-autoconfigure/src/test/java/com/linecorp/centraldogma/client/spring/CentralDogmaInitializationTimeoutTest.java
@@ -19,6 +19,7 @@ package com.linecorp.centraldogma.client.spring;
 import static org.assertj.core.api.Assertions.assertThat;
 
 import java.time.Duration;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
@@ -33,6 +34,9 @@ import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.test.context.ActiveProfiles;
 import org.springframework.test.context.junit.jupiter.SpringExtension;
 
+import com.google.common.util.concurrent.Uninterruptibles;
+
+import com.linecorp.armeria.client.retry.Backoff;
 import com.linecorp.armeria.common.HttpResponse;
 import com.linecorp.armeria.server.Server;
 import com.linecorp.centraldogma.client.CentralDogma;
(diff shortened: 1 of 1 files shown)


client/java-spring-boot-autoconfigure/src/test/java/com/linecorp/centraldogma/client/spring/CentralDogmaInitializationTimeoutTest.java
```diff
@@ -56,16 +61,31 @@ static class TestConfiguration {}
...
-                                (ctx, req) -> HttpResponse.delayed(HttpResponse.of("OK"),
-                                                                   Duration.ofSeconds(5)))
-                       .build();
-        server.start().join();
+        final int maxAttempts = 8;
+        for (int i = 1; i <= maxAttempts; i++) {
+            try {
+                final Server server =
+                        Server.builder()
+                              .http(TEST_SERVER_PORT)
```
jrhee17, line 73: Question) Is it possible to randomize the port number as well?

ikhoon: I have no idea how to randomize it at the moment. The port number is set in the following configuration.
https://github.com/line/centraldogma/blob/07bae6588f88600efa7d9de376e3ad854f16fc8d/client/java-spring-boot-autoconfigure/src/test/resources/config/application-initTimeout.yml#L3
I need to research if we can dynamically update the value.

jrhee17: Ah forgot about this. Fair enough 😅

jrhee17 (approved):
Thanks for the cleanup @ikhoon 👍

minwoox (approved):
I hope that too. 😆

PR #733, opened 2022-10-12
Increase the Git timeout of GitMirror to 60 seconds

ikhoon:
Motivation:

The timeout of Git is 10 seconds. It would be a sufficient value when the server is under normal workloads.  However, a Git server is under heavy loads, the Git client fails to read content in 10 seconds.

As GitMirror job is performed as a background task, increasing the timeout won't have any other bad effect on the CD server.

Modifications:

- Change the timeout of Git mirror client to 60 seconds from 10 seconds.

Result:

Reduce Git mirroring failures on slow Git Servers.


jrhee17 (approved):
Thanks @ikhoon ! 🙇 🙇 🙇

ikhoon:
Let me merge this PR after CI builds pass.

PR #735, opened 2022-10-13
Add metrics for mirroring

jrhee17:
**Motivation**

It is difficult to determine the state of mirroring currently.
I propose that we add metrics for mirroring with the following properties:

1) Add metrics for the executor which schedules mirroring. By doing so, we can gain visibility on how long processing all configured mirrors takes. (this is equivalent to the time taken for a single task)
2) Add timer metrics for each mirroring task, and a counter for whether each mirroring task succeeded or not. I propose that tags be configured per mirroring configuration. While this may increase the number of metrics, I don't think the impact is significant since by design, there can't be a large number of centraldogma servers (due to zookeeper quorum).

**Modifications**
- Pass `MeterRegistry` to `DefaultMirroringService`
- Add metrics for the executor which schedules mirroring.
- Add metrics for each monitoring tasks
- Extract a utility test class which creates a `Mirror` for convenience.

**Result**
- Better visibility on mirroring

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirroringService.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirroringService.java
@@ -39,9 +39,7 @@ import org.slf4j.LoggerFactory;
 import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
-import com.google.common.util.concurrent.ListenableScheduledFuture;
 import com.google.common.util.concurrent.ListeningExecutorService;
-import com.google.common.util.concurrent.ListeningScheduledExecutorService;
 import com.google.common.util.concurrent.MoreExecutors;
 
 import com.linecorp.centraldogma.server.MirrorException;
@@ -71,7 +69,7 @@ public final class DefaultMirroringService implements MirroringService {
     private final long maxNumBytesPerMirror;
 
     private volatile CommandExecutor commandExecutor;
-    private volatile ListeningScheduledExecutorService scheduler;
+    private volatile ScheduledExecutorService scheduler;
     private volatile ListeningExecutorService worker;
 
     private ZonedDateTime lastExecutionTime;
@@ -105,11 +103,10 @@ public final class DefaultMirroringService implements MirroringService {
 
         this.commandExecutor = requireNonNull(commandExecutor, "commandExecutor");
 
-        ScheduledExecutorService executorService = Executors.newSingleThreadScheduledExecutor(
+        final ScheduledExecutorService executorService = Executors.newSingleThreadScheduledExecutor(
                 new DefaultThreadFactory("mirroring-scheduler", true));
-        executorService = ExecutorServiceMetrics.monitor(meterRegistry, executorService,
+        scheduler = ExecutorServiceMetrics.monitor(meterRegistry, executorService,
                                                          "mirroringScheduler");
-        scheduler = MoreExecutors.listeningDecorator(executorService);
 
         // Use SynchronousQueue to prevent the work queue from growing infinitely
         // when the workers cannot handle the mirroring tasks fast enough.
@@ -128,19 +125,9 @@ public final class DefaultMirroringService implements MirroringService {
                     }
                 }));
 
-        final ListenableScheduledFuture<?> future = scheduler.scheduleWithFixedDelay(
+        scheduler.scheduleWithFixedDelay(
                 this::schedulePendingMirrors,
                 TICK.getSeconds(), TICK.getSeconds(), TimeUnit.SECONDS);
-
-        Futures.addCallback(future, new FutureCallback<Object>() {
-            @Override
-            public void onSuccess(@Nullable Object result) {}
-
-            @Override
-            public void onFailure(Throwable cause) {
-                logger.error("Git-to-CD mirroring scheduler stopped due to an unexpected exception:", cause);
-            }
-        }, MoreExecutors.directExecutor());
     }
 
     public synchronized void stop() {
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirroringService.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirroringService.java
@@ -25,6 +25,7 @@ import java.time.ZonedDateTime;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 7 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirroringTask.java
```diff
@@ -0,0 +1,71 @@
...
+    private final Iterable<Tag> tags;
+
+    MirroringTask(Mirror mirror, MeterRegistry meterRegistry) {
+        this.mirror = mirror;
+        this.meterRegistry = meterRegistry;
+        tags = generateTags(mirror);
+    }
+
+    private Counter counter(boolean success) {
+        return Counter.builder("mirror.result")
```
minwoox, line 55: How about creating success counter and failure counter only once when `GitMirror` is created and reusing it?

minwoox (commented):
Looks good 👍 
Left a small suggestion. 😉

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirroringService.java
```diff
@@ -100,8 +105,11 @@ public synchronized void start(CommandExecutor commandExecutor) {
 
         this.commandExecutor = requireNonNull(commandExecutor, "commandExecutor");
 
-        scheduler = MoreExecutors.listeningDecorator(Executors.newSingleThreadScheduledExecutor(
-                new DefaultThreadFactory("mirroring-scheduler", true)));
+        ScheduledExecutorService executorService = Executors.newSingleThreadScheduledExecutor(
+                new DefaultThreadFactory("mirroring-scheduler", true));
+        executorService = ExecutorServiceMetrics.monitor(meterRegistry, executorService,
+                                                         "mirroringScheduler");
+        scheduler = MoreExecutors.listeningDecorator(executorService);
```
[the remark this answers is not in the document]
jrhee17: I guess we don't need it if we don't care about being notified of the overall scheduling task failure.

https://github.com/line/centraldogma/blob/cfdb17f1ab6d93adba1e7ae2ebc4c2babf73783b/server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirroringService.java#L141

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirroringTask.java
jrhee17: I do acknowledge that registering a counter to a `meterRegistry` involves a `synchronized` call.
Having said this, I'm wondering if you had a location to define the counters in mind since both solution I can think of don't seem trivial 😅 

1) Bringing a reference to `meterRegistry` into `AbstractMirror`
2) Maintaining a separate mapping between `Mirror` -> `Counter`

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirroringService.java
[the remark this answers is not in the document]
minwoox: Oops, I completely missed that. 😅 
Let's rollback so that we can see the error log.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirroringTask.java
minwoox: I realized that the mirror configuration could be changed at the next mirroring so we cannot create the `counter` once and reuse it.
Also, there's a concurrent hash map inside with meter ID key so the counter is not created every time this is called unless the configuration is changed.
That being said, let's keep it as is. Sorry about it. 😅

minwoox (approved):
Thanks! 👍 👍 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirroringService.java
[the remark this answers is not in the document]
jrhee17: I see 😅   reverted!

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirroringTask.java
```diff
@@ -0,0 +1,71 @@
...
+        return Counter.builder("mirroring.result")
+                      .tags(tags)
+                      .tag("success", Boolean.toString(success))
+                      .register(meterRegistry);
+    }
+
+    void run(File workDir, CommandExecutor executor, int maxNumFiles, long maxNumBytes) {
+        try {
+            meterRegistry.timer("mirroring.task", tags)
+                         .record(() -> mirror.mirror(workDir, executor, maxNumFiles, maxNumBytes));
```
ikhoon, line 64: I'm worried this can create too many metrics because the cardinality of the tags can grow easily. How about manually recording only 1) failed mirrors or 2) slow mirrors which has executed longer than a thread hold? I think 1) is mandatory, but 2) could be optional and ignored at the moment.

jrhee17: If we only record certain events conditionally, I'm afraid we won't be able to distinguish between cases where:
1) The mirroring didn't run at all
2) The mirroring succeeded, but just isn't showing

I didn't think metric cardinality would be an issue mainly because
1) The number of mirroring tasks is likely to be limited
2) By design, CD relies on zookeeper quorum which limits the number of servers that can produce metrics.

Having said this, if you feel the number of metrics is still an issue, I guess a simpler solution would only to add metrics for the `mirroring-scheduler` and remove `MirroringTask` altogether. What do you think?

ikhoon: I see. Changing the metrics would be not a difficult task so let's keep it as it is.

server/src/test/java/com/linecorp/centraldogma/server/internal/mirror/MirroringTestUtils.java
```diff
@@ -0,0 +1,73 @@
...
+import com.linecorp.centraldogma.server.mirror.MirrorDirection;
+import com.linecorp.centraldogma.server.storage.project.Project;
+import com.linecorp.centraldogma.server.storage.repository.Repository;
+
+final class MirroringTestUtils {
+
+    static final Cron EVERY_MINUTE = new CronParser(
+            CronDefinitionBuilder.instanceDefinitionFor(CronType.QUARTZ)).parse("0 * * * * ?");
+
+    private MirroringTestUtils() {}
```
ikhoon, line 41: nit: Move the constructor below static methods?

ikhoon (approved):
Thanks, @jrhee17!

minwoox:
Nice work, @jrhee17! 👍

PR #746, opened 2022-11-08
Fix a wrong tarball name in publish-release.yml

ikhoon:
Motivation:

The filename of a a tarball which is uploaded to a release note should be `centraldogma-x.y.z.tgz`. But `refs/tags/` is prefixed to the name. https://github.com/line/centraldogma/actions/runs/3418769975/jobs/5691473376

Modifications:

- Use `github.ref_name` instead of `github.ref` to get a short ref name. https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

Result:

Correctly upload a built tarball using GitHub Actions.

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -46,6 +46,6 @@ jobs:
         uses: svenstaro/upload-release-action@v2
         with:
           repo_token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


minwoox (approved):
Thanks!

jrhee17 (approved):
Thanks @ikhoon ! 🙇 🙇 🙇

PR #748, opened 2022-11-10
Correctly copy CLI binaries to dist path

ikhoon:
Motivation:

`dogma` script expects the binaries to be in `dist/bin/native/` but `cli-version` was appended.
Hence, `dogma.{os}-{arch}` binaries were copied into
`dist/bin/native/{cli-version}/` and `dogma` script failed to run.

Modifications:

- Specify `cli-version` when copying CLI binaries.

Result:

Correctly run CLI binaries with the `dogma` script.


minwoox (approved):
Thanks! 👍

jrhee17 (approved):
Oops 😅    Thanks @ikhoon ! 🙇 🙇 🙇

PR #750, opened 2022-11-10
Improve the error message of a `TimeoutException` from `watcher.await…

minwoox:
…InitialValue()`

Motivation:
A `TimeoutException` is raised when using `watcher.awaitInitialValue(...)` if the target entry doesn't exist. We can give a hint that uses `errorOnEntryNotFound` in the exception message so that users who don't know the flag can give it a try if it's the problem.

Modifications:
- Provide a detailed error message when `Watcher.awaitInitialValue(...)` raises `TimeoutException`.

Result:

You can see the detailed error message when `Watcher.awaitInitialValue(...)` raises a `TimeoutException` because the target doesn't exist.

### DIFF
--- a/client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
+++ b/client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
@@ -137,9 +137,9 @@ public interface Watcher<T> extends AutoCloseable {
             final long initialTimeoutMillis = TimeUnit.MILLISECONDS.convert(timeout, unit);
             throw new TimeoutException("Failed to get the initial value in " + initialTimeoutMillis + " ms. " +
                                        "It's probably because the timeout value is too small or " +
-                                       "the target entry doesn't exist. Please consider using " +
-                                       "'errorOnEntryNotFound option' to get " +
-                                       EntryNotFoundException.class.getSimpleName() + " for that case.");
+                                       "the target entry doesn't exist. Please consider using the " +
+                                       "'errorOnEntryNotFound' option to get an " +
+                                       EntryNotFoundException.class.getSimpleName() + " for a missing entry.");
         } catch (ExecutionException e) {
             throw new Error(e);
         }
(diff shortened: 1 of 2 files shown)


client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
```diff
@@ -133,6 +133,13 @@ default Latest<T> awaitInitialValue(long timeout, TimeUnit unit) throws Interrup
         requireNonNull(unit, "unit");
         try {
             return initialValueFuture().get(timeout, unit);
+        } catch (TimeoutException e) {
+            final long initialTimeoutMillis = TimeUnit.MILLISECONDS.convert(timeout, unit);
+            throw new TimeoutException("Failed to get the initial value in " + initialTimeoutMillis + " ms. " +
+                                       "It's probably because the timeout value is too small or " +
+                                       "the target entry doesn't exist. Please consider using " +
+                                       "'errorOnEntryNotFound option' to get " +
```
KarboniteKream, line 141:
```suggestion
                                       "the 'errorOnEntryNotFound' option to get " +
```

KarboniteKream (approved):

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
```diff
@@ -133,6 +133,13 @@ default Latest<T> awaitInitialValue(long timeout, TimeUnit unit) throws Interrup
...
+                                       EntryNotFoundException.class.getSimpleName() + " for that case.");
```
jrhee17, line 142: Since `that case` is ambiguous

```suggestion
            throw new TimeoutException("Failed to get the initial value in " + initialTimeoutMillis + " ms. " +
                                       "It's probably because the timeout value is too small or " +
                                       "the target entry doesn't exist. Please consider using " +
                                       "the 'errorOnEntryNotFound' option to get a " +
                                       EntryNotFoundException.class.getSimpleName() + " for a missing entry.");
```

jrhee17 (approved):
Thanks @minwoox ! 🙇 👍 🙇

ikhoon (approved):
LGTM once other comments are addressed. 🙇‍♂️

client/java/src/main/java/com/linecorp/centraldogma/client/Watcher.java
minwoox: Thanks! 😉

PR #752, opened 2022-11-12
Fix equals() implementation in AbstractCommand (#751)

### LINKED ISSUE #751: The equals implementation in AbstractCommand is not correct
Here:
https://github.com/line/centraldogma/blob/c59d2a1defdb7eda3a27959eca0630fd261034cd/server/src/main/java/com/linecorp/centraldogma/server/command/AbstractCommand.java#L62

It says:
```java
        if (!(this instanceof AbstractCommand)) {
            return false;
        }
```

This condition is always false, as `this` is always `AbstractCommand`. It should be instead:
```java
        if (!(obj instanceof AbstractCommand)) {
            return false;
        }
```

It's interesting that IntelliJ IDEA tried to warn about this, but the warning was suppressed with `@SuppressWarnings("EqualsWhichDoesntCheckParameterClass")`. After the fix, the suppression would be unnecessary.

amaembo:
Motivation:
The `equals` method in `AbstractCommand` is not correctly implemented.

Modification:
- Implement `AbstractCommand.equals()` correctly.

Result:
- `ClassCastException` is not raised anymore when checking the equality of sub `AbstractCommand`s that don't override `equals()` method.
- Fixes #751

### DIFF
--- a/server/src/test/java/com/linecorp/centraldogma/server/command/CreateSessionCommandTest.java
+++ b/server/src/test/java/com/linecorp/centraldogma/server/command/CreateSessionCommandTest.java
@@ -17,6 +17,7 @@
 package com.linecorp.centraldogma.server.command;
 
 import static net.javacrumbs.jsonunit.fluent.JsonFluentAssert.assertThatJson;
+import static org.assertj.core.api.Assertions.assertThat;
 
 import java.io.ByteArrayOutputStream;
 import java.io.ObjectOutputStream;
@@ -77,5 +78,19 @@ class CreateSessionCommandTest {
             oos.writeObject(object);
             return Base64.getEncoder().encodeToString(baos.toByteArray());
         }
+    }
+
+    @Test
+    void testEquals() {
+        final Session session =
+                new Session("session-id-12345",
+                        "foo",
+                        Instant.EPOCH,
+                        Instant.EPOCH.plus(1, ChronoUnit.MINUTES),
+                        "abc");
+        final CreateSessionCommand command = new CreateSessionCommand(1234L,
+                new Author("foo", "bar@baz.com"),
+                session);
+        assertThat(command).isNotEqualTo("string");
     }
 }
--- a/server/src/main/java/com/linecorp/centraldogma/server/command/AbstractCommand.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/command/AbstractCommand.java
@@ -54,12 +54,11 @@ abstract class AbstractCommand<T> implements Command<T> {
     }
 
     @Override
-    @SuppressWarnings("EqualsWhichDoesntCheckParameterClass")
     public boolean equals(Object obj) {
         if (this == obj) {
             return true;
         }
-        if (!(this instanceof AbstractCommand)) {
+        if (!(obj instanceof AbstractCommand)) {
             return false;
         }
 


amaembo:
Some commands override equals and not affected, but others actually fail with ClassCastException, when comparing with unrelated object. Probably it would be better though to compare classes exactly, instead of instanceof. Like `if (obj == null || obj.getClass() != this.getClass()) return false;`. What do you think?

server/src/test/java/com/linecorp/centraldogma/server/command/CreateSessionCommandTest.java
```diff
@@ -78,4 +80,18 @@ private static String toSerializedBase64(Object object) throws Exception {
...
+        final Session session =
+                new Session("session-id-12345",
+                        "foo",
+                        Instant.EPOCH,
+                        Instant.EPOCH.plus(1, ChronoUnit.MINUTES),
+                        "abc");
+        CreateSessionCommand command = new CreateSessionCommand(1234L,
+                new Author("foo", "bar@baz.com"),
+                session);
+        assertFalse(command.equals("string"));
```
jrhee17, line 95:
```suggestion
        assertThat(command).isNotEqualTo("string")
```

jrhee17 (approved):
Thanks for finding/fixing this 👍 🙇 🚀

jrhee17:
> Probably it would be better though to compare classes exactly, instead of instanceof. 

Currently, all subclasses perform a check on the class type and the abstract class is package private so I guess it's fine to leave as is 😅

server/src/test/java/com/linecorp/centraldogma/server/command/CreateSessionCommandTest.java
jrhee17: Can you also take a look at lint failures?

minwoox (approved):
Thanks!
Please address @jrhee17's comment. 😄 
Could you also update the commit description?

ikhoon (approved):
Thanks, @amaembo!

server/src/test/java/com/linecorp/centraldogma/server/command/CreateSessionCommandTest.java
amaembo: I addressed the lint failures, thank you. I also updated to use `isNotEqualTo`, though I usually hesitate to do this. It's unclear whether `command.equals("string")` or `"string".equals(command)` (or both) will be called under the hood by testing framework. Here, we want to test the first so if testing framework actually invokes the second, then we would test `String.equals()` method instead. It's ok to use `isNotEqualTo` if you are sure that equality contract is correct, but here the contract is the thing we are testing. Test frameworks usually don't specify how exactly `isNotEqualTo`-like methods work, and behavior may change in future versions of testing frameworks.

I've posted a short twitter thread about this a while ago, you may [want to check](https://twitter.com/tagir_valeev/status/1265482450572136450). Normally, it's recommended to use [EqualsVerifier](https://github.com/jqno/equalsverifier) to check the equality contract, though this would be a much bigger pull request, and likely it will find more problems and I afraid I would not be able to fix all of them with zero experience in this project.

amaembo:
@minwoox updated the commit message. Is it better now?

amaembo:
@minwoox I've updated PR title and commit message, trying to mimic previous commits in master. Should I update anything else?

ikhoon:
> Should I update anything else?

Nope, it looks good.

server/src/test/java/com/linecorp/centraldogma/server/command/CreateSessionCommandTest.java
jrhee17: My original comment was because if we use `assertj`'s built in method,  we get a richer error message on why the test failed (as opposed to just "`true` isn't equal to `false`")

Having said this, I agree with your point as well. Let me also check out [equalsverifier](https://github.com/jqno/equalsverifier) 🙇

minwoox:
That looks good and thanks a lot for fixing this bug. :smile:

PR #756, opened 2022-11-17
Copy generated static files into the `server` module

ikhoon:
Motivation:

The `webapp` module should be served by a Central Domga server.

Modification:

- Add an `pages/index.tsx` to serve the initial html file.
- Add build tasks for the front and copy the resources to server.

Result:

You can now serve the new front webapp using a Central Dogma server.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
@@ -751,8 +751,15 @@ public class CentralDogma implements AutoCloseable {
                                    .build(),
                         "/index.html"));
             }
-            sb.serviceUnder("/",
+
+            sb.serviceUnder("/legacy-web",
                             FileService.builder(CentralDogma.class.getClassLoader(), "webapp")
+                                       .cacheControl(ServerCacheControl.REVALIDATED)
+                                       .build());
+
+            sb.serviceUnder("/",
+                            FileService.builder(CentralDogma.class.getClassLoader(),
+                                                "com/linecorp/centraldogma/webapp")
                                        .cacheControl(ServerCacheControl.REVALIDATED)
                                        .build());
         }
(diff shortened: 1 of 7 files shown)


webapp/package.json
```diff
@@ -6,7 +6,7 @@
     "test": "echo \"Error: no test specified\" && exit 1",
     "develop": "NEXT_PUBLIC_HOST=\"http://127.0.0.1:36462\" next dev",
     "mock": "NEXT_PUBLIC_HOST='' next dev",
-    "build": "next build",
+    "build": "NEXT_PUBLIC_HOST='' next build && next export -o build/web/",
```
ikhoon, line 9: Sometimes, I wanted to run `next` commands directly on the command line.
If `NEXT_PUBLIC_HOST=''` isn't set, `undefined` is prefixed to Ajax paths.
Should we use if-else condition or Elvis operator to avoid such cases?

webapp/pages/index.tsx
```diff
@@ -0,0 +1,24 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+const Index = () => {
+  // TODO(ikhoon):
+  //  - Move pages folder under 'src/' directory. https://nextjs.org/docs/advanced-features/src-directory
```
ikhoon, line 19: For better folder organization, it would be worth moving `pages/` directory into `src/pages/`

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -751,10 +751,17 @@ public String serviceName(ServiceRequestContext ctx) {
...
-            sb.serviceUnder("/",
+
+            sb.serviceUnder("/legacy-web",
                             FileService.builder(CentralDogma.class.getClassLoader(), "webapp")
                                        .cacheControl(ServerCacheControl.REVALIDATED)
                                        .build());
+
+            sb.serviceUnder("/",
+                            FileService.builder(CentralDogma.class.getClassLoader(),
+                                                "com/linecorp/centraldogma/webapp")
```
ikhoon, line 762: If a Central Dogma server starts with Gradle, the HTML and JS files are automatically added to classpath and served.

PR #757, opened 2022-11-18
Add more setup to integrate the webapp with the test server

rainy789:
## Motivation
Integrate the webapp with ShiroCentralDogmaTestServer and SimpleCentralDogmaTestServer.

## Modification
- Add a loading state to handle anonymous mode access.
- Update shiro login path.
- Moves `pages` to `src/pages` https://nextjs.org/docs/advanced-features/src-directory

## Result
- When the security_enabled flag is true, it will redirect to shiro login link for authentication. 
- When the security_enabled flag is false, it will allow access in anonymous mode.
- The webapp is still working as usual after moving `pages` directory in the root to `src/pages`


ikhoon:
Should we remove the following line to run Gradle build tasks successfully? 
https://github.com/line/centraldogma/blob/3e9c2ee4cafa19a00ac737572eec90c104d8fe93/webapp/build.gradle#L30

ikhoon:
There is an infinite loop in the login process. 😅

rainy789:
> There is an infinite loop in the login process. 😅

Hmm how do you repro that? Could you please delete .next and try again? 
Screen Shot 2022-11-18 at 14 44 54

SimpleCentralDogmaTestServer 
Screen Shot 2022-11-18 at 14 39 54

ShiroCentralDogmaTestServer 
-> blank page because the ```router.push(`${process.env.NEXT_PUBLIC_HOST}/link/auth/login/`);``` somehow changes the path to http://127.0.0.1:36462/web/auth/login/
Screen Shot 2022-11-18 at 14 43 24

rainy789:
For some reason, `router.pathname` will return a path without trailing slash.
Screen Shot 2022-11-18 at 14 59 39
--> router.pathname returns `http://127.0.0.1:36462/web/auth/login`

ikhoon:
>  --> router.pathname returns /web/auth/login

Right. it was my problem. 😅

ikhoon:
We might come up with a more declarative way to define whether a page should be authorized rather than matching path. 🤔 The URL will be changed in the future when we disable `trailingSlash`.

ikhoon (approved):
Thanks!

PR #759, opened 2022-11-18
Make a Next.js test app login with a Central Dogma server

ikhoon:
Motivation:

A dev server running on `3000` port redirects to the `http://DOGMA_HOST/link/auth/login` to get configured login page URL. The client expects the server to redirect to the dev server's login page if Shiro auth provider is configured. However, a Central Dogma server assumes the web client is hosted in the same origin and returns a relative path. As a result, the dev client loses a chance to render the login page and a Central Dogma hosted login page is rendered instead.
```
http://127.0.0.1:3000/ -> http://127.0.0.1:36462/link/auth/login ->
http://127.0.0.1:36462/web/auth/login
```
To request a login request from the dev server, the Central Dogma server should redirect to `http://127.0.0.1:3000/web/auth/login`.

Modifications:

- Prefix a host to the login path if `return_to` exists and redirect to the absolute URL.

Result:

A Next.js dev server can login with a Central Dogma server.

### DIFF
--- a/webapp/src/dogma/features/auth/Authorized.tsx
+++ b/webapp/src/dogma/features/auth/Authorized.tsx
@@ -48,7 +48,11 @@ export const Authorized = (props: { children: ReactNode }) => {
     return <>{props.children}</>;
   }
   if (typeof window !== 'undefined' && !auth.isAuthenticated) {
-    router.push(`${process.env.NEXT_PUBLIC_HOST}/link/auth/login/`);
+    if (process.env.NEXT_PUBLIC_HOST) {
+      router.push(`${process.env.NEXT_PUBLIC_HOST}/link/auth/login/?return_to=http://127.0.0.1:3000`);
+    } else {
+      router.push(`/link/auth/login/`);
+    }
   }
   return <></>;
 };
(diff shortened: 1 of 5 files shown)


webapp/src/dogma/features/auth/Authorized.tsx
```diff
@@ -48,7 +48,11 @@ export const Authorized = (props: { children: ReactNode }) => {
     return <>{props.children}</>;
   }
   if (typeof window !== 'undefined' && !auth.isAuthenticated) {
-    router.push(`${process.env.NEXT_PUBLIC_HOST}/link/auth/login/`);
+    if (process.env.NEXT_PUBLIC_HOST) {
+      router.push(`${process.env.NEXT_PUBLIC_HOST}/link/auth/login/?return_to=http://127.0.0.1:3000`);
```
rainy789, line 52: How about 
```
   return_to=${window.location.host}
```

ikhoon: Sounds good!

ikhoon: `window.location.host` does not have a protocol part.
After googling, `window.location.origin` fits our desire.
https://stackoverflow.com/a/25329802/1736581

PR #760, opened 2022-11-21
Decouple authentication routes from authorization routes

rainy789:
## Motivation
Currently login, user loading and session validation seem to get mixed up in authSlice.

## Modification
- Rewrite authSlice to separate handling the lifecycle actions from modifying the state.
- Move dispatch calls to suitable locations in the UIs.
- Update login, user reload and validation logic.
- Remove unused packages 
   - `qs`
   - `html-webpack-plugin`
   - `webpack-cli`
   - `webpack-dev-server`
- Update the logic for `<Authorized />`, replace isAuthentication / isAuthorized with simply checking for anonymous mode or the existence of `sessionId`.

## Result
- Each auth action now has a single responsibility.
- Error in authSlice is now handled. We can easily use them later for popup / display on the UIs.
- Hopefully a more understandable logic for authorized/protected pages.
  - In anonymous mode, any pages can be accessed.
  - In protected mode, the pages can only be accessed if a sessionId exists.
  - No restriction for access to the login page.


webapp/src/dogma/features/auth/authSlice.ts
```diff
@@ -16,130 +16,141 @@
...
-    });
-
-    if (userResponse.status === HttpStatusCode.Ok) {
-      // The current session ID is valid. Login is not required.
-      return { isAuthorized: true, user: userResponse.data as UserDto };
+export const validateSession = createAsyncThunk<UserSessionResponse>(
+  '/auth/validate_session',
+  async (_, { rejectWithValue }) => {
+    try {
+      await axios.get(`${process.env.NEXT_PUBLIC_HOST}/security_enabled`);
```
ikhoon, line 76: I don't see `security_enabled` can validate the current `sessionId`. It does not check the `sessionId` and returns 200 or 404 status.
`/api/v0/users/me` was added to check the current session is valid. If the API returns `401 Unauthorized`, we can assume the the session expires.

webapp/src/dogma/features/auth/authSlice.ts
```diff
@@ -16,130 +16,141 @@
...
-      removeSessionId();
-      return { isAuthorized: false };
-    }
-  }
-
-  // Should not reach here in the normal case.
-  return { isAuthorized: false };
-});
-
+const sessionId = typeof window !== 'undefined' && localStorage.getItem('sessionId');
```
ikhoon, line 87: We should not use `sessionId` until it is verified.

webapp/src/pages/web/auth/login.tsx
```diff
@@ -1,24 +1,29 @@
...
   const router = useRouter();
-  const user = useAppSelector((state) => state.auth.user);
+  const sessionId = useAppSelector((state) => state.auth.sessionId);
   useEffect(() => {
-    if (user) {
+    if (sessionId) {
       router.push('/');
     }
-  }, [router, user]);
+  }, [router, sessionId]);
```
ikhoon, line 15: If `isInAnonymousMode` is set to `true`, should we redirect to `/`?

webapp/src/pages/web/auth/login.tsx
```diff
@@ -1,24 +1,29 @@
 import Head from 'next/head';
 import { LoginForm } from 'dogma/features/auth/LoginForm';
-import { useAppSelector } from 'dogma/store';
+import { useAppDispatch, useAppSelector } from 'dogma/store';
 import { useEffect } from 'react';
 import { useRouter } from 'next/router';
+import { login } from 'dogma/features/auth/authSlice';
 
 const LoginPage = () => {
```
ikhoon, line 8: Suggestion: What do you think of adding a convention that confines filenames under `pages/` to ends `Page`? This convention make the filename more readable and easy to navigate with file browsers. 
For example: `login.tsx` -> `LoginPage.tsx`, `projects.tsx` -> `ProjectsPage.tsx`

webapp/src/dogma/features/auth/authSlice.ts
rainy789: `security_enabled` is only checking the mode (anonymous or not). 

The validation happens in the next step. 
```
.addCase(getUser.rejected, (state) => {
        state.user = null;
        state.sessionId = '';
        if (typeof window !== 'undefined') {
          localStorage.removeItem('sessionId');
        }
      });
```
Here if `getUser` fails then `sessionId` gets cleared.

webapp/src/pages/web/auth/login.tsx
rainy789: Good catch, thank you!

webapp/src/dogma/features/auth/authSlice.ts
ikhoon: Seems like we use this pattern to sequentially execute two thunks. https://stackoverflow.com/a/67028881/1736581

webapp/src/dogma/features/auth/authSlice.ts
rainy789: This one is only for the initial state.

webapp/src/dogma/features/auth/authSlice.ts
rainy789: > It looks good but I'm worried accessing localStorage might validate the basic rule of reducers.

Persisting the store is actually a common case, there's even a library for it https://www.npmjs.com/package/redux-persist

webapp/src/dogma/features/auth/Authorized.tsx
```diff
@@ -17,42 +17,29 @@
...
-      router.push(`${process.env.NEXT_PUBLIC_HOST}/link/auth/login/?return_to=${window.location.origin}`);
-    } else {
-      router.push(`/link/auth/login/`);
-    }
+  if (typeof window !== 'undefined') {
+    router.push(
+      process.env.NEXT_PUBLIC_HOST
+        ? `${process.env.NEXT_PUBLIC_HOST}/link/auth/login/?return_to=${window.location.origin}`
+        : `/link/auth/login/`,
+    );
```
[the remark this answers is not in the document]
rainy789: `const router = useRouter();` can only be called from inside a component. Will try to address this in a different PR.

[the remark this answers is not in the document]
ikhoon: > const router = useRouter(); can only be called from inside a component.

I understood. Please ignore the comment. It is a small piece of code and used in two paces.

webapp/src/dogma/features/auth/authSlice.ts
ikhoon: Makes sense.

webapp/src/dogma/features/auth/authSlice.ts
```diff
@@ -16,130 +16,142 @@
...
-  // TODO(ikhoon): Just use fetch API?
-  const response = await axios.get(`${process.env.NEXT_PUBLIC_HOST}/api/v0/users/me`);
-  return response.data as UserDto;
+export const getUser = createAsyncThunk('/auth/user', async (_, { getState, rejectWithValue }) => {
+  try {
+    const { auth } = getState() as { auth: AuthState };
+    if (!auth.sessionId) {
+      return rejectWithValue('Session id not provided');
+    }
+    const { data } = await axios.get(`${process.env.NEXT_PUBLIC_HOST}/api/v0/users/me`, {
```
ikhoon, line 27: How about adding a const value in a network utility class to return the hostname of the API server?
```
const DOGMA_HOST = process.env.NEXT_PUBLIC_HOST || "";
```

ikhoon:
Fixed checkstyle violations that I made in the previous PR. Additionally, `gradle eslint` task is added to run "npm run lint" in the build lifestyle.

webapp/src/dogma/features/auth/authSlice.ts
rainy789: > It looks good but I'm worried accessing `localStorage` might validate the basic rule of reducers.

Let me move them.

webapp/src/dogma/features/auth/authSlice.ts
rainy789: Updated the axios config instead.

webapp/src/dogma/features/auth/authSlice.ts
```diff
@@ -116,28 +133,34 @@ export const authSlice = createSlice({
...
+        state.sessionId = payload.access_token;
+      })
+      .addCase(login.rejected, (state) => {
+        state.sessionId = '';
+      })
+      .addCase(logout.fulfilled, (state) => {
+        state.isInAnonymousMode = true;
+        state.sessionId = '';
+        state.user = null;
+        state.ready = true;
```
ikhoon, line 146: I'm confused about the meaning of `ready`.
Does it mean the current session is validated?

webapp/src/dogma/features/auth/authSlice.ts
ikhoon: >  Should we also fix the following code to access sessionId from the redux store? The second parameter of prepareHeaders allows us to call getState().

This comment is not addressed. Generally, it would work well. But I prefer to use one of `localStorage.getItem('sessionId')` and `state.auth.sessionId` for consistency. 
As we decided to add `sessionId` to `auth` store, it would be worth using `state.auth.sessionId` for reading.

- `localStorage.getItem('sessionId')` is only used to read the initial value.
- `localStorage.setItem()` and `localStorage.removeItem()` are handled in `createThunkAsync`.
- Rest of them, should we use `state.auth.sessionId`?

ikhoon (commented):
Overall looking good. I left 2 comments.
We may consider the role `auth.ready` in a separate issue from this PR.

webapp/src/dogma/features/auth/authSlice.ts
rainy789: We need to wait until it's ready to check the state. For example, here https://github.com/line/centraldogma/blob/d4c6397baf7367d81677c90cdd5537dbf15ebe62/webapp/src/dogma/features/auth/Authorized.tsx#L41

ikhoon: Actually, I was wondering why `ready` is set to `true` for `logout.fulfilled`. From what I understand here, it needs to back to the initial state.

webapp/src/dogma/features/auth/authSlice.ts
rainy789: > 



> I'm OK if we only use `localStorage` as an auxiliary storage to get initial values. Otherwise, I think `auth.sessionId` should be used outside of reducers. You have fixed all other code but one thing seems to remain. Should we also fix the following code to access `sessionId` from the redux store? The second parameter of `prepareHeaders` allows us to call `getState()`.
> 
> https://github.com/line/centraldogma/blob/5b1514a3cc2f61e6bbdec8fe7911ddd1297d681c/webapp/src/dogma/features/api/apiSlice.ts#L24-L25

Somehow I missed it as it was in a separate file. Thanks for pointing it out. Updated.

webapp/src/dogma/features/auth/authSlice.ts
rainy789: For logout, I set it to `true` to bypass the loading state since we don't really need a loading indicator at this point. 
```
 if (!auth.ready) {
    return <p>Loading...</p>;
  }
```
The `ready` state is useful when we have a loading indicator. In that case we may display a progress bar / loading indicator while it's not `ready`.

rainy789: Actually let me remove them for now at it doesn't really matter at this point.

webapp/src/dogma/features/auth/authSlice.ts
```diff
@@ -116,28 +133,33 @@ export const authSlice = createSlice({
...
-          status.isAuthenticated = true;
-        }
+      .addCase(login.fulfilled, (state, { payload }) => {
+        state.sessionId = payload.access_token;
+      })
+      .addCase(login.rejected, (state) => {
+        state.sessionId = '';
+      })
+      .addCase(logout.fulfilled, (state) => {
+        state.isInAnonymousMode = true;
```
ikhoon, line 143: Shouldn’t it be?
```suggestion
        state.isInAnonymousMode = false;
```

ikhoon (approved):
Our code is getting cleaner! 👍🙇‍♂️

webapp/src/dogma/features/auth/authSlice.ts
rainy789: The initial state is `true`. It'll only become `false` if the call to `/security_enabled` api is successful.

PR #761, opened 2022-11-24
Display client error

rainy789:
## Motivation
Show client side error to the user.

## Modification
- Convert any caught error to string.
- Add a message reducer to handle any client side error.
- Dispatch caught auth error to the message reducer.
- Show an error toast in the UI.
- Default anonymous state to false.

## Result
- Error can be displayed using a component instead of the browser alert.
- Any caught / client side error can be dispatched to the message reducer. 
- The following error types are supported `'error' | 'info' | 'warning' | 'success' | 'loading'`.
- The app starts with false anonymous mode. 

Screen Shot 2022-11-24 at 13 56 20

### DIFF
--- a/webapp/src/dogma/common/components/ErrorWrapper.tsx
+++ b/webapp/src/dogma/common/components/ErrorWrapper.tsx
@@ -1,0 +1,20 @@
+import { ReactNode, useEffect } from 'react';
+import { useToast } from '@chakra-ui/react';
+import { useAppSelector } from 'dogma/store';
+
+export const ErrorWrapper = (props: { children: ReactNode }) => {
+  const { errorText, errorType } = useAppSelector((state) => state.message);
+  const toast = useToast();
+  useEffect(() => {
+    if (errorText) {
+      toast({
+        title: errorType,
+        description: errorText,
+        status: errorType,
+        duration: 10000,
+        isClosable: true,
+      });
+    }
+  });
+  return <>{props.children}</>;
+};
--- a/webapp/src/dogma/common/components/ErrorWrapper.tsx
+++ b/webapp/src/dogma/common/components/ErrorWrapper.tsx
@@ -3,18 +3,18 @@ import { useToast } from '@chakra-ui/react';
 import { useAppSelector } from 'dogma/store';
 
 export const ErrorWrapper = (props: { children: ReactNode }) => {
-  const { errorText, errorType } = useAppSelector((state) => state.message);
+  const { title, text, type } = useAppSelector((state) => state.message);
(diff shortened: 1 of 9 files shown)


webapp/src/dogma/features/auth/authSlice.ts
```diff
@@ -103,12 +101,10 @@ export const logout = createAsyncThunk('/auth/logout', async (_, { getState, rej
...
-  } catch (error) {
-    if (error.response && error.response.data.message) {
-      return rejectWithValue(error.response.data.message);
-    } else {
-      return rejectWithValue(error.message);
-    }
+  } catch (err) {
+    const error: string = ErrorHandler.handle(err);
+    dispatch(createMessageError(error));
+    return rejectWithValue(error);
```
ikhoon, line 107: Question: Would we create a middleware to display rejected values?
https://redux-toolkit.js.org/rtk-query/usage/error-handling#handling-errors-at-a-macro-level

webapp/src/dogma/common/components/ErrorWrapper.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { useToast } from '@chakra-ui/react';
+import { useAppSelector } from 'dogma/store';
+
+export const ErrorWrapper = (props: { children: ReactNode }) => {
+  const { errorText, errorType } = useAppSelector((state) => state.message);
+  const toast = useToast();
+  useEffect(() => {
+    if (errorText) {
+      toast({
+        title: errorType,
```
ikhoon, line 11: Status icons imply that an error type is warn or info.
How about exposing `title` only if a meaningful value is set?

webapp/src/dogma/features/auth/authSlice.ts
rainy789: Creating a rtk middleware involves more steps as we can't use ChakraUI `useToast` directly in the middleware.

webapp/src/dogma/common/components/ErrorWrapper.tsx
rainy789: Updated message slice to allow the caller to provide title, text and type.

rainy789:
I've updated auth state to the following.
image

webapp/src/dogma/features/auth/authSlice.ts
ikhoon: I see. We may create a new action and dispatch the action in a middleware. That said let's consider it later. It is crucial for this PR.

ikhoon (approved):
Thanks!

PR #768, opened 2022-12-07
Add a shortcut to the project detail page 

rainy789:
## Motivation
Add a shortcut to the project detail page (on the navbar). 

## Modification
- Add chakra react select
   - chakra-react-select https://github.com/csandman/chakra-react-select is a wrapper of react-select
   - react-select https://github.com/jedwatson/react-select
   - alternative https://github.com/koolamusic/chakra-ui-autocomplete but this one is using downshift, which looks more complex than react-select
- Add some fake data using fakerjs https://fakerjs.dev/
- Remove router from useEffect dependencies (causing infinite loop). 
- Combine chakra's link with next link https://github.com/line/centraldogma/pull/766#discussion_r1040993677
- Update mock apis to use dynamic routes.

## Result
Follow these instructions to view the result locally. 
- npm i
- npm run mock
- Use the navbar to navigate to any projects.


webapp/package.json
```diff
@@ -22,6 +22,7 @@
     "@reduxjs/toolkit": "^1.8.5",
     "@tanstack/react-table": "^8.7.0",
     "axios": "^1.1.3",
+    "chakra-react-select": "^4.4.2",
```
ikhoon, line 25: I'm not sure whether it is my local problem.
When I tried to start the app, it showed 
image
Clean install does not seems to work. 🤔

ikhoon: After I switched `npm` to `yarn`, the problem was fixed.

rainy789: These steps work for me 🤔 
```
1) git clone https://github.com/rainy789/centraldogma.git
2) cd centraldogma
3) git fetch origin
4) git checkout --track origin/new-webapp
5) cd webapp
6) npm i
7) npm run mock
```

webapp/src/dogma/common/components/Navbar.tsx
```diff
@@ -67,12 +65,37 @@ const NavLink = ({ link, children }: { link: string; children: ReactNode }) => (
...
+  const { user } = useAppSelector((state) => state.auth);
   const dispatch = useAppDispatch();
-  const router = useRouter();
+  const { data: projects } = useGetProjectsQuery();
+  const projectOptions: ProjectOptionType[] =
+    projects &&
+    projects.map((project: ProjectDto) => ({
+      value: project.name,
+      label: project.name,
+    }));
```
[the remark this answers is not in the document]
rainy789: I've set the default of the projects to an empty array instead.

ikhoon (approved):
I thought I merged this PR. 😅 The shortcut bar looks useful! 

Note: A page is redirected to the root page. We may need to handle it in the following work.

rainy789:
> Note: A page is redirected to the root page. We may need to handle it in the following work.

Found the issue. 😅  Will include the fix in the next PR.

PR #772, opened 2022-12-15
Add a keyboard shortcut for the project dropdown

rainy789:
## Motivation
Add a keyboard shortcut for the project dropdown on the navbar.

## Modification
- Update the dropdown icon to `/`.
- Auto-trigger the dropdown menu when it's on focus.
- Add an event listener for `/`.

## Result
Pressing forward slash `/` should trigger the project dropdown.

### DIFF
--- a/webapp/src/dogma/common/components/Navbar.tsx
+++ b/webapp/src/dogma/common/components/Navbar.tsx
@@ -13,7 +13,7 @@
  * License for the specific language governing permissions and limitations
  * under the License.
  */
-import { ReactNode, useEffect, useState } from 'react';
+import { ReactNode, useEffect, useState, useRef } from 'react';
 import {
   Avatar,
   Box,
@@ -39,7 +39,8 @@ import { logout } from 'dogma/features/auth/authSlice';
 import Router from 'next/router';
 import { useGetProjectsQuery } from 'dogma/features/api/apiSlice';
 import { ProjectDto } from 'dogma/features/project/ProjectDto';
-import { OptionBase, Select } from 'chakra-react-select';
+import { components, DropdownIndicatorProps, GroupBase, OptionBase, Select } from 'chakra-react-select';
+import { RxSlash } from 'react-icons/rx';
 
 interface TopMenu {
   name: string;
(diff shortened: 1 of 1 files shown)


webapp/src/dogma/common/components/Navbar.tsx
```diff
@@ -123,8 +148,11 @@ export const Navbar = () => {
             onChange={(option) => handleChange(option as ProjectOptionType)}
             placeholder="Jump to project ..."
             closeMenuOnSelect={true}
+            openMenuOnFocus={true}
```
ikhoon, line 151: Suggestion: Could we set a different color such as a darker one to the background color of the select input to emphasize it?

ikhoon: Otherwise, it is also good to set a distinguishable `borderColor` to `container` of the select box. 

I'm not 100% sure it is a good suggestion. 😅 UX and design are conundrums to me.

rainy789: Updated to the following.
Screen Shot 2022-12-16 at 10 59 32

ikhoon: Awesome!
By the way, `X` button is shown on the right without the initial placeholder when a page is newly rendered.
image

webapp/src/dogma/common/components/Navbar.tsx
```diff
@@ -123,8 +150,17 @@ export const Navbar = () => {
...
             isClearable={true}
             isSearchable={true}
+            ref={selectRef}
+            components={{ DropdownIndicator }}
+            chakraStyles={{
+              control: (baseStyles) => ({
+                ...baseStyles,
+                backgroundColor: colorMode === 'light' ? 'white' : 'whiteAlpha.50',
```
ikhoon, line 161: ❤️

webapp/src/dogma/common/components/Navbar.tsx
rainy789: 🙇  Fixed.

PR #773, opened 2022-12-16
Use `Flags.meterRegistry` as the default `MeterRegistry`

jrhee17:
**Motivation**

Currently, a `PrometheusMeterRegistry` is used as the default `MeterRegistry` for `centraldogma` server.
This makes it difficult to extend to add additional `MeterRegistry` to export metrics.
I propose the following three changes:

1) I propose that `Flags.meterRegistry()` is used as the default `MeterRegistry`, and `PrometheusMeterRegistry` is added to `Flags.meterRegistry` if it is a `CompositeMeterRegistry`.
2) Since `Flags#meterRegistry()` may be used by other threads within the jvm, I propose that the lifecycle of `CentralDogma#meterRegistry()` be independent from `CentralDogma` server's lifecycle.
This means that `CentralDogmaBuilder` accepts a `MeterRegistry`, and `CentralDogma` won't close the `MeterRegistry` on server shutdown.
3) Lastly, tests utilizing metrics may become unstable since metrics could be added to a shared `Flags#meterRegistry()`. Therefore, I propose that `CentralDogmaBuilder#meterRegistry()` is always set to a new `MeterRegistry` when a `CentralDogmaRule` is started up.

**Modifications**
- Modify to accept a `MeterRegistry` for `CentralDogmaBuilder`. The default value will be `Flags#meterRegistry`.
- `CentralDogma` accepts a default `MeterRegistry` and uses this to record metrics. A prometheus endpoint will be exposed only if the `meterRegistry` is a `PrometheusMeterRegistry` or `CompositeMeterRegistry`.
- `CentralDogma` now doesn't close the `MeterRegistry` as the lifecycle could be different.
- `CentralDogmaRuleDelegate` initializes a new `CompositeMeterRegistry` for each `CentralDogma` server started up to avoid flakiness/cleanup issues.

**Result**
- Users have more freedom to add a custom `MeterRegistry` to the main `CompositeMeterRegistry`


server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -320,13 +322,14 @@ private void doStart() throws Exception {
         ScheduledExecutorService purgeWorker = null;
         ProjectManager pm = null;
         CommandExecutor executor = null;
-        PrometheusMeterRegistry meterRegistry = null;
+        CompositeMeterRegistry meterRegistry = null;
         Server server = null;
         SessionManager sessionManager = null;
         try {
-            meterRegistry = PrometheusMeterRegistries.newRegistry();
+            meterRegistry = new CompositeMeterRegistry();
```
minwoox, line 329: Can't we just use `Flags.meterRegistry()` instead of creating `CompositeMeterRegistry` and adding it to the another `CompositeMeterRegistry`?

jrhee17: > Can't we just use Flags.meterRegistry() instead of creating CompositeMeterRegistry and adding it to the another CompositeMeterRegistry?

This was the first idea I had as well, before I started running into CI failures.
https://github.com/line/centraldogma/actions/runs/3719341734

Since integration tests continuously start up and shut down the server (and shutdown is asynchronous):
1. Tests that rely on using `meterRegistry` fail since a previous `meter` with a same name has been registered
2. An `OutOfMemoryException` is raised presumably due to `meterRegistry` not being cleaned up from previous tests.

I think what you have suggested can work if:
1) We make sure that between each test we wait until the `CentralDogma` is completely stopped synchronously
2) Between each test we clear the `Flags.meterRegistry`

I'm not sure if this is a good idea though 😅   I'm welcome to ideas.

minwoox: I should've read the description carefully. Sorry about it. 😓 
We didn't have any problem with Armeria so I thought it was also going to work.
So If we stop the server synchronously, we don't have any problem using the `Flags.meterRegitry()`. Is that correct?

jrhee17: > So If we stop the server synchronously, we don't have any problem using the Flags.meterRegitry(). Is that correct?

If we stop the server synchronously and clear the `meterRegistry`between each test I think it should work fine (there are also some tests which don't clean up the `CentralDogma` server so we need to handle those as well)

> We didn't have any problem with Armeria so I thought it was also going to work.

I thought so too 😅

jrhee17: Alternatively, we can also just remove tests that rely on checking metrics for correctness

e.g.
https://github.com/line/centraldogma/blob/be0426bf89418328c4ba11f2ac1940173583c32a/it/src/test/java/com/linecorp/centraldogma/it/PrematureClientFactoryCloseTest.java#L81-L84

minwoox: If so, I'm leaning toward modifying the tests because:
- It doesn't make sense to implement production code inefficiently because test code doesn't work
- It probably wouldn't take so long to complete tests even if we change to close synchronously compared to Armeria build time. 😆

jrhee17: I've made changes so that flakiness doesn't affect production code.

The main points were:
1) Add a setter method `CentralDogmaBuilder#meterRegistry(MeterRegistry)` so that users can specify the `MeterRegistry` to be used.
2) For the main implementation, use `Flags#meterRegistry()`.
3) For test implementations, initialize a new `CompositeMeterRegistry`

Please take another look 🙏

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -782,9 +782,24 @@ private static Function<? super HttpService, EncodingService> contentEncodingDec
...
         sb.meterRegistry(registry);
-        sb.service(METRICS_PATH, new PrometheusExpositionService(registry.getPrometheusRegistry()));
+
+        // expose the prometheus endpoint if the registry is either a PrometheusMeterRegistry or
+        // CompositeMeterRegistry
+        if (registry instanceof PrometheusMeterRegistry) {
+            final PrometheusMeterRegistry prometheusMeterRegistry = (PrometheusMeterRegistry) registry;
+            sb.service(METRICS_PATH,
+                       PrometheusExpositionService.of(prometheusMeterRegistry.getPrometheusRegistry()));
+        } else if (registry instanceof CompositeMeterRegistry) {
```
ikhoon, line 794: Does it make sense to use `PrometheusMeterRegistry` if the `CompositeMeterRegistry` contains it?
For example:
https://github.com/line/armeria/blob/7626562204bd4311a1205df9d733203324355a96/spring/boot2-autoconfigure/src/main/java/com/linecorp/armeria/spring/PrometheusSupport.java#L40-L55

testing/common/src/main/java/com/linecorp/centraldogma/testing/internal/CentralDogmaRuleDelegate.java
```diff
@@ -99,6 +100,10 @@ public final CompletableFuture<Void> startAsync(File dataDir) {
             }
         }
 
+        // specify a new meterRegistry since multiple instances writing to a single meterRegistry
+        // can be a source of flakiness
```
ikhoon, line 104: 👍

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -818,10 +833,6 @@ private void doStop() {
         this.pm = null;
         this.repositoryWorker = null;
         this.sessionManager = null;
-        if (meterRegistry != null) {
-            meterRegistry.close();
```
ikhoon, line 822: As of now, don't we close `meterRegistry` any more?

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
jrhee17: I think it's fine to leave as is since there is almost no difference in behavior.

For instance, assuming a `PrometheusMeterRegistry` already exists (`PrometheusMeterRegistry1`).
When the `centraldogma` server is started up, `CompositeMeterRegistry` will have (`PrometheusMeterRegistry1`, `PrometheusMeterRegistry2`).
At this point, whatever metrics have been added to `CompositeMeterRegistry` will also be synchronized to both children. (although it is possible each metric has different states for a little while, it will eventually become consistent).

Since the meters for both registries will eventually become the same, I thought it wasn't a big deal (although there may be slight performance degradation)

I actually think the current `PrometheusSupport` implementation is slightly flawed in that children `CompositeMeterReigstry`s aren't accounted for. (I guess that using [CompositeMeterRegistry#nonCompositeDescendants ](https://github.com/micrometer-metrics/micrometer/blob/38003f544f8a1ceb29d9bc634e964d2606f8bb39/micrometer-core/src/main/java/io/micrometer/core/instrument/composite/CompositeMeterRegistry.java#L49) would be cleanest, but unforunately it seems like this isn't available publicly)

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
jrhee17: I thought the lifecycle of the `MeterRegistry` is now independent from the `CentralDogma` server.

For instance, I was thinking if `Flags.meterRegistry` is used in `Application A` and `CentralDogma`, should `Flags.meterRegistry` be closed when `CentralDogma` is stopped although `ApplicationA` is running?

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
ikhoon: Understood. Let's keep it as is.

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
ikhoon: > MeterRegistry is now independent from the CentralDogma server since it is configurable from CentralDogmaBuilder.

It may be right or not. It is still possible for CentralDogma server to create a `PrometheusMeterRegistry` internally which should go along with the server.

jrhee17: Ideally, I think the best way would be to only remove all metrics that are related to `centraldogma`.
I think this would involve 
1) adding a prefix like `centraldogma.*` to all metrics
2) removing only metrics with the prefix `centraldogma.*`

Having said this, this is probably a breaking change and the scope of this PR would become a lot bigger.

> It is still possible for CentralDogma server to create a PrometheusMeterRegistry internally which should go along with the server.

I think it is possible to store the instantiated `PrometheusMeterRegistry` and remove it when closing the `CentralDogma` server. I think this will only remove the child registry and the meters will still be attached to the `CompositeMeterRegistry`.

jrhee17: [6086a8f](https://github.com/line/centraldogma/pull/773/commits/6086a8f4d1fa9a7a10978a48af61cf1bba360a08) added a commit for this.

Let me know if this is what you meant 🙏

ikhoon: Looks good! 🙇‍♂️

ikhoon (approved):
Thanks, @jrhee17!

minwoox (approved):
Thanks! 👍

PR #782, opened 2023-01-04
Add query by JSON path

rainy789:
## Motivation
Allow querying JSON file using JSON path. 

## Modification
- Add JSON path. https://github.com/dchester/jsonpath
- Add debounced JSON path component.
- Fix failing tests.

## Result
- Querying using JSON path is available in read/view mode.
- Unlike the old webapp, the query can be done on the frontend side thus no API calls are needed.

Screen Shot 2023-01-04 at 13 28 22


webapp/src/dogma/common/components/editor/JsonPath.tsx
```diff
@@ -0,0 +1,11 @@
+import { DebouncedInput } from 'dogma/common/components/table/DebouncedInput';
+import { Box, Text } from '@chakra-ui/react';
+
+export const JsonPath = ({ handleQuery }: { handleQuery: (value: string | number) => void }) => {
+  return (
+    <Box mb="4">
+      <Text mb="2">Query by JSON Path</Text>
```
ikhoon, line 7: `JSONPath` seems like an official term.
- https://goessner.net/articles/JsonPath/
- https://www.baeldung.com/guide-to-jayway-jsonpath
```suggestion
      <Text mb="2">Query by JSONPath</Text>
```

webapp/src/dogma/common/components/editor/JsonPath.tsx
```diff
@@ -0,0 +1,11 @@
...
+      <DebouncedInput type="text" value="" onChange={handleQuery} placeholder={`$..[0]`} />
```
ikhoon, line 8: While I tried to type some JSON path expressions, I got many error messages whenever I was in the middle of incomplete JSON path syntax.
I didn't see it as a good UX and it was annoying to me. 😅
What do you think of borrowing some ideas from https://jsonpath.com/ ?
Showing JSON expression tip looks good to me.

rainy789: Updated. 😅

ikhoon (approved):
The new UI and UX look so nice. Thanks always, @rainy789! 💯🙇‍♂️

PR #783, opened 2023-01-06
Add react-hook-form

rainy789:
## Motivation
Replace formik with react-hook-form. https://react-hook-form.com

## Modification
- Add react-hook-form and remove formik. Both are actually similar in functionality but react-hook-form doesn't have any dependencies at all. https://github.com/react-hook-form/react-hook-form/blob/master/package.json
- Update login form.
- Some clean up.

## Result
Screen Shot 2023-01-06 at 17 27 33

### DIFF
--- a/webapp/package.json
+++ b/webapp/package.json
@@ -66,6 +66,7 @@
     "tsconfig-paths": "^4.1.0",
     "typescript": "^4.8.3"
   },
+  "//" : "This is a workaround related to Typescript's node module resolution, which is used by RTK query.",
   "exports": {
     "./query/react": "./dist/query/react/index.js",
     "./query": "./dist/query/index.js",
--- a/webapp/package.json
+++ b/webapp/package.json
@@ -25,12 +25,12 @@
     "axios": "^1.1.3",
     "chakra-react-select": "^4.4.2",
     "date-fns": "^2.29.3",
-    "formik": "^2.2.9",
     "framer-motion": "^6.5.1",
     "jsonpath": "^1.1.1",
     "next": "^13.0.2",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
+    "react-hook-form": "^7.41.3",
     "react-icons": "^4.7.1",
     "react-redux": "^8.0.4"
   },
@@ -65,5 +65,10 @@
     "ts-node": "^10.9.1",
     "tsconfig-paths": "^4.1.0",
     "typescript": "^4.8.3"
+  },
+  "exports": {
+    "./query/react": "./dist/query/react/index.js",
+    "./query": "./dist/query/index.js",
+    ".": "./dist/index.js"
   }
 }
--- a/webapp/src/dogma/features/auth/LoginForm.tsx
+++ b/webapp/src/dogma/features/auth/LoginForm.tsx
@@ -46,7 +46,7 @@ export const LoginForm = () => {
       <Box bg={colorMode === 'light' && 'white'} p={6} rounded="md" w={64}>
         <form onSubmit={handleSubmit(onSubmit)}>
           <VStack spacing={4} align="flex-start">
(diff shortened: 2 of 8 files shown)


webapp/src/dogma/features/auth/LoginForm.tsx
```diff
@@ -13,41 +13,64 @@
...
-              </FormControl>
-              <Button type="submit">Submit</Button>
-            </VStack>
-          </Form>
-        </Formik>
+    <Flex bg={colorMode === 'light' && 'gray.100'} align="center" justify="center" h="100vh">
+      <Box bg={colorMode === 'light' && 'white'} p={6} rounded="md" w={64}>
+        <form onSubmit={handleSubmit(onSubmit)}>
+          <VStack spacing={4} align="flex-start">
+            <FormControl isRequired isInvalid={errors.username ? true : false}>
```
ikhoon, line 49: Question: Do we need both `isInvalid={...}` and `isRequired`?

ikhoon (approved):
It looks good!

webapp/src/dogma/features/auth/LoginForm.tsx
rainy789: I've removed isInvalid so we're relying on HTML5 form validation instead. 

Screen Shot 2023-01-09 at 09 53 18

webapp/package.json
```diff
@@ -65,5 +65,10 @@
     "ts-node": "^10.9.1",
     "tsconfig-paths": "^4.1.0",
     "typescript": "^4.8.3"
+  },
+  "exports": {
+    "./query/react": "./dist/query/react/index.js",
+    "./query": "./dist/query/index.js",
+    ".": "./dist/index.js"
```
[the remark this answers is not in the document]
ikhoon: Do you think we should leave a comment on why we added this workaround?
I found that we use the follow patten to leave comments on package.json
```
"//": "this is a comment",
```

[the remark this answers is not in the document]
rainy789: This is a workaround related to Typescript's node module resolution, which used by RTK query.

webapp/package.json
```diff
@@ -65,5 +65,10 @@
     "ts-node": "^10.9.1",
     "tsconfig-paths": "^4.1.0",
     "typescript": "^4.8.3"
+  },
+  "exports": {
+    "./query/react": "./dist/query/react/index.js",
+    "./query": "./dist/query/index.js",
```
ikhoon, line 71: My suggestion was to add the comment on the `package.json` so that other maintainers understand the historical reason for the workaround in the future. 😄
```suggestion
  "//" : "This is a workaround related to Typescript's node module resolution, which is used by RTK query.",
  "exports": {
    "./query/react": "./dist/query/react/index.js",
    "./query": "./dist/query/index.js",
```

webapp/src/dogma/features/auth/LoginForm.tsx
ikhoon: Looks nice.

ikhoon:
Regarding the lint error on Windows, I am not 100% sure but I guess the error message might be related to https://stackoverflow.com/questions/53516594/why-do-i-keep-getting-eslint-delete-cr-prettier-prettier
https://github.com/prettier/eslint-plugin-prettier/issues/219#issuecomment-610262026

rainy789:
> Regarding the lint error on Windows, I am not 100% sure but I guess the error message might be related to https://stackoverflow.com/questions/53516594/why-do-i-keep-getting-eslint-delete-cr-prettier-prettier [prettier/eslint-plugin-prettier#219 (comment)](https://github.com/prettier/eslint-plugin-prettier/issues/219#issuecomment-610262026)

Thanks for sharing. 👍 Let's try it.

ikhoon:
Let me merge this PR and let's consider the lint error as a separate issue. 😀

PR #785, opened 2023-01-09
Update eslint config

rainy789:
## Motivation
Properly setup eslint, prettier and typescript plugins for nextjs.

## Modification
- Remove eslint package as it comes bundled with nextjs. https://nextjs.org/docs/basic-features/eslint
- Use strict mode `next/core-web-vitals`  https://nextjs.org/docs/basic-features/eslint#core-web-vitals
- The following default setting is not in use https://github.com/line/centraldogma/blob/new-webapp/settings/eslint/eslintrc.js
- Fix some lint related error.
- Add typescript plugin.
- Remove the following plugins.
```
    "eslint-config-airbnb": "^19.0.4",
    "eslint-config-airbnb-typescript": "^17.0.0",
    "eslint-plugin-eslint-comments": "^3.2.0",
    "eslint-plugin-import": "^2.26.0",
    "eslint-plugin-json": "^3.1.0",
    "eslint-plugin-jsx-a11y": "^6.5.1",
    "eslint-plugin-next": "^0.0.0",
```
We may add them back as required.

## Result
- Simplified eslint config.


webapp/src/dogma/features/history/HistoryList.tsx
```diff
@@ -11,7 +11,7 @@ export type HistoryListProps<Data extends object> = {
   data: Data[];
   projectName: string;
   repoName: string;
-  handleTabChange: Function;
+  handleTabChange: (index: number) => void;
```
ikhoon, line 14: Looks better!

ikhoon (approved):
Thanks for cleaning up!

PR #787, opened 2023-01-11
Update history 

rainy789:
## Motivation
Update history to use v1 API. 

Version 0: [https://github.com/line/centraldogma/blob/1ceed8a869d364e2b6fd12ae1bbbc0fb0c83f292[…]ntraldogma/server/internal/admin/service/RepositoryService.java](https://github.com/line/centraldogma/blob/1ceed8a869d364e2b6fd12ae1bbbc0fb0c83f292/server/src/main/java/com/linecorp/centraldogma/server/internal/admin/service/RepositoryService.java#L170-L184)
Version 1: [https://github.com/line/centraldogma/blob/e46e977e8c2836047e44f7389b00da41412a44a5[…]linecorp/centraldogma/server/internal/api/ContentServiceV1.java](https://github.com/line/centraldogma/blob/e46e977e8c2836047e44f7389b00da41412a44a5/server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java#L319-L332)

## Modification
- Update API endpoints (and remove trailing `/`)
- Update history.
- Memoize file and history tables. https://react-table-v7.tanstack.com/docs/quick-start#define-columns

## Result
The history for a newly created project is now available.

Screen Shot 2023-01-11 at 15 24 16

### DIFF
--- a/webapp/src/dogma/features/api/apiSlice.ts
+++ b/webapp/src/dogma/features/api/apiSlice.ts
@@ -81,7 +81,7 @@ export const apiSlice = createApi({
         `/v1/projects/${projectName}/repos/${repoName}/files/revisions/${revision}/${filePath}?queryType=IDENTITY`,
     }),
     getHistoryByProjectAndRepoName: builder.query<HistoryDto[], GetFilesByProjectAndRepoName>({
-      query: ({ projectName, repoName }) => `/v1/projects/${projectName}/repos/${repoName}/commits/`,
+      query: ({ projectName, repoName }) => `/v1/projects/${projectName}/repos/${repoName}/commits`,
     }),
   }),
 });
--- a/webapp/src/dogma/features/api/apiSlice.ts
+++ b/webapp/src/dogma/features/api/apiSlice.ts
@@ -64,7 +64,7 @@ export const apiSlice = createApi({
       invalidatesTags: ['Project'],
     }),
     getMetadataByProjectName: builder.query<ProjectMetadataDto, string>({
-      query: (projectName) => `/v1/projects/${projectName}/`,
+      query: (projectName) => `/v1/projects/${projectName}`,
     }),
     getReposByProjectName: builder.query<RepoDto[], string>({
       query: (projectName) => `/v1/projects/${projectName}/repos`,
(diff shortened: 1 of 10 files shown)


webapp/src/dogma/features/api/apiSlice.ts
```diff
@@ -81,7 +81,7 @@ export const apiSlice = createApi({
         `/v1/projects/${projectName}/repos/${repoName}/files/revisions/${revision}/${filePath}?queryType=IDENTITY`,
     }),
     getHistoryByProjectAndRepoName: builder.query<HistoryDto[], GetFilesByProjectAndRepoName>({
-      query: ({ projectName, repoName }) => `/v1/projects/${projectName}/repos/${repoName}/history`,
+      query: ({ projectName, repoName }) => `/v1/projects/${projectName}/repos/${repoName}/commits/`,
```
ikhoon, line 84: What do you think of making `getHistoryByProjectAndRepoName` taking `revision` and `maxCommits`? We may browse old commits when `Older` button is clicked.

The default value of `maxCommits` is 100 which would be large for a web application. 
https://github.com/line/centraldogma/blob/e9e82d1c087f1a0ec4f5e5cc9e6609f2130f015e/server/src/main/java/com/linecorp/centraldogma/server/storage/repository/Repository.java#L66
How about setting the value to 20 or 30?

rainy789: Question: how are going to handle the pagination? I don't see `page` param in the API, is it going to be handled in the frontend?

rainy789: I've added a pagination for the table instead. 

Screen Shot 2023-01-13 at 15 01 21

ikhoon: Wow! It is so fast.

ikhoon: It looks to handle the pagination on the client side.
Some repositories may have very long histories whose old versions may not be shown without relying on the server-side pagination API. 🤔
To show the histories from head revision, we wouod use negative revision numbers.
For example, the first 20 revisions would be -1 to -20 and the next page should be -21 to -40.

Anyway, the UI design looks good and convenient. This PR is itself good. 👍

ikhoon (approved):
Looks great! 🤩💯

PR #789, opened 2023-01-20
Add breadcrumbs

rainy789:
## Motivation
Allow the user to go back to the top level. This is especially helpful when there are nested directories.

## Modification
- Add breadcrumbs component.
- Add breadcrumbs to the top of each page.

## Result
- `npm run mock`
- Navigate to the file/directory.

Screen Shot 2023-01-20 at 10 19 37

### DIFF
--- a/webapp/src/pages/app/projects/[projectName]/repos/[repoName]/files/[revision]/[[...path]]/index.tsx
+++ b/webapp/src/pages/app/projects/[projectName]/repos/[repoName]/files/[revision]/[[...path]]/index.tsx
@@ -3,6 +3,7 @@ import { Box, Flex, Heading, Tag, Tooltip } from '@chakra-ui/react';
 import { useGetFileContentQuery } from 'dogma/features/api/apiSlice';
 import { useRouter } from 'next/router';
 import FileEditor from 'dogma/common/components/editor/FileEditor';
+import { Breadcrumbs } from 'dogma/common/components/Breadcrumbs';
 
 const FileContentPage = () => {
   const router = useRouter();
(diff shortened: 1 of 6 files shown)


webapp/src/pages/app/projects/[projectName]/repos/[repoName]/files/[revision]/[[...path]]/index.tsx
```diff
@@ -22,8 +23,12 @@ const FileContentPage = () => {
   }
   return (
     <Box p="2">
+      <Breadcrumbs path={router.asPath} omitIndexList={[0, 3, 5, 6]} suffixes={{ 4: '/list/head' }} />
```
Comment (ikhoon) [inline webapp/src/pages/app/projects/[projectName]/repos/[repoName]/files/[revision]/[[...path]]/index.tsx:26]: Not sure it is a good idea. 😅 
We may generalize `Breadcrumbs` to use in various situations. The number-based `omitIndexList` is unlikely to be used in complex cases.

Just random thoughts:
```tsx
// Create breakcrumbs for all path segments by default.
<Breadcrumbs path={router.asPath}/>

// Remove a segment or transform it into another.
<Breadcrumbs path={router.asPath}, 
  filter={(previousSegments, currentSegment, index) => {
    if (currentSegment === "ignorable") {
       return false;
    } 
    if (currentSegment === "replacable") {
        return "alternative";
    }
    ...
  }
}/>

<Breadcrumbs path={router.asPath}, 
             toLink={(segments) => segments.join("/") + "/list/head") />
```

I read the examples after I wrote. It does not look convenient to use. 😅
We may optimize the component later when we need such cases.

Comment (ikhoon) [reply webapp/src/pages/app/projects/[projectName]/repos/[repoName]/files/[revision]/[[...path]]/index.tsx:26]: Let's keep it as is. I don't think it is widely used in various cases.

ikhoon (approved):
Thanks, 🥖🍞

Issue #56, opened 2017-10-11
Pagination when browsing history


trustin:
No pagination has been implemented in the history view of the administrative UI. This can cause a problem as the number of commits increases.

ikhoon:
This has been fixed by #793

PR #793, opened 2023-01-25
Add controlled pagination for commit history

rainy789:
## Motivation
Address #56 

## Modification
- Add server side pagination based on the following summary. cc @ikhoon 
  - The first low of the first page should start with the latest revision aka. HEAD.
  -  If  a page renders 20 revisions
      - The first page starts with HEAD and ends with HEAD - 20.
      - The second page starts with HEAD - 21 and ends with HEAD - 40
      - If the number of commits returned by server is less then 20, we could assume it is the last page.
- Add `msw` and `whatwg-fetch` for API endpoint mocking. https://mswjs.io/

## Result
- `npm run mock`
- Navigate to the repository page
- Click on the history tab
Screen Shot 2023-01-25 at 15 24 05

### DIFF
--- a/webapp/src/dogma/features/history/HistoryList.tsx
+++ b/webapp/src/dogma/features/history/HistoryList.tsx
@@ -1,26 +1,20 @@
-import { ColumnDef, createColumnHelper } from '@tanstack/react-table';
+import { PaginationState, createColumnHelper } from '@tanstack/react-table';
 import { HistoryDto } from 'dogma/features/history/HistoryDto';
-import { DynamicDataTable } from 'dogma/common/components/table/DynamicDataTable';
-import { Badge, Box, Button, HStack } from '@chakra-ui/react';
-import NextLink from 'next/link';
+import { Badge, Box, HStack } from '@chakra-ui/react';
 import { ChakraLink } from 'dogma/common/components/ChakraLink';
-import { FaHistory } from 'react-icons/fa';
 import { DateWithTooltip } from 'dogma/common/components/DateWithTooltip';
-import { useMemo } from 'react';
+import { useMemo, useState } from 'react';
+import { useGetHistoryQuery } from 'dogma/features/api/apiSlice';
+import { DynamicDataTable } from 'dogma/common/components/table/DynamicDataTable';
 
-export type HistoryListProps<Data extends object> = {
-  data: Data[];
+export type HistoryListProps = {
   projectName: string;
   repoName: string;
   handleTabChange: (index: number) => void;
+  totalRevision: number;
 };
 
-const HistoryList = <Data extends object>({
-  data,
-  projectName,
-  repoName,
-  handleTabChange,
-}: HistoryListProps<Data>) => {
+const HistoryList = ({ projectName, repoName, handleTabChange, totalRevision }: HistoryListProps) => {
   const columnHelper = createColumnHelper<HistoryDto>();
   const columns = useMemo(
     () => [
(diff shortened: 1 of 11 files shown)


webapp/src/dogma/features/history/HistoryList.tsx
```diff
@@ -49,21 +43,48 @@ const HistoryList = <Data extends object>({
...
+  const {
+    data = [],
+    isLoading,
+    error,
+  } = useGetHistoryQuery({
+    projectName,
+    repoName,
+    //  revision starts from -1, for example for pageSize=20
+    //  The first page  /projects/{projectName}/repos/{repoName}/commits/-1?to=20
+    //  The second page /projects/{projectName}/repos/{repoName}/commits/-21?to=20
```
ikhoon, line 69: Sorry, my previous example was incorrect. 😅 `to` is not the size of commits but a pointer to a specific revision.

Let's say we have two parameters `pageNumber` and `pageSize` and the initial page starts with `1`. We can derive `revision` and `to` parameters with the following formula.
- `revision`: `((pageNumber - 1) * pageSize * -1) - 1`
- `to`: `revision - pageSize + 1`

The final queries are: 
- The first page: `/projects/{projectName}/repos/{repoName}/commits/-1?to=-20`
  - revision = ((1 - 1) * 20 * -1) - 1 = -1
  - to = -1 - 20 + 1 = -20
- The second page: `/projects/{projectName}/repos/{repoName}/commits/-21?to=-40`
  - revision = ((2 - 1) * 20 * -1) - 1 = -21
  - to = -21 - 20 + 1 = -40
- The N page: ...

rainy789: Updated to the following.
```
    //  revision starts from -1, for example for pageSize=20
    //  The first page  /projects/{projectName}/repos/{repoName}/commits/-1?to=-20
    //  The second page /projects/{projectName}/repos/{repoName}/commits/-21?to=-40
    revision: -pageIndex * pageSize - 1,
    to: Math.max(-totalRevision, -(pageIndex + 1) * pageSize),
```

webapp/src/dogma/features/history/HistoryList.tsx
```diff
@@ -49,21 +43,48 @@ const HistoryList = <Data extends object>({
...
-      }),
     ],
     [columnHelper, handleTabChange, projectName, repoName],
   );
-  return <DynamicDataTable columns={columns as ColumnDef<Data>[]} data={data} pagination={true} />;
+  const [{ pageIndex, pageSize }, setPagination] = useState<PaginationState>({
+    pageIndex: 0,
+    pageSize: 10,
+  });
+  const pagination = useMemo(
```
ikhoon, line 53: Question: Is `useMemo` efficient for this sample case?

webapp/src/pages/app/projects/[projectName]/repos/[repoName]/list/[revision]/[[...path]]/index.tsx
```diff
@@ -169,10 +163,10 @@ cat ${project}/${repo}${path}`;
           </TabPanel>
           <TabPanel>
             <HistoryList
```
Comment (ikhoon) [inline webapp/src/pages/app/projects/[projectName]/repos/[repoName]/list/[revision]/[[...path]]/index.tsx:165]: It is not related to this PR but I found out that the pagination isn't preserved when a page is refreshed or on history back. It might be useful if we provide a unique link for a page when a user wants to share a specific history page or visit several revisions to check changes.

ikhoon (approved):
Very nice! 💯📃

webapp/src/dogma/features/history/HistoryList.tsx
rainy789: This is just in case the same (pageIndex, pageSize) gets accessed multiple times.

webapp/src/pages/app/projects/[projectName]/repos/[repoName]/list/[revision]/[[...path]]/index.tsx
Comment (rainy789) [reply webapp/src/pages/app/projects/[projectName]/repos/[repoName]/list/[revision]/[[...path]]/index.tsx:165]: That will require some kind of pagination link for the datatable.

webapp/src/dogma/features/history/HistoryList.tsx
ikhoon: Did you intend to return the same instance for `{ pageIndex, pageSize }`?
`pagination` looks like a simple JSON object.

webapp/src/pages/app/projects/[projectName]/repos/[repoName]/list/[revision]/[[...path]]/index.tsx
Comment (ikhoon) [reply webapp/src/pages/app/projects/[projectName]/repos/[repoName]/list/[revision]/[[...path]]/index.tsx:165]: Agreed.

webapp/src/dogma/features/history/HistoryList.tsx
rainy789: Found the example https://tanstack.com/table/v8/docs/examples/react/pagination-controlled. It looks like `useMemo` is used to stabilise the table between rerendering.

rainy789: Without useMemo
Screen Shot 2023-01-28 at 00 42 36

With useMemo 
Screen Shot 2023-01-28 at 00 34 11

ikhoon: TIL. 📝

PR #794, opened 2023-01-30
Display application tokens

rainy789:
## Motivation
Display all tokens.

## Modification
- Add `getTokens` endpoint to `apiSlice`.
- Add NewToken component.
- When `secret` is available, allow viewing/copying the token.
- Add some mocked data.

## Result
- `npm run mock`
- Navigate to `http://localhost:3000/app/settings/tokens`

Screen Shot 2023-01-30 at 08 23 18

### DIFF
--- a/webapp/src/dogma/common/components/Navbar.tsx
+++ b/webapp/src/dogma/common/components/Navbar.tsx
@@ -180,8 +180,9 @@ export const Navbar = () => {
                 <Avatar name={user.login} size="sm" />
               </MenuButton>
               <MenuList>
-                <MenuItem>Application tokens</MenuItem>
-                <MenuItem>Add ...</MenuItem>
+                <MenuItem>
+                  <RouteLink href="/app/settings/tokens">Application tokens</RouteLink>
+                </MenuItem>
                 <MenuDivider />
                 <MenuItem
                   onClick={async () => {
--- a/webapp/src/dogma/common/components/table/DataTableClientPagination.tsx
+++ b/webapp/src/dogma/common/components/table/DataTableClientPagination.tsx
@@ -10,9 +10,9 @@ import {
   useReactTable,
 } from '@tanstack/react-table';
 import { DataTable } from 'dogma/common/components/table/DataTable';
-import { DebouncedInput } from 'dogma/common/components/table/DebouncedInput';
+import { Filter } from 'dogma/common/components/table/Filter';
 import { PaginationBar } from 'dogma/common/components/table/PaginationBar';
-import { useMemo, useState } from 'react';
+import { useState } from 'react';
 
 export type DataTableClientPaginationProps<Data extends object> = {
   data: Data[];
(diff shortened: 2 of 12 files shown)


webapp/src/dogma/common/components/Navbar.tsx
```diff
@@ -180,8 +180,9 @@ export const Navbar = () => {
                 <Avatar name={user.login} size="sm" />
               </MenuButton>
               <MenuList>
-                <MenuItem>Application tokens</MenuItem>
-                <MenuItem>Add ...</MenuItem>
+                <MenuItem>
+                  <RouteLink href="/app/settings/tokens">Application tokens</RouteLink>
```
ikhoon, line 184: Should we use `ChakraLink`?

webapp/src/pages/api/v1/tokens.ts
```diff
@@ -0,0 +1,43 @@
...
+};
+
+const tokenList: TokenDto[] = [];
+
+const makeData = (len: number) => {
+  for (let i = len; i > 0; i--) {
+    tokenList.push(newToken(i));
+  }
+};
+makeData(500);
```
ikhoon, line 26: When 500 tokens are rendered, the response time for scroll, filter and so on looks slow.
🤔
Is it possible to apply an infinity-style scroll.

webapp/src/dogma/common/components/Navbar.tsx
rainy789: `ChakraLink` shows an underline on hover. We may not need that on the nav menu.

ikhoon: 👍 Thanks for the explanation.

webapp/src/pages/api/v1/tokens.ts
rainy789: Added the client side pagination. The only thing is resetting the filter needs to be done manually (using a button).

Screen Shot 2023-02-01 at 14 05 27

ikhoon: Wow. You make it so fast and the page is rendered fast as well.

I was wondering if it is possible to reset the filter when all search keywords are removed.
Additionally, I found out that the pagination moved back to the first page when I click next page after filtering tokens.

webapp/src/dogma/common/components/table/DataTableClientPagination.tsx
```diff
@@ -0,0 +1,76 @@
...
+      <datalist id={column.id + 'list'}>
+        {sortedUniqueValues.slice(0, 5000).map((value: string | number) => (
+          <option value={value} key={value} />
+        ))}
+      </datalist>
+      <Flex gap={2}>
+        <DebouncedInput
+          type="text"
+          value={(columnFilterValue ?? '') as string}
+          onChange={(value) => value && column.setFilterValue(value)}
```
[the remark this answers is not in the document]
rainy789: If we do this the pagination (navigating to another page) won't work 😅

webapp/src/dogma/common/components/table/DataTableClientPagination.tsx
```diff
@@ -0,0 +1,76 @@
...
+export type DynamicDataTableProps<Data extends object> = {
+  data: Data[];
+  columns: ColumnDef<Data>[];
+  clientPagination?: boolean;
+};
+
+export const DataTableClientPagination = <Data extends object>({
+  data,
+  columns,
+  clientPagination = false,
```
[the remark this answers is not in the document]
rainy789: I've removed `clientPagination` since we can use `DynamicDataTable` component (and not use the controlled pagination) instead.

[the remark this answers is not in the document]
rainy789: In summary
- No pagination -> use `DynamicDataTable`
- Server side pagination -> use `DynamicDataTable` and specify the following props

```
  pagination?: { pageIndex: number; pageSize: number };
  setPagination?: Dispatch<SetStateAction<PaginationState>>;
  pageCount?: number;
```
- Client side pagination -> use `DataTableClientPagination`

webapp/src/pages/api/v1/tokens.ts
rainy789: In order for the pagination to not move back to the first page, we need the following line `value && column.setFilterValue(value)}`. 
Unfortunately, because of this work around we have to reset the filter by clicking the reset button.

rainy789: Found another way that doesn't require a reset button. 😅 Removed the button.

ikhoon: Nice! 😀

webapp/src/dogma/common/components/table/DataTableClientPagination.tsx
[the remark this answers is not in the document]
ikhoon: It makes sense.

ikhoon (approved):
It is so nice! 💯 ⭐

PR #802, opened 2023-02-14
Allow to set ZooKeeper quorumListenOnAllIPs

### LINKED ISSUE #753: Replication configuration additionalProperties
Hello,

I've been trying to composing my conf/dogma.json for replication following the wiki
(https://line.github.io/centraldogma/setup-configuration.html)

I would like to use `clientPortAddress` property of zookeeper in additionalProperties field,
but the field isn't added in `/opt/centraldogma/data/_zookeepr/config.properties`
ex) `"additionalProperties": { "clientPortAddress":"123.123.123.123" },`

Is there any solution to use `clientPortAddress` property to zookeeper config?

***
EDIT

Additionally, I attached the legal server format.
https://zookeeper.apache.org/doc/r3.5.3-beta/zookeeperReconfig.html#ch_reconfig_format
> Starting with 3.5.0 the clientPort and clientPortAddress configuration parameters should no longer be used. Instead, this information is now part of the server keyword specification, which becomes as follows:
server.\<positive id> = \<address1>:\<port1>:\<port2>[:role];[\<client port address>:]\<client port>

The zookeeper config file format which is made by dogma doesn't look like the legal format.

minwoox:
Motivation:
Users might want to set ZooKeeper `quorumListenOnAllIPs` that make ZooKeeper listen for connections on all available IP addresses.
https://zookeeper.apache.org/doc/r3.5.7/zookeeperAdmin.html

Modification:
- Propage `quorumListenOnAllIPs` that is configured using `additionalProperties` to ZooKeeper configuration.

Result:
- Close #753 
- You can now set ZooKeeper `quorumListenOnAllIPs` property to accept connections from all available IP addresses.


ikhoon:
Would it be worth listing it to https://github.com/line/centraldogma/blob/c8fc35ac9d2c3306a70e1cdca5470cc57125ad32/site/src/sphinx/setup-configuration.rst?plain=1#L349-L352

ikhoon (approved):
Thanks!

jrhee17 (approved):
Thanks @minwoox 🙇 👍 🙇

minwoox:
> Would it be worth listing it to

Thanks! Updated. 😉

ikhoon:
Very nice!

PR #805, opened 2023-02-15
Add project token mutation

rainy789:
## Motivation
Allow adding/deleting project tokens.

## Modification
- Add new token. Allow filter on the token dropdown. 
- Add delete token.
- Remove `RepoTokenList.test` due to side effect `useDeleteTokenMemberMutation`. For now, please rely on integration tests instead. 

## Result
- `npm run develop`
- Generate a token on `app/settings/tokens`
- Add a token on the project metadata page.

Screen Shot 2023-02-15 at 23 21 08


webapp/src/dogma/features/api/apiSlice.ts
```diff
@@ -134,6 +134,27 @@ export const apiSlice = createApi({
...
+      }),
+      invalidatesTags: ['Metadata'],
+    }),
+    deleteTokenMember: builder.mutation<void, DeleteRepoMemberDto>({
+      query: ({ projectName, id }) => ({
+        url: `/v1/metadata/${projectName}/tokens/${id}`,
+        method: 'DELETE',
+        headers: {
+          'Content-type': 'application/json; charset=UTF-8',
+        },
```
ikhoon, line 154:
```suggestion
```

webapp/src/dogma/features/repo/NewRepoToken.tsx
```diff
@@ -0,0 +1,128 @@
...
+}
+
+type FormData = {
+  appId: string;
+  role: string;
+};
+
+export const NewRepoToken = ({ projectName }: { projectName: string }) => {
+  const { data = [] } = useGetTokensQuery();
+  const tokenOptions: TokenOptionType[] = data
```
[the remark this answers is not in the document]
ikhoon: Not related to this PR but I found another null access error that happens in non-authentication mode.
image

ikhoon (commented):
The basic function looks good.

webapp/src/dogma/features/repo/NewRepoToken.tsx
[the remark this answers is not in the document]
rainy789: Good catch. 💯 Updated.

ikhoon:
If `npm run develop` is run with `SimpleCentralDomgaTestServer`, we can get the following error.
- Auth data is not given.
  image
- `/ap1/v1/tokens` will return 404 not found and it is regarded as an error. 404 is an expected status because `TokenService` is bound to the server when an auth provider is configured. 
image 

https://github.com/line/centraldogma/blob/a619e3551e2d275bb6fc174f573324161447ca6c/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java#L707-L714

If you want, I'm okay to handle them in a separate issue.

webapp/src/dogma/features/repo/NewRepoToken.tsx
```diff
@@ -0,0 +1,128 @@
...
+  value: string;
+  label: string;
+}
+
+type FormData = {
+  appId: string;
+  role: string;
+};
+
+export const NewRepoToken = ({ projectName }: { projectName: string }) => {
```
ikhoon, line 37: `RepoToken` is not a term we commonly use and it is a token used at a project level.  How about `AppToken` although other tabs start with `Repo` otherwise `ProjectToken`?

```suggestion
export const NewAppToken = ({ projectName }: { projectName: string }) => {
```

webapp/src/dogma/features/repo/RepoTokenList.tsx
```diff
@@ -1,28 +1,34 @@
...
+import { useDeleteTokenMemberMutation } from 'dogma/features/api/apiSlice';
+import { DeleteMember } from 'dogma/features/repo/DeleteMember';
 
 export type RepoTokenListProps<Data extends object> = {
   data: Data[];
+  projectName: string;
 };
 
-const RepoTokenList = <Data extends object>({ data }: RepoTokenListProps<Data>) => {
+const RepoTokenList = <Data extends object>({ data, projectName }: RepoTokenListProps<Data>) => {
```
ikhoon, line 16: How about `AppTokenList`?

rainy789:
>
> If you want, I'm okay to handle them in a separate issue.

Let's handle this separately.

ikhoon:
Sounds good.

webapp/src/dogma/features/repo/RepoTokenList.tsx
rainy789: Let's handle naming separately as well. I think we may need to update the other tabs.

ikhoon:
Speaking of `NewAppToken`, should we update the file name `NewRepoToken.tsx` and `import` statements as well? 😉

ikhoon (approved):
Very nice! 👍⭐️

Issue #659, opened 2021-12-24
Add useful factory methods to `PathPattern`


ikhoon:
`PathPattern` is introduced in #651 to create complex path patterns easily.
We will provide more common patterns as factory methods.
```java
// *.json or /**/foo.txt
PathPattern.endsWith("json");
PathPattern.endsWith("foo.txt");

// /foo/bar/**
PathPattern.under("/foo/bar"); // or 
PathPattern.startsWith("/foo/bar");

// Compose the existing path patterns
// /foo/bar/**/*.json
PathPattern.startsWith("/foo/bar").endsWith("json");
// or 
PathPattern.builder()
           .startsWith("/foo/bar")
           .endsWith("json")
           .build();


// *.json and /foo/bar/**
PathPattern.of(PatternPattern.of("json"), PathPattern.startsWith("/foo/bar"));
```

PR #809, opened 2023-02-22
Path pattern builder

noodlze:
Motivation:
Addresses https://github.com/line/centraldogma/issues/659.
 - Add `PathPattern#of` accepting  `PathPattern... pathPatterns` 
 ```// *.json and /foo/bar/**
PathPattern.of(PatternPattern.of("json"), PathPattern.startsWith("/foo/bar"));
```
- Add `PathPatternBuilder` to provide a way to combination many patterns into one.
The following initial options have been added:
```java
startsWith(String dirPath)

contains(String dirPath)

endsWith(String filename) 
//or
extension(String extension)
```
 
Modifications:
- `PathPatternOption`, `PathPathOptions` -> define and manage the options available.
- `PathPatternBuilder` -> responsible for composing one pathPattern from the added options.
   - Can only use one of  the options `endsWith` or `extension` ; currently, option added after will override previous 

Result:
```java
PathPattern.of(PatternPattern.of("json"), PathPattern.startsWith("/foo/bar"));

PathPattern.builder().startsWith("/foo/bar").build(); // "/foo/bar/**"

PathPattern.builder()
                    .startsWith("/foo/bar")
                    .endsWith("foo.txt")
                     .build(); // "/foo/bar/**/foo.txt"
                           
 PathPattern.builder()
                      .startsWith("/foo")
                      .contains("/bar/")
                      .extension("json")
                      .build() // "/foo/**/bar/**/*.json"

 PathPattern.builder()
                     .startsWith("/foo")
                     .endsWith("qux.json")
                     .extension("json")
                     .build() // "/foo/**/*.json"
```


common/src/main/java/com/linecorp/centraldogma/common/PathPatternBuilder.java
```diff
@@ -0,0 +1,137 @@
...
+    public PathPatternBuilder startsWith(String dirPath) {
+        put(PathPatternOptions.STARTS_WITH_OPTION, dirPath);
+        return this;
+    }
+
+    /**
+     * Adds {@link PathPatternOptions#CONTAINS_OPTION}.
+     */
+    public PathPatternBuilder contains(String dirPath) {
+        put(PathPatternOptions.CONTAINS_OPTION, dirPath);
```
ikhoon, line 77: We may allow multiple `contains` unlike other methods such as `startWith()` and `enddWith()`.

- `startsWith()` supports only one path. We can simply set the value as a string field.
`endsWith()` and `extension()` are mutually exclusive. We can make the methods override the same field.
```java
public final class PathPatternBulider {

    @Nullable
    private String startPattern;
    @Nullable
    private String endPattern;
    @Nullable
    private List<String> innerPatterns;

    public PathPatternBuilder startsWith(String dirPath) {
        // validation and normalize ...
        this.startPattern = dirPath;
    }


    public PathPatternBuilder extension(String extension) {
        // validation and normalize ...
        this.endPattern = normalize(extension);
    }
}
```

ikhoon (changes requested):

common/src/main/java/com/linecorp/centraldogma/common/PathPatternBuilder.java
```diff
@@ -0,0 +1,137 @@
...
+     * Returns a newly-created {@link PathPattern} based on the options of this builder.
+     */
+    public PathPattern build() {
+        checkArgument(!options.isEmpty(), "Requires at least one pattern to build in PathPatternBuilder");
+
+        if (options.size() == 1) {
+            return options.entrySet().iterator().next().getValue().getPathPattern();
+        }
+        // given the same precedence, option added after overrides previous one
+        final Map<Integer, PathPatternOption> optionByPrecedence = options.values()
```
noodlze, line 112: Hi, for this comment https://github.com/line/centraldogma/pull/809/files#r1115291121 :
 the reason I use `Integer` instead of primitive` int` for precedence is because I use Map to group options by precedence. If there are `contains` and `endsWith`, only one is allowed --> this part will select only one of the two based on which is added by user.

ikhoon: I see. However, `PathPatternOption.getPrecedence()` already returns `int`.

noodlze: Noted! Will fix `PathPatternOption.getPrecedence()`.

noodlze:
> We may allow multiple `contains` unlike other methods such as `startWith()` and `enddWith()`.
> 
> * `startsWith()` supports only one path. We can simply set the value as a string field.
>   `endsWith()` and `extension()` are mutually exclusive. We can make the methods override the same field.
> 
> ```java
> public final class PathPatternBulider {
> 
>     @Nullable
>     private String startPattern;
>     @Nullable
>     private String endPattern;
>     @Nullable
>     private List<String> innerPatterns;
> 
>     public PathPatternBuilder startsWith(String dirPath) {
>         // validation and normalize ...
>         this.startPattern = dirPath;
>     }
> 
> 
>     public PathPatternBuilder extension(String extension) {
>         // validation and normalize ...
>         this.endPattern = normalize(extension);
>     }
> }
> ```

As a result of moving the pattern build order logic from `PathPatternOption` to `PathPatternBuilder` , `precedence`  info stored in `PathPatternOption` is no longer needed.
Also, it's harder to determine whether `endsWith`/`extension` option has been added more than once with this new structure as adding either will simply override any existing value stored in `endPattern`. It's easy to do so for `startPattern` since there is only one option (`startsWith`). So in the end, for consistency, I found it simpler to just allow adding same option twice  to the builder; when we add the same option, it will just override existing.

common/src/main/java/com/linecorp/centraldogma/common/PathPatternBuilder.java
```diff
@@ -0,0 +1,126 @@
...
+import javax.annotation.Nullable;
+
+import com.google.common.collect.ImmutableSet;
+
+/**
+ * Builds a new {@link PathPattern}.
+ *
+ * <h2>Example</h2>
+ * <pre>{@code
+ * final PathPattern factory =
```
ikhoon, line 34:
```suggestion
 * final PathPattern pathPattern =
```

common/src/main/java/com/linecorp/centraldogma/common/PathPatternBuilder.java
```diff
@@ -0,0 +1,126 @@
...
+ *                    .build();
+ * }</pre>
+ */
+public final class PathPatternBuilder {
+    @Nullable
+    private PathPatternOption startPattern;
+    private final List<PathPatternOption> innerPatterns = new ArrayList<>();
+    @Nullable
+    private PathPatternOption endPattern;
+
```
ikhoon, line 48: A package-private constructor is necessary to prevent users from directly creating a new instance.
```suggestion
    PathPatternBuilder() {}
```

common/src/main/java/com/linecorp/centraldogma/common/PathPatternBuilder.java
```diff
@@ -0,0 +1,126 @@
...
+    public PathPatternBuilder endsWith(String filename) {
+        endPattern = PathPatternOptions.ENDS_WITH.apply(filename);
+        return this;
+    }
+
+    /**
+     * Adds {@link PathPatternOptions#EXTENSION}.
+     */
+    public PathPatternBuilder extension(String extension) {
+        endPattern = PathPatternOptions.EXTENSION.apply(extension);
```
ikhoon, line 61: Could we check null for all builder methods?
```suggestion
        requireNonNull(extension, "extension");
        endPattern = PathPatternOptions.EXTENSION.apply(extension);
```

common/src/main/java/com/linecorp/centraldogma/common/PathPatternBuilder.java
```diff
@@ -0,0 +1,126 @@
...
+    private static String combine(String left, String right) {
+        checkArgument(left.endsWith("/**"), "left should end with \"/**\"");
+        checkArgument(right.startsWith("/**/"), "right should start with \"/**/\"");
+        return left + right.substring(3);
+    }
+
+    /**
+     * Compose one pathPattern from a list of {@code patterns}.
+     */
+    private static String combine(Iterable<PathPattern> patterns) {
```
ikhoon, line 93: No reason to use `Iterable` for a private method.
```suggestion
    private static String combine(List<PathPattern> patterns) {
```

common/src/main/java/com/linecorp/centraldogma/common/PathPatternBuilder.java
```diff
@@ -0,0 +1,126 @@
...
+        final Iterator<PathPattern> iter = patterns.iterator();
+        String combinedPattern = iter.next().patternString();
```
ikhoon, line 95: Should we use `StringBuilder` for effiency?

common/src/main/java/com/linecorp/centraldogma/common/PathPatternBuilder.java
```diff
@@ -0,0 +1,126 @@
...
+            combinedPattern = combine(combinedPattern, iter.next().patternString());
+        }
+        return combinedPattern;
+    }
+
+    /**
+     * Returns a newly-created {@link PathPattern} based on the options of this builder.
+     */
+    public PathPattern build() {
+        final List<PathPatternOption> options = new ArrayList<>();
```
ikhoon, line 106: Code style: We prefer `ImmutableList.Builder`.

common/src/main/java/com/linecorp/centraldogma/common/PathPatternBuilder.java
```diff
@@ -0,0 +1,126 @@
...
+
+        checkArgument(!options.isEmpty(), "Requires at least one pattern to build in PathPatternBuilder");
+
+        if (options.size() == 1) {
+            return options.get(0).pathPattern();
+        }
+
+        final List<PathPattern> patterns = options.stream()
+                                                  .map(PathPatternOption::pathPattern)
+                                                  .collect(Collectors.toList());
```
ikhoon, line 123:
```suggestion
                                                  .collect(toImmutableList());
```

common/src/main/java/com/linecorp/centraldogma/common/PathPatternOption.java
```diff
@@ -0,0 +1,42 @@
...
+ * under the License.
+ */
+package com.linecorp.centraldogma.common;
+
+import java.util.function.Function;
+
+/**
+ * A {@link PathPatternBuilder} option.
+ */
+public class PathPatternOption {
```
ikhoon, line 23:
```suggestion
final class PathPatternOption {
```

common/src/main/java/com/linecorp/centraldogma/common/PathPatternOption.java
```diff
@@ -0,0 +1,42 @@
...
+    PathPatternOption(String pattern,
+                      Function<String, PathPattern> pathPatternCreator) {
+        this.pattern = pattern;
+        this.pathPatternCreator = pathPatternCreator;
+    }
+
+    /**
+     * Returns the {@link PathPattern} of the option.
+     */
+    public PathPattern pathPattern() {
```
ikhoon, line 39:
```suggestion
    PathPattern pathPattern() {
```

common/src/main/java/com/linecorp/centraldogma/common/PathPatternOptions.java
```diff
@@ -0,0 +1,95 @@
...
+    /**
+     * Appends "&#47;**" to {@code dirPath}.
+     * Returns the path pattern for matching all file(s) under {@code dirPath}.
+     */
+    public static final Function<String, PathPatternOption> STARTS_WITH = pattern ->
+            new PathPatternOption(pattern,
+                                  dirPath -> {
+                                      checkArgument(Util.isValidDirPath(dirPath), "dir");
+                                      return new DefaultPathPattern(ImmutableSet.of(
+                                              dirPath + (dirPath.endsWith("/") ? "" : "/") + "**"));
```
ikhoon, line 46: `DefaultPathPattern` does not hold patterns as `Set` so it looks inefficient to create a `Set`. What do you think of creating a constructor for a single string?

common/src/test/java/com/linecorp/centraldogma/common/PathPatternBuilderTest.java
```diff
@@ -0,0 +1,75 @@
...
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.linecorp.centraldogma.common;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.jupiter.api.Test;
+
+public class PathPatternBuilderTest {
```
ikhoon, line 22: `public` is not required for JUnit 5.
```suggestion
class PathPatternBuilderTest {
```

common/src/test/java/com/linecorp/centraldogma/common/PathPatternBuilderTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @Test
+    void testSingleOption() {
+        assertThat(PathPattern.builder().startsWith("/foo/bar").build()
+                              .patternString()).isEqualTo("/foo/bar/**");
+        assertThat(PathPattern.builder().endsWith("json").build()
+                              .patternString()).isEqualTo("/**/json");
```
ikhoon, line 29: Many users refer to the test cases to see the usage.
It would be helpful to add more realistic inputs.
```suggestion
        assertThat(PathPattern.builder().endsWith("foo.json").build()
                              .patternString()).isEqualTo("/**/foo.json");
```

common/src/main/java/com/linecorp/centraldogma/common/PathPatternOptions.java
```diff
@@ -0,0 +1,95 @@
...
+
+    /**
+     * Prepends "&#47;**&#47;" to {@code filename}.
+     * Returns the path pattern for matching file(s) ending in {@code filename}.
+     */
+    public static final Function<String, PathPatternOption> ENDS_WITH = pattern ->
+            new PathPatternOption(pattern,
+                                  filename -> {
+                                      checkArgument(Util.isValidFileName(filename), "filename");
+                                      return new DefaultPathPattern(ImmutableSet.of(filename));
```
ikhoon, line 71: Should we leave the comment on the line?
```suggestion
                                      // `/**` is added by the constructor of `DefaultPathPattern`
                                      return new DefaultPathPattern(ImmutableSet.of(filename));
```

ikhoon (commented):
Overall looks good. Left minor comments.

common/src/main/java/com/linecorp/centraldogma/common/PathPatternBuilder.java
noodlze: There is a check inside PathPatternOptions.EXTENSION method: 
```java
checkArgument(isValidFileExtension(extension), "extension");
```
which is why I did not add any checks for builder methods in PathPatternBuilder. 
Would you still like me to check for null in builder methods ?

common/src/main/java/com/linecorp/centraldogma/common/PathPattern.java
```diff
@@ -36,6 +37,13 @@
  */
 public interface PathPattern {
 
+    /**
+     *Returns a newly created {@link PathPatternBuilder}.
```
ikhoon, line 41:
```suggestion
     * Returns a newly created {@link PathPatternBuilder}.
```

common/src/main/java/com/linecorp/centraldogma/common/PathPatternBuilder.java
ikhoon: I think so. Because the cause of the exception is different. I'd like to raise `NullPointerException` if the given value is null.

common/src/main/java/com/linecorp/centraldogma/common/PathPatternBuilder.java
```diff
@@ -0,0 +1,125 @@
...
+        if (startPattern != null) {
+            optionsBuilder.add(startPattern);
+        }
+        optionsBuilder.addAll(innerPatterns);
+        if (endPattern != null) {
+            optionsBuilder.add(endPattern);
+        }
+        final ImmutableList<PathPatternOption> options = optionsBuilder.build();
+
+        checkArgument(!options.isEmpty(), "Requires at least one pattern to build in PathPatternBuilder");
```
ikhoon, line 114: If a method validates a value that the method directly takes, `IllegalArgumentException` would be fine. `options` is created from the values given before. In that case, we normally use `IllegalStateException`.
```suggestion
        checkState(!options.isEmpty(), "Requires at least one pattern to build in PathPatternBuilder");
```

common/src/main/java/com/linecorp/centraldogma/common/PathPatternOptions.java
```diff
@@ -0,0 +1,93 @@
...
+
+    private static final Pattern FILE_EXTENSION_PATTERN = Pattern.compile("^.{0,1}[0-9a-z]+$");
+
+    private PathPatternOptions() {}
+
+    /**
+     * Appends "&#47;**" to {@code dirPath}.
+     * Returns the path pattern for matching all file(s) under {@code dirPath}.
+     */
+    public static final Function<String, PathPatternOption> STARTS_WITH = pattern ->
```
ikhoon, line 39:
```suggestion
    static final Function<String, PathPatternOption> STARTS_WITH = pattern ->
```

common/src/main/java/com/linecorp/centraldogma/common/PathPatternOptions.java
```diff
@@ -0,0 +1,93 @@
...
+                                      checkArgument(Util.isValidDirPath(dirPath), "dir");
+                                      return new DefaultPathPattern(
+                                              dirPath + (dirPath.endsWith("/") ? "" : "/") + "**");
+                                  });
+
+    /**
+     * Prepends and appends "&#47;**" to target {@code dirPath}.
+     * Returns the path pattern for matching all file(s) containing {@code dirPath}.
+     */
+    public static final Function<String, PathPatternOption> CONTAINS = pattern ->
```
ikhoon, line 51:
```suggestion
    static final Function<String, PathPatternOption> CONTAINS = pattern ->
```

common/src/main/java/com/linecorp/centraldogma/common/PathPatternOptions.java
```diff
@@ -0,0 +1,93 @@
...
+                                      } else { // add ending slash to dirPath
+                                          return new DefaultPathPattern("/**" + dirPath + "/**");
+                                      }
+                                  });
+
+    /**
+     * Prepends "&#47;**&#47;" to {@code filename}.
+     * Returns the path pattern for matching file(s) ending in {@code filename}.
+     */
+    public static final Function<String, PathPatternOption> ENDS_WITH = pattern ->
```
ikhoon, line 66:
```suggestion
    static final Function<String, PathPatternOption> ENDS_WITH = pattern ->
```

common/src/main/java/com/linecorp/centraldogma/common/PathPatternOptions.java
```diff
@@ -0,0 +1,93 @@
...
+                                      checkArgument(Util.isValidFileName(filename), "filename");
+                                      // `/**` is added by the constructor of `DefaultPathPattern`
+                                      return new DefaultPathPattern(filename);
+                                  });
+
+    /**
+     * Prepends "&#47;**&#47;*" to {@code extension}.
+     * Returns the path pattern for matching file(s) ending in {@code extension}.
+     */
+    public static final Function<String, PathPatternOption> EXTENSION = pattern ->
```
ikhoon, line 78:
```suggestion
    static final Function<String, PathPatternOption> EXTENSION = pattern ->
```

ikhoon (approved):
Nice work! 🚀:tada:

common/src/main/java/com/linecorp/centraldogma/common/PathPattern.java
```diff
@@ -50,6 +58,14 @@ static PathPattern of(String... patterns) {
         return of(ImmutableSet.copyOf(requireNonNull(patterns, "patterns")));
     }
 
+    /**
+     * Creates a path pattern with the {@code pathPatterns}.
+     */
+    static PathPattern of(PathPattern... pathPatterns) {
+        requireNonNull(pathPatterns, "pathPatterns");
```
trustin, line 65: How about:
- throwing an IAE if `pathPatterns.length == 0`; and
- returning `pathPatterns[0]` if `pathPatterns.length == 1`?

common/src/main/java/com/linecorp/centraldogma/common/PathPattern.java
```diff
@@ -62,6 +78,34 @@ static PathPattern of(Iterable<String> patterns) {
...
+     * Returns the path pattern for file(s) that contains {@code dirPath}.
+     */
+    static PathPattern contains(String dirPath) {
+        return builder().contains(dirPath).build();
+    }
+
+    /**
+     * Returns the path pattern for file(s) with {@code extension}.
+     */
+    static PathPattern extension(String extension) {
```
trustin, line 105: `hasExtension` for consistency?

common/src/main/java/com/linecorp/centraldogma/common/PathPatternBuilder.java
```diff
@@ -0,0 +1,130 @@
...
+    public PathPatternBuilder endsWith(String filename) {
+        requireNonNull(filename, "filename");
+        endPattern = PathPatternOptions.ENDS_WITH.apply(filename);
+        return this;
+    }
+
+    /**
+     * Adds {@link PathPatternOptions#EXTENSION}.
+     */
+    public PathPatternBuilder extension(String extension) {
```
trustin, line 64: `hasExtension`

common/src/main/java/com/linecorp/centraldogma/common/PathPatternBuilder.java
```diff
@@ -0,0 +1,130 @@
...
+    @Nullable
+    private PathPatternOption startPattern;
+    private final List<PathPatternOption> innerPatterns = new ArrayList<>();
+    @Nullable
+    private PathPatternOption endPattern;
+
+    PathPatternBuilder() {}
+
+    /**
+     * Adds {@link PathPatternOptions#ENDS_WITH}.
```
trustin, line 53: `PathPatternOptions` is package private and thus cannot be used in public API Javadoc. Please write Javadoc without referring to it for all methods in this class. For example:

```
Ensures the last name component matches the specified {@code filename}.
For example, `endWith("foo.txt")` will match `/foo.txt`, `/alice/foo.txt` and
`/alice/bob/foo.txt`, but not `/barfoo.txt`.
```

Please note that the Javadoc contains some examples for a user to understand better.

common/src/main/java/com/linecorp/centraldogma/common/PathPatternOptions.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.regex.Pattern;
+
+import com.linecorp.centraldogma.internal.Util;
+
+/**
+ * A set of {@link PathPatternOption}s.
+ */
+final class PathPatternOptions {
+
+    private static final Pattern FILE_EXTENSION_PATTERN = Pattern.compile("^.{0,1}[0-9a-z]+$");
```
trustin, line 31: - `.` match any character. Should be `\\.`
  - Add more test case that covers this loophole. It shouldn't match `"가txt"`.
- What about upper-case extensions?
  - Add more test case.

common/src/main/java/com/linecorp/centraldogma/common/PathPattern.java
```diff
@@ -50,6 +59,17 @@ static PathPattern of(String... patterns) {
         return of(ImmutableSet.copyOf(requireNonNull(patterns, "patterns")));
     }
 
+    /**
+     * Creates a path pattern with the {@code pathPatterns}.
+     */
+    static PathPattern of(PathPattern... pathPatterns) {
+        checkArgument(pathPatterns.length > 0, "pathPatterns");
```
trustin, line 66:
```suggestion
        requireNonNull(pathPatterns, "pathPatterns");
        checkArgument(pathPatterns.length > 0, "pathPatterns is empty.");
```

common/src/main/java/com/linecorp/centraldogma/common/PathPattern.java
```diff
@@ -50,6 +59,17 @@ static PathPattern of(String... patterns) {
...
+        if (pathPatterns.length == 1) {
+            return pathPatterns[0];
+        }
+        return new DefaultPathPattern(ImmutableList.copyOf(pathPatterns));
```
trustin, line 70:
```suggestion
        return new DefaultPathPattern(ImmutableSet.copyOf(pathPatterns));
```

common/src/main/java/com/linecorp/centraldogma/common/PathPatternBuilder.java
```diff
@@ -0,0 +1,154 @@
...
+ *                    .hasExtension("json")
+ *                    .build();
+ * }</pre>
+ */
+public final class PathPatternBuilder {
+    @Nullable
+    private PathPatternOption startPattern;
+    private final List<PathPatternOption> innerPatterns = new ArrayList<>();
+    @Nullable
+    private PathPatternOption endPattern;
```
trustin, line 48: Can't we just use `PathPattern` here?

common/src/main/java/com/linecorp/centraldogma/common/PathPatternOption.java
```diff
@@ -0,0 +1,42 @@
...
+ * under the License.
+ */
+package com.linecorp.centraldogma.common;
+
+import java.util.function.Function;
+
+/**
+ * A {@link PathPatternBuilder} option.
+ */
+final class PathPatternOption {
```
trustin, line 23: I'm not sure if this class is needed at all. Why do we need to create a `PathPattern` lazily? Can't we just create it as early as possible?

common/src/main/java/com/linecorp/centraldogma/common/PathPattern.java
noodlze: Previously, a new constructor for `DefaultPathPattern(List<PathPattern> verifiedPatterns)` was introduced that would skip the validatePathPattern step. 
I changed the argument type from `List<PathPattern>` -> varargs `PathPattern...`.

common/src/main/java/com/linecorp/centraldogma/common/PathPatternOption.java
noodlze: I have merged `PathPatternOptions` and `PathPatternOption` with `PathPatternBuilder`

common/src/test/java/com/linecorp/centraldogma/common/PathPatternBuilderTest.java
```diff
@@ -0,0 +1,87 @@
...
+                              .contains("/foo")
+                              .contains("/bar")
+                              .build()
+                              .patternString()).isEqualTo("/**/foo/**/bar/**");
+        assertThat(PathPattern.builder()
+                              .startsWith("/foo/bar")
+                              .startsWith("/override")
+                              .hasExtension("json")
+                              .build()
+                              .patternString()).isEqualTo("/override/**/*.json");
```
trustin, line 85: I think this can be quite confusing to users. Should we always override, even for `contains()`?

noodlze: Hi @trustin , 
Your point about differing behavior for `contains`(combining allowed) vs other options (combining not allowed , will override) is valid.
I also understand @ikhoon  point's that, when considered in isolation, `contains` can be combined (ref https://github.com/line/centraldogma/pull/809#discussion_r1115309420 ).
Personally, I am on the fence, with no strong opinion.
Perhaps @minwoox , and @jrhee17 could share their opinion :)

hyrahul:
Hi all — I came across this PR while looking for a first contribution to Central Dogma. It looks like the PR was mostly approved back in March 2023, with the main open question being @trustin’s comment about `contains()` behaving differently from the other methods.

@noodlze, are you still planning to continue with this? If not, I’d be happy to take it over.

I also looked at how `AbstractCentralDogmaBuilder` handles similar cases:

- `host()` accumulates values, which is documented as “Adds...”.
- `useTls()` and `profileResources()` overwrite previous values when called again, without any guard.
- `checkState()` is used when two different methods conflict, such as `host()` and `profile()`.

Based on that, I think we can follow the same approach here:

- Keep the current accumulating behavior for `contains()`.
- Keep the current overwrite behavior for `startsWith()`, `endsWith()`, and `hasExtension()`.
- Add clearer Javadocs to make these behaviors explicit.
- The main issue is that `endsWith()` and `hasExtension()` currently update the same underlying constraint without checking for a conflict. Calling both can silently overwrite one another, similar to the `host()` / `profile()` case.
- I’d suggest adding a `checkState()` between those two methods.

Let me know if this approach makes sense, or if you’d prefer to make all four methods fully consistent instead.

PR #811, opened 2023-02-23
Allow configuring CORS origins in config file

### LINKED ISSUE #754: Provide a way to configure CORS policies
so that web applications hosted in a different origin can call Central Dogma REST API.

baont:
This is my implementation for the issue https://github.com/line/centraldogma/issues/754.

Motivation
Provide a way to configure CORS policies, so that web applications hosted in a different origin can call Central Dogma REST API.

Modifications
- Add `corsAllowedOrigins` to `CentralDogmaConfig`
- Add Cors decorator when setting up HttpApi

Result:
- Close #754
- You can now configure CORS policies in the config file.

### DIFF
--- a/it/src/test/java/com/linecorp/centraldogma/it/CorsPolicyTest.java
+++ b/it/src/test/java/com/linecorp/centraldogma/it/CorsPolicyTest.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2020 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.linecorp.centraldogma.it;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import com.linecorp.armeria.client.WebClient;
+import com.linecorp.armeria.common.AggregatedHttpResponse;
+import com.linecorp.armeria.common.HttpHeaderNames;
+import com.linecorp.armeria.common.ResponseHeaders;
+import com.linecorp.centraldogma.server.CentralDogmaBuilder;
+import com.linecorp.centraldogma.testing.junit.CentralDogmaExtension;
+
+class CorsPolicyTest {
+
+    @RegisterExtension
+    static final CentralDogmaExtension dogma = new CentralDogmaExtension() {
+        @Override
+        protected void configure(CentralDogmaBuilder builder) {
+            builder.corsAllowedOrigins("SomeOrigin","SomeOtherOrigin");
+        }
+    };
+
+    @Test
+    void testCorsHeadersAvailable() throws Exception {
+        final WebClient client = dogma.httpClient();
+
+        final AggregatedHttpResponse res = client.blocking().prepare()
+                           .header(HttpHeaderNames.ORIGIN.toString(), "SomeOrigin")
+                           .header(HttpHeaderNames.ACCESS_CONTROL_REQUEST_METHOD.toString(), "GET")
+                                                 .options("/api/v1/projects").execute();
+        final ResponseHeaders headers = res.headers();
+        Assertions.assertTrue(headers.contains("access-control-allow-origin", "SomeOrigin"));
+        Assertions.assertTrue(headers.contains("access-control-allow-credentials", "true"));
+        Assertions.assertTrue(headers.contains("access-control-max-age", "1800"));
+    }
+
+    @Test
+    void testCorsHeadersUnAvailable() throws Exception {
+        final WebClient client = dogma.httpClient();
+
+        final AggregatedHttpResponse res = client.blocking().prepare()
+                                                 .header(HttpHeaderNames.ORIGIN.toString(),
+                                                         "SomeOriginNotIncluded")
+                                                 .header(HttpHeaderNames.ACCESS_CONTROL_REQUEST_METHOD
+                                                                 .toString(), "GET")
+                                                 .options("/api/v1/projects").execute();
+        final ResponseHeaders headers = res.headers();
+        Assertions.assertFalse(headers.contains("access-control-allow-origin", "SomeOrigin"));
+        Assertions.assertFalse(headers.contains("access-control-allow-credentials", "true"));
+        Assertions.assertFalse(headers.contains("access-control-max-age", "1800"));
+    }
+}
--- a/it/src/test/java/com/linecorp/centraldogma/it/CorsPolicyTest.java
+++ b/it/src/test/java/com/linecorp/centraldogma/it/CorsPolicyTest.java
@@ -16,7 +16,11 @@
 
 package com.linecorp.centraldogma.it;
 
-import org.junit.jupiter.api.Assertions;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.entry;
+
+import java.util.Arrays;
+
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.RegisterExtension;
 
@@ -25,7 +29,10 @@ import com.linecorp.armeria.common.AggregatedHttpResponse;
 import com.linecorp.armeria.common.HttpHeaderNames;
 import com.linecorp.armeria.common.ResponseHeaders;
 import com.linecorp.centraldogma.server.CentralDogmaBuilder;
+import com.linecorp.centraldogma.server.CorsConfig;
 import com.linecorp.centraldogma.testing.junit.CentralDogmaExtension;
+
+import io.netty.util.AsciiString;
 
 class CorsPolicyTest {
 
@@ -33,7 +40,7 @@ class CorsPolicyTest {
     static final CentralDogmaExtension dogma = new CentralDogmaExtension() {
         @Override
         protected void configure(CentralDogmaBuilder builder) {
-            builder.corsAllowedOrigins("SomeOrigin","SomeOtherOrigin");
+            builder.cors(new CorsConfig(Arrays.asList("SomeOrigin", "AnotherOrigin"), 1800));
         }
     };
 
@@ -41,29 +48,35 @@ class CorsPolicyTest {
     void testCorsHeadersAvailable() throws Exception {
         final WebClient client = dogma.httpClient();
 
-        final AggregatedHttpResponse res = client.blocking().prepare()
-                           .header(HttpHeaderNames.ORIGIN.toString(), "SomeOrigin")
-                           .header(HttpHeaderNames.ACCESS_CONTROL_REQUEST_METHOD.toString(), "GET")
-                                                 .options("/api/v1/projects").execute();
+        final AggregatedHttpResponse res =
+                client.blocking()
+                      .prepare()
+                      .header(HttpHeaderNames.ORIGIN.toString(), "SomeOrigin")
+                      .header(HttpHeaderNames.ACCESS_CONTROL_REQUEST_METHOD, "GET")
+                      .options("/api/v1/projects")
+                      .execute();
         final ResponseHeaders headers = res.headers();
-        Assertions.assertTrue(headers.contains("access-control-allow-origin", "SomeOrigin"));
-        Assertions.assertTrue(headers.contains("access-control-allow-credentials", "true"));
-        Assertions.assertTrue(headers.contains("access-control-max-age", "1800"));
+
+        assertThat(headers).contains(entry(AsciiString.of("access-control-allow-origin"), "SomeOrigin"));
+        assertThat(headers).contains(entry(AsciiString.of("access-control-allow-credentials"), "true"));
+        assertThat(headers).contains(entry(AsciiString.of("access-control-max-age"), "1800"));
     }
 
     @Test
     void testCorsHeadersUnAvailable() throws Exception {
         final WebClient client = dogma.httpClient();
 
-        final AggregatedHttpResponse res = client.blocking().prepare()
-                                                 .header(HttpHeaderNames.ORIGIN.toString(),
-                                                         "SomeOriginNotIncluded")
-                                                 .header(HttpHeaderNames.ACCESS_CONTROL_REQUEST_METHOD
-                                                                 .toString(), "GET")
-                                                 .options("/api/v1/projects").execute();
+        final AggregatedHttpResponse res =
+                client.blocking()
+                      .prepare()
+                      .header(HttpHeaderNames.ORIGIN.toString(), "SomeOriginNotIncluded")
+                      .header(HttpHeaderNames.ACCESS_CONTROL_REQUEST_METHOD, "GET")
+                      .options("/api/v1/projects")
+                      .execute();
         final ResponseHeaders headers = res.headers();
-        Assertions.assertFalse(headers.contains("access-control-allow-origin", "SomeOrigin"));
-        Assertions.assertFalse(headers.contains("access-control-allow-credentials", "true"));
-        Assertions.assertFalse(headers.contains("access-control-max-age", "1800"));
+
+        assertThat(headers.get("access-control-allow-origin")).isNull();
+        assertThat(headers.get("access-control-allow-credentials")).isNull();
+        assertThat(headers.get("access-control-max-age")).isNull();
     }
 }
--- a/it/src/test/java/com/linecorp/centraldogma/it/CorsPolicyTest.java
+++ b/it/src/test/java/com/linecorp/centraldogma/it/CorsPolicyTest.java
@@ -40,7 +40,7 @@ class CorsPolicyTest {
     static final CentralDogmaExtension dogma = new CentralDogmaExtension() {
         @Override
         protected void configure(CentralDogmaBuilder builder) {
-            builder.cors(new CorsConfig(ImmutableList.of("SomeOrigin", "AnotherOrigin"), 1800));
+            builder.cors(new CorsConfig(ImmutableList.of("SomeOrigin"), 1800));
         }
     };
 
--- a/it/src/test/java/com/linecorp/centraldogma/it/CorsPolicyTest.java
+++ b/it/src/test/java/com/linecorp/centraldogma/it/CorsPolicyTest.java
@@ -19,10 +19,10 @@ package com.linecorp.centraldogma.it;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.entry;
 
-import java.util.Arrays;
-
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.RegisterExtension;
+
+import com.google.common.collect.ImmutableList;
 
 import com.linecorp.armeria.client.WebClient;
 import com.linecorp.armeria.common.AggregatedHttpResponse;
@@ -40,7 +40,7 @@ class CorsPolicyTest {
     static final CentralDogmaExtension dogma = new CentralDogmaExtension() {
         @Override
         protected void configure(CentralDogmaBuilder builder) {
-            builder.cors(new CorsConfig(Arrays.asList("SomeOrigin", "AnotherOrigin"), 1800));
+            builder.cors(new CorsConfig(ImmutableList.of("SomeOrigin", "AnotherOrigin"), 1800));
         }
     };
 
@@ -51,7 +51,7 @@ class CorsPolicyTest {
         final AggregatedHttpResponse res =
                 client.blocking()
                       .prepare()
-                      .header(HttpHeaderNames.ORIGIN.toString(), "SomeOrigin")
+                      .header(HttpHeaderNames.ORIGIN, "SomeOrigin")
                       .header(HttpHeaderNames.ACCESS_CONTROL_REQUEST_METHOD, "GET")
                       .options("/api/v1/projects")
                       .execute();
@@ -69,7 +69,7 @@ class CorsPolicyTest {
         final AggregatedHttpResponse res =
                 client.blocking()
                       .prepare()
-                      .header(HttpHeaderNames.ORIGIN.toString(), "SomeOriginNotIncluded")
+                      .header(HttpHeaderNames.ORIGIN, "SomeOriginNotIncluded")
                       .header(HttpHeaderNames.ACCESS_CONTROL_REQUEST_METHOD, "GET")
                       .options("/api/v1/projects")
                       .execute();
--- a/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
@@ -512,7 +512,6 @@ public class CentralDogma implements AutoCloseable {
         sb.clientAddressSources(cfg.clientAddressSourceList());
         sb.clientAddressTrustedProxyFilter(cfg.trustedProxyAddressPredicate());
 
-        // Config cors policy
         configCors(sb, config().corsConfig());
 
         cfg.numWorkers().ifPresent(
--- a/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
@@ -70,6 +70,7 @@ import com.linecorp.armeria.common.Flags;
 import com.linecorp.armeria.common.HttpData;
 import com.linecorp.armeria.common.HttpHeaderNames;
 import com.linecorp.armeria.common.HttpHeaders;
+import com.linecorp.armeria.common.HttpMethod;
 import com.linecorp.armeria.common.HttpRequest;
 import com.linecorp.armeria.common.HttpResponse;
 import com.linecorp.armeria.common.HttpStatus;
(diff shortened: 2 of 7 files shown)


server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -762,6 +772,25 @@ public String serviceName(ServiceRequestContext ctx) {
...
+        }
+
+        CorsServiceBuilder corsServiceBuilder;
+        if (allowedOrigins.contains("*")) {
+            corsServiceBuilder = CorsService.builderForAnyOrigin();
+        } else {
+            corsServiceBuilder = CorsService.builder(allowedOrigins);
+        }
+        return corsServiceBuilder
+                       .allowRequestMethods(HttpMethod.GET, HttpMethod.POST, HttpMethod.OPTIONS)
```
ikhoon, line 787: Central Dogma REST API also uses `DELETE` and `PATCH`. `HttpMethod.knownMethod()` for simplicity.

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -762,6 +772,25 @@ public String serviceName(ServiceRequestContext ctx) {
         }
     }
 
+    private static Function<? super HttpService, CorsService> configCorsDecorator(List<String> allowedOrigins) {
+        if (allowedOrigins == null || allowedOrigins.isEmpty()) {
+            return null;
```
ikhoon, line 777: We use `@Nullable` if a type can be null.
```suggestion
    @Nullable
    private static Function<? super HttpService, CorsService> configCorsDecorator(List<String> allowedOrigins) {
        if (allowedOrigins == null || allowedOrigins.isEmpty()) {
            return null;
```

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -681,6 +684,13 @@ protected HttpResponse doGet(ServiceRequestContext ctx, HttpRequest req) {
         // Enable content compression for API responses.
         decorator = decorator.andThen(contentEncodingDecorator());
 
+        // Config cors policy
+        Function<? super HttpService, CorsService> corsDecorator =
+                configCorsDecorator(config().corsAllowedOrigins());
+        if (corsDecorator != null) {
+            decorator = decorator.andThen(corsDecorator);
+        }
```
ikhoon, line 692: How about taking `decorator` as a parameter of `configCorsDecorator()` method?
```
maybeConfigCors(config().corsAllowedOrigins(), decorator);
```

baont: i don't think that would work, `andThen` returns a new function, not altering the input function

it/src/test/java/com/linecorp/centraldogma/it/CorsPolicyTest.java
```diff
@@ -0,0 +1,54 @@
...
+    void updateWriteQuota() throws Exception {
+        WebClient client = dogma.httpClient();
+
+        QueryParams params =
+                QueryParams.builder()
+                           .set(HttpHeaderNames.ORIGIN.toString(), "SomeOrigin")
+                           .set(HttpHeaderNames.ACCESS_CONTROL_REQUEST_METHOD.toString(), "GET")
+                           .build();
+        final AggregatedHttpResponse res = client
+                .options("/api/v1/projects", params).aggregate().join();
```
jrhee17, line 51: I think the issue is we're setting query parameters instead of the headers.

e.g.
```
final AggregatedHttpResponse res = client.blocking().prepare()
                                         .header(HttpHeaderNames.ORIGIN.toString(), "SomeOrigin")
                                         .header(HttpHeaderNames.ACCESS_CONTROL_REQUEST_METHOD.toString(), "GET")
                                         .options("/api/v1/projects").execute()
```

server/src/test/java/com/linecorp/centraldogma/server/CorsPolicyTest.java
```diff
@@ -0,0 +1,102 @@
...
+import com.linecorp.armeria.common.QueryParams;
+import com.linecorp.armeria.common.SessionProtocol;
+import com.linecorp.centraldogma.server.auth.AuthProviderFactory;
+import com.linecorp.centraldogma.testing.internal.FlakyTest;
+import com.linecorp.centraldogma.testing.internal.TemporaryFolderExtension;
+import com.linecorp.centraldogma.testing.internal.auth.TestAuthMessageUtil;
+import com.linecorp.centraldogma.testing.internal.auth.TestAuthProviderFactory;
+
+@FlakyTest
+class CorsPolicyTest {
```
jrhee17, line 46: I don't think we need to check if replicated mode also supports CORS since the code path is nearly identical 😅 
Sorry, but what do you think of removing this test?

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
baont: done

it/src/test/java/com/linecorp/centraldogma/it/CorsPolicyTest.java
baont: thanks, the fix works

server/src/test/java/com/linecorp/centraldogma/server/CorsPolicyTest.java
baont: test is removed

it/src/test/java/com/linecorp/centraldogma/it/CorsPolicyTest.java
```diff
@@ -0,0 +1,69 @@
...
+        final WebClient client = dogma.httpClient();
+
+        final AggregatedHttpResponse res = client.blocking().prepare()
+                           .header(HttpHeaderNames.ORIGIN.toString(), "SomeOrigin")
+                           .header(HttpHeaderNames.ACCESS_CONTROL_REQUEST_METHOD.toString(), "GET")
+                                                 .options("/api/v1/projects").execute();
+        final ResponseHeaders headers = res.headers();
+        Assertions.assertTrue(headers.contains("access-control-allow-origin", "SomeOrigin"));
+        Assertions.assertTrue(headers.contains("access-control-allow-credentials", "true"));
+        Assertions.assertTrue(headers.contains("access-control-max-age", "1800"));
```
ikhoon, line 51: Code style: Prefer to use AssertJ like `assertThat(...).isTrue()`.

it/src/test/java/com/linecorp/centraldogma/it/CorsPolicyTest.java
```diff
@@ -0,0 +1,69 @@
...
+    };
+
+    @Test
+    void testCorsHeadersAvailable() throws Exception {
+        final WebClient client = dogma.httpClient();
+
+        final AggregatedHttpResponse res = client.blocking().prepare()
+                           .header(HttpHeaderNames.ORIGIN.toString(), "SomeOrigin")
+                           .header(HttpHeaderNames.ACCESS_CONTROL_REQUEST_METHOD.toString(), "GET")
+                                                 .options("/api/v1/projects").execute();
```
ikhoon, line 47: - Could we add a line break after `res =` and line `.options()`?
- `.toString()` is redundant since `.header()` takes `CharSequence`.
```suggestion
        final AggregatedHttpResponse res = 
                client.blocking()
                      .prepare()
                      .header(HttpHeaderNames.ORIGIN, "SomeOrigin")
                      .header(HttpHeaderNames.ACCESS_CONTROL_REQUEST_METHOD, "GET")
                      .options("/api/v1/projects")
                      .execute();
```

it/src/test/java/com/linecorp/centraldogma/it/CorsPolicyTest.java
```diff
@@ -0,0 +1,69 @@
...
+    @Test
+    void testCorsHeadersUnAvailable() throws Exception {
+        final WebClient client = dogma.httpClient();
+
+        final AggregatedHttpResponse res = client.blocking().prepare()
+                                                 .header(HttpHeaderNames.ORIGIN.toString(),
+                                                         "SomeOriginNotIncluded")
+                                                 .header(HttpHeaderNames.ACCESS_CONTROL_REQUEST_METHOD
+                                                                 .toString(), "GET")
+                                                 .options("/api/v1/projects").execute();
```
ikhoon, line 63: Ditto

it/src/test/java/com/linecorp/centraldogma/it/CorsPolicyTest.java
```diff
@@ -0,0 +1,69 @@
...
+        final ResponseHeaders headers = res.headers();
+        Assertions.assertFalse(headers.contains("access-control-allow-origin", "SomeOrigin"));
+        Assertions.assertFalse(headers.contains("access-control-allow-credentials", "true"));
+        Assertions.assertFalse(headers.contains("access-control-max-age", "1800"));
```
ikhoon, line 67: Use `assertThat(...).isFalse()`?

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -762,6 +771,20 @@ public String serviceName(ServiceRequestContext ctx) {
...
+    private static Function<? super HttpService, CorsService> configCorsDecorator(List<String> allowedOrigins) {
+        if (allowedOrigins == null || allowedOrigins.isEmpty()) {
+            return null;
+        }
+
+        return CorsService.builder(allowedOrigins)
+                       .allowRequestMethods(HttpMethod.knownMethods())
+                       .allowAllRequestHeaders(true)
+                       .allowCredentials()
+                       .maxAge(1800)
```
ikhoon, line 784: Where did `1800` come from? We may expose the value as a configuration and could note the default value in the documentation.

I prefer to follow the default value of Chrome which is the most popular browser although I'm using Firefox. 🦊
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Max-Age#directives

ikhoon: The JSON configuration may need changing like:
```json
"cors": {
   // It should be nice if we both support string and array of strings.
   "allowedOrigins": "*",
   "allowedOrigins": ["foo.com", "bar.com"],
   "maxAge": 1800  // optional. If not set, 7200 is used by default?
}
```

it/src/test/java/com/linecorp/centraldogma/it/CorsPolicyTest.java
```diff
@@ -0,0 +1,69 @@
...
+    @Test
+    void testCorsHeadersAvailable() throws Exception {
+        final WebClient client = dogma.httpClient();
+
+        final AggregatedHttpResponse res = client.blocking().prepare()
+                           .header(HttpHeaderNames.ORIGIN.toString(), "SomeOrigin")
+                           .header(HttpHeaderNames.ACCESS_CONTROL_REQUEST_METHOD.toString(), "GET")
+                                                 .options("/api/v1/projects").execute();
+        final ResponseHeaders headers = res.headers();
+        Assertions.assertTrue(headers.contains("access-control-allow-origin", "SomeOrigin"));
```
jrhee17, line 49: We prefer assertj because it provides richer debug messages on failures.
Ditto for the other assertions

```suggestion
        assertThat(headers).contains(entry(AsciiString.of("access-control-allow-origin"), "SomeOrigin"));
```

it/src/test/java/com/linecorp/centraldogma/it/CorsPolicyTest.java
```diff
@@ -0,0 +1,69 @@
...
+        final WebClient client = dogma.httpClient();
+
+        final AggregatedHttpResponse res = client.blocking().prepare()
+                                                 .header(HttpHeaderNames.ORIGIN.toString(),
+                                                         "SomeOriginNotIncluded")
+                                                 .header(HttpHeaderNames.ACCESS_CONTROL_REQUEST_METHOD
+                                                                 .toString(), "GET")
+                                                 .options("/api/v1/projects").execute();
+        final ResponseHeaders headers = res.headers();
+        Assertions.assertFalse(headers.contains("access-control-allow-origin", "SomeOrigin"));
```
jrhee17, line 65: I think it would be a strong test if we confirm that the header doesn't exist at all
Ditto for the other asssertions.

```suggestion
        assertThat(headers.get("access-control-allow-origin")).isNull();
```

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
baont: i follow spring default configuration at [https://github.com/spring-projects/spring-framework/blob/3854861a8a71da97a8275c76ea81ad04025da9ae/spring-web/src/main/java/org/springframework/web/cors/CorsConfiguration.java#L499](url). Will make this configurable as you suggest

jrhee17 (commented):
Looks good!     I'm guessing with the next iteration of reviews this PR should be good for merging 👍

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -681,6 +683,13 @@ protected HttpResponse doGet(ServiceRequestContext ctx, HttpRequest req) {
         // Enable content compression for API responses.
         decorator = decorator.andThen(contentEncodingDecorator());
 
+        // Config cors policy
+        final Function<? super HttpService, CorsService> corsDecorator =
```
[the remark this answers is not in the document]
baont: that makes sense, i will make the decorator global

it/src/test/java/com/linecorp/centraldogma/it/CorsPolicyTest.java
```diff
@@ -0,0 +1,82 @@
...
+    };
+
+    @Test
+    void testCorsHeadersAvailable() throws Exception {
+        final WebClient client = dogma.httpClient();
+
+        final AggregatedHttpResponse res =
+                client.blocking()
+                      .prepare()
+                      .header(HttpHeaderNames.ORIGIN.toString(), "SomeOrigin")
```
ikhoon, line 54:
```suggestion
                      .header(HttpHeaderNames.ORIGIN, "SomeOrigin")
```

server/src/main/java/com/linecorp/centraldogma/server/CorsConfig.java
```diff
@@ -0,0 +1,90 @@
...
+                    "allowedOrigins: " + allowedOrigins +
+                    " (expected: either a string or an array of strings)");
+        }
+        if (this.allowedOrigins.isEmpty()) {
+            throw new IllegalArgumentException(
+                    "allowedOrigins: " + allowedOrigins +
+                    " (expected: the list of origins must not be empty)");
+        }
+
+        this.maxAge = maxAge == null || maxAge <= 0 ? DEFAULT_MAX_AGE : maxAge;
```
ikhoon, line 63: `-1` or `0` could mean an unlimited age. If we don't provide the unlimited age, I prefer disallowing `0` and a negative number and throwing `IllegalArgumentException`.
What do you think of using the default value if `maxAge` is null?

server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaBuilder.java
```diff
@@ -532,6 +536,14 @@ public CentralDogmaBuilder meterRegistry(MeterRegistry meterRegistry) {
         return this;
     }
 
+    /**
+     * Sets config for CORS policy.
+     */
+    public CentralDogmaBuilder cors(CorsConfig corsConfig) {
+        this.corsConfig = corsConfig;
```
ikhoon, line 543:
```suggestion
        requireNonNull(corsConfig, "corsConfig");
        this.corsConfig = corsConfig;
```

site/src/sphinx/setup-configuration.rst
```diff
@@ -48,11 +48,15 @@ defaults:
...
+        "requestQuota": 5,
         "timeWindowSeconds": 1
       },
       "accessLogFormat": "common",
-      "authentication": null
+      "authentication": null,
+      "cors": {
+         "allowedOrigins": ["http://example1.com", "http://example2.com"],
+         "maxAge": 1800
+      }
```
ikhoon, line 59: CORS is an optional feature that users should explicitly configure to use.
Some users copy the configuration without deep thought. For that reason, I'd like to make it null here. 
```suggestion
      "cors": null
```

server/src/test/java/com/linecorp/centraldogma/server/CentralDogmaConfigTest.java
```diff
@@ -240,4 +241,124 @@ void maxRemovedRepositoryAgeMillis_withNegativeValue() {
...
+            assertThat(cfg.corsConfig().maxAge()).isEqualTo(1200);
+        } else {
+            fail("corsConfig is null");
+        }
+    }
+
+    @Test
+    void corsConfig_withNullAllowOrigins() throws Exception {
+        assertThatThrownBy(() ->
+                                   Jackson.readValue("{\n" +
```
ikhoon, line 316: nit: Join two lines?

server/src/test/java/com/linecorp/centraldogma/server/CentralDogmaConfigTest.java
```diff
@@ -240,4 +241,124 @@ void maxRemovedRepositoryAgeMillis_withNegativeValue() {
...
+                                                     "  }\n" +
+                                                     '}',
+                                                     CentralDogmaConfig.class)
+        ).hasCauseInstanceOf(IllegalArgumentException.class);
+    }
+
+    @Test
+    void corsConfig_withEmptyAllowOrigins() throws Exception {
+        assertThatThrownBy(() ->
+                                   Jackson.readValue("{\n" +
```
ikhoon, line 342: nit: Join two lines?

ikhoon (commented):
Mostly looks good. Left only minor comments.

server/src/main/java/com/linecorp/centraldogma/server/CorsConfig.java
baont: After searching a bit, cache can not be unlimited since the browser caches per requests, not the entire domain. If cache is unlimited, browser would run out of RAM. I also could not find a reference to when to set cache to 0 for CORS. To conclude, i will disallow <=0 values and provide set the default cache time (7200s) if user does not set (aka maxAge is null).

site/src/sphinx/setup-configuration.rst
baont: i agree

server/src/main/java/com/linecorp/centraldogma/server/CorsConfig.java
```diff
@@ -0,0 +1,98 @@
...
+        if (maxAge == null) {
+            this.maxAge = DEFAULT_MAX_AGE;
+            return;
+        }
+        if (maxAge <= 0) {
+            throw new IllegalArgumentException(
+                    "maxAge: " + maxAge +
+                    " (expected: maxAge must be positive)");
+        }
+        this.maxAge = maxAge;
```
ikhoon, line 71: It is a bit awkward for me to use `return` statement in the conductor. If a new field is added and it is added to the later of the block, the value could not be set.
```java

if (maxAge == null) {
    maxAge = DEFAULT_MAX_AGE;
}
if (maxAge <= 0) {
    throw new IllegalArgumentException(
            "maxAge: " + maxAge +
            " (expected: maxAge must be positive)");
}
// Set the value after setting the default value and validation.
this.maxAge = maxAge;
```

ikhoon (approved):
Thanks, @baont! 🚀💯
Left a minor comment.

server/src/main/java/com/linecorp/centraldogma/server/CorsConfig.java
baont: thank, i will remove the `return`

site/src/sphinx/setup-configuration.rst
```diff
@@ -229,6 +230,19 @@ Core properties
...
 
+- ``cors``
+
+  - specify ``null`` to disable CORS entirely.
+  - ``allowedOrigins`` (string or string array)
+
+    - the list of origins which are allowed a lenient CORS policy. If the literal value ``*`` is specified then
+      all origins are allowed.
+
+  - ``maxAge`` (integer)
```
jrhee17, line 241: Can we also specify the units for the configuration? 

```suggestion
  - ``maxAgeSeconds`` (integer)
```

server/src/main/java/com/linecorp/centraldogma/server/CorsConfig.java
```diff
@@ -0,0 +1,97 @@
...
+     * Returns the list of origins which are allowed a CORS policy.
+     */
+    @JsonProperty
+    public List<String> allowedOrigins() {
+        return allowedOrigins;
+    }
+
+    /**
+     * Returns how long in seconds the results of a preflight request can be cached.
+     * If unspecified, the default of {@value #DEFAULT_MAX_AGE} is returned.
```
jrhee17, line 83: `DEFAULT_MAX_AGE` is private anyways so it won't show up in the javadocs

```suggestion
     * If unspecified, the default of {@code 7200} seconds is returned.
```

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -762,6 +771,21 @@ public String serviceName(ServiceRequestContext ctx) {
         }
     }
 
+    @Nullable
+    private static Function<? super HttpService, CorsService> configCorsDecorator(CorsConfig corsConfig) {
```
jrhee17, line 775:
```suggestion
    private static Function<? super HttpService, CorsService> configCorsDecorator(@Nullable CorsConfig corsConfig) {
```

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -762,6 +771,21 @@ public String serviceName(ServiceRequestContext ctx) {
...
+        if (corsConfig == null || corsConfig.allowedOrigins() == null ||
+            corsConfig.allowedOrigins().isEmpty()) {
```
jrhee17, line 777: I thought `corsConfig.allowedOrigins` couldn't be `null` nor `empty` due to validations in the constructor of `corsConfig`

```suggestion
        if (corsConfig == null) {
```

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -510,6 +512,13 @@ private Server startServer(ProjectManager pm, CommandExecutor executor,
         sb.clientAddressSources(cfg.clientAddressSourceList());
         sb.clientAddressTrustedProxyFilter(cfg.trustedProxyAddressPredicate());
 
+        // Config cors policy
```
jrhee17, line 515: This is optional, maybe it's just me but it feels cleaner to make a simple method call.
I think as-is is fine though

e.g.
```
private static void maybeConfigureCors(ServerBuilder sb, @Nullable CorsConfig corsConfig) {
    if (corsConfig == null || corsConfig.allowedOrigins() == null ||
        corsConfig.allowedOrigins().isEmpty()) {
        return;
    }

    sb.decorator(CorsService.builder(corsConfig.allowedOrigins())
                            .allowRequestMethods(HttpMethod.knownMethods())
                            .allowAllRequestHeaders(true)
                            .allowCredentials()
                            .maxAge(corsConfig.maxAge())
                            .newDecorator());
}
```

site/src/sphinx/setup-configuration.rst
```diff
@@ -229,6 +230,19 @@ Core properties
...
+
+    - how long in seconds the results of a preflight request can be cached, if not specified then the default
```
jrhee17, line 243:
```suggestion
    - how long in seconds the results of a preflight request can be cached. If not specified then the default
```

server/src/main/java/com/linecorp/centraldogma/server/CorsConfig.java
```diff
@@ -0,0 +1,97 @@
+/*
+ * Copyright 2018 LINE Corporation
```
jrhee17, line 2:
```suggestion
 * Copyright 2023 LINE Corporation
```

jrhee17 (commented):
I think this will be my last review.  Looking really nice 👍   Left only nit comments 🙇

server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaBuilder.java
```diff
@@ -532,6 +536,14 @@ public CentralDogmaBuilder meterRegistry(MeterRegistry meterRegistry) {
         return this;
     }
 
+    /**
+     * Sets config for CORS policy.
```
jrhee17, line 540:
```suggestion
     * Sets CORS related configurations.
```

server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaBuilder.java
```diff
@@ -126,6 +126,10 @@ public final class CentralDogmaBuilder {
     private int timeWindowSeconds;
     private MeterRegistry meterRegistry = Flags.meterRegistry();
 
+    // Cors config
```
jrhee17, line 129:
```suggestion
```

jrhee17 (approved):
Thanks for your patience @baont ! Looks good to me 👍 🚀 💯

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -510,6 +512,9 @@ private Server startServer(ProjectManager pm, CommandExecutor executor,
...
```
trustin, line 515: This comment is a kind of repetitive. The method name `configCors` itsels tells it all.
```suggestion
```

trustin (approved):
Excellent work, @baont! 🙇

jrhee17:
gentle ping @minwoox 😄

jrhee17:
gentle ping @minwoox 😄

minwoox (approved):
Sorry about the late review. Thanks a lot for adding this configuration! 😄

minwoox:
Thanks a lot, @baont! 🎉 🎉 🎉

PR #814, opened 2023-03-01
Fix to route a link to the specified page with client-side rendering

ikhoon:
Motivation:

Path patterns are used folder structures in Next.js such as '/app/projects/[projectNames]'. If '/app/projects/myProj' is requested to Central Dogma server, the server fails to find
'app/projects/myProj/index.html' file and returns 404 not found. This does not happen if a client app starts with `npm run xxx` commands because the links are handled by Next.js server and only REST APIs are requested to a Central Dogma server.

Modifications:

- Return `index.html` file as a fallback if the `FileService` for new webapp fails to find a file.
- Trigger Next.js router to route to the desired page when a page is loaded for the first time.

Result:

You can now visit a Central Dogma webapp with a deep link.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
@@ -386,13 +386,13 @@ public class CentralDogma implements AutoCloseable {
                 logger.info("Starting plugins on the leader replica ..");
                 pluginsForLeaderOnly
                         .start(cfg, pm, exec, meterRegistry, purgeWorker).handle((unused, cause) -> {
-                    if (cause == null) {
-                        logger.info("Started plugins on the leader replica.");
-                    } else {
-                        logger.error("Failed to start plugins on the leader replica..", cause);
-                    }
-                    return null;
-                });
+                            if (cause == null) {
+                                logger.info("Started plugins on the leader replica.");
+                            } else {
+                                logger.error("Failed to start plugins on the leader replica..", cause);
+                            }
+                            return null;
+                        });
             }
         };
 
@@ -744,15 +744,25 @@ public class CentralDogma implements AutoCloseable {
                 sb.service(LOGOUT_PATH, authProvider.webLogoutService());
             }
 
+            // Folder names contain path patterns such as `[projectName]` which FileService can't infer from
+            // the request path. Return `index.html` as a fallback so that Next.js client router handles the
+            // path patterns.
             final HttpService fallbackFileService = HttpFile.of(CentralDogma.class.getClassLoader(),
-                                                         "com/linecorp/centraldogma/webapp/index.html")
-                                                     .asService();
-            sb.serviceUnder("/",
-                            FileService.builder(CentralDogma.class.getClassLoader(),
-                                                "com/linecorp/centraldogma/webapp")
-                                       .cacheControl(ServerCacheControl.REVALIDATED)
-                                       .build()
-                                       .orElse(fallbackFileService));
+                                                                "com/linecorp/centraldogma/webapp/index.html")
+                                                            .asService();
+            sb.serviceUnder("/app", FileService.builder(CentralDogma.class.getClassLoader(),
+                                                        "com/linecorp/centraldogma/webapp/app")
+                                               .cacheControl(ServerCacheControl.REVALIDATED)
+                                               .build().orElse(fallbackFileService));
+
+            // Service all web resources except for '/app'
+            sb.route()
+              .pathPrefix("/")
+              .exclude("/app")
+              .build(FileService.builder(CentralDogma.class.getClassLoader(),
+                                         "com/linecorp/centraldogma/webapp")
+                                .cacheControl(ServerCacheControl.REVALIDATED)
+                                .build());
 
             sb.serviceUnder("/legacy-web",
                             FileService.builder(CentralDogma.class.getClassLoader(), "webapp")
--- a/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
@@ -755,7 +755,7 @@ public class CentralDogma implements AutoCloseable {
                                                .cacheControl(ServerCacheControl.REVALIDATED)
                                                .build().orElse(fallbackFileService));
 
-            // Service all web resources except for '/app'
+            // Serve all web resources except for '/app'.
             sb.route()
               .pathPrefix("/")
               .exclude("/app")
--- a/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
@@ -744,11 +744,15 @@ public class CentralDogma implements AutoCloseable {
                 sb.service(LOGOUT_PATH, authProvider.webLogoutService());
             }
 
+            final HttpService fallbackFile = HttpFile.of(CentralDogma.class.getClassLoader(),
+                                                         "com/linecorp/centraldogma/webapp/index.html")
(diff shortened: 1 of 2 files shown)


ikhoon:
@rainy789 Could you take a look if available? :wink:

webapp/src/pages/_app.tsx
```diff
@@ -20,8 +20,18 @@ type AppPropsWithLayout = AppProps & {
...
 
-const DogmaApp = ({ Component, pageProps }: AppPropsWithLayout) => {
+const DogmaApp = ({Component, pageProps}: AppPropsWithLayout) => {
   const router = useRouter();
+  useEffect(() => {
+    // Path patterns are used folder structures in Next.js such as '/app/projects/[projectNames]'.
+    // If '/app/projects/myProj' is requested to Central Dogma server, the server fails to find
+    // 'app/projects/myProj/index.html' file and returns 'index.html' as a fallback and the landing page is
+    // rendered instead. As a workaround, this triggers Next.js router to route to the desired page when a page
+    // is loaded for the first time.
```
jrhee17, line 30: > is loaded for the first time.

It seems like `router.asPath` is always pushed every time a state change is made. I'm wondering if this is the intended behavior?

jrhee17 (approved):
I don't want to be a blocker so approving 😄 

Also, when I tried running the webapp I got stuck at the following page. Any pointers would be helpful 🙏 
Screen Shot 2023-03-03 at 1 10 27 PM

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -744,11 +744,15 @@ public String serviceName(ServiceRequestContext ctx) {
...
+            final HttpService fallbackFile = HttpFile.of(CentralDogma.class.getClassLoader(),
+                                                         "com/linecorp/centraldogma/webapp/index.html")
+                                                     .asService();
             sb.serviceUnder("/",
                             FileService.builder(CentralDogma.class.getClassLoader(),
                                                 "com/linecorp/centraldogma/webapp")
                                        .cacheControl(ServerCacheControl.REVALIDATED)
-                                       .build());
+                                       .build()
+                                       .orElse(fallbackFile));
```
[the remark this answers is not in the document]
ikhoon: My explanation was insufficient. The actual HTML file path is `/app/projects/[projectName]/index.html` but the request path for the file is `/app/projects/myProj`.
FileService tries to find a matched file in `/app/projects/myProj` and `/app/projects/myProj/index.html`. So no page is rendered.

> Question) I think sometimes returning a 404 makes sense. 

There is a 404 page on the client-side. Next.js router may redirect to 404 if there are no matched routes.

webapp/src/pages/_app.tsx
ikhoon: This should be called only once when the web application is initialized.
Is this repeatedly called?

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -744,11 +744,15 @@ public String serviceName(ServiceRequestContext ctx) {
                 sb.service(LOGOUT_PATH, authProvider.webLogoutService());
             }
 
+            final HttpService fallbackFile = HttpFile.of(CentralDogma.class.getClassLoader(),
```
trustin, line 747: `fallbackFileService`?

webapp/src/pages/_app.tsx
```diff
@@ -20,8 +20,18 @@ type AppPropsWithLayout = AppProps & {
   Component: NextPageWithLayout;
 };
 
-const DogmaApp = ({ Component, pageProps }: AppPropsWithLayout) => {
+const DogmaApp = ({Component, pageProps}: AppPropsWithLayout) => {
   const router = useRouter();
+  useEffect(() => {
+    // Path patterns are used folder structures in Next.js such as '/app/projects/[projectNames]'.
```
trustin, line 26: Maybe rephrase?

ikhoon:
> Also, when I tried running the server with the built webapp I got stuck at the following page. Any pointers would be helpful 🙏

How did you start the test server?

ikhoon:
I found a bug that `secury_enabled` is handled by the file service.
We may need to set `app` as a context path for the file service.

ikhoon:
The bug has been fixed by [`90ce304` (#814)](https://github.com/line/centraldogma/pull/814/commits/90ce30461bcf9ca70007f5ef6b63afbf6abc1fa9)

trustin (approved):

webapp/src/pages/_app.tsx
```diff
@@ -22,6 +22,18 @@ type AppPropsWithLayout = AppProps & {
...
+  useEffect(() => {
+    // Next.js uses a path pattern to dynamically match the request path to a specific HTML file.
+    // For example, `/app/projects/[projectNames]/index.html' is generated to render `/app/projects/myProj`
+    // page. `FileService` serving the static resources for Central Dogma webapp does not understand the
+    // path pattern syntax in the folder names. If `/ap/projects/myProj` is requested, the FileService will try
+    // to find '/app/projects/myProj/index.html' as is, and 'index.html' is returned as a fallback. As a
+    // workaround, this triggers Next.js router to route to the desired page when a page is loaded for the
+    // first time.
+    if (router.asPath !== '/') {
+      Router.push(router.asPath);
```
rainy789, line 34:
```suggestion
      router.push(router.asPath);
```

webapp/src/pages/_app.tsx
```diff
@@ -4,8 +4,8 @@ import { Provider } from 'react-redux';
 import { ChakraProvider, theme } from '@chakra-ui/react';
 import { Authorized } from 'dogma/features/auth/Authorized';
 import { NextPage } from 'next';
-import { ReactElement, ReactNode } from 'react';
-import { useRouter } from 'next/router';
+import { ReactElement, ReactNode, useEffect } from 'react';
+import Router, { useRouter } from 'next/router';
 import { Layout } from 'dogma/common/components/Layout';
```
rainy789, line 9: We can also get `router` from `router = useRouter()`, they are actually similar. I think the difference is the non global `router` will return a different path when we call `router.asPath`.

rainy789:
> * Return `index.html` file as a fallback if the `FileService` for new webapp fails to find a file.

Question: can this be fixed by adding a trailing slash? https://nextjs.org/docs/api-reference/next.config.js/exportPathMap#adding-a-trailing-slash

webapp/src/pages/_app.tsx
jrhee17: I'm not sure, but aren't effects called every time a state is changed?

ikhoon: `[]` is set to the dependency of `useEffect()` so it should be run once.
https://css-tricks.com/run-useeffect-only-once/#aa-the-trick-is-that-useeffect-takes-a-second-parameter

webapp/src/pages/_app.tsx
ikhoon: `replace` looks better. Let me try to use it.

ikhoon:
It may be not related to a trailing slash. As you know, the HTML file for a project is stored at `/app/products/[projectName]/index.html`. If I directly type, click or refresh `https://mydogma.apps.com/app/products/my-project`, a Next.js server interprets the URL and then serves `/app/products/[projectName]/index.html` correctly from the file system.

However, Armeria's `FileService` does not understand the path pattern (`[projectName]`), so it just tries to find `/app/products/my-project` file or `/app/products/my-project/index.html` from the file system which does not exist. 

I set `app/index.html` as a fallback and trigger re-routing in the client-side when the page is first loaded.

trustin (approved):
Thanks, @ikhoon !

PR #816, opened 2023-03-02
Do not source system- and user- Git settings / Fix `GitMirrorAuthTest`

trustin:
Motivation:

We can't test Git-over-HTTPS authentication with `GitMirrorAuthTest` because of the following:

- JGit silently sources the custom configurations, breaking our expected behaviors:
  - `~/.gitconfig`
  - `~/.config/jgit`
  - `~/.jgitconfig`
  - `/etc/gitconfig`
- It's not convenient to pass Git username and password via system properties.

Modifications:

- Added `IsolatedSystemReader` that prevents reading the system- and user- Git settings so that Central Dogma doesn't behave differently between environments.
- Revamped `GitMirrorAuthTest` so that it always uses `github.com/line/centraldogma-authtest.git` as its test repository.
  - The Git-over-SSH test now always runs as a part of the build.
  - The Git-over-HTTPS test requires sensitive information, so it still has to run manually, although we could at least run it as a part of GitHub action (but not in a PR build).
- Updated `GitMirrorAuthTest` so that it pulls GitHub username and password via environment variables rather than system properties.

Result:

- Central Dogma doesn't behave differently between environments.
- `GitMirrorAuthTest` now always runs the Git-over-SSH test.
- `GitMirrorAuthTest` now uses environment variables rather than JVM system properties to retrieve GitHub username and password.


taiga-nada (approved):
LGTM. Thank you.

it/src/test/resources/com/linecorp/centraldogma/it/mirror/git/test.pem
```diff
@@ -0,0 +1,20 @@
+-----BEGIN EC PRIVATE KEY-----
```
jrhee17, line 1: Question) Just to make sure, is this a read-only deploy key?

jrhee17 (approved):
Left two super minor nits.  Thanks @trustin 👍 🙇 👍

ikhoon (approved):
Thanks, @trustin! 🙇‍♂️

server/src/main/java/com/linecorp/centraldogma/server/internal/IsolatedSystemReader.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    private static final class EmptyConfig extends FileBasedConfig {
+        EmptyConfig(Config base, FS fs) {
+            super(base, null, fs);
+        }
+
+        @Override
+        public void load() {
+        }
```
thachlp, line 101: How do you think adding a comment explaining `// Do nothing because of X and Y.`?

it/src/test/java/com/linecorp/centraldogma/it/mirror/git/GitMirrorAuthTest.java
```diff
@@ -62,76 +53,32 @@ protected void configure(CentralDogmaBuilder builder) {
...
      */
     @Nullable
-    private static final String GIT_PUBLIC_KEY = System.getProperty("git.publicKey");
-
-    /**
-     * The passphrase of the private key file.
-     */
-    private static final String GIT_PASSPHRASE = System.getProperty("git.passphrase");
+    private static final String GITHUB_PASSWORD =
+            System.getenv("GITHUB_CD_AUTHTEST_PASSWORD");
```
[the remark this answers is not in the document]
trustin: `user.name` is JDK standard property which is always given, whereas `user.password` is not. 😉 Let me leave it as it is.

it/src/test/resources/com/linecorp/centraldogma/it/mirror/git/test.pem
trustin: Yes, it's read-only.

server/src/main/java/com/linecorp/centraldogma/server/internal/IsolatedSystemReader.java
trustin: Sounds like a good idea.
```suggestion
            // Do nothing because we don't want to load anything from external sources.
        }
```

minwoox (approved):
Much simplified. Thanks!

server/src/main/java/com/linecorp/centraldogma/server/internal/IsolatedSystemReader.java
thachlp: Thanks

Issue #710, opened 2022-08-16
Display the original commit message when mirroring


minwoox:
Currently, the commit message of a commit that is pushed by the mirroring task looks like this:
```
Mirror fbee605, git+ssh://github.com/xxx.git#master to the repository 'foo'
```
It does not contain the original commit message which is useful to users. It would be nice if we can display the original commit message.
We can retrieve the message via:
```java
Iterable<RevCommit> call = git.log().setMaxCount(1).call();
```

minwoox:
We might display the message using the [detail](https://github.com/line/centraldogma/blob/master/server/src/main/java/com/linecorp/centraldogma/server/command/AbstractPushCommand.java#L79-L81) field.
However, the current UI does not display it so we should revamp the UI beforehand.

trustin:
There might be more than 1 commit involved in a mirroring attempt. Should we show every commit message (up to N commits)?

minwoox:
> Should we show every commit message (up to N commits)?

That's a good idea.

trustin:
At second thought, we probably should only show the latest one only, given we're going to use shallow clone and shallow fetch (see #801).

thachlp:
From my understanding, here is the pr: https://github.com/line/centraldogma/pull/818
Please help review 🤗

trustin:
Closed via #818 Thanks, @thachlp!

PR #818, opened 2023-03-07
Display the original commit message when mirroring

thachlp:
Related: https://github.com/line/centraldogma/issues/710
### Motivation
Currently, we construct commit messages while mirroring by:
> "Mirror " + abbrId + ", " + remoteRepoUri() + '#' + remoteBranch()+  " to the repository '" + localRepo().name()

So, we would like to use the original message which is useful for users. 
### Modification
1. Get a commit message from RevWalk:
> RevCommit commit = revWalk.parseCommit(objectIdOfCommit);
2. Generate a detail message including: author's name, email, time, and message. 
### Result
We display the original commit instead while mirroring.

### DIFF
--- a/it/src/test/java/com/linecorp/centraldogma/it/mirror/git/GitMirrorTest.java
+++ b/it/src/test/java/com/linecorp/centraldogma/it/mirror/git/GitMirrorTest.java
@@ -26,6 +26,7 @@ import java.io.File;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Paths;
+import java.util.List;
 
 import javax.annotation.Nullable;
 
@@ -56,6 +57,7 @@ import com.google.common.io.Files;
 import com.linecorp.centraldogma.client.CentralDogma;
 import com.linecorp.centraldogma.common.CentralDogmaException;
 import com.linecorp.centraldogma.common.Change;
+import com.linecorp.centraldogma.common.Commit;
 import com.linecorp.centraldogma.common.Entry;
 import com.linecorp.centraldogma.common.PathPattern;
 import com.linecorp.centraldogma.common.Revision;
@@ -177,9 +179,20 @@ class GitMirrorTest {
         addToGitIndex(".gitkeep", "");
         addToGitIndex("first/light.txt", "26-Aug-2014");
         addToGitIndex("second/son.json", "{\"release_date\": \"21-Mar-2014\"}");
-        git.commit().setMessage("Add the release dates of the 'Infamous' series").call();
+        git.commit().setMessage("Add the release dates of the 'Infamous' series")
+           .setAuthor("Mirror", "mirror@localhost.localdomain")
+           .call();
 
         mirroringService.mirror().join();
+        final List<Commit> commits = client.getHistory(projName, REPO_FOO, Revision.HEAD, Revision.INIT,
+                                                       PathPattern.all(), 1)
+                                           .join();
+        assertThat(commits).isNotEmpty();
+        final String detail = commits.get(0).detail();
+        assertThat(detail).isNotEmpty()
+                          .contains("Author", "Date")
+                          .contains("Mirror", "mirror@localhost.localdomain")
+                          .contains("Add the release dates of the 'Infamous' series");
 
         //// Make sure a new commit is added.
         final Revision rev3 = client.normalizeRevision(projName, REPO_FOO, Revision.HEAD).join();
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/GitMirror.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/GitMirror.java
@@ -59,6 +59,7 @@ import org.eclipse.jgit.lib.PersonIdent;
 import org.eclipse.jgit.lib.Ref;
 import org.eclipse.jgit.lib.RefUpdate;
 import org.eclipse.jgit.lib.RefUpdate.Result;
+import org.eclipse.jgit.revwalk.RevCommit;
 import org.eclipse.jgit.revwalk.RevWalk;
 import org.eclipse.jgit.transport.FetchResult;
 import org.eclipse.jgit.transport.RefSpec;
@@ -212,6 +213,7 @@ public final class GitMirror extends AbstractMirror {
 
         final Map<String, Change<?>> changes = new HashMap<>();
         final String summary;
+        final String detail;
 
         try (GitWithAuth git = openGit(workDir)) {
             final String headBranchRefName = getHeadBranchRefName(git);
@@ -246,12 +248,12 @@ public final class GitMirror extends AbstractMirror {
                 // Add mirror_state.json.
                 changes.put(mirrorStatePath, Change.ofJsonUpsert(
                         mirrorStatePath, "{ \"sourceRevision\": \"" + id.name() + "\" }"));
-
                 // Construct the log message and log.
                 summary = "Mirror " + abbrId + ", " + remoteRepoUri() + '#' + remoteBranch() +
                           " to the repository '" + localRepo().name() + '\'';
+                final RevCommit headCommit = revWalk.parseCommit(id);
+                detail = generateCommitDetail(headCommit);
                 logger.info(summary);
-
                 long numFiles = 0;
                 long numBytes = 0;
                 while (treeWalk.next()) {
@@ -327,7 +329,7 @@ public final class GitMirror extends AbstractMirror {
 
             executor.execute(Command.push(
                     MIRROR_AUTHOR, localRepo().parent().name(), localRepo().name(),
-                    Revision.HEAD, summary, "", Markup.PLAINTEXT, changes.values())).join();
+                    Revision.HEAD, summary, detail, Markup.PLAINTEXT, changes.values())).join();
         }
     }
 
@@ -427,6 +429,17 @@ public final class GitMirror extends AbstractMirror {
 
         // We should not reach here, but if we do, fall back to 'refs/heads/master'.
         return HEAD_REF_MASTER;
+    }
+
+    private static String generateCommitDetail(RevCommit headCommit) {
+        final PersonIdent authorIdent = headCommit.getAuthorIdent();
+        return "Remote commit:\n" +
+               "- SHA: " + headCommit.name() + '\n' +
+               "- Subject: " + headCommit.getShortMessage() + '\n' +
+               "- Author: " + authorIdent.getName() + " <" +
+               authorIdent.getEmailAddress() + "> \n" +
+               "- Date: " + authorIdent.getWhen() + "\n\n" +
+               headCommit.getFullMessage();
     }
 
     @Nullable


server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/GitMirror.java
```diff
@@ -240,10 +240,14 @@ protected void mirrorRemoteToLocal(File workDir, CommandExecutor executor,
...
-
                 // Construct the log message and log.
-                summary = "Mirror " + abbrId + ", " + remoteRepoUri() + '#' + remoteBranch() +
-                          " to the repository '" + localRepo().name() + '\'';
+                final String headCommitMessage = revWalk.parseCommit(id).getFullMessage();
+                if (!headCommitMessage.isEmpty()) {
+                    summary = headCommitMessage;
+                } else {
+                    summary = "Mirror " + abbrId + ", " + remoteRepoUri() + '#' + remoteBranch() +
+                              " to the repository '" + localRepo().name() + '\'';
```
ikhoon, line 249: What do you thinking of fully appending the upstream's commit message to the detail of the mirroring summary. For example
```
Mirror abbrId, remote to the repository local

The commit information of 'abbrId':
Author:  John Doe <johne.doe@gmail.com>
Date:    Thu Mar 2 19:43:28 2023 +0900
Message: Set xxx to yyy to enable zzz feature ...
```

thachlp: Thanks for the suggestion. 
image
As I understand (correct me if I am wrong), we change `2` to meaning message, from my side, I think the full message is enough.
How do you think @trustin @jrhee17 @minwoox

ikhoon: The UI for commit messages will be improved to show the message in detail. REST API could be used to get the commit message of a revision.

trustin: I think we don't need to change the summary but just add the author, date and commit message from the remote into the detail.

trustin: As you might have noticed, Central Dogma's `Commit` object has `summary` and `detail` field. Generate the string @ikhoon suggested and put it into `Commit.detail`.

it/src/test/java/com/linecorp/centraldogma/it/mirror/git/LocalToRemoteGitMirrorTest.java
```diff
@@ -349,6 +350,14 @@ void cannotMirrorInternalRepositories(String localRepo) {
                 .hasMessageContaining("invalid localRepo:");
     }
 
+    @Test
+    void localToRemoteEmptyMessage() throws Exception {
+        pushMirrorSettings("/local", "/remote", null);
+        final RevCommit call = git.commit().setMessage("").call();
+        mirroringService.mirror().join();
+        assertThat(call.getFullMessage()).isEmpty();
+    }
```
trustin, line 359: We're implementing this for remote-to-local cases only. Please remove this.

it/src/test/java/com/linecorp/centraldogma/it/mirror/git/GitMirrorTest.java
```diff
@@ -419,6 +419,14 @@ void cannotMirrorToInternalRepositories(String localRepo) {
                 .hasMessageContaining("invalid localRepo:");
     }
 
+    @Test
+    void remoteToLocalEmptyMessage() throws Exception {
+        pushMirrorSettings("/local", "/remote", null);
+        final RevCommit call = git.commit().setMessage("").call();
+        mirroringService.mirror().join();
+        assertThat(call.getFullMessage()).isEmpty();
+    }
```
trustin, line 428: This test you wrote basically does the following:

1. Add a commit to the remote Git repository.
2. Run mirroring.
3. You check the commit message of the commit you pushed in (1).

As you might have noticed, the step (3) doesn't make sense. You have to check the local Central Dogma repository has the commit that contains the message copied from the commit you added in the step (1), by using `client`.

Try to understand the test `remoteToLocal()` first.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/GitMirror.java
thachlp: Seem I misunderstand the requirement, let me check, thank you.

thachlp: Added info to detail field

it/src/test/java/com/linecorp/centraldogma/it/mirror/git/LocalToRemoteGitMirrorTest.java
thachlp: Removed

it/src/test/java/com/linecorp/centraldogma/it/mirror/git/GitMirrorTest.java
thachlp: I updated the test

thachlp (commented):
@trustin @ikhoon Please help me review this pr again 🤩

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/GitMirror.java
```diff
@@ -395,6 +398,15 @@ private GitWithAuth openGit(File workDir) throws Exception {
         }
     }
 
+    private static String generateCommitDetail(String abbrId, RevCommit headCommit) {
+        final String newLine = System.lineSeparator();
```
minwoox, line 402: I think we need to use `\n` instead of the line separator so that the detail does not have a different message depending on the OS.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/GitMirror.java
```diff
@@ -395,6 +398,15 @@ private GitWithAuth openGit(File workDir) throws Exception {
...
+        return "The commit information of " + abbrId + ": " + newLine +
```
minwoox, line 403: Perhaps we don't need to add `abbrId` because we already have it in the summary? cc @ikhoon

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/GitMirror.java
```diff
@@ -395,6 +398,15 @@ private GitWithAuth openGit(File workDir) throws Exception {
...
+               "Author: " + headCommit.getAuthorIdent().getName() + " <" +
+                       headCommit.getAuthorIdent().getEmailAddress() + "> " + newLine +
+               "Date: " + headCommit.getAuthorIdent().getWhen() + newLine +
```
minwoox, line 406: How about aligning?
```suggestion
               "Author: " + headCommit.getAuthorIdent().getName() + " <" +
                       headCommit.getAuthorIdent().getEmailAddress() + "> " + newLine +
               "Date:   " + headCommit.getAuthorIdent().getWhen() + newLine +
```

minwoox: Let's extract as `final PersonIdent authorIdent = headCommit.getAuthorIdent();` because `getAuthorIdent()` has some cost internally to parse.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/GitMirror.java
```diff
@@ -240,12 +242,13 @@ protected void mirrorRemoteToLocal(File workDir, CommandExecutor executor,
...
                         mirrorStatePath, "{ \"sourceRevision\": \"" + id.name() + "\" }"));
-
                 // Construct the log message and log.
                 summary = "Mirror " + abbrId + ", " + remoteRepoUri() + '#' + remoteBranch() +
                           " to the repository '" + localRepo().name() + '\'';
+                final RevCommit headCommit = revWalk.parseCommit(id);
+                detail = generateCommitDetail(abbrId, headCommit);
                 logger.info(summary);
-
+                logger.info(detail);
```
minwoox, line 251: How about not logging the detail?
I'm worried about the case where the detail is so long that it might give a burden to the system.

it/src/test/java/com/linecorp/centraldogma/it/mirror/git/GitMirrorTest.java
```diff
@@ -180,6 +182,13 @@ void remoteToLocal() throws Exception {
         git.commit().setMessage("Add the release dates of the 'Infamous' series").call();
 
         mirroringService.mirror().join();
+        final List<Commit> commits = client.getHistory(projName, REPO_FOO, Revision.HEAD, Revision.INIT,
+                                                       PathPattern.all(), 1)
+                                                       .join();
```
minwoox, line 187: indentation:
```suggestion
                                           .join();
```

minwoox (commented):
Thanks! Left minor comments. 😄

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/GitMirror.java
thachlp: Thanks @minwoox, as I understand from https://docs.oracle.com/javase/9/docs/api/java/lang/System.html#lineSeparator--
We should use `System.lineSeparator()`, it will auto-generate line separator on Unix or Window. 
I am just concerned that if we use `\n`, how will it run on Window because line separator on Window is `\r\n`?
Feel free to correct me if I am wrong 🚀

minwoox: If the file has different separators depending on the OS, that would cause certain problems. So we usually use just `\n` and it's automatically handleed via `core.autocrlf`. 😄 
This [link](https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_formatting_and_whitespace) explains well about the problem. 😄

thachlp: Ah, ok, I understand opposite. 
Many thanks😊

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/GitMirror.java
thachlp: Nice find 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/GitMirror.java
thachlp: Agree, removed.

it/src/test/java/com/linecorp/centraldogma/it/mirror/git/GitMirrorTest.java
thachlp: Done

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/GitMirror.java
thachlp: Thanks, @minwoox, I removed!

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/GitMirror.java
```diff
@@ -395,6 +398,16 @@ private GitWithAuth openGit(File workDir) throws Exception {
...
     }
 
+    private static String generateCommitDetail(RevCommit headCommit) {
+        final String newLine = "\n";
+        final PersonIdent authorIdent = headCommit.getAuthorIdent();
+        return "The commit information of " + ": " + newLine +
+               "Author: " + authorIdent.getName() + " <" +
+               authorIdent.getEmailAddress() + "> " + newLine +
+               "Date: " + authorIdent.getWhen() + newLine +
+               "Message: " + headCommit.getFullMessage();
```
minwoox, line 408: - I think we can just use the new line character.
- How about adding another new line before the message and removing `Message: `?
```suggestion
        final PersonIdent authorIdent = headCommit.getAuthorIdent();
        return "The commit information" + '\n' +
               "Author: " + authorIdent.getName() + " <" + authorIdent.getEmailAddress() + "> " + '\n' +
               "Date:   " + authorIdent.getWhen() + '\n' + '\n' +
               headCommit.getFullMessage();
```

minwoox (approved):
I left a small suggestion. Thanks a lot, @thachlp! 😄

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/GitMirror.java
thachlp: @minwoox I agree with removing `Message` but why two new lines? 🤔
The result will be
```
    Author: ...
    Date: ...

    <Message>

Comment (thachlp) [reply server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/GitMirror.java:408]: Ah I see the template you suggest when I try `git log` ☺️

Comment (minwoox) [reply server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/GitMirror.java]: I was inspired by git log. 😄 
image

Comment (thachlp) [reply server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/GitMirror.java]: Yes, I see, thanks.

File: server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/GitMirror.java [hunk 10]
```diff
@@ -395,6 +398,15 @@ private GitWithAuth openGit(File workDir) throws Exception {
         }
     }
 
+    private static String generateCommitDetail(RevCommit headCommit) {
+        final PersonIdent authorIdent = headCommit.getAuthorIdent();
+        return "The commit information of " + ": " + '\n' +
```
ikhoon, line 403: - It seems `information of <somethinig>` is missing.
- No need to append `\n` separately. 
```suggestion
        return "The commit information of " + ":\n" +
```

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/GitMirror.java
```diff
@@ -395,6 +398,15 @@ private GitWithAuth openGit(File workDir) throws Exception {
...
+               "Author: " + authorIdent.getName() + " <" +
+               authorIdent.getEmailAddress() + "> " + '\n' +
+               "Date: " + authorIdent.getWhen() + '\n' + '\n' +
```
ikhoon, line 406:
```suggestion
               authorIdent.getEmailAddress() + ">\n" +
               "Date: " + authorIdent.getWhen() + "\n\n" +
```

it/src/test/java/com/linecorp/centraldogma/it/mirror/git/GitMirrorTest.java
```diff
@@ -180,6 +182,13 @@ void remoteToLocal() throws Exception {
         git.commit().setMessage("Add the release dates of the 'Infamous' series").call();
 
         mirroringService.mirror().join();
+        final List<Commit> commits = client.getHistory(projName, REPO_FOO, Revision.HEAD, Revision.INIT,
+                                                       PathPattern.all(), 1)
+                                           .join();
+        assertThat(commits).isNotEmpty();
+        final String detail = commits.get(0).detail();
+        assertThat(detail).isNotEmpty()
+                          .contains("Author", "Date");
```
ikhoon, line 191: It would be helpful if we can assert the content of `Author` and `Data` to detect unexpected regressions in the future.

trustin: and the original commit message as well, please.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/GitMirror.java
trustin: How about this format:

```
Remote commit:
- SHA: <sha256>
- Subject: <the first line>
- Author: name <email>
- Date: ...
```

thachlp: Hi @ikhoon, as @minwoox suggest in comment, `abbrId` has been in the summary already, so we don't need it in detail, how do you think?

it/src/test/java/com/linecorp/centraldogma/it/mirror/git/GitMirrorTest.java
thachlp: Thanks @trustin, @ikhoon, seem I see the concern:
1) From `RevCommit` we have `PersonIdent`, but on push, we still send the `MIRROR_AUTHOR = new Author("Mirror", "mirror@localhost.localdomain")` Is it correct?
2) If 1) was wrong, seems `Author` and `Date` are dynamic, so how can we assert the content?

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/GitMirror.java
thachlp: Yes, I agree with your suggestion, let me update this. Thank you @trustin cc @ikhoon

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/GitMirror.java
thachlp: Thanks @ikhoon

trustin (approved):
Thanks a lot, @thachlp! Exactly what we wanted :+1:

ikhoon (approved):
Looks nice! Thanks, @thachlp. 👍 👍

jrhee17 (approved):
Looks good! Thanks @thachlp ! 🙇 👍 🙇

trustin:
Thanks for your patience and hard work, @thachlp! 🙇 🚀

PR #821, opened 2023-03-13
Add HTTP-to-CD mirroring

### LINKED ISSUE #154: HTTP API to CD mirroring
It would be very useful to set up a simple mirror that mirrors the result of an HTTP query into a repository, just like IFTTT.

trustin:
Motivation:

It would be very useful to set up a simple mirror that mirrors the
result of an HTTP query into a repository, just like IFTTT.

Modifications:

- Added support for the following two new URL schemes:
  - `http`
  - `https`
- Renamed `Mirror.remotePath` to `remoteSubpath` to better reflect what
  it actually does.
- Made `Mirror.remoteSubpath` nullable because HTTP-to-CD mirror doesn't
  need it.
- Added `HttpMirror` and `HttpMirrorTest`.

Result:

- A user can mirror a certain HTTP endpoint into a file in Central Dogma.
- Closes #154

Future works:

- Add support for basic and token-based authentication
- Add documentation

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/HttpMirror.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/HttpMirror.java
@@ -1,0 +1,86 @@
+/*
+ * Copyright 2023 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.linecorp.centraldogma.server.internal.mirror;
+
+import java.io.File;
+import java.net.URI;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+
+import org.slf4j.Logger;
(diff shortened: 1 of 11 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/HttpMirror.java
```diff
@@ -0,0 +1,86 @@
...
+        }
+
+        final String content = res.content(charset);
+        final String summary = "Mirror " + remoteRepoUri() + " to the repository '" + localRepo().name() + '\'';
+        logger.info(summary);
+
+        executor.execute(Command.push(
+                MIRROR_AUTHOR, localRepo().parent().name(), localRepo().name(),
+                Revision.HEAD, summary, "", Markup.PLAINTEXT,
+                Change.ofTextUpsert(localPath(), content))).join();
```
minwoox, line 84: How about raising an exception before creating HttpMirror if `localPath` is a directory?

minwoox (commented):
Looks good all in all.
We also need to fix the test here: https://github.com/line/centraldogma/blob/120597ac3b63bcbe74d529f4f0d1912a2bf5d5a6/server/src/test/java/com/linecorp/centraldogma/server/internal/mirror/MirroringTaskTest.java#L51
Thanks!

server/src/main/java/com/linecorp/centraldogma/server/mirror/MirrorSchemes.java
```diff
@@ -50,5 +50,15 @@ public final class MirrorSchemes {
      */
     public static final String SCHEME_GIT_FILE = "git+file";
 
+    /**
+     * {@code "http"}.
+     */
+    public static final String SCHEME_HTTP = "http";
+
+    /**
+     * {@code "http"}.
```
juchanei, line 59: I think this annotation has a typo!
```suggestion
     * {@code "https"}.
```

PR #826, opened 2023-03-21
Auto-detect the default branch when mirroring a Git repository

trustin:
Motivation:

Central Dogma currently chooses the branch `master` when a user doesn't specify the branch. However, not all repositories' default branches are `master`. They are sometimes `main` or even something else.

Modifications:

- Set `null` to `remoteBranch` when a user didn't specify a branch.
- Auto-detect the default branch before mirroring when `remoteBranch` is `null`.

Result:

- Less surprising behavior.
- Fixes the failing `GitMirrorAuthTest`.

### DIFF
--- a/.github/workflows/actions_build.yml
+++ b/.github/workflows/actions_build.yml
@@ -2,7 +2,7 @@ name: CI
 on:
   push:
     branches:
-      - master
+      - main
     tags-ignore:
       # The release versions will be verified by 'publish-release.yml'
       - centraldogma-*
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,7 +20,7 @@ Please [contact us](dl_oss_dev@linecorp.com) if you need the CCLA (corporate con
 
 ### Code of conduct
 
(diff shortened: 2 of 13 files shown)


minwoox (approved):
Thanks! 👍

thachlp (commented):
I see we change the default branch `master` to `main`, I think it's time to change all `master` references to `main`. 
Example in class `LocalToRemoteGitMirrorTest`
How do you think?

ikhoon (approved):
Very nice!

trustin:
> I see we change the default branch `master` to `main`, I think it's time to change all `master` references to `main`. 
> Example in class `LocalToRemoteGitMirrorTest`
> How do you think?

We could do that probably for tests, but not for the places such as `GitRepository` under `src/main`, because it could break many things.

thachlp:
> > I see we change the default branch `master` to `main`, I think it's time to change all `master` references to `main`.
> > Example in class `LocalToRemoteGitMirrorTest`
> > How do you think?
> 
> We could do that probably for tests, but not for the places such as `GitRepository` under `src/main`, because it could break many things.

Thanks for explaining to me.

thachlp (approved):

jrhee17 (approved):
Thanks @trustin 👍 👍 👍

PR #835, opened 2023-04-12
Remove JGit 5 dependencies when building server

minwoox:
Motivation:
We use JGit 6 library for the SSH connection in mirroring. So we need to exclude JGit 5 dependencies.

Modifications:
- Exclude JGit 5 dependencies in the `copyLib` task.

Result:
- Central Dogma servers run correctly without dependency conflict.

### DIFF
--- a/dist/build.gradle
+++ b/dist/build.gradle
@@ -56,12 +56,18 @@ task copyLib(group: 'Build',
              dependsOn: project(':server').tasks.jar,
              type: Copy) {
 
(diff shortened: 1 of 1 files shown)


ikhoon (approved):
🙀

jrhee17 (approved):
The bugfix itself looks good to me 👍

ikhoon:
Too many CI build failed. It looks flaky but let me restart them.

minwoox:
Thanks for the quick review. 😉

PR #836, opened 2023-04-14
Extract mirroring implementation from the server module

minwoox:
Motivation:
Previously, the server module was using jgit5 and jgit6 libraries together. It's because jgit6 requires JDK 11 as a minimum, so we needed jgit5 for JDK 8 client.
However, that might cause a dependency conflict problem such as #835. If we extract jgit6 specific API and implementations to a new module, we can avoid this problem.

Modifications:
- Add `MirrorProvider` interface that creates a `Mirror`.
  - `MirrorProvider`s are discovered via SPI.
- Created a new module called `server-mirror-git` and added `GitMirrorProvider`.
- jgit6 is used if `javaTestVersion` is greater or equal to 11.
  - jgit5 is sued if `javaTestVersion` is 8.
- Clean up dependencies in build.gradle.
- (Breaking)
  - The `type` property, which was useless, in the mirroring configuration is now removed.
  - Removed `server-git6` module.

Result:
- The server module is now correctly built and tested with the proper JGit libraries depending on the JDK test version.


mirror/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMetaRepository.java
trustin, line 1: I'm not sure if it makes sense to move `DefaultMetaRepository` out of `:server`. How about introducing some SPI so `:mirror` can register itself as a handler for a certain set of URI schemes?

settings.gradle
```diff
@@ -9,24 +9,29 @@ includeWithFlags ':bom', 'bom'
...
 includeWithFlags ':client:java-armeria-legacy',             'java', 'publish', 'relocate'
-includeWithFlags ':client:java-spring-boot2-autoconfigure',  'java', 'publish', 'relocate', 'no_aggregation'
-includeWithFlags ':client:java-spring-boot2-starter',        'java', 'publish', 'relocate', 'no_aggregation'
+includeWithFlags ':client:java-spring-boot2-autoconfigure', 'java', 'publish', 'relocate', 'no_aggregation'
+includeWithFlags ':client:java-spring-boot2-starter',       'java', 'publish', 'relocate', 'no_aggregation'
 includeWithFlags ':client:java-spring-boot3-autoconfigure', 'java17', 'publish', 'relocate'
 includeWithFlags ':client:java-spring-boot3-starter',       'java17', 'publish', 'relocate'
 includeWithFlags ':common',                                 'java', 'publish', 'shade', 'trim'
 includeWithFlags ':common-legacy',                          'java', 'publish', 'relocate'
+includeWithFlags ':mirror',                                 'java17', 'publish', 'relocate'
```
trustin, line 18: How about renaming to `:server-mirror-git` and making it focus on `GitMirror` implementation only?

mirror/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMetaRepository.java
minwoox: That's a good suggestion. Addressed.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/DefaultMetaRepository.java
```diff
@@ -172,18 +202,16 @@ private static List<MirrorCredential> loadCredentials(Map<String, Entry<?>> entr
     }
 
     @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
-    @JsonSubTypes({
-            @Type(value = SingleMirrorConfig.class, name = "single"),
-            @Type(value = MultipleMirrorConfig.class, name = "multiple")
-    })
+    @JsonSubTypes(@Type(value = SingleMirrorConfig.class, name = "single"))
     private abstract static class MirrorConfig {
```
minwoox, line 206: Will combine `MirrorConfig` and `SingleMirrorConfig` together after @ikhoon's work for mirroring UI is done.

server-mirror-git/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirrorProvider.java
```diff
@@ -0,0 +1,86 @@
...
+
+import java.net.URI;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import com.linecorp.centraldogma.server.mirror.Mirror;
+import com.linecorp.centraldogma.server.mirror.MirrorContext;
+import com.linecorp.centraldogma.server.mirror.MirrorProvider;
+
+public final class DefaultMirrorProvider implements MirrorProvider {
```
trustin, line 36: `GitMirrorProvider`?

server-mirror-git/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirrorProvider.java
```diff
@@ -0,0 +1,86 @@
...
+
+        final URI remoteUri = context.remoteUri();
+        final String scheme = remoteUri.getScheme();
+        if (scheme == null) {
+            throw new IllegalArgumentException("no scheme in remoteUri: " + remoteUri);
+        }
+
+        //noinspection EnhancedSwitchMigration
+        switch (scheme) {
+            case SCHEME_DOGMA: {
```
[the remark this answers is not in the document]
minwoox: Added `CentralDogmaMirrorProvider` and moved it back to the `server` module. 😅

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/DefaultMetaRepository.java
minwoox: Combine them in this PR instead of waiting for @ikhoon's work. 😉

server-mirror-git/src/test/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMetaRepositoryWithMirrorTest.java
```diff
@@ -138,32 +138,28 @@ void testInvalidMirrors() {
...
 
     @Test
-    void testSingleTypeMirror() {
+    void testMirror() {
         metaRepo.commit(Revision.HEAD, 0, Author.SYSTEM, "",
                         Change.ofJsonUpsert(
                                 PATH_MIRRORS,
                                 "[{" +
                                 "  \"enabled\": true," +
-                                "  \"type\": \"single\"," +
```
trustin, line 147: Can we at least keep one entry with `"type:": "single"` to ensure the backward compatibility?

server/src/main/java/com/linecorp/centraldogma/server/command/AbstractPushCommand.java
```diff
@@ -81,7 +81,7 @@ public String detail() {
     }
 
     /**
-     * Returns the {@link Markup} of the {@link #detail()}.
+     * Returns the {@link Markup} of the {@code detail()}.
```
trustin, line 84: Maybe revert? This class still have `#detail()`

server-mirror-git/src/test/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMetaRepositoryWithMirrorTest.java
minwoox: It's here. 😉  https://github.com/line/centraldogma/pull/836/files#diff-42445cf029e7cfb1d5a9f00593163dd61061d62631ab68e3f8200e56a626870dR233-R235

server/src/main/java/com/linecorp/centraldogma/server/command/AbstractPushCommand.java
minwoox: Revert and make `AbstractPushCommand` `public` so that Javadoc is correctly rendered.

trustin (approved):
:+1:

minwoox:
Let me merge this and continue to work on #830. Please leave a comment here if there's something I need to fix. I will fix it in #830 together. 😉

server/build.gradle
```diff
@@ -40,9 +40,13 @@ dependencies {
     // jCommander
     optionalImplementation libs.jcommander
 
-    // jGit
-    implementation libs.jgit
-    implementation libs.jgit.ssh.jsch
+    // Hacky way to use different jGit versions depending on the test Java version.
+    // We will use jgit6 after we drop the support for Java 8 client.
+    if (project.ext.testJavaVersion >= 11) {
```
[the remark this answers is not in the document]
ikhoon: By the way, I thought, an info message for the current JGit version would be useful.
```java
static {
    InputStream is = null;
    try {
        is = SystemReader.class.getClassLoader().getResourceAsStream("META-INF/maven/org.eclipse.jgit/org.eclipse.jgit/pom.properties");
        final Properties props = new Properties();
        props.load(is);
        final Object jgitVersion = props.get("version");
        if (jgitVersion != null) {
            logger.info("Using JGit: {}", jgitVersion);
        }
    } catch (IOException e) {
        logger.debug("Failed to read JGit version", e);
    } finally {
        if (is != null) {
            try {
                is.close();
            } catch (IOException e) {
                // ignore
            }
        }
    }
}
```
Let me send a PR for that.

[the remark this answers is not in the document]
minwoox: > Yes, it works.

Does it work when we specify `-PtestJavaVersion=8`? 🤔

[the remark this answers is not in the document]
ikhoon: - When `-PtestJavaVersion=8` was set, JGit 5 was chosen.
```sh
❯ gw clean :client:java-armeria:test -PtestJavaVersion=8 
❯ ag "Using JGit"  
java-armeria/build/test-results/test/TEST-com.linecorp.centraldogma.client.armeria.ArmeriaCentralDogmaTest.xml
57:2023-08-23 21:01:05.980 [INFO ](c.l.c.s.i.s.r.g.GitRepository) [repository-worker-6-1] Using JGit: 5.13.0.202109080827-r
```
- When `-PtestJavaVersion=8 -PminimumJavaVersion=11` is set, no tests are run since `testJavaVersion` is lower than `targerJavaVersion`. `minimumJavaVersion` raises `targerJavaVersion` to 11.
```sh
❯ gw clean :client:java-armeria:test -PtestJavaVersion=8 -PminimumJavaVersion=11

Skipping tests for :client:java-armeria since the testJavaVersion(8) is smaller than the targetJavaVersion(11)
```
- If we set `PtestJavaVersion` to 11, the client tests can be run with JGit 6.
```sh
~/src/centraldogma/client log-jgit-version*
❯ gw clean :client:java-armeria:test -PtestJavaVersion=11 -PminimumJavaVersion=11

java-armeria/build/reports/tests/test/classes/com.linecorp.centraldogma.client.armeria.ArmeriaCentralDogmaTest.html
146:2023-08-23 21:07:18.011 [INFO ](c.l.c.s.i.s.r.g.GitRepository) [repository-worker-6-1] Using JGit: 6.6.0.202305301015-r
```

This is the change used to test.
```diff
diff --git server/build.gradle server/build.gradle
index 7f828d61..a2b0f2b1 100644
--- server/build.gradle
+++ server/build.gradle
@@ -42,7 +42,7 @@ dependencies {
 
     // Hacky way to use different jGit versions depending on the test Java version.
     // We will use jgit6 after we drop the support for Java 8 client.
-    if (project.ext.testJavaVersion >= 11) {
+    if (project.ext.targetJavaVersion >= 11) {
         implementation libs.jgit6
     } else {
         implementation libs.jgit
```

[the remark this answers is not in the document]
minwoox: Anyway, it's too late so let's keep discussing tomorrow. 😉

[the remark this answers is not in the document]
ikhoon: We don’t need to set an extra properties since we publish the server module with JGit 5.
JGit 6 will be chosen only mininumJavaVersion is higher than 11.

When packing a tarball, JGit 5 is excluded and JGit is added instead.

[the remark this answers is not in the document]
minwoox: Ah ha, now I understand. Well actually, publishing the server module with JGit 6 was my intention but I guess I was wrong because users would still need the server module for testing. Let me make a change, then. Thanks!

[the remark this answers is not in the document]
minwoox: The changed is pushed: https://github.com/line/centraldogma/pull/830/commits/51184221667e659c21f009cf0abae90b9e2363b0
PTAL. 🙇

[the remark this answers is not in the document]
ikhoon: It looks 👍!

PR #837, opened 2023-04-19
Do not run auto-GC against any repositories

trustin:
Motivation:

Our default jGit settings did not disable auto-GC. Auto-GC is not desirable for our use case because:

- It triggers GC at unexpected point of time, affecting performance.
- Since jGit 6.5.0, jGit acquires a filesystem lock before performing GC
  and registers a JVM shutdown hook that cleans the lock file up.
  This new behavior may be OK for short-living processes such as CLI
  tools, but it leads to memory leak for us.
  See: https://github.com/eclipse-mirrors/org.eclipse.jgit/blob/2e3f12a0fc63deba80e725bfa985c0c5bd31de99/org.eclipse.jgit/src/org/eclipse/jgit/internal/storage/file/GC.java#L1769

Modifications:

- Added a new class `JGitUtil` that sets the common Git configuration
  properties for all JGit `Config`s
- Made sure the desired default settings are enforced for all Git
  repositories, including `GitRepository` and `GitMirror`
- Optimized `EmptyConfig` so it can be used as a singleton
- Added test cases that make sure the default settings have all the
  necessary properties
- Added `CentralDogmaExtension.dataDir()` so that a user can access the
  location of the data directory, which is useful for checking the
  internal state of Central Dogma

Result:

- Automatic GC is never triggered for all repositories including both
  Central Dogma repositories and mirrored Git repositories.
- The memory leak, that occurs when jGit 6.5 is used, is gone.
- A user can now know the location of the data directory with
  `dataDir()` when using `CentralDogmaExtension`.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/JGitUtil.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/JGitUtil.java
@@ -38,7 +38,6 @@ import java.io.IOException;
 import java.util.Objects;
 
 import org.eclipse.jgit.lib.Config;
-import org.eclipse.jgit.lib.ConfigConstants;
 import org.eclipse.jgit.lib.StoredConfig;
 
 public final class JGitUtil {
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/JGitUtil.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/JGitUtil.java
@@ -1,0 +1,107 @@
+/*
+ * Copyright 2023 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.linecorp.centraldogma.server.internal;
+
+import static java.util.Objects.requireNonNull;
+import static org.eclipse.jgit.lib.ConfigConstants.CONFIG_COMMIT_SECTION;
+import static org.eclipse.jgit.lib.ConfigConstants.CONFIG_CORE_SECTION;
+import static org.eclipse.jgit.lib.ConfigConstants.CONFIG_DIFF_SECTION;
+import static org.eclipse.jgit.lib.ConfigConstants.CONFIG_GC_SECTION;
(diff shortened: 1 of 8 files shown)


ikhoon (approved):
Thanks for quickly fixing the issue. 🙇‍♂️

server/src/main/java/com/linecorp/centraldogma/server/internal/JGitUtil.java
```diff
@@ -0,0 +1,106 @@
...
+        // Disable rename detection which we do not use.
+        updated |= set(config, CONFIG_DIFF_SECTION, CONFIG_KEY_RENAMES, "false");
+
+        // Disable auto-GC by default because of the following reasons:
+        // - GC can take long time affecting performance.
+        // - jGit's GC task has memory leak which adds a JVM shutdown hook for each GC task.
+        //   See https://github.com/eclipse-mirrors/org.eclipse.jgit/blob/2e3f12a0fc63deba80e725bfa985c0c5bd31de99/org.eclipse.jgit/src/org/eclipse/jgit/internal/storage/file/GC.java#L1769
+        updated |= set(config, CONFIG_GC_SECTION, CONFIG_KEY_AUTO, "0");
+        updated |= set(config, CONFIG_GC_SECTION, CONFIG_KEY_AUTOPACKLIMIT, "0");
+        updated |= set(config, CONFIG_RECEIVE_SECTION, CONFIG_KEY_AUTOGC, "false");
```
jrhee17, line 86:
```suggestion
```

Probably not too important, but I believe since `needGc()` is always false thanks to the settings above and `automatic == true` for auto gc anyways, this setting isn't needed. (there are other code paths calling `FileRepository#autoGC` anyways)

https://github.com/eclipse-mirrors/org.eclipse.jgit/blob/f1a9adf7da3cd130a0233aad3f31e2917317c13a/org.eclipse.jgit/src/org/eclipse/jgit/internal/storage/file/GC.java#L275

I'm fine with leaving this config if it makes sense semantically though

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/GitWithAuth.java
```diff
@@ -108,6 +109,8 @@ private static Repository repo(File repoDir) throws IOException {
             if (!repo.getObjectDatabase().exists()) {
                 repo.create(true);
             }
+
+            JGitUtil.applyDefaultsAndSave(repo.getConfig());
```
jrhee17, line 113: 😱   Thanks 👍

jrhee17 (approved):
Looks really nice 👍    Thanks @trustin 🙇 👍 🚀

minwoox (approved):
Looks great, thanks!

server/src/main/java/com/linecorp/centraldogma/server/internal/JGitUtil.java
trustin: Yeah, this setting is not really needed for this specific case. I'm just trying to be pre-cautious. Let me leave this as it is.

Issue #647, opened 2021-11-29
Should we deprecate MultipleMirrorConfig?


minwoox:
We have [SingleMirrorConfig](https://github.com/line/centraldogma/blob/centraldogma-0.52.6/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/DefaultMetaRepository.java#L216) and [MultipleMirrorConfig](https://github.com/line/centraldogma/blob/centraldogma-0.52.6/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/DefaultMetaRepository.java#L256) for the mirroring setup.
I believe that `MultipleMirrorConfig` is currently not used because it's not documented on our web and a user can configure multiple mirroring setups using multiple `SingleMirrorConfig`.
I'm wondering if we need to deprecate `MultipleMirrorConfig` to reduce the complexity of the module.

ikhoon:
It seems no one uses `MultipleMirrorConfig` since the documentation says to use `single`.

Additionally, I found it difficult to update a mirror configuration with REST API while implementing the mirror UI with the current data format. #50 There is no unique ID in a mirror configuration. As a result, an update request should use the index in the array at `mirrors.json`.

If `MultipleMirrorConfig` is considered, the UI and index lookup could be complex for the unused feature. Let me move `MultipleMirrorConfig` in the mirroring UI issue together.

ikhoon:
Resolved by #838

PR #838, opened 2023-04-20
Add Mirroring and Credential UI

### LINKED ISSUE #50: UI for editing mirror settings
Creating and editing mirror settings require manual JSON editing which is time-consuming and error-prone. We should provide a dedicated UI for a meta repository so that a user cannot make a mistake when configuring a mirror.

Also, for some reason, when a bad mirror setting leaked into a meta repository, the UI should provide an easy way to figure out what's wrong with his/her setting.

ikhoon:
Motivation:

Creating and editing mirror settings require manual JSON editing which is time-consuming and error-prone. If a dedicated UI validates inputs and suggests default values such as mirrorable repositories and credentials, users will easily configure a mirror setting without making a mistake.

Related: #50

Modifications:

- Server:
  - Added dedicated REST APIs for mirror configurations and credential information. - Creating, listing, and updating operations are supported. 
    - Mirror ID and credential ID ared used as a key to update the existing configuration.   
  - Added `enabled` property to `MirrorCredential` so that users can disable unused credentials.
  - Added `id` property to `Mirror` so that users distinguish a mirror from others.
  - Removed support for multiple mirrors that are currently not used. #647
  - Annotated `@JsonProperties` to the getters of `MirrorCredential` and its subclasses to serialize `MirrorCredential` to JSON as a REST API response.
- Webapp:
  - The following web pages were newly added for mirroring and credentials. 
  - Mirroring: 
    - Create: `/app/projects/{projectName}/mirrors/new`
    - Update: `/app/projects/{projectName}/mirrors/{id}/edit`
    - List: `/app/projects/{projectName}/mirrors`
    - View: `/app/projects/{projectName}/mirrors/{id}`
  - Credential: 
    - Create: `/app/projects/{projectName}/credentials/new`
    - Update: `/app/projects/{projectName}/credentials/{index}/edit`
    - List: `/app/projects/{projectName}/credentials`
    - View: `/app/projects/{projectName}/credentials/{index}`
  - Switched query-based tab URL for metadata to path-based to easily create breadcrumbs based on the path.
    - The old paths: `/app/projects/metadata/{projectName}?tab=repositories|permissions}members|tokens|mirror`
    - The new paths:
      - `/app/projects/{projectName}/metadata`
      - `/app/projects/{projectName}/permissions` 
      - `/app/projects/{projectName}/permissions/repos/{repoName}` 
      - `/app/projects/{projectName}/members` 
      - `/app/projects/{projectName}/tokens` 
      - `/app/projects/{projectName}/mirrors` 
      - `/app/projects/{projectName}/credentials`
  - Updated `Breadcrumbs` to replace a specific path segment name with another to display more meaningful information for breadcrumbs rather than showing an index number.

Result:

- You can now use a web UI to set up a Git-to-CD mirror.
- Closes #50


ikhoon:
Screenshots for the new UI
- Credential form
  credential-form
- Credential list
  credential-list
- Credential view
  credential-view
- Mirror form
  mirror-form
- Mirror list
  mirror-list
- Mirror view
  mirror-view

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/DefaultMetaRepository.java
```diff
@@ -68,307 +55,147 @@ public class DefaultMetaRepository extends RepositoryWrapper implements MetaRepo
...
 
         return builder.build();
     }
-
-    @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
-    @JsonSubTypes({
-            @Type(value = SingleMirrorConfig.class, name = "single"),
-            @Type(value = MultipleMirrorConfig.class, name = "multiple")
-    })
-    private abstract static class MirrorConfig {
```
ikhoon, line 179: `MultipleMirrorConfig` is removed and `SingleMirrorConfig` is migrated to `MirrorConfig.java`

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/credential/AbstractMirrorCredential.java
```diff
@@ -27,45 +28,83 @@
...
     }
 
     private static Set<Pattern> validateHostnamePatterns(@Nullable Iterable<Pattern> hostnamePatterns) {
-        if (hostnamePatterns == null) {
+        if (hostnamePatterns == null || Iterables.isEmpty(hostnamePatterns)) {
+            return ImmutableSet.of();
+        }
+        if (Iterables.size(hostnamePatterns) == 1 && Iterables.getFirst(hostnamePatterns, null) == null) {
+            // Web browsers may send an empty array `[]` as a value of hostnamePatterns
+            // which is converted into `[null]`.
```
jrhee17, line 71: Question) Is it possible for browsers to clean the input before sending such parameters? This feels unintuitive since the Web UI isn't the only code-path which uses this logic.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/MirroringServiceV1.java
```diff
@@ -0,0 +1,152 @@
...
+                             mirror.id(),
+                             projectName,
+                             mirror.schedule().asString(),
+                             mirror.direction().name(),
+                             mirror.localRepo().name(),
+                             mirror.localPath(),
+                             remoteRepoUri.getScheme(),
+                             remoteRepoUri.getHost() + remoteRepoUri.getPath(),
+                             mirror.remotePath(),
+                             firstNonNull(mirror.remoteBranch(), "master"),
```
jrhee17, line 147: I'm not sure why we need to fallback to `master`.  If I remember correctly, @trustin added a capability to detect the default branch by default if this value is null, and `master` is no longer the de-facto if due to `main`.

Is this necessary for the UI?

server/src/main/java/com/linecorp/centraldogma/server/mirror/MirrorUtil.java
```diff
@@ -59,15 +61,15 @@ public static String normalizePath(String path) {
...
+     * - remotePath:    /some-path
      * - remoteBranch:  master
      *
      * <p>e.g. dogma://foo.com/bar/qux.dogma is split into:
      * - remoteRepoUri: dogma://foo.com/bar/qux.dogma
      * - remotePath:    / (default)
      * - remoteBranch:  {@code defaultBranch}
      */
-    static String[] split(URI remoteUri, String suffix) {
+    public static String[] splitRemoteUri(URI remoteUri, String suffix, @Nullable String defaultBranch) {
```
jrhee17, line 72: Like mentioned in another comment, I'm unsure why this change is necessary 😅  Would you mind explaining/updating in the PR description?

```suggestion
    public static String[] splitRemoteUri(URI remoteUri, String suffix) {
```

server/src/main/java/com/linecorp/centraldogma/server/internal/api/CredentialServiceV1.java
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * GET /projects/{projectName}/credentials?id={id}
+     *
+     * <p>Returns the credential for the ID in the project.
+     */
+    // TODO(ikhoon): Automatically inject the non-optional query parameters to Route.matchesParam().
+    @MatchesParam("id")
+    @Get("/projects/{projectName}/credentials")
+    public CompletableFuture<MirrorCredential> getCredentialById(@Param String projectName, @Param String id) {
```
jrhee17, line 81: Question) Are `credentialId`s unique? I wasn't able to find any checks for duplicates, and I think users can still manually modify the `credentials.json` file anyways

I wasn't able to check the UI, but I'm wondering what happens if
1) a credential doesn't have an id
2) a credential has a duplicate id

ikhoon:
Please don't review the client part. The change is too huge that is hard to review. 
It would be okay to q/a the features in a browser at the moment.

I will send a request to the main branch away. You can check the front-end code on the final PR if you want.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/DefaultMetaRepository.java
```diff
@@ -68,307 +55,147 @@ public class DefaultMetaRepository extends RepositoryWrapper implements MetaRepo
...
-                    throw new RepositoryMetadataException(PATH_MIRRORS + " contains null.");
-                }
-                mirrors.addAll(c.toMirrors(parent(), credentials));
+        int index = 0;
+        for (JsonNode m : mirrorsJson) {
+            final MirrorConfig c = Jackson.treeToValue(m, MirrorConfig.class);
+            if (c == null) {
+                throw new RepositoryMetadataException(PATH_MIRRORS + " contains null.");
+            }
+            final Mirror mirror = c.toMirror(parent(), credentials, index++);
```
jrhee17, line 162: Question) What do you think of just persistently assigning an id? I understand that previously we didn't have mirrors, but I think it's possible to migrate to an object based schema instead of a mirror based schema and enforce ids and create an arbitrary id if missing.

With the current setup, I fear that users might make a mistake if multiple users modify mirror for the same project. i.e. Two users may try to modify a mirror configuration concurrently, and the index mapping might change.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/DefaultMetaRepository.java
jrhee17: I'd feel better if we migrate the internally used `MultipleMirrorConfig` before merging this PR (or at least before the next release so that the next release driver isn't suprised)

ikhoon:
Minwoo also worked for `MultipleMirrorConfig` removal at 
https://github.com/line/centraldogma/pull/836#discussion_r1168435030
#836 may be released before this. Otherwise, I will handle it separately.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/credential/AbstractMirrorCredential.java
ikhoon: It is possible. I can inject a custom logic to covert an empty array to null for `hostnamePatterns.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/DefaultMetaRepository.java
```diff
@@ -68,307 +55,147 @@ public class DefaultMetaRepository extends RepositoryWrapper implements MetaRepo
...
+            return UnmodifiableFuture.completedFuture(mirrorsAndCredentials);
         }
     }
 
-    private Set<Mirror> loadMirrors(int rev) {
-        // TODO(trustin): Asynchronization
-        final Map<String, Entry<?>> entries =
-                find(new Revision(rev), PATH_CREDENTIALS_AND_MIRRORS, Collections.emptyMap()).join();
+    private CompletableFuture<Map.Entry<List<Mirror>, List<MirrorCredential>>> mirrorsAndCredentials(int rev) {
+        return find(new Revision(rev), PATH_CREDENTIALS_AND_MIRRORS, Collections.emptyMap()).thenApply(
```
jrhee17, line 120:
```suggestion
        return find(new Revision(rev, PATH_CREDENTIALS_AND_MIRRORS).thenApply(
```

jrhee17 (commented):
Maybe I couldn't find it, but do we also provide a way to delete a mirror/credential? Or is `enabled` the equivalent of this?

common/src/main/java/com/linecorp/centraldogma/internal/jsonpatch/AddOperation.java
```diff
@@ -86,10 +89,16 @@ public AddOperation(@JsonProperty("path") final JsonPointer path,
     }
 
     @Override
-    JsonNode apply(final JsonNode node) {
+    JsonNode apply(JsonNode node) {
         if (path.toString().isEmpty()) {
             return valueCopy();
         }
+        if (node.isNull() && Strings.isNullOrEmpty(path.head().getMatchingProperty()) &&
```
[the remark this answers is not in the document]
jrhee17: I see, makes sense 👍

ikhoon:
I don't think we need to provide an API for removal because of the limitation of JSON patch that can update an array with its index. An index for an item should be consistent to provide a deep link for a mirror view and edit. Additionally, the size of Git mirrors and credentials isn't growing up to 100 or 1000. The `enabled` flag would be enough to manage those scales of data.

As a following work, we will hide `mirrors.json` and `credentials.json` from the file navigator of web UI and REST API retrieval so that users don't modify `mirrors.json` and `credentials.json` manually.

If removal for the mirror and credential information is really required, we can choose one of the following:
- Migrate the array data format to object key-value format. 
- Add a custom JSON patch operation to update an array by a key field.

As you know, key / ID fields for `mirrors` and `credentials` is optional. A migration strategy is necessary to fill nullable IDs to implement the ideas.

trustin:
> As you know, key / ID fields for `mirrors` and `credentials` is optional. A migration strategy is necessary to fill nullable IDs to implement the ideas.

Agreed. Maybe we could auto-generate IDs and commit the changes automatically when it's read first time?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/MirroringServiceV1.java
```diff
@@ -0,0 +1,152 @@
...
+    @Get("/projects/{projectName}/mirrors/{index}")
+    public CompletableFuture<MirrorDto> getMirror(@Param String projectName, @Param int index) {
+        checkArgument(index >= 0, "index: %s (expected: >= 0)", index);
+
+        return projectManager().get(projectName).metaRepo().mirrors(true).thenApply(mirrors -> {
+            if (index >= mirrors.size()) {
+                throw new EntryNotFoundException(
+                        "No such mirror at the index " + index + " in " + projectName);
+            }
+            return convertToMirrorDto(projectName, mirrors.get(index));
```
jrhee17, line 118: It seems like we're just using the index and not actually comparing `mirrors.index` with `index` from the parameters.
What do you think of just removing the index from the DTO altogether for simplicity if we know it's not going to change?

I didn't think there is any sorting from any point (since this is an array of json objects, there probably isn't a natural sorting order anyways)

jrhee17 (commented):
I see, I guess I didn't read the PR description carefully and didn't catch the intention 😅 
I think there should be no problem with the current approach if the `mirrors.json`, `credentials.json` is not editable directly by end-users 👍 

I'm unsure how this will be rolled out with respect to the UI migration though (at the same time? vs. after UI migration is complete?)

server/src/main/java/com/linecorp/centraldogma/server/internal/api/CredentialServiceV1.java
```diff
@@ -0,0 +1,138 @@
...
+import com.linecorp.centraldogma.server.mirror.MirrorCredential;
+import com.linecorp.centraldogma.server.storage.project.ProjectManager;
+
+/**
+ * Annotated service object for managing credential service.
+ */
+@ProducesJson
+@RequiresRole(roles = ProjectRole.OWNER)
+@ExceptionHandler(HttpApiExceptionHandler.class)
+public class CredentialServiceV1 extends AbstractService {
```
trustin, line 51:
```suggestion
public final class CredentialServiceV1 extends AbstractService {
```

server/src/main/java/com/linecorp/centraldogma/server/internal/api/MirroringServiceV1.java
trustin: I guess we can leave it as `null` if a user specified `null` (= use default branch).

server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java
```diff
@@ -976,6 +986,95 @@ public CompletableFuture<Token> findTokenBySecret(String secret) {
...
+        final String summary = "Update the mirror '" + mirrorDto.id() + '\'';
+        final MirrorConfig mirrorConfig = converterToMirrorConfig(mirrorDto);
+        final JsonNode jsonNode = convertToJsonNodeAndRemoveIndex(mirrorConfig);
+        final Change<JsonNode> change = Change.ofJsonPatch(
+                PATH_MIRRORS, asJsonArray(new ReplaceOperation(JsonPointer.compile("/" + index), jsonNode)));
+
+        return metadataRepo.push(projectName, Project.REPO_META, author, summary, change);
+    }
+
+    private static MirrorConfig converterToMirrorConfig(MirrorDto mirrorDto) {
```
trustin, line 1023:
```suggestion
    private static MirrorConfig convertToMirrorConfig(MirrorDto mirrorDto) {
```

server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java
```diff
@@ -976,6 +986,95 @@ public CompletableFuture<Token> findTokenBySecret(String secret) {
...
+        final String remoteUri =
+                mirrorDto.remoteScheme() + "://" + mirrorDto.remoteUrl() +
+                MirrorUtil.normalizePath(mirrorDto.remotePath()) + '#' + mirrorDto.remoteBranch();
```
trustin, line 1026: `#branch` should be omitted if `remoteBranch` is `null`

trustin (commented):
What do you think about changing the directory layout of mirrors and credentials? e.g.

```
- mirrors
  - <mirror-id>.json
  - ...
- credentials
  - <credentials-id>.json
  - ...
```

At startup, we could:

1. perform automatic migration from `mirrors.json` and `credentials.json` to the new layout; and
2. rename `mirrors.json` and `credentials.json` to `mirrors.json.bak` and `credentials.json.bak` (so that the migration doesn't repeat.)

Then, all mirrors and credentials will have their unique IDs and we don't need to use integral index which is confusing.

I think this can be done as a follow-up task, though. (or we can do the migration first?) What do you think?

common/src/main/java/com/linecorp/centraldogma/internal/Jackson.java
```diff
@@ -77,7 +79,8 @@ public final class Jackson {
         compactMapper.enable(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS);
         prettyMapper.enable(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS);
 
-        registerModules(new SimpleModule().addSerializer(Instant.class, InstantSerializer.INSTANCE)
+        registerModules(new Jdk8Module(),
+                        new SimpleModule().addSerializer(Instant.class, InstantSerializer.INSTANCE)
                                           .addDeserializer(Instant.class, InstantDeserializer.INSTANT));
```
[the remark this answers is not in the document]
ikhoon: Armeria's default object mapper uses SPI to load all available modules. The new Java 8 API and time API are declared as dependencies. 
https://github.com/line/armeria/blob/219ee3758b232ca9c47e03258a0a62f4cc4841c3/core/src/main/java/com/linecorp/armeria/internal/common/DefaultJacksonObjectMapperProvider.java#L42-L43
https://github.com/line/armeria/blob/70296b99b4277eef75da93ec85eeb26beec8d39e/core/build.gradle#L107-L108

ikhoon:
I confirmed that the basic operations were performed correctly after merging the new mirror API.

The changes of this PR are `+7,780 −3,298` which is not easy for maintainers to review. Most of them are related to UI such as rendering, design and API fetching. As this PR is not sent to the main branch, I think it is safe to merge this PR and continue to work.

There are still some remaining works to do to finish this webapp project. I will request other maintainers for Q/A tests when all features are ready. That may be a faster way to find bugs than reviewing the code.

If there are differing opinions on the design of the code, we may handle them in the f/u PRs.

PR #839, opened 2023-04-21
Do not read user SSH config / Fix GitMirrorAuthTest

ikhoon:
Motivation:

When JGit creates a session JSch, it obtains user's configuration first.
https://github.com/eclipse-mirrors/org.eclipse.jgit/blob/master/org.eclipse.jgit.ssh.jsch/src/org/eclipse/jgit/transport/ssh/jsch/OpenSshConfig.java#L69-L77
Although the SSH config is not actually used to Git mirror, JSch tries to parse the SSH identity files.

JSch does not support Ed25519 public-key signature algorithm. If Ed25519 key is used configured in a user's SSH key, `KeyPair.load()` raises `JSchException: invalid privatekey ...`.

Consequently, GitMirrorAuthTest fails due to the exception.

Modifications:

- Set an empty `OpenSshConfig` to `JschConfigSessionFactory` when a Git mirroring repository is initialized to disable the default SSH config file lookup.

Result:

- The Central Dogma operates consistently across different environments.
- `GitMirrorAuthTest` no longer fails when Ed25519 key is configured in the SSH config file.

### DIFF
--- a/server-jgit6/src/main/java/com/linecorp/centraldogma/server/internal/mirror/Git6WithAuth.java
+++ b/server-jgit6/src/main/java/com/linecorp/centraldogma/server/internal/mirror/Git6WithAuth.java
@@ -23,6 +23,7 @@ import org.eclipse.jgit.api.FetchCommand;
 import org.eclipse.jgit.api.TransportCommand;
 import org.eclipse.jgit.transport.SshTransport;
 import org.eclipse.jgit.transport.ssh.jsch.JschConfigSessionFactory;
+import org.eclipse.jgit.transport.ssh.jsch.OpenSshConfig;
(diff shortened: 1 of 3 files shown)


ikhoon:
I found it difficult to test if the SSH config file is initialized.
I manually tested both jgit5 and jgit6 and they worked correctly.
Let me know if there is a good idea to verify this.

trustin (approved):
:+1:

jrhee17 (approved):
Double checked that deleting the file while the application is running doesn't break behavior 👍   Thanks @ikhoon ! 🙇 👍 🙇

minwoox (approved):
Thanks!

PR #852, opened 2023-07-03
Try migrating auth-web

jrhee17:
**Modifications**

- `setup-java` version is upgraded
  - Unsure why, but windows refuses to find the correct java installation path if two java versions are used
  - https://github.com/line/centraldogma/pull/853
  - https://github.com/line/centraldogma/actions/runs/5452700579/jobs/9920562254?pr=853#step:6:96
```
Directory 'D:\a\centraldogma\centraldogma\C:hostedtoolcachewindowsJava_Temurin-Hotspot_jdk17.0.7-7x64' (system property 'org.gradle.java.installations.paths') used for java installations does not exist
Directory 'D:\a\centraldogma\centraldogma\C:hostedtoolcachewindowsJava_Temurin-Hotspot_jdk11.0.19-7x64' (system property 'org.gradle.java.installations.paths') used for java installations does not exist
...
Execution failed for task ':common:compileJava'.
> Error while evaluating property 'javaCompiler' of task ':common:compileJava'.
   > Failed to calculate the value of task ':common:compileJava' property 'javaCompiler'.
      > No matching toolchains found for requested specification: {languageVersion=17, vendor=any, implementation=vendor-specific}.
         > No locally installed toolchains match (see https://docs.gradle.org/8.1.1/userguide/toolchains.html#sec:auto_detection) and toolchain download repositories have not been configured (see https://docs.gradle.org/8.1.1/userguide/toolchains.html#sub:download_repositories).
```

- Removed usage of yarn in `server-auth:webapp`
  - Configuration files are mostly copied over from armeria's `:docs-client` module
  - Verified that the login/logout pages are displayed correctly
  - Unified npm, node versions to match that of the `:server` module
  - `.gz` compressed files are served by the `FileService` serving `server-auth:webapp`
- Fix build failures from `*.lock` files
  - `server` doesn't copy over `.lock` files (it seems like gradle keeps a lock)


.github/workflows/actions_build.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-12, windows-latest]
-        java: [11]
+        java: [17]
```
jrhee17, line 31:
```suggestion
        java: [11]
```

ikhoon (approved):
Thanks, @jrhee17!

server/build.gradle
```diff
@@ -126,6 +129,11 @@ task copyWebApp(type: Copy) {
     doFirst {
         delete(dst)
     }
+
+    // gradle holds a lock even after the buildWebapp task is completed
```
minwoox, line 133: Question: which lock file is it?

server-auth/webapp/webpack.config.ts
```diff
@@ -0,0 +1,135 @@
...
+  template: './public/index.html',
+  hash: true,
+}));
+
+plugins.push(new optimize.LimitChunkCountPlugin({
+  maxChunks: 1,
+}));
+
+plugins.push(new DefinePlugin({
+  'process.env': '{}',
```
minwoox, line 121: Don't we need this?
`'process.env.WEBPACK_DEV': JSON.stringify(process.env.WEBPACK_DEV),`

server-auth/webapp/build.gradle
```diff
@@ -46,7 +17,14 @@ tasks.assemble.dependsOn tasks.buildWebApp
...
 node {
-    version = '16.1.0'
-    yarnVersion = '1.22.10'
+    version = '16.14.0'
+    npmVersion = '8.5.2'
     download = true
+    npmInstallCommand = "ci"
+
+    // Change the cache location under Gradle user home directory so that it's cached by CI.
+    if (System.getenv('CI') != null) {
```
minwoox, line 26: You mean cached by CI or not cached by CI?

minwoox (commented):
Thanks! Left nits and questions

server/build.gradle
minwoox: How about specifying the file?
reference: https://github.com/Kotlin/kotlinx-datetime/blob/master/core/build.gradle.kts#L295-L301

server-auth/webapp/webpack.config.ts
jrhee17: I ran into issues with `blueprint` when I tried this.
Honestly, this module will be deprecated in the new UI, so I'd rather not put too much time into this 😅

minwoox: No problem then. Let's skip it. 😉

minwoox (approved):
Thanks! 👍 👍 👍

trustin (approved):
:heart: Thanks a lot for cleaning this up, @jrhee17!

PR #584, opened 2021-04-12
Add Snappy and Dropwizard metric dependency.

minwoox:
Motivation:
Snappy and Dropwizard dependencies are removed from ZooKeeper 3.6.0
https://github.com/apache/zookeeper/commit/b787770b9920a7ae99f9ed2dbe583ece1b2665cc#diff-237e154d86dc584115184b6d4b8bfda7ef88fe934a67ae796c6856badc0cacbf
The test didn't fail because curator-test has the dependencies.

Modifications:
- Add Snappy 1.1.7
- Add Dropwizard metric 3.2.5

Result:
- No more `ClassNotFoundException`

PR #856, opened 2023-07-17
Add Snappy dependency

minwoox:
Motivation:
Snappy is removed from ZooKeeper 3.6.0
https://github.com/apache/zookeeper/commit/b787770b9920a7ae99f9ed2dbe583ece1b2665cc#diff-237e154d86dc584115184b6d4b8bfda7ef88fe934a67ae796c6856badc0cacbf Our test didn't fail because curator-test has the dependencies. Related #584

Modifications:
- Add Snappy 1.1.10.1

Result:
- No more `ClassNotFoundException`

### DIFF
--- a/dependencies.toml
+++ b/dependencies.toml
@@ -49,6 +49,9 @@ quartz = "2.3.2"
 shadow-gradle-plugin = "7.1.2"
 shiro = "1.3.2"
 slf4j = "1.7.36"
+# Ensure that we use the same Snappy version as what Curator depends on.
+# See: https://github.com/apache/curator/blob/master/pom.xml
+snappy = "1.1.10.1"
 sphinx = "2.10.1"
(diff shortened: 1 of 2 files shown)


dependencies.toml
```diff
@@ -306,6 +309,10 @@ version.ref = "slf4j"
 module = "org.slf4j:log4j-over-slf4j"
 version.ref = "slf4j"
 
+[libraries.snappy]
+module = "org.xerial.snappy:snappy-java"
```
ikhoon, line 313: Add this as `implementation` dependency of `:server` module?

jrhee17 (approved):
Looks good once @ikhoon 's comment is addressed 👍

dependencies.toml
minwoox: Yes, 😓

PR #864, opened 2023-08-21
Update cron-utils to 9.2.1

minwoox:
I'm not sure why we haven't upgrade cron-utils to the latest. There's no reason not to upgrade it.

cron-utils: 5.0.5 -> 9.2.1


server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/AbstractMirror.java
```diff
@@ -98,9 +101,14 @@ public final ZonedDateTime nextExecutionTime(ZonedDateTime lastExecutionTime) {
...
-        final ZonedDateTime next =
-                executionTime.nextExecution(lastExecutionTime.minus(jitterMillis, ChronoUnit.MILLIS));
-        return next.plus(jitterMillis, ChronoUnit.MILLIS);
+        final Optional<ZonedDateTime> next = executionTime.nextExecution(
+                lastExecutionTime.minus(jitterMillis, ChronoUnit.MILLIS));
+        if (next.isPresent()) {
+            return next.get().plus(jitterMillis, ChronoUnit.MILLIS);
+        }
+        throw new IllegalArgumentException(
+                "no next execution time for " + CRON_DESCRIPTOR.describe(schedule) + ", lastExecutionTime: " +
```
ikhoon, line 110: nit: Should we log `localRepo` as well to inform the errors to users.

ikhoon (approved):
Thanks!

ikhoon (approved):
Thanks!

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/AbstractMirror.java
ikhoon: Makes sense.

jrhee17 (approved):
Thanks! 👍 👍 👍 
I think this would be a good release to add monitoring dashboards for mirroring as well. I'll add this before this release

trustin (approved):
Thanks for keeping it up-to-date!

PR #868, opened 2023-08-23
Log the current JGit version

ikhoon:
The JGit version between 5 and 6 is determined dynamically by the current Java version. If the Java version is lower than Java 11, JGit 5 is used.

Therefore, if the current JGit version is logged, it should be helpful to easily check what JGit version is used.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
@@ -27,6 +27,7 @@ import static org.eclipse.jgit.lib.ConfigConstants.CONFIG_KEY_REPO_FORMAT_VERSIO
 
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
 import java.lang.reflect.Field;
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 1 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
```diff
@@ -136,6 +139,18 @@ class GitRepository implements Repository {
...
+        final String jgitPomProperties = "META-INF/maven/org.eclipse.jgit/org.eclipse.jgit/pom.properties";
+        try (InputStream is = SystemReader.class.getClassLoader().getResourceAsStream(jgitPomProperties)) {
+            final Properties props = new Properties();
+            props.load(is);
+            final Object jgitVersion = props.get("version");
+            if (jgitVersion != null) {
+                logger.info("Using JGit: {}", jgitVersion);
+            }
+        } catch (IOException e) {
+            logger.debug("Failed to read JGit version", e);
```
minwoox, line 151: warn would be better?

minwoox (approved):
Thanks!

jrhee17 (approved):
👍 👍 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
ikhoon: Debug level used in case it is intentionally removed while shadowing tasksby users.

trustin (approved):

PR #867, opened 2023-08-23
Upgrade dependencies for 0.62.0

jrhee17:
Note: I wasn't sure how CLA works for https://github.com/line/centraldogma/pull/865, so I just also included armeria upgrades in this PR.

- Armeria 1.24.3 → 1.25.1
- Guava 32.0.1-jre → 32.1.2-jre
- Hibernate Validator 8.0.0.Final → 8.0.1.Final
- Micrometer 1.11.1 → 1.11.3
- Spring Boot 3 3.1.0 → 3.1.2

PR #869, opened 2023-08-24
Remove JSch from server module

minwoox:
Motivation:
Using Apache MINA instead of JSch is recommended by JGit team.

Modification:
- Use Apache MINA instead of JSch for server module.
  - JSch is now only used for testing.

Result:
- JSch gone.


dependencies.toml
```diff
@@ -3,7 +3,7 @@
 #     If its classes are exposed in Javadoc, update offline links as well.
 #
 [versions]
-armeria = "1.25.1"
+armeria = "1.24.3"
```
minwoox, line 6: Will change once 1.25.1 is released.

minwoox:
I believe the build will succeed after https://github.com/line/centraldogma/pull/867 is merged.

dist/build.gradle
```diff
@@ -90,8 +90,8 @@ task copyLib(group: 'Build',
         FileCollection jgits = layout.files { libDir.listFiles() }.filter { File file ->
             file.name.contains("jgit")
         }
-        // org.eclipse.jgit-6.x and org.eclipse.jgit.ssh.jsch-6.x are required.
-        if (jgits.size() != 2 || !jgits.any { File file -> file.name.startsWith("org.eclipse.jgit-6") }) {
+        // org.eclipse.jgit-6.x is required.
+        if (jgits.size() != 1 || !jgits[0].name.startsWith("org.eclipse.jgit-6")) {
```
jrhee17, line 94:
```suggestion
        if (jgits.size() != 1 || !jgits.first().name.startsWith("org.eclipse.jgit-6")) {
```

jrhee17 (approved):
Thanks @minwoox 🙇 👍 🙇

ikhoon (approved):
Thanks!

dist/build.gradle
```diff
@@ -90,8 +90,8 @@ task copyLib(group: 'Build',
         FileCollection jgits = layout.files { libDir.listFiles() }.filter { File file ->
             file.name.contains("jgit")
         }
-        // org.eclipse.jgit-6.x and org.eclipse.jgit.ssh.jsch-6.x are required.
-        if (jgits.size() != 2 || !jgits.any { File file -> file.name.startsWith("org.eclipse.jgit-6") }) {
+        // org.eclipse.jgit-6.x is required.
+        if (jgits.size() != 1 || !jgits.first().name.startsWith("org.eclipse.jgit-6")) {
             throw new IllegalStateException("Unexpected JGit JARs: $jgits.files")
```
ikhoon, line 95: By the way, could we move `startsWith("org.eclipse.jgit-6")` to `FileCollection`'s filter and check only if the size of `jgits` is one?

ikhoon: Ah, it seems intentionally used for leaving an exception message for the jar files.
Please ignore the commend above.

trustin (approved):
:+1:

PR #872, opened 2023-08-25
Exclude `slf4j2` from `cron-utils`

jrhee17:
Screenshot 2023-08-25 at 6 24 47 PM


minwoox (approved):
Thanks! I've also checked that slf4j isn't included.

ikhoon (approved):
👍

PR #873, opened 2023-08-26
Downgrade cron-utils to 9.2.0

minwoox:
Motivation:
We have upgraded cron-utils to 9.2.1 to address [CVE-2021-41269](https://github.com/advisories/GHSA-p9m8-27x8-rg87). The 9.2.1 version uses slf4j2 that must be used with logback 1.3 or 1.4. Because we use logback 1.2 version in Armeria, we need to exclude slf4j2: https://github.com/line/centraldogma/pull/872.
Just excluding slf4j should be okay because cron-utils isn't using any APIs that are introduced after slf4j 2.0: https://github.com/search?q=repo%3Ajmrozanec%2Fcron-utils+slf4j&type=code
However, there's no guarantee that cron-utils won't use the new APIs in the future. So, I think we should stop upgrading it until there's another CVE is found or Armeria uses higher version of Logback and Slf4j.\

While, I'm working on this I found out that cron-utils 9.2.0, which is one micro version eariler, uses Slf4j 1.x which is compatible with Armeria so it's better to use that version.

Modification:
- Downgrade cron-utils to 9.2.0

Result:
- Resovle dependency conflict for server module.

### DIFF
--- a/dist/src/conf/logback.xml
+++ b/dist/src/conf/logback.xml
@@ -37,7 +37,7 @@
   <logger name="com.github.benmanes.caffeine.cache" level="ERROR" />
   <logger name="com.linecorp" level="INFO" />
   <logger name="com.linecorp.centraldogma" level="DEBUG" />
-  <logger name="com.linecorp.centraldogma.server.internal.mirror" level="INFO" />
+  <logger name="com.linecorp.centraldogma.server.internal.mirror.GitWithAuth$DefaultGitSshdSessionFactory" level="INFO" />
   <logger name="com.linecorp.armeria.logging.access" level="INFO" additivity="false">
     <appender-ref ref="ACCESS"/>
   </logger>
--- a/dist/src/conf/logback.xml
+++ b/dist/src/conf/logback.xml
@@ -37,6 +37,7 @@
   <logger name="com.github.benmanes.caffeine.cache" level="ERROR" />
   <logger name="com.linecorp" level="INFO" />
   <logger name="com.linecorp.centraldogma" level="DEBUG" />
+  <logger name="com.linecorp.centraldogma.server.internal.mirror" level="INFO" />
   <logger name="com.linecorp.armeria.logging.access" level="INFO" additivity="false">
     <appender-ref ref="ACCESS"/>
   </logger>
--- a/dependencies.toml
+++ b/dependencies.toml
@@ -10,7 +10,8 @@ bouncycastle = "1.76"
 caffeine = "2.9.3"
 checkstyle = "10.3.3"
 curator = "5.5.0"
-cron-utils = "9.2.1"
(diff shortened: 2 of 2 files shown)


minwoox:
I have checked that there's no SLF4J2 in the classpath for the server.

dist/src/conf/logback.xml
```diff
@@ -37,6 +37,7 @@
   <logger name="com.github.benmanes.caffeine.cache" level="ERROR" />
   <logger name="com.linecorp" level="INFO" />
   <logger name="com.linecorp.centraldogma" level="DEBUG" />
+  <logger name="com.linecorp.centraldogma.server.internal.mirror" level="INFO" />
```
minwoox, line 40: I found out that `GitSshdSessionFactory` from Apache MINA, which is Inherited by the `DefaultGitSshdSessionFactory` Central Dogma class, produces verbose debug outputs.
`GitSshdSessionFactory` creates the logger with its class name which is `DefaultGitSshdSessionFactory` in this case so they are all logged by the `<logger name="com.linecorp.centraldogma" level="DEBUG" />`.
```
2023-08-26 00:38:06.468 [DEBUG](c.l.c.s.i.m.GitWithAuth$DefaultGitSshdSessionFactory$1) [mirroring-worker-6-1] Connecting to ...
2023-08-26 00:38:06.537 [DEBUG](c.l.c.s.i.m.GitWithAuth$DefaultGitSshdSessionFactory$1) [mirroring-worker-6-1] Connected to ...
2023-08-26 00:38:06.537 [DEBUG](c.l.c.s.i.m.GitWithAuth$DefaultGitSshdSessionFactory$1) [mirroring-worker-6-1] Authenticating: ClientSessionImpl[git@...]
.....
6-1] Stopped SshClient[3c45db89]
```

trustin: Can we extract the class (`...$1`) into a top-level class and adjust the log level for that specific class only?

jrhee17 (approved):
👍 👍 👍

dist/src/conf/logback.xml
minwoox: There are two separate `DefaultGitSshdSessionFactory` anonymous classes in the file.
Let me just use the class specifically:
` <logger name="com.linecorp.centraldogma.server.internal.mirror.GitWithAuth$DefaultGitSshdSessionFactory" level="INFO" />`

ikhoon (approved):
Thanks! 🙇‍♂️

dist/src/conf/logback.xml
trustin: SGTM

minwoox:
Thanks for the review and sorry for causing the issue. 😅

PR #874, opened 2023-08-28
Record metrics when INIT revision or an old revision is used

minwoox:
Motivation:
Before we apply https://github.com/line/centraldogma/pull/681 that removes old commits, it would be nice if we could find any usages that access data with the INIT Revision or revisions that are more than 5000 (default minimum number of commit retentions) commits ahead from the head.

Modifications:
- Add a counter that is increased when INIT Revision or revisions that are more than 5000 ahead from the head are used.

Result:
- Track the number of usages.
- This commit will be reverted after we find the usage and fix it.

### DIFF
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -22,5 +22,7 @@ if (tasks.findByName('trimShadedJar')) {
         keep "class com.linecorp.centraldogma.internal.shaded.caffeine.** { *; }"
         // Prevent ProGuard from removing all enum values from Option because otherwise it becomes a non-enum class.
         keep "class com.linecorp.centraldogma.internal.shaded.jsonpath.Option { *; }"
+
+        dontnote
     }
 }
--- a/common/build.gradle
+++ b/common/build.gradle
@@ -23,6 +23,7 @@ if (tasks.findByName('trimShadedJar')) {
         // Prevent ProGuard from removing all enum values from Option because otherwise it becomes a non-enum class.
         keep "class com.linecorp.centraldogma.internal.shaded.jsonpath.Option { *; }"
 
+        // Reduces the verbosity of ProGuardTask when running in parallel.
         dontnote
     }
 }
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/api/RepositoryServiceV1.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/api/RepositoryServiceV1.java
@@ -213,13 +213,13 @@ public class RepositoryServiceV1 extends AbstractService {
         if (normalized.major() == 1) {
             ctx.log().whenComplete().thenAccept(
                     log -> ctx.meterRegistry()
-                              .counter("init.revisions", generateTags(projectName, repoName, log).build())
+                              .counter("revisions.init", generateTags(projectName, repoName, log).build())
                               .increment());
         }
         if (head.major() - normalized.major() >= 5000) {
             ctx.log().whenComplete().thenAccept(
                     log -> ctx.meterRegistry()
-                              .summary("old.revisions",
+                              .summary("revisions.old",
                                        generateTags(projectName, repoName, log)
                                                .add(Tag.of("init", Boolean.toString(normalized.major() == 1)))
                                                .build())
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/api/RepositoryServiceV1.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/api/RepositoryServiceV1.java
@@ -16,6 +16,7 @@
 
 package com.linecorp.centraldogma.server.internal.api;
 
+import static com.google.common.base.MoreObjects.firstNonNull;
 import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.linecorp.centraldogma.server.internal.api.HttpApiUtil.checkUnremoveArgument;
 import static com.linecorp.centraldogma.server.internal.api.HttpApiUtil.returnOrThrow;
@@ -28,9 +29,11 @@ import java.util.concurrent.CompletableFuture;
 import javax.annotation.Nullable;
 
 import com.fasterxml.jackson.databind.JsonNode;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 
 import com.linecorp.armeria.common.HttpStatus;
+import com.linecorp.armeria.common.logging.RequestLog;
 import com.linecorp.armeria.server.ServiceRequestContext;
 import com.linecorp.armeria.server.annotation.Consumes;
 import com.linecorp.armeria.server.annotation.Delete;
@@ -57,6 +60,8 @@ import com.linecorp.centraldogma.server.metadata.User;
 import com.linecorp.centraldogma.server.storage.project.Project;
 import com.linecorp.centraldogma.server.storage.project.ProjectManager;
 import com.linecorp.centraldogma.server.storage.repository.Repository;
+
+import io.micrometer.core.instrument.Tag;
 
 /**
  * Annotated service object for managing repositories.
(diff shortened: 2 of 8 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/api/RepositoryServiceV1.java
```diff
@@ -186,8 +191,41 @@ public CompletableFuture<RepositoryDto> patchRepository(@Param String repoName,
...
+        final Revision normalized = repository.normalizeNow(revision);
+        final Revision head = repository.normalizeNow(Revision.HEAD);
+        increaseCounterIfOldRevisionUsed(ctx, repository.parent().name(), repository.name(), normalized, head);
+    }
+
+    public static void increaseCounterIfOldRevisionUsed(
+            ServiceRequestContext ctx, String projectName, String repoName,
+            Revision normalized, Revision head) {
+        if (normalized.major() == 1 || head.major() - normalized.major() >= 5000) {
+            ctx.log().whenComplete().thenAccept(log -> {
```
jrhee17, line 214: Question) Do we want to differentiate between cases using `1` vs. `>5000`?
Imagining that the counter goes up, I wonder if we want to know which case needs handling.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/RepositoryServiceV1.java
```diff
@@ -186,8 +191,41 @@ public CompletableFuture<RepositoryDto> patchRepository(@Param String repoName,
...
+        }
+    }
+
+    private static List<Tag> generateTags(String projectName, String repoName,
+                                          RequestLog log, Revision normalized, Revision head) {
+        return ImmutableList.of(Tag.of("project", projectName),
+                                Tag.of("repo", repoName),
+                                Tag.of("service", firstNonNull(log.serviceName(), "none")),
+                                Tag.of("method", log.name()),
+                                Tag.of("normalized", Integer.toString(normalized.major())),
```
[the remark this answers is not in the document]
minwoox: I have added the metric with the assumption that this will be short-lived and there will not be many requests that are caught by the condition.
However, there's no reason not to decrease the cardinality so let me use the distribution. Thanks!

server/src/main/java/com/linecorp/centraldogma/server/internal/thrift/CentralDogmaServiceImpl.java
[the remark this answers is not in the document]
minwoox: I've meant it but there's no harm if I add it to the method. Left me add it as well. 😉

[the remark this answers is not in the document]
minwoox: I found out that I've mistakenly added to `mergeFiles` in `ContentServiceV1`. 😓

server/src/main/java/com/linecorp/centraldogma/server/internal/api/RepositoryServiceV1.java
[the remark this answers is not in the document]
jrhee17: > I have added the metric with the assumption that this will be short-lived and there will not be many requests that are caught by the condition.

I agree, but I also thought this code will also need to go to production env. to confirm our hypothesis 😄

jrhee17 (approved):
Looks good to me 👍   Thanks @minwoox ! 🙇 👍 🚀

common/build.gradle
```diff
@@ -22,5 +22,7 @@ if (tasks.findByName('trimShadedJar')) {
         keep "class com.linecorp.centraldogma.internal.shaded.caffeine.** { *; }"
         // Prevent ProGuard from removing all enum values from Option because otherwise it becomes a non-enum class.
         keep "class com.linecorp.centraldogma.internal.shaded.jsonpath.Option { *; }"
+
+        dontnote
```
minwoox, line 26: This reduces the verbosity of ProGuardTask when running in parallel.
The original comment: https://github.com/line/armeria/pull/2172/files#r332367951

trustin: Please add a line comment about it.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/RepositoryServiceV1.java
```diff
@@ -186,8 +191,48 @@ public CompletableFuture<RepositoryDto> patchRepository(@Param String repoName,
...
+        increaseCounterIfOldRevisionUsed(ctx, repository.parent().name(), repository.name(), normalized, head);
+    }
+
+    public static void increaseCounterIfOldRevisionUsed(
+            ServiceRequestContext ctx, String projectName, String repoName,
+            Revision normalized, Revision head) {
+        if (normalized.major() == 1) {
+            ctx.log().whenComplete().thenAccept(
+                    log -> ctx.meterRegistry()
+                              .counter("init.revisions", generateTags(projectName, repoName, log).build())
```
trustin, line 216: revisions.init?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/RepositoryServiceV1.java
```diff
@@ -186,8 +191,48 @@ public CompletableFuture<RepositoryDto> patchRepository(@Param String repoName,
...
+                              .increment());
+        }
+        if (head.major() - normalized.major() >= 5000) {
+            ctx.log().whenComplete().thenAccept(
+                    log -> ctx.meterRegistry()
+                              .summary("old.revisions",
```
trustin, line 222: revisions.old?

jrhee17 (approved):
New changes still look good 👍

server/src/test/java/com/linecorp/centraldogma/server/MetricsTest.java
```diff
@@ -55,23 +59,27 @@ void metrics() {
...
+                                                                 .errorOnEntryNotFound(false);
+        jsonNodeWatchRequest.start().join();
+        content = dogma.httpClient().get("/monitor/metrics").aggregate().join().contentUtf8();
         assertThat(content).contains("com.linecorp.centraldogma.server.internal.api.WatchContentServiceV1");
+        assertThat(content).doesNotContain("init_revisions");
+
+        // Trigger old revision recording
+        jsonNodeWatchRequest.start(Revision.INIT).join();
+        content = dogma.httpClient().get("/monitor/metrics").aggregate().join().contentUtf8();
+        assertThat(content).contains("init_revisions");
```
trustin, line 83: Can we also add a similar test for `old_revisions`? Should check if the meters with `init=true` and `init=false` both exists.

minwoox: All addressed. PTAL. 😉

ikhoon (approved):
Thanks, @minwoox

server/src/main/java/com/linecorp/centraldogma/server/internal/api/RepositoryServiceV1.java
```diff
@@ -186,8 +191,48 @@ public CompletableFuture<RepositoryDto> patchRepository(@Param String repoName,
...
+        final Revision normalized = repository.normalizeNow(revision);
+        final Revision head = repository.normalizeNow(Revision.HEAD);
+        increaseCounterIfOldRevisionUsed(ctx, repository.parent().name(), repository.name(), normalized, head);
+    }
+
+    public static void increaseCounterIfOldRevisionUsed(
+            ServiceRequestContext ctx, String projectName, String repoName,
+            Revision normalized, Revision head) {
+        if (normalized.major() == 1) {
+            ctx.log().whenComplete().thenAccept(
```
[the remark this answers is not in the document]
minwoox: Thanks! All addressed. 😉

minwoox:
Thanks for reviewing. 😉

PR #875, opened 2023-08-28
Use strict versions for SFL4J and Logback

ikhoon:
Motivation:

There was an issue with slf4j 2.0 being used by a transitive dependency. Due to Gradle's resolution strategy, we didn't get any notices that a higher version was picked when our declaration.

I propose to use `strict` for SLF4J and Logback to avoid unexpected resolution. If a strict version is declared in `dependencies.toml`, it overrides non-strict versions. If two strict versions are detected, it fails to resolve dependencies. We need to resolve the conflicts by overriding Gradle `resolutionStrategy`.

Modifications:

- Use strict version scheme for SLF4J and Logback
- Fore Spring Boot 3 integrations to use SLF4J 2.x and Logback 1.4.x

Result:

Prevent unwanted versions through strict version control.


minwoox:
This doesn't prevent the use of a third party that uses Slf4j 2.x. For example, if we use cron-utils with 9.2.1, the build succeeds. Can we prevent that?

ikhoon:
A strict version takes precedence over other versions. When I tested, SLF4J 1.7.x was picked when cron-utils with 9.2.1 was used.

```
$ ./gradlew :server:dependencies -q
...
+--- com.cronutils:cron-utils:9.2.1
|    \--- org.slf4j:slf4j-api:2.0.7 -> 1.7.36

```

minwoox:
Oops, what I meant was preventing from using the third-party library that uses the incompatible transitive dependency.
Let's assume that the cron-utils uses `logger.atInfo().log..` API that is introduced after Slf4J 2.+.
If we force to use SLF4J 1.X and the `logger.atInfo().log..` is invoked with SLF4J 1.x, `NoSuchMethodError` will be raised in runtime.

So what I was suggesting was printing or recording somewhere if a dependency resolution happens, so that we can check if there's an unexpected library was introduced. Let me think about this more if we have a better solution. 🤔

ikhoon:
> If we force to use SLF4J 1.X and the logger.atInfo().log.. is invoked with SLF4J 1.x, NoSuchMethodError will be raised in runtime.

I think it should be checked by test suites. If a third-party library uses an API that only exists in SLFJ 2.x, we can't use the library. Even if version conflicts are logged, it might be difficult to check if a library uses an incompatible API.

minwoox:
> I think it should be checked by test suites.

Yeah, probably. One thing we have to keep in mind is that the test runtime classpath is different from the runtime classpath.
Anyway, we can also fix it later if we find any problem regarding the dependency conflict. So I'm okay with this change. 👍

minwoox (approved):
Thanks a lot, @ikhoon! 👍 👍 👍

ikhoon:
I understood. It is non-trivial to fully control the dependencies without auto resolution. 
`failOnVersionConflict()` is enabled in `resolutionStrategy`, I had no idea how to resolve all conflicts.

Speaking of tests, `strictly` version set to `implemetation` overrides the dependencies in test classpath since test elements extend `api` or `implementation`

client/java-spring-boot3-autoconfigure/build.gradle
```diff
@@ -7,3 +7,11 @@ dependencies {
...
     testRuntimeOnly libs.hibernate.validator8
 }
+
+configurations.configureEach {
+    resolutionStrategy {
+        force libs.slf4j2.api.get()
+        force libs.slf4j2.jul.to.slf4j.get()
+        force libs.logback14.get()
+    }
+}
```
trustin, line 17: Is it possible to move this block inside the `if` block you added to `/build.gradle`?

jrhee17 (approved):
Changes look straightforward enough!
I wish `gradle` provided a "`failOnVersionConflict` for specific versions" option, but I guess it's not available from a quick search

client/java-spring-boot3-autoconfigure/build.gradle
ikhoon: Good idea.

client/java-spring-boot3-autoconfigure/build.gradle
```diff
@@ -7,3 +7,11 @@ dependencies {
     testImplementation libs.spring.boot3.starter.test
     testRuntimeOnly libs.hibernate.validator8
 }
+
+configurations.configureEach {
+    resolutionStrategy {
+        force libs.slf4j2.api.get()
+        force libs.slf4j2.jul.to.slf4j.get()
```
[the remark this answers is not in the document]
ikhoon: It should be. Sorry, I'm super late. 😅

PR #876, opened 2023-09-04
Fix Bouncy Castle version conflict in SAML module

minwoox:
Motivation:
- Armeria saml uses Bouncy Castle jdk15on which conflicts with the version jdk18on that CD uses

Modification:
- Exclude Bouncy Castle jdk15on from SAML module

Result:
- Fixed version conflict

### DIFF
--- a/server-auth/saml/build.gradle
+++ b/server-auth/saml/build.gradle
@@ -1,4 +1,8 @@
 dependencies {
     implementation project(':server')
-    implementation libs.armeria.saml
+    implementation(libs.armeria.saml) {
+        exclude group: 'org.bouncycastle'
+    }
+
+    implementation libs.bouncycastle.bcprov
 }
--- a/server-auth/saml/build.gradle
+++ b/server-auth/saml/build.gradle
@@ -1,7 +1,7 @@
 dependencies {
     implementation project(':server')
     implementation(libs.armeria.saml) {
-        exclude group: 'org.bouncycastle'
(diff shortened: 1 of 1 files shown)


server-auth/saml/build.gradle
```diff
@@ -1,4 +1,8 @@
 dependencies {
     implementation project(':server')
-    implementation libs.armeria.saml
+    implementation(libs.armeria.saml) {
+        exclude group: 'org.bouncycastle'
+    }
+
+    implementation libs.bouncycastle.bcprov
```
jrhee17, line 7: nit; I think it should be possible to add a `strictly` since this is the only version that will be used across all modules

jrhee17 (approved):
Thanks!

server-auth/saml/build.gradle
minwoox: Yeah, we can probably add it later. 😉 
But, the current problem is that they are using different artifacts.
e.g. `org.bouncycastle:bcprov-jdk15on` and `org.bouncycastle:bcprov-jdk18on`

ikhoon (approved):
Thanks for the quick fix!

server-auth/saml/build.gradle
minwoox: I've added the module as well in order to be more specific. 😉

minwoox:
Thanks for the quick review. 😉

PR #878, opened 2023-09-05
Reuse BouncyCastleRandom

minwoox:
Motivation:
`BouncyCastleRandom`, which creates a `SecureRandom` inside, was created whenever an SSH client is created. That blocked the thread because the instance lacks entropy for getting the seed. Instead, we could create a BouncyCastleRandome once and reuse it.

Modifications:
- Create BouncyCastleRandom once and reuse it

Result:
- Mirroring doesn't take longer anymore due to an SSH client creation.


jrhee17 (approved):

ikhoon (approved):
Thanks, @minwoox!

minwoox:
Thank for reviewing. 😉

PR #880, opened 2023-09-06
Introduce new mirroring and credential settings format and REST API

ikhoon:
Motivation:

The ID for mirroring and credential configurations is optional, so I found it difficult to safely update a configuration with REST API. If a user updates a file manually on UI or commit API, the REST API may update a wrong configuration without a unique ID.

@trustin suggested changing the directory layout to store a configuration to a file with a unique ID. https://github.com/line/centraldogma/pull/838#pullrequestreview-1410657085

This PR has three major changes:
- Migrate the old mirror and credential settings to the new layout.
  ```
  - mirrors
    - <mirror-id>.json
    - ...
  - credentials
    - <credentials-id>.json
    - ...
  ```
- Add a migration job to automatically migrate the old settings to the new format when a server starts. The old files are renamed by adding `.bak` suffix. e,g. `mirrors.json` -> `mirrors.json.bak`, `credentials.json` -> `credentials.json.bak`.
- Add REST API for mirroring and credential configurations. This is a necessary task to add mirror UI. #838 

Modifications:

- Add `MirroringMigrationService` that is executed when a server starts and scan all `/mirrors.json` and `/credentials.json` in the meta repo of projects and migrate them to the new format.
  - "id" is a required property in each configuration. Human-readable random words are used to create a unique ID.
    - Mirror ID format: `mirror-<projectName>-<localRepo>-<shortWord>`
    - Credential ID format: `credential-<projectName>-<shortWord>`
    - `short_wordlist.txt` is used as the word database. 
- Change `Mirror` and related classes to have `id`, `enabled` as required fields.
- Add `CredentailServiceV1` and `MirrorServiceV1` to serve REST API for CRU.
  - Create, read, and update operations are implemented in `DefaultMetaRepository`.
- Add `RepositoryUri` to represent a repository-specific URI such as a Git repository URL.
- Add `MirrorDto` to serialize a mirroring configuration. `Mirror` represents a mirroring task, so `Mirror` is not suitable for serialization.
  - `MirrorCredential` is used as is instead of creating a new DTO.
- Migrated all mirroring tests to use the new configuration format.
- Updated site documentation with the new format.

Result:

- Mirroring and credential settings have been updated to the new formats.
- You can now access and modify mirroring and credential resources using the REST API.


server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/credential/PublicKeyMirrorCredential.java
```diff
@@ -57,26 +58,30 @@ public PublicKeyMirrorCredential(@JsonProperty("id") @Nullable String id,
...
 
-        super(id, hostnamePatterns);
+        super(id, enabled, "public_key", hostnamePatterns);
 
         this.username = requireNonEmpty(username, "username");
         this.publicKey = requireNonEmpty(publicKey, "publicKey");
         requireNonEmpty(privateKey, "privateKey");
         // privateKey is converted into a list of Strings that will be used as an input of
         // KeyPairResourceLoader.loadKeyPairs(...)
-        this.privateKey = ImmutableList.copyOf(NEWLINE_SPLITTER.splitToList(privateKey));
```
minwoox, line 67: Could you remove `NEWLINE_SPLITTER`?

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/credential/PublicKeyMirrorCredential.java
```diff
@@ -57,26 +58,30 @@ public PublicKeyMirrorCredential(@JsonProperty("id") @Nullable String id,
...
+        this.privateKey = privateKey;
         this.passphrase = maybeDecodeBase64(passphrase, "passphrase");
+        this.passphraseString = passphrase;
```
minwoox, line 70: nit:
```suggestion
        passphraseString = passphrase;
```

minwoox (commented):
Looks great all in all. 👍 👍 👍

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -173,8 +176,22 @@ public class CentralDogma implements AutoCloseable {
...
     private static final Logger logger = LoggerFactory.getLogger(CentralDogma.class);
 
+    private static final boolean GIT_MIRROR_ENABLED;
+
     static {
         Jackson.registerModules(new SimpleModule().addSerializer(CacheStats.class, new CacheStatsSerializer()));
+
+        boolean gitMirrorEnabled = false;
+        try {
+            Class.forName(CentralDogma.class.getPackage().getName() + ".internal.mirror.GitMirror");
```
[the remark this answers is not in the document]
ikhoon: Let me use https://github.com/line/centraldogma/pull/883 instead when the PR is merged.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/MirroringServiceV1.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    /**
+     * GET /projects/{projectName}/mirrors
+     *
+     * <p>Returns the list of the mirrors in the project.
+     */
+    @Get("/projects/{projectName}/mirrors")
+    public CompletableFuture<List<MirrorDto>> listMirrors(@Param String projectName) {
+        return metaRepo(projectName).mirrors(true).thenApply(mirrors -> {
```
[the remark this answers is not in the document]
ikhoon: Possible but this REST API is designed for UI. Let's add the query parameter later if there is a demand for that.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirroringMigrationService.java
```diff
@@ -0,0 +1,276 @@
...
+            id = generateRandomIdForMirror(project.name(), mirror);
+        } else {
+            id = idNode.asText();
+        }
+        id = uniquify(id, mirrorIds);
+        mirror.put("id", id);
+
+        fillCredentialId(mirror, credentials);
+        if (mirror.get("schedule") == null) {
+            mirror.put("schedule", DEFAULT_SCHEDULE);
```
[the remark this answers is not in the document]
ikhoon: I hesitated to change the value because, currently, the default schedule is 1 minute if no schedule is set. If we change the default schedule to 5 or 10 minutes, that could impact users.

If necessary, I prefer handling the breaking change in a separate PR to easily know the change from the PR title.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirroringMigrationService.java
```diff
@@ -0,0 +1,276 @@
...
+               .commit(Revision.HEAD, System.currentTimeMillis(), Author.SYSTEM,
+                       "Migrate the mirror " + id + " in '" + PATH_LEGACY_MIRRORS + "' into '" + jsonFile +
+                       "'.",
+                       Change.ofJsonUpsert(jsonFile, mirror))
+               .get();
+    }
+
+    private static void fillCredentialId(ObjectNode mirror, List<MirrorCredential> credentials) {
+        final JsonNode credentialId = mirror.get("credentialId");
+        if (credentialId != null) {
```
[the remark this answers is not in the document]
ikhoon: Central Dogma doesn't have transactions so some data is updated during migration, it could remain in `.bak`. I changed to call `rename()` right after reading metadata which users prevent updating the configuration from the JSON patch. 

But I have no idea how to prevent JSON upsert operation. We may need to stop replication just before rolling updates.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/credential/AbstractMirrorCredential.java
```diff
@@ -16,56 +16,73 @@
...
 
-    AbstractMirrorCredential(@Nullable String id, @Nullable Iterable<Pattern> hostnamePatterns) {
-        this.id = id;
+    AbstractMirrorCredential(String id, @Nullable Boolean enabled, String type,
+                             @Nullable Iterable<Pattern> hostnamePatterns) {
+        this.id = requireNonNull(id, "id");
+        this.enabled = firstNonNull(enabled, true);
+        // JsonTypeInfo is ignored when serializing collections.
+        // As a workaround, manually set the type hint to serialize.
+        this.type = requireNonNull(type, "type");
```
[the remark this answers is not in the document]
ikhoon: Let me handle it in the following PR.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/DefaultMetaRepository.java
```diff
@@ -16,142 +16,291 @@
...
 
-    private Set<Mirror> loadMirrors(int rev) {
-        // TODO(trustin): Asynchronization
-        final Map<String, Entry<?>> entries =
-                find(new Revision(rev), PATH_CREDENTIALS_AND_MIRRORS, Collections.emptyMap()).join();
+    @Override
+    public CompletableFuture<Revision> updateMirror(MirrorDto mirrorDto, Author author) {
+        validateMirror(mirrorDto);
+        final String summary = "Update the mirror '" + mirrorDto.id() + '\'';
+        return commitMirror(mirrorDto, author, summary);
```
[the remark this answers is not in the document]
ikhoon: It would be not a good idea to change the ID of a mirror. I checked if the mirror exists before committing although that is not an atomic operation.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirroringMigrationService.java
```diff
@@ -0,0 +1,297 @@
...
+        final int rand = ThreadLocalRandom.current().nextInt(shortWords.size());
+        final String shortWord = shortWords.get(rand);
+        return "credential-" + projectName + '-' + shortWord;
+    }
+
+    private static String uniquify(String id, Set<String> existingIds) {
+        while (existingIds.contains(id)) {
+            id += '-' + Integer.toHexString(ThreadLocalRandom.current().nextInt());
+        }
+        return id;
```
jrhee17, line 282: nit; although the probability is low, it seems like the current implementation keeps adding hex strings which I don't think makes sense

```suggestion
        String tmp = id;
        while (existingIds.contains(tmp)) {
            tmp = id + '-' + Integer.toHexString(ThreadLocalRandom.current().nextInt());
        }
        return tmp;
```

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirroringMigrationService.java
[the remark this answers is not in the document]
jrhee17: > But I have no idea how to prevent JSON upsert operation. We may need to stop replication just before rolling updates.

Instead of the current approach of updating from a Leader directly, would it be possible to issue a custom command (i.e. `MigrateMirrorCommand`) ?

If we take a write lock while migrating, I guess there will be minimal issues regarding inconsistency.

jrhee17 (commented):
I didn't get a chance to look at the services, but overall looks nice. Left some questions 🙇

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirroringMigrationService.java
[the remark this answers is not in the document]
minwoox: Yeah, we need to handle this separately. Let me create an issue for this.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirroringMigrationService.java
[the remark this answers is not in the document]
minwoox: > We may need to stop replication just before rolling updates.

That's a good solution. 👍 

> would it be possible to issue a custom command (i.e. MigrateMirrorCommand)

That would also work. 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirroringMigrationService.java
```diff
@@ -0,0 +1,297 @@
...
+
+        final ArrayNode mirrors = getMetaData(project, PATH_LEGACY_MIRRORS);
+        if (mirrors == null) {
+            return false;
+        }
+
+        // Back up the old mirrors.json file and don't use it anymore.
+        rename(project, PATH_LEGACY_MIRRORS, PATH_LEGACY_MIRRORS_BACKUP);
+
+        final List<MirrorCredential> credentials = project.metaRepo().credentials()
```
jrhee17, line 108: Question) Say that the `mirrors.json` file is migrated, but this line throws an exception.
Do we manually revert the file before trying again?

Would it make sense to rename the file at the last step instead?

In the interest of making `migrateMirrors` idempotent, what do you think of assigning a deterministic mirrorId as well? (e.g. using a hash)

common/src/main/java/com/linecorp/centraldogma/internal/jsonpatch/AddOperation.java
```diff
@@ -86,10 +89,16 @@ public AddOperation(@JsonProperty("path") final JsonPointer path,
     }
 
     @Override
-    JsonNode apply(final JsonNode node) {
+    JsonNode apply(JsonNode node) {
         if (path.toString().isEmpty()) {
             return valueCopy();
         }
+        if (node.isNull() && Strings.isNullOrEmpty(path.head().getMatchingProperty()) &&
```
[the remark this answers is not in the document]
ikhoon: JSON patch specification does not consider if the target node is absent.
Let me revert this change since the code is no longer used because we changed the format of mirroring and credentials.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirroringMigrationService.java
ikhoon: Nice catch!

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirroringMigrationService.java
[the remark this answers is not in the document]
ikhoon: > Instead of the current approach of updating from a Leader directly, would it be possible to issue a custom command (i.e. MigrateMirrorCommand) ?

It would be possible if we can 1) make the mirror ID an credential ID deterministic and 2) all replica has the same revision for all repositories.

One possible scenario is:

1. Stop writing for all replicas.
2. Wait until all pending logs are replayed.
3. Execute a special command `MigrateMirrorCommand` that should be written to the replication logs even if `writable` is false.
4. Trigger `MirroringMigrationService` and all replica produces identical results for migration.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/MirroringServiceV1.java
```diff
@@ -0,0 +1,136 @@
...
+import com.linecorp.centraldogma.server.metadata.ProjectRole;
+import com.linecorp.centraldogma.server.mirror.Mirror;
+import com.linecorp.centraldogma.server.storage.project.ProjectManager;
+import com.linecorp.centraldogma.server.storage.repository.MetaRepository;
+
+/**
+ * Annotated service object for managing mirroring service.
+ */
+@ProducesJson
+@RequiresRole(roles = ProjectRole.OWNER)
```
ikhoon, line 47: TODO: Dynamically check the permission of `meta`. #898

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/DefaultMetaRepository.java
```diff
@@ -16,142 +16,291 @@
...
-        final JsonNode mirrorsJson = (JsonNode) entries.get(PATH_MIRRORS).content();
-        if (!mirrorsJson.isArray()) {
-            throw new RepositoryMetadataException(
-                    PATH_MIRRORS + " must be an array: " + mirrorsJson.getNodeType());
-        }
+    private static void validateMirror(MirrorDto mirror) {
+        checkArgument(!Strings.isNullOrEmpty(mirror.id()), "Mirror ID is empty");
+        final Cron schedule = MirrorConfig.CRON_PARSER.parse(mirror.schedule());
+        final CronField secondField = schedule.retrieve(CronFieldName.SECOND);
+        checkArgument(!secondField.getExpression().asString().contains("*"),
```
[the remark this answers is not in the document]
ikhoon: Good suggestion. Let me control the value in the UI PR.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirroringMigrationService.java
ikhoon: Implemented rollback logic that removes all new migrated files and reverts the backup files.

ikhoon:
I found a bug in updating server's status during rolling restart. Let me send a separate PR to handle it first.

ikhoon:
I will proceed with this PR after #921 is merged.

ikhoon:
This PR is now ready for review again.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/CredentialServiceV1.java
```diff
@@ -0,0 +1,115 @@
...
+     * <p>Update the existing credential.
+     */
+    @Put("/projects/{projectName}/credentials")
+    @ConsumesJson
+    public CompletableFuture<PushResultDto> updateCredential(@Param String projectName,
+                                                             MirrorCredential credential, Author author) {
+        return createOrUpdate(projectName, credential, author, true);
+    }
+
+    private CompletableFuture<PushResultDto> createOrUpdate(String projectName,
```
jrhee17, line 102: Question) Is there a reason we separate create from update? (e.g. is it easier to express in the UI?)  Given the distributed nature of CD, I think I usually prefer that we provide an upsert method instead which is why I'm asking. (e.g. the semantics of `update` may not hold if two concurrent requests go through).

common/src/main/java/com/linecorp/centraldogma/internal/api/v1/PushResultDto.java
```diff
@@ -34,9 +35,15 @@ public class PushResultDto {
...
-    public PushResultDto(Revision revision, long commitTimeMillis) {
+    @JsonCreator
+    public PushResultDto(@JsonProperty("revision") Revision revision,
+                         @JsonProperty("pushedAt") Instant pushedAt) {
         this.revision = requireNonNull(revision, "revision");
-        pushedAt = ISO_INSTANT.format(Instant.ofEpochMilli(commitTimeMillis));
+        this.pushedAt = ISO_INSTANT.format(pushedAt);
+    }
+
+    public PushResultDto(Revision revision, long commitTimeMillis) {
```
jrhee17, line 45: nit; for consistent naming

```suggestion
    public PushResultDto(Revision revision, long pushedAt) {
```

jrhee17 (commented):
Went through the controller layer for now, looks good overall 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/api/MirroringServiceV1.java
```diff
@@ -0,0 +1,146 @@
...
+                          .collect(toImmutableList());
+        });
+    }
+
+    /**
+     * GET /projects/{projectName}/mirrors/{id}
+     *
+     * <p>Returns the mirror of the ID in the project mirror list.
+     */
+    @RequiresReadPermission(repository = Project.REPO_META)
```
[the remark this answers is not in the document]
ikhoon: I thought method annotation had higher precedence. If so, it needs to be fixed.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/CredentialServiceV1.java
ikhoon: I wanted to implement a more restful API.
- `create` needs `POST` method and `201 Created` status is returned.
- `update` uses `PUT` method to replace the whole content and 200 status will be returned instead.

Aside from the REST endpoint, the internal goes through the same code path.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirroringMigrationService.java
```diff
@@ -0,0 +1,428 @@
...
+        logger.info("Mirrors and credentials migration has been completed. (took: {} ms.)",
+                    stopwatch.elapsed().toMillis());
+
+        shortWords = null;
+    }
+
+    private boolean wasMigrated() throws Exception {
+        for (Project project : projectManager.list().values()) {
+            final MetaRepository repository = project.metaRepo();
+            if (getMetaData(repository, PATH_LEGACY_MIRRORS_BACKUP) != null) {
```
jrhee17, line 151: Question on the null check here - am I understanding correctly that if a legacy `mirrors.json`/`credentials.json` doesn't exist for a project then no backup file will exist?

i.e. If there is a repository without mirroring configured, will this method always return false?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/CredentialServiceV1.java
jrhee17: > Aside from the REST endpoint, the internal goes through the same code path.

Understood. I personally I prefer simplicity but I think it's fine to keep as is then.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirroringMigrationService.java
ikhoon: > if a legacy mirrors.json/credentials.json doesn't exist for a project then no backup file will exist?

Correct. no backup file won't be created if no mirror is configured.
https://github.com/line/centraldogma/blob/a80425c2780047e86ce05b04e226af08801192fd/server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirroringMigrationService.java#L164-L167

> i.e. If there is a repository without mirroring configured, will this method always return false?

This method returns `false` but nothing will be performed.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/CredentialServiceV1.java
ikhoon: `update` is not necessary for this API, but a consistent convention for all REST endpoints would be better. 

Users may get confused if only some resources provide the PUT operation.
As REST API continues to be added to CD, there will be a high possibility of using PUT in the future.
- POST is a non-idempotent operation but `PUT` is so retryable. 
- We may add more validations to `PUT` if some fields should be immutable.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirroringMigrationService.java
```diff
@@ -0,0 +1,428 @@
...
+            }
+            try {
+                migrateMirror(repository, (ObjectNode) mirror, mirrorIds, credentials);
+            } catch (Exception e) {
+                logger.warn("Failed to migrate a mirror config: {} (project: {})", mirror,
+                            repository.parent().name(), e);
+            }
+        }
+        // Back up the old mirrors.json file and don't use it anymore.
+        rename(repository, PATH_LEGACY_MIRRORS, PATH_LEGACY_MIRRORS_BACKUP, false);
```
jrhee17, line 187: Proposal) I'm curious if the old `meta/[mirrors|credentials].json` will be visible to users in the new mirroring UI. If not, what do you think of doing the rename/backup migrations after a few versions to ensure stability?

I realize that changes to the new mirroring/credentials won't be reflected in the old mirrors (unless we pre-implement and release some logic). I still think it's better than nothing and provides us an escape hatch if something goes wrong in the new release (e.g. it may not necessarily be due to this feature, but an armeria bug may force us to revert versions)

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirroringMigrationService.java
jrhee17: > This method returns false but nothing will be performed.

I see. I just wanted to make sure the following lines won't be executed after a migration is successfully completed.

https://github.com/line/centraldogma/blob/f0184f023f35208aa91949502439e7e3824a9a40/server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirroringMigrationService.java#L100-L108

server/src/main/java/com/linecorp/centraldogma/server/internal/api/CredentialServiceV1.java
jrhee17: > a consistent convention for all REST endpoints would be better.

I'm not strong on the definition/usefulness of REST. I just wanted to make clear the reasoning for adding this logic in case we need to revisit later.

server/src/main/java/com/linecorp/centraldogma/server/storage/project/InternalProjectInitializer.java
```diff
@@ -30,28 +31,52 @@
...
         final long creationTimeMillis = System.currentTimeMillis();
-        initializeInternalProject(executor, creationTimeMillis, INTERNAL_PROJECT_DOGMA);
+        try {
+            executor.execute(createProject(creationTimeMillis, Author.SYSTEM, INTERNAL_PROJECT_DOGMA))
+                    .get();
+        } catch (Throwable cause) {
+            final Throwable peeled = Exceptions.peel(cause);
+            if (peeled instanceof ReadOnlyException) {
+                // The executor has stopped right after starting up.
+                return;
```
jrhee17, line 70: Question to make sure I'm understanding correctly. Does this scenario occur because of the server status update? If a server doesn't have the necessary internal files but an exception isn't thrown here, could a CD server exist without the `dogma` project?

jrhee17 (commented):
Went through all of the files and looks good overall 👍   Just left a couple comments on whether migration can be done safely and we have a clear rollback plan. Other parts look good to me 👍

jrhee17 (commented):
Forgot to ask one more question, what do you think of allowing `null` ids when creating mirrors and credentials from a usability perspective. CD server could generate random ids when no id is specified.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirroringMigrationService.java
```diff
@@ -0,0 +1,428 @@
...
+        for (JsonNode mirror : mirrors) {
+            if (!mirror.isObject()) {
+                logger.warn("A mirror config must be an object: {} (project: {})", mirror,
+                            repository.parent().name());
+                continue;
+            }
+            try {
+                migrateMirror(repository, (ObjectNode) mirror, mirrorIds, credentials);
+            } catch (Exception e) {
+                logger.warn("Failed to migrate a mirror config: {} (project: {})", mirror,
```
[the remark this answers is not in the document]
ikhoon: Agreed. Let me log the exception and rethrow it.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirroringMigrationService.java
ikhoon: Hmm. It's a good idea, but it's not easy to apply. What if a bug is discovered a certain amount of time after the new mirroring configuration is exposed to users?

If an update has already been made in the new file, converting to the old file does not seem to be a good option. The migration itself should be verified through testing, and for bugs caused by external libraries, it would be better to revert the version only rather than reverting the all changes.

Speaking of the release strategy, what do you think of releasing a new version only for this change to minimize side effects?

server/src/main/java/com/linecorp/centraldogma/server/storage/project/InternalProjectInitializer.java
ikhoon: > Does this scenario occur because of the server status update?

Right. `"server-status.properties"` in a server was `WRITABLE` but `READ_ONLY` was applied from reply logs.

> could a CD server exist without the dogma project?

No, impossible. However, a server's status mode can be changed after the server starts successfully. I don't think it will be a problem for new servers because the existing data in a replica should be copied to the new ones before starting them.

ikhoon:
>  what do you think of allowing null ids when creating mirrors and credentials from a usability perspective.

ID is required to update the existing data. Previously, users updated the whole data such as `mirrors.json` or `credentials.json` at once. In the new UI, each configuration is updated separately, so we need ID to find the source file.

ikhoon:
> CD server could generate random ids when no id is specified.

I missed this point. A mirror configuration can refer to a credential. So `credential`'s ID will be used as a foreign key. ID is an identifier in the listing paging.

After I think about it in many ways, I still think users have to create and manage their own IDs.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirroringMigrationService.java
ikhoon: `/mirror-migration-job.json` is added to check if a migration has been performed before.
`/mirror-migration-job.json` will be written even if there are no mirror and credential configurations.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirroringMigrationService.java
jrhee17: > Hmm. It's a good idea, but it's not easy to apply. What if a bug is discovered a certain amount of time after the new mirroring configuration is exposed to users?

I see. I was thinking there is a higher probability that a bug is discovered right away rather than a long time. This comment also assumed that mirroring configurations are usually not updated often.
 
I understood we prefer the current approach - if a roll back to the previous version is needed, no mirrors will be executed.

jrhee17:
> After I think about it in many ways, I still think users have to create and manage their own IDs.

Sure, I'm not sure what the UI will look like. If there is a good way for users to know in advance the used ids, I guess it will be fine.

jrhee17 (approved):
I assume the changes will be released together with the new mirroring UI.  Looks good to me once the build passes 👍 👍 👍

ikhoon:
> I assume the changes will be released together with the new mirroring UI

This PR is going to be released before the mirroring UI. It may be difficult to detect bugs and handle them if this change was released along with the mirroring UI, so I wanted to release it separately.

Do you have any concerns about that?

jrhee17:
> It may be difficult to detect bugs and handle them if this change was released along with the mirroring UI, so I wanted to release it separately.

> Do you have any concerns about that?

I assume this comment is for the record since we already talked about this.
I misread the code thinking reading/writing to the new mirrors/credentials files is not allowed anymore. Releasing separately sounds good 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirroringMigrationService.java
```diff
@@ -0,0 +1,456 @@
...
+        commandExecutor.execute(Command.updateServerStatus(ServerStatus.WRITABLE))
+                       .get(1, TimeUnit.MINUTES);
+        logger.info("Mirrors and credentials migration has been completed. (took: {} ms.)",
+                    stopwatch.elapsed().toMillis());
+
+        shortWords = null;
+    }
+
+    private void logMigrationJob() throws Exception {
+        final ImmutableMap<String, Instant> data = ImmutableMap.of("timestamp", Instant.now());
```
minwoox, line 155: How about also adding the total number of the projects that have been migrated so we can easily know if the migration is successfully done?

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirroringMigrationService.java
```diff
@@ -0,0 +1,456 @@
...
+            remoteMigrationJobLog();
+        }
+    }
+
+    private void rollbackMigration(MetaRepository repository, String targetDirectory, String originalFile,
+                                   String backupFile) throws Exception {
+        // Delete all files in the target directory
+        final Map<String, Entry<?>> entries = repository.find(Revision.HEAD, targetDirectory + "**")
+                                                        .get();
+        for (String name : entries.keySet()) {
```
minwoox, line 264: How about removing the entries at once?

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirroringMigrationService.java
```diff
@@ -0,0 +1,456 @@
...
+                                                            Jackson.writeValueAsString(data));
+        final Command<CommitResult> command =
+                Command.push(Author.SYSTEM, InternalProjectInitializer.INTERNAL_PROJECT_DOGMA,
+                             Project.REPO_DOGMA, Revision.HEAD,
+                             "Migration of mirrors and credentials has been done", "",
+                             Markup.PLAINTEXT, change);
+        executeCommand(command);
+    }
+
+    private void remoteMigrationJobLog() throws Exception {
```
minwoox, line 166: remove?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/DefaultMetaRepository.java
```diff
@@ -73,90 +82,232 @@ public org.eclipse.jgit.lib.Repository jGitRepository() {
...
+    }
+
+    @Override
+    public CompletableFuture<Command<CommitResult>> createPushCommand(MirrorDto mirrorDto, Author author,
+                                                                      boolean update) {
+        validateMirror(mirrorDto);
+        if (update) {
+            final String summary = "Update the mirror '" + mirrorDto.id() + '\'';
+            return mirror(mirrorDto.id()).thenApply(mirror -> {
+                // Perform the update operation only if the mirror exists.
```
minwoox, line 241: Isn't this called even when the mirror wasn't there?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/CredentialServiceV1.java
```diff
@@ -0,0 +1,116 @@
...
+    /**
+     * POST /projects/{projectName}/credentials
+     *
+     * <p>Creates a new credential.
+     */
+    @RequiresWritePermission(repository = Project.REPO_META)
+    @Post("/projects/{projectName}/credentials")
+    @ConsumesJson
+    @StatusCode(201)
+    public CompletableFuture<PushResultDto> createCredential(@Param String projectName,
```
minwoox, line 85: So users always have to create the credential first before creating a mirror configuration.
Would it be worth if we provide an API that commits both together at once ?

common/src/main/java/com/linecorp/centraldogma/internal/Jackson.java
```diff
@@ -77,7 +78,8 @@ public final class Jackson {
         compactMapper.enable(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS);
         prettyMapper.enable(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS);
 
-        registerModules(new SimpleModule().addSerializer(Instant.class, InstantSerializer.INSTANCE)
+        registerModules(new Jdk8Module(),
```
minwoox, line 81: Which one made you add this module?

minwoox (commented):
Basically looks great. 👍 👍 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/DefaultMetaRepository.java
ikhoon: If no mirror is found, `EntryNotFoundException` will be raised. Did you mean something else?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/CredentialServiceV1.java
ikhoon: A credential can be used in multiple mirror configs so the relation is 1(credential):N(mirror).

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/DefaultMetaRepository.java
minwoox: I missed that. 😓 Please forget about the comment.

common/src/main/java/com/linecorp/centraldogma/internal/Jackson.java
ikhoon: It was added to serialize `Optional` in `MirrorCredential` but I decided to change the return type. Let me remove it.
https://github.com/line/centraldogma/pull/880/files#diff-3fcccba8a8cdf811e08f7c99cf837dbcc5c6c5164ff32caeecf9db1d395a1ee0R68

server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
```diff
@@ -438,33 +442,56 @@ public <T> CompletableFuture<MergedEntryDto<T>> mergeFiles(
...
+                    }
+                }
+            }
+
+            // TODO(ikhoon): Disallow creating a mirror with the commit API. Mirroring REST API should be used
+            //               to validate the input.
             final Optional<String> notAllowedLocalRepo =
                     Streams.stream(changes)
-                           .filter(change -> DefaultMetaRepository.PATH_MIRRORS.equals(change.path()))
+                           .filter(change -> isMetaFile(change.path()))
```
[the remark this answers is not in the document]
ikhoon: Changed to check only mirror files.

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -764,6 +794,16 @@ protected HttpResponse doGet(ServiceRequestContext ctx, HttpRequest req) {
         sb.annotatedService(API_V1_PATH_PREFIX,
                             new RepositoryServiceV1(executor, mds), decorator,
                             v1RequestConverter, jacksonRequestConverterFunction, v1ResponseConverter);
+
+        if (GIT_MIRROR_ENABLED) {
+            sb.annotatedService(API_V1_PATH_PREFIX,
+                                new MirroringServiceV1(projectApiManager, executor), decorator,
+                                v1RequestConverter, jacksonRequestConverterFunction, v1RequestConverter);
```
minwoox, line 801: nit: v1ResponseConverter

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -764,6 +794,16 @@ protected HttpResponse doGet(ServiceRequestContext ctx, HttpRequest req) {
...
+            sb.annotatedService(API_V1_PATH_PREFIX,
+                                new CredentialServiceV1(projectApiManager, executor), decorator,
+                                v1RequestConverter, jacksonRequestConverterFunction, v1RequestConverter);
```
minwoox, line 804: `v1ResponseConverter`

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirroringMigrationService.java
```diff
@@ -0,0 +1,464 @@
...
+                ex0.addSuppressed(mirrorException);
+                throw new MirrorMigrationException("Failed to rollback the mirror migration:", ex0);
+            }
+            throw mirrorException;
+        }
+
+        // Exit read-only mode.
+        commandExecutor.execute(Command.updateServerStatus(ServerStatus.WRITABLE))
+                       .get(1, TimeUnit.MINUTES);
+        logger.info("Mirrors and credentials migration has been completed. (took: {} ms.)",
```
minwoox, line 151: Can we also add `numMigratedProjects` to this log?

minwoox (approved):
Looks excellent. 
Thanks a lot, @ikhoon! 👍 👍 👍

common/build.gradle
```diff
@@ -7,6 +7,7 @@ dependencies {
     implementation libs.jackson.core
     implementation libs.jackson.databind
     implementation libs.jackson.datatype.jsr310
+    implementation libs.jackson.datatype.jdk8
```
minwoox, line 10: We also need to remove this line.

minwoox:
@ikhoon Could you check the failure?
https://github.com/line/centraldogma/actions/runs/9609269292/job/26503477206?pr=880#step:6:783
@trustin Could you take a look at this PR?

minwoox:
Let me merge this to avoid conflicts with https://github.com/line/centraldogma/pull/965
Thanks a lot, @ikhoon 👍 👍 👍

PR #881, opened 2023-09-07
Check remote revision before fetching when mirroring

### LINKED ISSUE #879: Investigate increased mirroring duration
After migrating to Apache MINA from JSch, the maximum mirroring duration is increased to 2 seconds which was 1 second before.
This can be a problem when there are multiple mirroring configurations that should be run at short intervals.

minwoox:
Motivation:
If there are no commits from the last mirroring, the mirroring service does not have to fetch from the remote.
Because commits are infrequently made, we can just use `git-ls-remote`, which is way cheaper than `git-fetch-pack`, to check the remote head revision.

When mirroring, an SSH session might be established a couple of times. Establishing an SSH session is a highly resource-intensive task. It would be nice if we could reduce it.

Modifications:
- Check the remote revision before fetching when mirroring if there's a new commit.
- Split `GitMirror` into:
  - `AbstractGitMirror` that is the common ancestor of `GitMirror` and `SshGitMirror`.
  - `SshGitMirror` for SSH mirroring and `GitMirror` for the rest of them.
- Share SSH client and session for a mirror job.

Result:
- Partially fix #879
- The average mirroring duration is now decreased approximately by 50%.

### DIFF
--- a/server-mirror-git/src/main/java/com/linecorp/centraldogma/server/internal/mirror/AbstractGitMirror.java
+++ b/server-mirror-git/src/main/java/com/linecorp/centraldogma/server/internal/mirror/AbstractGitMirror.java
@@ -1,0 +1,721 @@
+/*
+ * Copyright 2017 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.linecorp.centraldogma.server.internal.mirror;
+
+import static com.linecorp.centraldogma.server.storage.repository.FindOptions.FIND_ALL_WITHOUT_CONTENT;
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+import java.io.ByteArrayInputStream;
+import java.io.File;
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.function.Consumer;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import javax.annotation.Nullable;
+
+import org.eclipse.jgit.api.FetchCommand;
+import org.eclipse.jgit.api.LsRemoteCommand;
+import org.eclipse.jgit.api.PushCommand;
+import org.eclipse.jgit.api.RemoteSetUrlCommand;
+import org.eclipse.jgit.api.RemoteSetUrlCommand.UriType;
+import org.eclipse.jgit.api.TransportCommand;
+import org.eclipse.jgit.api.errors.GitAPIException;
+import org.eclipse.jgit.dircache.DirCache;
+import org.eclipse.jgit.dircache.DirCacheBuilder;
+import org.eclipse.jgit.dircache.DirCacheEditor;
+import org.eclipse.jgit.dircache.DirCacheEditor.DeletePath;
+import org.eclipse.jgit.dircache.DirCacheEditor.PathEdit;
+import org.eclipse.jgit.dircache.DirCacheEntry;
+import org.eclipse.jgit.ignore.IgnoreNode;
+import org.eclipse.jgit.ignore.IgnoreNode.MatchResult;
+import org.eclipse.jgit.lib.CommitBuilder;
+import org.eclipse.jgit.lib.Constants;
+import org.eclipse.jgit.lib.FileMode;
+import org.eclipse.jgit.lib.ObjectId;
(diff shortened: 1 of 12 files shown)


server-mirror-git/src/main/java/com/linecorp/centraldogma/server/internal/mirror/SshGitMirror.java
```diff
@@ -0,0 +1,257 @@
...
+            GitSshdSessionFactory sessionFactory) {
+        return command -> {
+            // Need to set the credentials provider, otherwise NPE is raised when creating GitSshdSession.
+            command.setCredentialsProvider(NoopCredentialsProvider.INSTANCE);
+            command.setTransportConfigCallback(transport -> {
+                final SshTransport sshTransport = (SshTransport) transport;
+                sshTransport.setSshSessionFactory(sessionFactory);
+            });
+        };
+    }
```
ikhoon, line 229: How about moving this method to `DefaultGitSshdSessionFactory` which may be simpler?
```java
class DefaultGitSshdSessionFactory {
  ...

  void configureCommand(TransportCommand<?, ?> command) {
    // Need to set the credentials provider, otherwise NPE is raised when creating GitSshdSession.
    command.setCredentialsProvider(NoopCredentialsProvider.INSTANCE);
    command.setTransportConfigCallback(transport -> {
        final SshTransport sshTransport = (SshTransport) transport;
        sshTransport.setSshSessionFactory(this);
    });
  }
}
```

ikhoon (commented):
Overall, looks good. Left minor suggestions.

server-mirror-git/src/main/java/com/linecorp/centraldogma/server/internal/mirror/SshGitMirror.java
minwoox: That's a good suggestion. Addressed. 😉

server-mirror-git/src/main/java/com/linecorp/centraldogma/server/internal/mirror/AbstractGitMirror.java
```diff
@@ -131,283 +136,285 @@ public GitMirror(Cron schedule, MirrorDirection direction, MirrorCredential cred
...
+            return findHeadBranchRef(git, headBranchRefName, refs);
         }
 
         // Otherwise, we need to figure out which branch we should fetch.
         // Fetch the remote reference list to determine the default branch.
-        final Collection<Ref> refs = git.lsRemote()
-                                        .setTags(false)
-                                        .setTimeout(GIT_TIMEOUT_SECS)
-                                        .call();
+        final Collection<Ref> refs = lsRemote(git, configurator, false);
```
jrhee17, line 382: Question) I know this is just a refactor, but I'm curious if automatic branch detection actually works with the new logic. If heads is false, IIUC `refs/heads/*` aren't fetched which means branches aren't fetched.

I have no problem if this works, but wanted to point out just in case it was somehow missed

jrhee17 (commented):
Overall looks good! Left some minor questions 🙇

jrhee17 (approved):
🍀 🍀 🍀

server-mirror-git/src/main/java/com/linecorp/centraldogma/server/internal/mirror/AbstractGitMirror.java
```diff
@@ -131,283 +136,285 @@ public GitMirror(Cron schedule, MirrorDirection direction, MirrorCredential cred
...
+        }
 
-                final String abbrId = reader.abbreviate(id).name();
-                if (id.name().equals(localSourceRevision)) {
-                    logger.info("Repository '{}' already at {}, {}#{}", localRepo().name(), abbrId,
-                                remoteRepoUri(), remoteBranch());
-                    return;
-                }
+        final PushCommand push = git.push();
+        configurator.accept(push);
```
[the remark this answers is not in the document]
minwoox: > So is it necessary that we always call configurator.accept for each command from now on?

Yes, 😉 

> I think I prefer the original style where Git automatically configures authentication for each command.

The reason why I've changed is to share the `client` and `session` for ls-remote, fetch and push commands.
Because the client and session must be closed after use, I think it's hard to use it with the old style.

server-mirror-git/src/main/java/com/linecorp/centraldogma/server/internal/mirror/AbstractGitMirror.java
minwoox: > If heads is false, IIUC refs/heads/* aren't fetched which means branches aren't fetched.

It's fetched. 😉 
```
// head true:
centraldogma-line git:(master) git ls-remote --heads
From https://git.linecorp.com/minu-song/centraldogma-line.git
6fa34705f21f9b032e71b0a14069ec1d456ea84f	refs/heads/add_actions
0be19a8460f987a1eea8dd621c32d45c1a144678	refs/heads/master

// head false:
centraldogma-line git:(master) git ls-remote
From https://git.linecorp.com/minu-song/centraldogma-line.git
0be19a8460f987a1eea8dd621c32d45c1a144678	HEAD
6fa34705f21f9b032e71b0a14069ec1d456ea84f	refs/heads/add_actions
0be19a8460f987a1eea8dd621c32d45c1a144678	refs/heads/master
1cd207f84ce0ce83e459bf2d355593f0ac44e17b	refs/heads/use_Javet
e73e4e573cd6183a00bb1fdf8f85d7b69f821dc7	refs/tags/centraldogma-line-0.14.0.0
3be144a526e31513cbc720b8cc8a63a2db7867c4	refs/tags/centraldogma-line-0.14.1.0
...
```

ikhoon (approved):
Thanks, @minwoox ❤️

minwoox:
Thanks for reviewing. 😉

trustin (approved):
:+1:

PR #883, opened 2023-09-13
Provide a way to configure server for a plugin

minwoox:
Motivation:
Currently, there's no way to configure the Armeria server outside of the server module. We can perhaps do this by providing DI, SPI, or adding a method to `Plugin`.

Modification:
- Add `AllReplicasPlugin` class for configuring Central Dogma server

Result:
- You can now add configurations to the Central Dogma server for your plugin using `AllReplicasPlugin.init()` method.


jrhee17 (approved):
Changes look good to me 👍  Thanks @minwoox 🙇 👍 🙇

server/src/main/java/com/linecorp/centraldogma/server/plugin/PluginInitContext.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+public final class PluginInitContext {
+
+    private final CentralDogmaConfig config;
+    private final ServerBuilder serverBuilder;
+
+    /**
+     * Creates a new instance.
+     */
+    public PluginInitContext(CentralDogmaConfig config, ServerBuilder serverBuilder) {
```
ikhoon, line 35:
```suggestion
    PluginInitContext(CentralDogmaConfig config, ServerBuilder serverBuilder) {
```

ikhoon (approved):
Thanks, @minwoox!

server/src/main/java/com/linecorp/centraldogma/server/plugin/PluginInitContext.java
```diff
@@ -0,0 +1,53 @@
...
+
+import static java.util.Objects.requireNonNull;
+
+import com.linecorp.armeria.server.ServerBuilder;
+import com.linecorp.centraldogma.server.CentralDogmaConfig;
+
+/**
+ * A context that is used to pass when calling {@link AllReplicasPlugin#init(PluginInitContext)}.
+ */
+public final class PluginInitContext {
```
jrhee17, line 27: Question) I was thinking we would need at least `ProjectManager`, `CommandExecutor` will be needed to build services which can write to `CentralDogma` storage. Is there no need for these values? (it looks like they are available at init time).

I actually expect we may also want access to `MeterRegistry`, and actually all values in `PluginContext` eventually.

jrhee17 (approved):

server/src/main/java/com/linecorp/centraldogma/server/plugin/PluginInitContext.java
minwoox: Oops, I thought somehow they are not created when the init method is called. 😅 
Let me fix that. 😉

minwoox: Fixed. PTAL. 😉

server/src/main/java/com/linecorp/centraldogma/server/plugin/PluginInitContext.java
minwoox: Oops, I cannot remove the public modifier because the class is in a different package.

jrhee17 (approved):
Sorry about the late check! Looks good to me! 🙇 👍 🚀

minwoox:
Let me merge this for now to keep working on the control plane.

PR #884, opened 2023-09-14
Try to configure authentication for command from `GitWithAuth`

jrhee17:
**Motivation**

- Goes well with the name `GitWithAuth` which implies that authentication is done from the object implicitly
- We don't have to pass a `Consumer` around for each location where we want to execute a git command

### DIFF
--- a/server-mirror-git/src/main/java/com/linecorp/centraldogma/server/internal/mirror/AbstractGitMirror.java
+++ b/server-mirror-git/src/main/java/com/linecorp/centraldogma/server/internal/mirror/AbstractGitMirror.java
@@ -37,9 +37,6 @@ import java.util.stream.Stream;
 
 import javax.annotation.Nullable;
 
-import org.eclipse.jgit.api.FetchCommand;
-import org.eclipse.jgit.api.LsRemoteCommand;
-import org.eclipse.jgit.api.PushCommand;
 import org.eclipse.jgit.api.RemoteSetUrlCommand;
 import org.eclipse.jgit.api.RemoteSetUrlCommand.UriType;
 import org.eclipse.jgit.api.TransportCommand;
@@ -225,11 +222,11 @@ abstract class AbstractGitMirror extends AbstractMirror {
             updateRef(gitRepository, revWalk, headBranchRefName, nextCommitId);
         }
 
-        final PushCommand push = git.push();
-        push.setRefSpecs(new RefSpec(headBranchRefName))
-            .setAtomic(true)
-            .setTimeout(GIT_TIMEOUT_SECS)
-            .call();
+        git.push()
+           .setRefSpecs(new RefSpec(headBranchRefName))
+           .setAtomic(true)
+           .setTimeout(GIT_TIMEOUT_SECS)
+           .call();
     }
 
     void mirrorRemoteToLocal(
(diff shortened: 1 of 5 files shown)


server-mirror-git/src/main/java/com/linecorp/centraldogma/server/internal/mirror/SshGitMirror.java
```diff
@@ -87,52 +88,42 @@ final class SshGitMirror extends AbstractGitMirror {
...
-        ClientSession session = null;
-        try (GitWithAuth git = openGit(workDir)) {
-            sshClient = createSshClient();
-            session = createSession(sshClient, git);
-            final DefaultGitSshdSessionFactory sessionFactory =
-                    new DefaultGitSshdSessionFactory(sshClient, session);
-            mirrorRemoteToLocal(git, executor, maxNumFiles, maxNumBytes, sessionFactory::configureCommand);
+        final SshClient sshClient = createSshClient();
+        final URIish remoteUri = remoteUri();
+        final ClientSession session = createSession(sshClient, remoteUri);
```
minwoox, line 112: We need to close the `sshClient` even if an exception is raised while creating `ClientSession`.

server-mirror-git/src/main/java/com/linecorp/centraldogma/server/internal/mirror/AbstractGitMirror.java
```diff
@@ -228,20 +226,18 @@ dirCache, new InsertText(mirrorStatePath.substring(1), // Strip the leading '/'.
         }
 
         final PushCommand push = git.push();
-        configurator.accept(push);
         push.setRefSpecs(new RefSpec(headBranchRefName))
```
minwoox, line 229: Could you revert this as well?
```java
git.push().setRefSpecs(...)
```

minwoox (commented):
I thought configuring commands right away was more intuitive but this also looks good. 👍

server-mirror-git/src/main/java/com/linecorp/centraldogma/server/internal/mirror/AbstractGitMirror.java
```diff
@@ -136,21 +135,21 @@ abstract class AbstractGitMirror extends AbstractMirror {
         }
     }
 
-    GitWithAuth openGit(File workDir, String jGitUri,
-                        URIish remoteUri) throws IOException, URISyntaxException, GitAPIException {
```
minwoox, line 140: That was my mistake. 😓

server-mirror-git/src/test/java/com/linecorp/centraldogma/server/internal/mirror/GitMirrorTest.java
```diff
@@ -29,12 +29,32 @@
...
+        @Override
+        public void close() {
+            System.out.println("close");
+        }
+    }
+
+    @Test
+    void asdf() {
+        try (SimpleClass sc = new SimpleClass()) {
+            System.out.println("Hello");
```
minwoox: cruft. 😆

server-mirror-git/src/main/java/com/linecorp/centraldogma/server/internal/mirror/SshGitMirror.java
```diff
@@ -111,14 +115,17 @@ protected void mirrorLocalToRemote(File workDir, int maxNumFiles, long maxNumByt
...
         ClientSession session = null;
-        try (GitWithAuth git = openGit(workDir)) {
+        try {
             sshClient = createSshClient();
-            session = createSession(sshClient, git);
+            session = createSession(sshClient, remoteUri);
             final DefaultGitSshdSessionFactory sessionFactory =
                     new DefaultGitSshdSessionFactory(sshClient, session);
-            mirrorRemoteToLocal(git, executor, maxNumFiles, maxNumBytes, sessionFactory::configureCommand);
+            try (GitWithAuth git = openGit(workDir, remoteUri, sessionFactory::configureCommand)) {
```
minwoox, line 126: How about using the previous style so that we can use only one try with resources block?

jrhee17: `openGit` has a dependency on `sessionFactory`. Do you mean initializing the variable as `null` and closing in the `finally` block manually?

minwoox: Ah ha, the parameters of `openGit` methods are now changed and I forgot about it. 😅 
I'm fine with the current approach. 👍

minwoox (approved):
Thanks for cleaning this up! 👍

ikhoon (approved):
Thanks, @jrhee17!

PR #885, opened 2023-09-15
Close an SSH session when an auth fails

minwoox:
Motivation:
An SSH session wasn't closed when an exception was raised while calling `session.auth()`

Modification:
- Close an SSH session when an exception is raised while calling `session.auth()`

Result:
- An SSH session is closed when an exception is raised while calling `sesison.auth()`

### DIFF
--- a/server-mirror-git/src/main/java/com/linecorp/centraldogma/server/internal/mirror/SshGitMirror.java
+++ b/server-mirror-git/src/main/java/com/linecorp/centraldogma/server/internal/mirror/SshGitMirror.java
@@ -53,6 +53,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.cronutils.model.Cron;
+import com.google.common.annotations.VisibleForTesting;
 
 import com.linecorp.centraldogma.server.MirrorException;
 import com.linecorp.centraldogma.server.command.CommandExecutor;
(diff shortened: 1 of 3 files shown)


ikhoon (approved):
Thanks for the quick fix! 🙇‍♂️

server-mirror-git/src/main/java/com/linecorp/centraldogma/server/internal/mirror/SshGitMirror.java
```diff
@@ -170,27 +145,37 @@ private SshClient createSshClient() {
...
+                                       sshClient))
+                               .getSession();
             session.auth().verify(GitModuleProperties.AUTH_TIMEOUT.getRequired(session));
             logger.trace("The session established: {}", session);
             return session;
-        } catch (IOException e) {
-            throw new RuntimeException(e);
+        } catch (Throwable t) {
+            if (session != null) {
+                session.close(true);
```
jrhee17, line 176: Understood that this is the actual point that is expected to fix the start/stop logs

jrhee17 (approved):
Looks good! Thanks @minwoox 🙇 👍 🙇

server-mirror-git/src/main/java/com/linecorp/centraldogma/server/internal/mirror/SshGitMirror.java
minwoox: That is correct. 👍

PR #890, opened 2023-11-02
Provide ConfigValueConverter to hide secret properties in Central Dog…

minwoox:
…ma configuration file.

Motivation:
We can enhance the security of the Central Dogma configuration by allowing users to specify sensitive secret values using a variable in the configuration file and converting them using the `ConfigValueConverter`. This will help in safeguarding sensitive information and prevent accidental exposure of secrets.

Modifications:
- Added `ConfigValueConverter` that can be loaded via SPI.
  - `replication.secret`, `tls.keyCertChain`, `tls.key` and `tls.keyPassword` can now be converted using the converter.

Result:
- You can now easily specify secret values with variables, in the configuration file, that are converted using `ConfigValueConverter`. This enhances the security of Central Dogma configuration, reducing the risk of sensitive information exposure.
- (Deprecation) `tls.keyCertChainFile` and `tls.keyFile` in the configuration are deprecated.
  - Use `tls.keyCertChain` and `tls.key` with the `file:` prefix or `ConfigValueConverter` instead.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
@@ -33,6 +33,7 @@ import static java.util.Objects.requireNonNull;
 
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
 import java.net.InetSocketAddress;
 import java.util.Collections;
 import java.util.List;
@@ -501,7 +502,10 @@ public class CentralDogma implements AutoCloseable {
             try {
                 final TlsConfig tlsConfig = cfg.tls();
                 if (tlsConfig != null) {
-                    sb.tls(tlsConfig.keyCertChainFile(), tlsConfig.keyFile(), tlsConfig.keyPassword());
+                    try (InputStream keyCertChainInputStream = tlsConfig.keyCertChainInputStream();
+                         InputStream keyInputStream = tlsConfig.keyInputStream()) {
+                        sb.tls(keyCertChainInputStream, keyInputStream, tlsConfig.keyPassword());
+                    }
                 } else {
                     logger.warn(
                             "Missing TLS configuration. Generating a self-signed certificate for TLS support.");
--- a/server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaConfig.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaConfig.java
@@ -36,11 +36,16 @@ import java.io.File;
 import java.io.IOException;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
+import java.util.ServiceLoader;
 import java.util.function.Predicate;
 
 import javax.annotation.Nullable;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.core.JsonGenerator;
@@ -59,6 +64,7 @@ import com.fasterxml.jackson.databind.util.StdConverter;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableList.Builder;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Streams;
 
 import com.linecorp.armeria.common.HttpHeaderNames;
 import com.linecorp.armeria.common.SessionProtocol;
@@ -74,6 +80,44 @@ import io.netty.util.NetUtil;
  * {@link CentralDogma} server configuration.
  */
 public final class CentralDogmaConfig {
+
+    private static final Logger logger = LoggerFactory.getLogger(CentralDogmaConfig.class);
+
+    private static final List<ConfigValueConverter> CONFIG_VALUE_CONVERTERS;
+
+    static {
+        final ArrayList<ConfigValueConverter> configValueConverters = new ArrayList<>();
+        Streams.stream(ServiceLoader.load(ConfigValueConverter.class)).forEach(configValueConverters::add);
+        configValueConverters.add(DefaultConfigValueConverter.INSTANCE);
+        CONFIG_VALUE_CONVERTERS = ImmutableList.copyOf(configValueConverters);
+        logger.debug("Available {}s: {}", ConfigValueConverter.class.getName(), CONFIG_VALUE_CONVERTERS);
+    }
+
+    @Nullable
+    static String convertValue(@Nullable String value) {
+        if (value == null) {
+            return null;
+        }
+
+        final int index = value.indexOf(':');
+        if (index < 0) {
+            // no prefix
+            return value;
+        }
+
+        final String prefix = value.substring(0, index);
+        final String rest = value.substring(index + 1);
+
+        for (ConfigValueConverter converter : CONFIG_VALUE_CONVERTERS) {
+            if (converter.supportedPrefixes().contains(prefix)) {
+                return converter.convert(prefix, rest);
+            }
+        }
+
+        // return the value as is instead of rest because the value might contain ':' without prefix.
+        // e.g. "@#$:@#$_is_not_prefix"
+        return value;
+    }
 
     private final File dataDir;
 
--- a/server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaConfig.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaConfig.java
@@ -110,8 +110,12 @@ public final class CentralDogmaConfig {
         logger.debug("Available {}s: {}", ConfigValueConverter.class.getName(), sb);
     }
 
+    /**
+     * Converts the specified {@code value} using {@link ConfigValueConverter} if the specified {@code value}
+     * starts with a prefix followed by a colon {@code ':'}.
+     */
     @Nullable
-    static String convertValue(@Nullable String value, String propertyName) {
+    public static String convertValue(@Nullable String value, String propertyName) {
         if (value == null) {
             return null;
         }
--- a/server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaConfig.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaConfig.java
@@ -38,6 +38,8 @@ import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.ServiceLoader;
 import java.util.function.Predicate;
@@ -63,6 +65,7 @@ import com.fasterxml.jackson.databind.node.JsonNodeType;
 import com.fasterxml.jackson.databind.util.StdConverter;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableList.Builder;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Streams;
 
@@ -83,14 +86,28 @@ public final class CentralDogmaConfig {
 
     private static final Logger logger = LoggerFactory.getLogger(CentralDogmaConfig.class);
 
-    private static final List<ConfigValueConverter> CONFIG_VALUE_CONVERTERS;
+    private static final Map<String, ConfigValueConverter> CONFIG_VALUE_CONVERTERS;
 
     static {
         final ArrayList<ConfigValueConverter> configValueConverters = new ArrayList<>();
         Streams.stream(ServiceLoader.load(ConfigValueConverter.class)).forEach(configValueConverters::add);
         configValueConverters.add(DefaultConfigValueConverter.INSTANCE);
-        CONFIG_VALUE_CONVERTERS = ImmutableList.copyOf(configValueConverters);
-        logger.debug("Available {}s: {}", ConfigValueConverter.class.getName(), CONFIG_VALUE_CONVERTERS);
+        final ImmutableMap.Builder<String, ConfigValueConverter> builder = ImmutableMap.builder();
+        for (ConfigValueConverter configValueConverter : configValueConverters) {
+            configValueConverter.supportedPrefixes()
+                                .forEach(prefix -> builder.put(prefix, configValueConverter));
+        }
+        CONFIG_VALUE_CONVERTERS = ImmutableMap.copyOf(builder.buildOrThrow());
+        final StringBuilder sb = new StringBuilder();
+        sb.append('{');
+        CONFIG_VALUE_CONVERTERS.entrySet().stream().sorted(Entry.comparingByKey()).forEach(
+                entry -> sb.append(entry.getKey())
+                           .append('=')
+                           .append(entry.getValue().getClass().getSimpleName()).append(", "));
+        sb.setLength(sb.length() - 2);
+        sb.append('}');
+
+        logger.debug("Available {}s: {}", ConfigValueConverter.class.getName(), sb);
     }
 
     @Nullable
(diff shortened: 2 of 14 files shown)


server/src/main/java/com/linecorp/centraldogma/server/TlsConfig.java
```diff
@@ -15,55 +15,117 @@
...
     }
 
     /**
      * Returns a private key file which is created with the given {@code keyFilePath}.
+     *
+     * @deprecated Use {@link #keyInputStream()}.
      */
+    @Nullable
     @JsonProperty
+    @Deprecated
```
jrhee17, line 90: Note: I was originally confused since I thought this didn't need to be deprecated (e.g. it may be useful to read an encrypted file and decrypt on-the-fly)

jrhee17 (approved):
Left some concerns, but overall looks good. Thanks @minwoox 🙇 👍 🙇

server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaConfig.java
```diff
@@ -75,6 +81,44 @@
...
+            return null;
+        }
+
+        final int index = value.indexOf(':');
+        if (index < 0) {
+            // no prefix
+            return value;
+        }
+
+        final String prefix = value.substring(0, index);
```
ikhoon, line 108: Should we early return if `prefix` is empty?

server/src/main/java/com/linecorp/centraldogma/server/DefaultConfigValueConverter.java
```diff
@@ -0,0 +1,51 @@
...
+
+import com.google.common.collect.ImmutableList;
+
+enum DefaultConfigValueConverter implements ConfigValueConverter {
+    INSTANCE;
+
+    private static final String FILE = "file";
+
+    // TODO(minwoox): Add more prefixes such as classpath, url, etc.
+    private static final List<String> SUPPORTED_PREFIX = ImmutableList.of(FILE);
```
ikhoon, line 32: Should we also support `plain` in case a value starts with `file` or some prefixes which are actually part of the data, e.g. `plain:file:;!abc`? Although it would rarely happen.

server/src/main/java/com/linecorp/centraldogma/server/DefaultConfigValueConverter.java
```diff
@@ -0,0 +1,51 @@
...
+    @Override
+    public List<String> supportedPrefixes() {
+        return SUPPORTED_PREFIX;
+    }
+
+    @Override
+    public String convert(String prefix, String value) {
+        if (!FILE.equals(prefix)) {
+            throw new IllegalArgumentException("unsupported prefix: " + prefix + ", value: " + value);
+        }
```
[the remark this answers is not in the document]
minwoox: Alritht, let me remove it.

server/src/main/java/com/linecorp/centraldogma/server/DefaultConfigValueConverter.java
```diff
@@ -0,0 +1,51 @@
...
+
+        try {
+            return new String(Files.readAllBytes(Paths.get(value)), StandardCharsets.UTF_8);
+        } catch (IOException e) {
+            throw new RuntimeException("failed to read a file: " + value, e);
+        }
```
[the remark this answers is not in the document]
minwoox: Because we have the `supportedPrefixes()` method, I prefer to just throw an exception instead of chaining.
But, please let me know if you are strong on this.

server/src/main/java/com/linecorp/centraldogma/server/TlsConfig.java
```diff
@@ -15,55 +15,117 @@
...
-    public TlsConfig(@JsonProperty(value = "keyCertChainFile", required = true) File keyCertChainFile,
-                     @JsonProperty(value = "keyFile", required = true) File keyFile,
+    public TlsConfig(@JsonProperty(value = "keyCertChainFile") @Nullable File keyCertChainFile,
+                     @JsonProperty(value = "keyFile") @Nullable File keyFile,
+                     @JsonProperty(value = "keyCertChain") @Nullable String keyCertChain,
+                     @JsonProperty(value = "key") @Nullable String key,
                      @JsonProperty("keyPassword") @Nullable String keyPassword) {
-        this.keyCertChainFile = requireNonNull(keyCertChainFile, "keyCertChainFile");
-        this.keyFile = requireNonNull(keyFile, "keyFile");
+        checkArgument(keyCertChainFile != null || keyCertChain != null,
```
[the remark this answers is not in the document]
minwoox: That's a good suggestion.

server/src/main/java/com/linecorp/centraldogma/server/TlsConfig.java
minwoox: That's what the converter is for. 😉
```
"tls.key" : "encryption:myKeyVariable"

class NClavisConvigValueConverter implements ConfigValueConverter {
  public string convert(String prefix, String value) {
    assert prefix.equals("encryption");
    if ("myKeyVariable".equals(value)) {
      // Read a specific encrypted file and decrypt it using thrid-party service.
    }
    ...
  }
}
```

server/src/main/java/com/linecorp/centraldogma/server/TlsConfig.java
```diff
@@ -15,55 +15,117 @@
...
+    @MustBeClosed
+    public InputStream keyInputStream() {
+        return inputStream(keyFile, key, "key");
+    }
+
+    private static InputStream inputStream(@Nullable File file,
+                                           @Nullable String property, String propertyName) {
+        if (file != null) {
+            try {
+                return Files.newInputStream(file.toPath());
```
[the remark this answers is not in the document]
minwoox: It calls `ServerBuilder.tls()` which does the job in it so I've just created an inputstream. But let me do that there's no reason not to do that.

server/src/main/java/com/linecorp/centraldogma/server/TlsConfig.java
```diff
@@ -75,10 +137,11 @@ public String keyPassword() {
 
     @Override
     public String toString() {
-        return MoreObjects.toStringHelper(this)
+        return MoreObjects.toStringHelper(this).omitNullValues()
                           .add("keyCertChainFile", keyCertChainFile)
                           .add("keyFile", keyFile)
-                          .add("keyPassword", keyPassword)
+                          .add("keyCertChain", keyCertChain)
+                          .add("key", key)
```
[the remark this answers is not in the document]
minwoox: Oops. 😉

server/src/main/java/com/linecorp/centraldogma/server/DefaultConfigValueConverter.java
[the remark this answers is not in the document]
ikhoon: What I meant was to add `throws Exception` signature to the method instead of using `RuntimeException`. 
```java
String convert(String prefix, String value) throws Exception;
```

server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaConfig.java
```diff
@@ -75,6 +81,44 @@
...
+
+        for (ConfigValueConverter converter : CONFIG_VALUE_CONVERTERS) {
+            if (converter.supportedPrefixes().contains(prefix)) {
+                return converter.convert(prefix, rest);
+            }
+        }
+
+        // return the value as is instead of rest because the value might contain ':' without prefix.
+        // e.g. "@#$:@#$_is_not_prefix"
+        return value;
```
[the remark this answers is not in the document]
jrhee17: > What's the benefit of throwing an exception you think

Say that `asdf:<encrypted value>`  is used but the user forgets to specify an SPI.
For inputs that are not well defined, the encrypted value might be used as-is without the user being aware.

server/src/main/java/com/linecorp/centraldogma/server/DefaultConfigValueConverter.java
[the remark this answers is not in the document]
ikhoon: On second thought, it seems okay to raise an exception implicitly. Please ignore this comment then.

server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaConfig.java
ikhoon: > Do you mean a String that starts with :?

Right, I thought about that case.

server/src/main/java/com/linecorp/centraldogma/server/ConfigValueConverter.java
```diff
@@ -0,0 +1,37 @@
...
+
+import java.util.List;
+
+/**
+ * A converter that converts a value of certain configuration properties in {@link CentralDogmaConfig}.
+ * Currently, it is used to convert a value of {@code replication.secret},
+ * {@code tls.keyCertChain} and {@code tls.key} properties.
+ * Implement this interface and register it via SPI to convert a value of the properties.
+ */
+public interface ConfigValueConverter {
```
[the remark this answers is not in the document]
jrhee17: I was imagining it might be cool if some files can be marked as "encryptable" and users can also resue this scheme. We may also reuse this to places like `mirrors.json`.
However, I guess we can consider it later if needed

site/src/sphinx/setup-configuration.rst
jrhee17, line 1: Doesn't have to be done now but we may want to mark which properties are "encryptable".
It might seem confusing for users that some properties use the provided scheme while other don't.

minwoox: > but we may want to mark which properties are "encryptable".

Was going to update it after the PR is sufficiently completed. 😉

server/src/main/java/com/linecorp/centraldogma/server/ConfigValueConverter.java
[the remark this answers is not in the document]
minwoox: That's a good idea and I think we can probably use this class or need another interface for that.
Let's do it later when we need it. 😉

server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaConfig.java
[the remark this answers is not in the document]
minwoox: Ah ha, that's a good suggestion. One of the reasons that I didn't throw an exception is that a user already contains `:` in their secret so that irrelevant exception is raised.
However, because there are not so many users who run Central Dogma, I think we can make a breaking change and provide the documentation about the prefix. 😉

server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaConfig.java
minwoox: Updated in the above if condition. 😉

server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaConfig.java
```diff
@@ -75,6 +81,42 @@
...
+
+    private static final List<ConfigValueConverter> CONFIG_VALUE_CONVERTERS;
+
+    static {
+        final ArrayList<ConfigValueConverter> configValueConverters = new ArrayList<>();
+        Streams.stream(ServiceLoader.load(ConfigValueConverter.class)).forEach(configValueConverters::add);
+        configValueConverters.add(DefaultConfigValueConverter.INSTANCE);
+        CONFIG_VALUE_CONVERTERS = ImmutableList.copyOf(configValueConverters);
+        logger.debug("Available {}s: {}", ConfigValueConverter.class.getName(), CONFIG_VALUE_CONVERTERS);
+    }
```
trustin, line 94: - Can we keep a `Map<String /* prefix */, ConfigValueConverter>` and fail when duplicate prefixes are found?
- We could then log a `Map<String /* prefix */, String /* class name *>` instead of ugly `className@12345`.

server/src/main/java/com/linecorp/centraldogma/server/DefaultConfigValueConverter.java
```diff
@@ -0,0 +1,52 @@
...
+import com.google.common.collect.ImmutableList;
+
+enum DefaultConfigValueConverter implements ConfigValueConverter {
+    INSTANCE;
+
+    private static final String PLAIN = "plain";
+    private static final String FILE = "file";
+
+    // TODO(minwoox): Add more prefixes such as classpath, url, etc.
+    private static final List<String> SUPPORTED_PREFIX = ImmutableList.of(PLAIN, FILE);
```
trustin, line 33:
```suggestion
    private static final List<String> SUPPORTED_PREFIXES = ImmutableList.of(PLAIN, FILE);
```

server/src/main/java/com/linecorp/centraldogma/server/DefaultConfigValueConverter.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @Override
+    public List<String> supportedPrefixes() {
+        return SUPPORTED_PREFIX;
+    }
+
+    @Override
+    public String convert(String prefix, String value) {
+        if (PLAIN.equals(prefix)) {
```
trustin, line 42: nit: Could use `switch-case`?

trustin (commented):
Just a few nits :-)

server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaConfig.java
minwoox: That's a good suggestion. All addressed. 👍

server/src/main/java/com/linecorp/centraldogma/server/DefaultConfigValueConverter.java
```diff
@@ -0,0 +1,56 @@
...
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.List;
+
+import com.google.common.collect.ImmutableList;
+
+enum DefaultConfigValueConverter implements ConfigValueConverter {
+    INSTANCE;
+
+    private static final String PLAIN = "plain";
```
trustin, line 29: Should this be `plaintext`? :thinking:

trustin (approved):
Nice work, @minwoox! :bow:

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -501,7 +502,10 @@ private Server startServer(ProjectManager pm, CommandExecutor executor,
             try {
                 final TlsConfig tlsConfig = cfg.tls();
                 if (tlsConfig != null) {
-                    sb.tls(tlsConfig.keyCertChainFile(), tlsConfig.keyFile(), tlsConfig.keyPassword());
+                    try (InputStream keyCertChainInputStream = tlsConfig.keyCertChainInputStream();
+                         InputStream keyInputStream = tlsConfig.keyInputStream()) {
+                        sb.tls(keyCertChainInputStream, keyInputStream, tlsConfig.keyPassword());
+                    }
```
ikhoon, line 508: Q) Don't We need `ConfigValueConverter` for the passwords in SAML properties?  https://github.com/line/centraldogma/blob/be15612884080517614bb1d468ffb102514a4c30/server-auth/saml/src/main/java/com/linecorp/centraldogma/server/auth/saml/SamlAuthProperties.java#L186-L195

ikhoon: It seems there is a TODO for that. https://github.com/line/centraldogma/pull/890/files#diff-1a885e58bb68ec2d7203b8bf406e7ef91752da64cce733cf52c6f0df9e51929cR104
Then, ignore this comment.

ikhoon (approved):
Thanks, @minwoox!

server/src/main/java/com/linecorp/centraldogma/server/DefaultConfigValueConverter.java
minwoox: Fixed. 😉

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
minwoox: Just applied the passwords in SAML. PTAL. 😉

PR #898, opened 2023-11-27
Fix to enable changing the permission of internal repositories.

minwoox:
Motivation:
Internal repositories are initially created with the [internal](https://github.com/line/centraldogma/blob/0cf273402d70773e0aab1e2b858111a1c1c0f22c/server/src/main/java/com/linecorp/centraldogma/server/metadata/PerRolePermissions.java#L61) permission, and I previously restricted the ability to modify this permission. However, users may have valid reasons to adjust the permission settings.

Modifications:
- Implemented functionality to change the permission of internal repositories.
  - Note: The guest role remains restricted from having any permission on internal repositories.

Result:
- You can now have the capability to modify the permission settings for internal repositories, excluding the guest role.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java
@@ -407,8 +407,11 @@ public class MetadataService {
         requireNonNull(perRolePermissions, "perRolePermissions");
 
         if (Project.isReservedRepoName(repoName)) {
-            throw new UnsupportedOperationException(
-                    "can't update the per role permission for internal repository: " + repoName);
+            final Set<Permission> guest = perRolePermissions.guest();
+            if (!guest.isEmpty()) {
+                throw new UnsupportedOperationException(
+                        "can't give a permission to guest for internal repository: " + repoName);
+            }
         }
 
         final JsonPointer path = JsonPointer.compile("/repos" + encodeSegment(repoName) +
--- a/server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java
@@ -406,11 +406,16 @@ public class MetadataService {
         requireNonNull(repoName, "repoName");
         requireNonNull(perRolePermissions, "perRolePermissions");
 
-        if (Project.isReservedRepoName(repoName)) {
+        if (Project.REPO_DOGMA.equals(repoName)) {
+            throw new UnsupportedOperationException(
+                    "Can't update the per role permission for internal repository: " + repoName);
+        }
+
+        if (Project.REPO_META.equals(repoName)) {
             final Set<Permission> guest = perRolePermissions.guest();
             if (!guest.isEmpty()) {
                 throw new UnsupportedOperationException(
-                        "can't give a permission to guest for internal repository: " + repoName);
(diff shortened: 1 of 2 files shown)


server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java
```diff
@@ -407,8 +407,11 @@ public CompletableFuture<Revision> updatePerRolePermissions(Author author,
         requireNonNull(perRolePermissions, "perRolePermissions");
 
         if (Project.isReservedRepoName(repoName)) {
-            throw new UnsupportedOperationException(
-                    "can't update the per role permission for internal repository: " + repoName);
+            final Set<Permission> guest = perRolePermissions.guest();
+            if (!guest.isEmpty()) {
+                throw new UnsupportedOperationException(
+                        "can't give a permission to guest for internal repository: " + repoName);
```
ikhoon, line 413: For consistency with the permission of `internalPermissions`, should we disallow giving permissions to `MEMBER` for internal repositories?
https://github.com/line/centraldogma/blob/0cf273402d70773e0aab1e2b858111a1c1c0f22c/server/src/main/java/com/linecorp/centraldogma/server/metadata/PerRolePermissions.java#L55-L56

minwoox: I believed we had previously discussed this matter and reached a consensus to delegate permission-granting authority to the owner for members. 😉 
`internalPermissions` is assigned to internal repositories upon creation, and the owner has the discretion to change it at will. While I'm not strongly opposed, I see no compelling reason to restrict granting permissions to members.

ikhoon: Understood.
I was confused that the permission could be also granted to `dogma` repo.

ikhoon (approved):
Thanks, @minwoox!

server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java
jrhee17, line 413: Since I believe this exception will be directly exposed to the end user

```suggestion
                        "Can't give a permission to guest for internal repository: " + repoName);
```

jrhee17: Technically, I believe allowing MEMBER access to `dogma` can potentially allow MEMBERs to "kick" OWNERs out of projects.
I don't really know the distinction of owner and member though. If the meaning is simply to add an extra layer of authority, I guess the current change is fine.

jrhee17 (approved):
👍 👍 👍

server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java
minwoox: Ah, got it. The dogma repo is restricted from being accessed by any user, so I initially thought this would suffice: https://github.com/line/centraldogma/blob/0cf273402d70773e0aab1e2b858111a1c1c0f22c/server/src/main/java/com/linecorp/centraldogma/server/internal/api/auth/RequiresPermissionDecorator.java#L78-L83.

However, for clarity, I believe it's beneficial to explicitly allow changes only to the meta repository.

minwoox:
Thanks for reviewing. 😉

PR #907, opened 2023-12-26
Enable admin access to the internal `dogma` project.

minwoox:
Motivation:
@trustin suggested utilizing the internal `dogma` project for the control plane. However, it's not easy to use the project because it isn't accessible through UI even for an admin.
The `ProjectManager`, which is used in the `ProjectServiceV1`, is wrapped with the `SafeProjectManager` that blocks to access the `dogma` project. (A Plugin uses the `ProjectManager` directly so no problem from the plugin side.)

Modifications:
- Grant admins to access to the internal `dogma` project via REST API and web UI.
- Rename `SafeProjectManager` to `ProjectApiManager`.
  - It's also used for Thrift service.

Result:
- Administrators can now access the internal `dogma` project through both REST API and the web UI.


server/src/main/java/com/linecorp/centraldogma/server/internal/storage/project/SafeProjectManager.java
```diff
@@ -49,19 +51,19 @@ public void close(Supplier<CentralDogmaException> failureCauseSupplier) {
...
     public boolean exists(String name) {
-        validateProjectName(name);
+        validateProjectName(name, true);
         return delegate().exists(name);
     }
 
     @Override
     public Project get(String name) {
-        validateProjectName(name);
+        validateProjectName(name, true);
```
jrhee17, line 60: In short, if a plugin (i.e. ControlPlanePlugin) wants to modify an internal `dogma` project, do we need to set a `ServiceRequestContext` before calling methods?

I'm wondering if it's easier to...
- Just validate the (dogma && admin) condition at the API level (we already seem to be doing this via. `RequiresRole`, `RequiresAdministrator`, `RequiresPermission` decorators)
- Hard code disallowing deleting/purging the internal `dogma` project at this layer

jrhee17 (commented):
Looks good overall. Left a minor comment 🙇

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/project/SafeProjectManager.java
minwoox: > In short, if a plugin (i.e. ControlPlanePlugin) wants to modify an internal dogma project, do we need to set a ServiceRequestContext before calling methods?

Not exactly. 😄 The plugin will use `ProjectManager` instead of this `SafeProjectManager` so a `RequestContext` isn't needed.
This change is for REST API which is used from UI. With this feature, we can easily check the repository through UI if the configuration is stored as expected. Let me add this to the commit description. 

 > we already seem to be doing this via. RequiresRole, RequiresAdministrator, RequiresPermission decorators
 
 We can do that but I'm not sure if it makes it simpler. For list() method, it includes the dogma project if it's called by an admin:
 ```java
public Map<String, Project> list() {
    final Map<String, Project> list = delegate().list();
    final Map<String, Project> ret = new LinkedHashMap<>(list.size());
    for (Map.Entry<String, Project> entry : list.entrySet()) {
        if (isValidProjectName(entry.getValue().name(), true)) {
            ret.put(entry.getKey(), entry.getValue());
        }
    }
    return Collections.unmodifiableMap(ret);
}
 ```
Doing this with the decorators, isn't simple I think. 🤔

jrhee17: > The plugin will use ProjectManager instead of this SafeProjectManager so a RequestContext isn't needed.

I see, I didn't know that 👍 

> We can do that but I'm not sure if it makes it simpler.

I see, the concern I had with the current approach is:
- Admin permission is checked at multiple places, making it complex and difficult to maintain
- We need to remember to add `validateProjectName` for every API we add in the future

Having said this, if you believe the current approach is most efficient I trust your judgement

jrhee17 (approved):
👍 👍 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/project/SafeProjectManager.java
minwoox: Ah ha, I get your point. Let me encapsulate it by putting the logic that validates the dogma project in the `SafeProjectManager`.

minwoox: I've renamed `SafeProjectManager` to `ProjectApiManager` and gathered the validating logic in it. It's now also used in the Thrift service as well. PTAL. 🙇

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/project/ProjectApiManager.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    public static ImmutableMap<String, Project> listProjectsWithoutDogma(Map<String, Project> projects) {
+        final Map<String, Project> result = new LinkedHashMap<>(projects.size() - 1);
+        for (Map.Entry<String, Project> entry : projects.entrySet()) {
+            if (!INTERNAL_PROJECT_DOGMA.equals(entry.getKey())) {
+                result.put(entry.getKey(), entry.getValue());
+            }
+        }
+        return ImmutableMap.copyOf(result);
```
ikhoon, line 80: nit: To avoid additional copies. 
```suggestion
        return Collections.unmodifiableMap(result);
```

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/project/ProjectApiManager.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    public CompletableFuture<Void> createProject(String projectName, Author author) {
+        checkInternalDogmaProject(projectName, "create");
+        return commandExecutor.execute(Command.createProject(author, projectName));
+    }
+
+    private static void checkInternalDogmaProject(String projectName, String methodName) {
+        if (INTERNAL_PROJECT_DOGMA.equals(projectName)) {
+            throw new IllegalArgumentException("Cannot " + methodName + ' ' + projectName);
```
ikhoon, line 94: For readability. 
```suggestion
    private static void checkInternalDogmaProject(String projectName, String operation) {
        if (INTERNAL_PROJECT_DOGMA.equals(projectName)) {
            throw new IllegalArgumentException("Cannot " + operation + ' ' + projectName);
```

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/project/ProjectApiManager.java
```diff
@@ -0,0 +1,128 @@
...
+                                      Command.removeProject(author, projectName)));
+    }
+
+    public CompletableFuture<Void> purgeProject(String projectName, Author author) {
+        checkInternalDogmaProject(projectName, "purge");
+        return commandExecutor.execute(Command.purgeProject(author, projectName));
+    }
+
+    public CompletableFuture<Revision> unremoveProject(String projectName, Author author) {
+        checkInternalDogmaProject(projectName, "patch");
```
ikhoon, line 116: Is `patch` used intentionally? Wrote a comment in case of typo.
```suggestion
        checkInternalDogmaProject(projectName, "unremove");
```

ikhoon (approved):
Looks good! 👍

minwoox:
Thanks! All addressed. 😉

ikhoon:
Would you mind resolving the conflicts?

minwoox:
> Would you mind resolving the conflicts?

Did it. 😆

PR #909, opened 2023-12-27
Fix to show project creators correctly

### LINKED ISSUE #908: Incorrect Project Creator
The project creators are linked to the creator of the `dogma` repository, which is always the `system`.
https://github.com/line/centraldogma/blob/0fd267dc34209dd60c4ed924bebb047d02550f9a/server/src/main/java/com/linecorp/centraldogma/server/storage/project/Project.java#L62
https://github.com/line/centraldogma/blob/0cf273402d70773e0aab1e2b858111a1c1c0f22c/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/project/DefaultProject.java#L131

minwoox:
Motivation:
The current creator of a project is the creator of the internal `dogma` repository, consistently set as `System`. We should show the correct creator which is stored in the `metadata.json` file.

A noteworthy observation is that only the email of the creator is stored in `metadata.json` instead of the complete `Author` information. A temporary solution involves restoring the `Author` using the email, with plans to revisit this approach if issues arise.

Modifications:
- Fix to show project creators correctly using information stored in the `metadata.json` file.

Result:
- Close #908 
- The project creator is now correctly retrieved based on the information stored in the `metadata.json` file.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/project/DefaultProject.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/project/DefaultProject.java
@@ -29,14 +29,19 @@ import javax.annotation.Nullable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.fasterxml.jackson.core.JsonParseException;
+import com.fasterxml.jackson.databind.JsonMappingException;
+import com.fasterxml.jackson.databind.JsonNode;
 import com.google.common.collect.ImmutableMap;
 
 import com.linecorp.centraldogma.common.Author;
 import com.linecorp.centraldogma.common.CentralDogmaException;
 import com.linecorp.centraldogma.common.Change;
+import com.linecorp.centraldogma.common.Entry;
 import com.linecorp.centraldogma.common.Markup;
 import com.linecorp.centraldogma.common.ProjectExistsException;
 import com.linecorp.centraldogma.common.ProjectNotFoundException;
+import com.linecorp.centraldogma.common.Query;
 import com.linecorp.centraldogma.common.RepositoryExistsException;
 import com.linecorp.centraldogma.common.Revision;
 import com.linecorp.centraldogma.internal.Jackson;
@@ -61,6 +66,8 @@ public class DefaultProject implements Project {
     private static final Logger logger = LoggerFactory.getLogger(DefaultProject.class);
 
     private final String name;
+    private final long creationTimeMillis;
+    private final Author author;
     final RepositoryManager repos;
     private final AtomicReference<MetaRepository> metaRepo = new AtomicReference<>();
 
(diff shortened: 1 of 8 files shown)


server/src/main/java/com/linecorp/centraldogma/server/metadata/UserAndTimestamp.java
```diff
@@ -89,6 +89,13 @@ public String timestamp() {
         return timestampAsText;
     }
 
+    /**
+     * Returns the epoch milliseconds of the {@link #timestamp()}.
+     */
+    public long epochMillis() {
```
jrhee17, line 95: nit; `timestampMillis` may be a better name since it represents both the field and how it is represented

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/project/DefaultProject.java
```diff
@@ -82,6 +89,14 @@ public class DefaultProject implements Project {
         boolean success = false;
         try {
             createReservedRepos(System.currentTimeMillis());
+            final UserAndTimestamp creation = metadataCreation();
+            if (creation != null) {
```
jrhee17, line 93: nit; personally I prefer that the branching is done on `REPO_DOGMA` for clarity, but I guess functionally the same is achieved in this block. No problem to leave as-is.

```
    if (REPO_DOGMA.equals(name)) {
        creationTimeMillis = repos.get(REPO_DOGMA).creationTimeMillis();
        author = repos.get(REPO_DOGMA).author();
    } else {
        final UserAndTimestamp creation = metadataCreation(); // fallbacks to Author.SYSTEM
        creationTimeMillis = creation.epochMillis();
        author = Author.ofEmail(creation.user());
    }
```

jrhee17 (approved):
I understood this PR doesn't have a real user impact (I couldn't find any notion of project author from the UI). Looks good to me 👍 🙇 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/project/DefaultProject.java
minwoox: I've used the current logic to handle the case when creating metadata fails. (I'm not sure if there's a chance for that event to occur though.)
https://github.com/line/centraldogma/blob/81958a67ac630ee5b306a0229f5b73ce27e6e5d6/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/project/DefaultProject.java#L149-L150
If it fallbacks to Author.SYSTEM in that case, it has to create `Author` again event if it can get it from `repos.get(REPO_DOGMA).author()`. So let me leave it as is. 😉

minwoox:
> I understood this PR doesn't have a real user impact (I couldn't find any notion of project author from the UI).

That is correct for now. 👍

ikhoon (approved):
Looks good!

PR #910, opened 2024-01-10
Configure JacksonRequestConverterFunction with the default ObjectMapper without any additional configuration

minwoox:
Motivation:
The server currently utilizes `DefaultJacksonObjectMapperProvider` to create an `ObjectMapper`. The requirement is for the `ObjectMapper` to allow null values during JSON conversion. However, the presence of jackson-module-scala in the classpath configures the `ObjectMapper` to disallow null values.

Modification:
- Configure `JacksonRequestConverterFunction` with the default `ObjectMapper` without any additional configuration.

Result:
- The server now correctly converts JSON requests to instances.


jrhee17 (approved):
Looks straightforward 👍  Thanks @minwoox ! 🙇 👍 🙇

server/src/main/java/com/linecorp/centraldogma/server/internal/ArmeriaJacksonObjectMapperProvider.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+package com.linecorp.centraldogma.server.internal;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import com.linecorp.armeria.common.JacksonObjectMapperProvider;
+
+public final class ArmeriaJacksonObjectMapperProvider implements JacksonObjectMapperProvider {
+
+    @Override
```
ikhoon, line 24: I don't see that a custom `JacksonObjectMapperProvider` with SPI is a good idea for a module that can be used as a library. Unfortunately, the scope of `ArmeriaJacksonObjectMapperProvider` is not limited to Central Dogma. `ArmeriaJacksonObjectMapperProvider` may override user's `JacksonObjectMapperProvider`.
What do you think of creating a custom `Jackson{Request,Response}ConverterFunction` instead? It needs more changes though.

minwoox: I thought it was okay because this class is in the server module and users do not have the server module in their environment.
However, they can have a problem because this can be used in the test. 😓  Let me fix that. Thanks!

ikhoon (approved):
Thanks!

minwoox:
Thanks for reviewing. 😉

PR #921, opened 2024-02-21
Update the whole cluster status via REST API

### LINKED ISSUE #912: Use `UpdateServerStatusCommand` for updating writable status in `AdministrativeSevice`
When `writable` status is updated via `AdministrativeSevice.updateStatus()`, the status isn't propagated to other cluster members.
All servers will have the same `writable` status by issuing `UpdateServerStatusCommand` in `AdministrativeSevice`.

ikhoon:
Motivation:

When the writable status is updated via `AdministrativeSevice.updateStatus()`, the status isn't propagated to other cluster members. As a result, they will have inconsistent status. #912

Additionally, the status isn't persistent. If one replica is restarted, the status will be reset. That means the Central Dogma server can't be started with read-only mode. It makes it difficult to serve data in read-only mode after restart if there are replication problems.

This feature also helps to scale out the Central Dogma server by adding read-only nodes when the server receives heavy read queries. If more features are implemented, new data can be replicated while serving in read-only mode.

Modifications:

- Replication is now controlled by `UpdateServerStatusCommand`.
- `<data-dir>/server-status.properties` file will be created to store a server status when it is updated via `UpdateServerStatusCommand`.
  - The file is used to determine whether the server is writable and replicating when the server is started.
  - As the server status is persistent, we should use REST API or the file to restore the status. REST API is preferred, however, if replication is disabled, it can't be recovered via REST API. There is no way to propagate new status because ZooKeeper servers are also stopped. In this case, 1) the status file should be updated manually before restarting servers. Otherwise, 2) each replica's status can be updated by specifying `scope=LOCAL` to the status API.
- `ServerStatusManager` is added to atomically update and read the status file.
- `CommandExecutorStatusManager` is a helper class to handle a start request and a stop request.
  - A start request being processed is canceled if a stop request is received and vice versa.  
- `AdministrativeService` uses `UpdateServerStatusCommand` to propagate the received server status to other replicas.
- Add `ReadOnlyException` that is raised when data is pushed in read-only mode.
- Testing) `CentralDogmaReplicationExtension` is added to `testing-internal` for easier testing of Central Dogma features in replication mode.

Result:

- You can update the status of all servers equally using the Administrative REST API.
- You can now start Central Dogma servers in read-only mode.
- Fixes #912


server/src/main/java/com/linecorp/centraldogma/server/management/ServerStatus.java
```diff
@@ -0,0 +1,101 @@
...
+
+    /**
+     * Creates a new instance with the specified properties.
+     */
+    @JsonCreator
+    public ServerStatus(@JsonProperty("writable") @Nullable Boolean writable,
+                        @JsonProperty("replicating") @Nullable Boolean replicating) {
+        writable = firstNonNull(writable, true);
+        replicating = firstNonNull(replicating, true);
+        if (writable && !replicating) {
```
minwoox, line 49: I often find it confusing to determine which combination of writable and replicable is possible. What do you think about adding an enum instead?
This would eliminate confusion for both us and users, ensuring that incorrect combinations are not input.

```java
enum ServerStatus {
  READABLE(false, false),
  REPLICABLE(false, true),
  WRITABLE(true, true);
}
```

ikhoon: Good idea. I prefer renaming some of them for clarity. How about?
```java
enum ServerStatus {
  READONLY(false, false),
  REPLICATION_ONLY(false, true),
  WRITABLE(true, true);
}
```

minwoox: That looks perfect. 💯 💯 💯

server/src/main/java/com/linecorp/centraldogma/server/command/CommandExecutorStatusManager.java
```diff
@@ -0,0 +1,235 @@
...
+                    logger.debug("Cancelled the previous replication stop request");
+                    break;
+                case NONE:
+                    if (replicating()) {
+                        return;
+                    }
+                    replicatingRequest = ReplicatingRequest.START;
+                    break;
+            }
+            startFuture = executor.start();
```
jrhee17, line 158: Question) What do you think of simply waiting for start/stops inside the `synchronized` block?

```suggestion
            startFuture = executor.start().get();
```

The reasoning is this I think `executor.start()` and `executor.stop()` can be executed concurrently with the current logic.
Looking through `ZooKeeperCommandExecutor` it seems like there are multiple local states, and I'm not sure how thread safe the class is.

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -431,23 +435,29 @@ private CommandExecutor startCommandExecutor(
...
+                executor = new StandaloneCommandExecutor(pm, repositoryWorker, statusManager, sessionManager,
                                                          cfg.writeQuotaPerRepository(),
                                                          onTakeLeadership, onReleaseLeadership);
                 break;
             default:
                 throw new Error("unknown replication method: " + replicationMethod);
         }
 
+        final ServerStatus initialServerStatus = statusManager.serverStatus();
+        executor.setWritable(initialServerStatus.writable());
```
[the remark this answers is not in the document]
ikhoon: `ProjectInitializer.initializeInternalProject()` is added after this method returns.
https://github.com/line/centraldogma/blob/48586150dbf1370e3739b7fda7ea00cef47aa092/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java#L378-L381
If `writable` is `false`, the internal repositories are not created. It is an undesirable status. 

However, this is an administrative operation to make a repository read-only that new servers do not need because they are empty. So it would be safe that all new servers must have `writable == true`.

server/src/main/java/com/linecorp/centraldogma/server/command/AbstractCommandExecutor.java
```diff
@@ -111,7 +114,7 @@ public final <T> CompletableFuture<T> execute(Command<T> command) {
             // Reject all commands except for AdministrativeCommand when the replica is in read-only mode.
             // AdministrativeCommand is allowed because it is used to change the read-only mode or migrate
             // metadata under maintenance mode.
-            throw new IllegalStateException("running in read-only mode. command: " + command);
+            throw new ReadOnlyException("running in read-only mode. command: " + command);
```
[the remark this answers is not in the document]
ikhoon: Good point. It is a bug. 
- `isStarted() == false`: reject all commands even for `AdministrativeCommand`
- `isStarted() == true && writable == false`: reject non-administrave commands

server/src/main/java/com/linecorp/centraldogma/server/command/CommandExecutorStatusManager.java
ikhoon: That is a point of consideration. Initially, as you said, I tried to execute `start` and stop atomically and I didn't want to add this complex state machine.

Migrating the code, I found the following issue that we should call `execute.stop()` to cancel the start process. https://github.com/line/centraldogma/pull/301
We allow the behavior only when a server is being started. Let me try to add a tweak for that.

server/src/main/java/com/linecorp/centraldogma/server/command/StandaloneCommandExecutor.java
```diff
@@ -361,7 +369,10 @@ private CompletableFuture<Void> removeSession(RemoveSessionCommand c) {
     }
 
     private CompletableFuture<Void> updateServerStatus(UpdateServerStatusCommand c) {
-        setWritable(c.writable());
-        return CompletableFuture.completedFuture(null);
+        return CompletableFuture.supplyAsync(() -> {
+            serverStatusManager.updateStatus(c.writable(), c.replicating());
+            statusManager().updateStatus(c);
```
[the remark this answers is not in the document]
ikhoon: I added a sequential executor to guarantee the order of executions.

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -742,6 +748,9 @@ private synchronized void replayLogs(long targetRevision) {
 
                 updateLastReplayedRevision(nextRevision);
                 info.lastReplayedRevision = nextRevision;
+                if (command instanceof UpdateServerStatusCommand) {
```
minwoox, line 751: this command is executed at line 734. Should we execute it twice?

minwoox (commented):
Basically looks great! 👍 👍 👍 
Left some suggestions. 😉

server/src/main/java/com/linecorp/centraldogma/server/command/UpdateServerStatusCommand.java
```diff
@@ -29,26 +31,45 @@
...
+     * Creates a new instance with the specified properties.
+     */
     @JsonCreator
-    UpdateServerStatusCommand(@JsonProperty("timestamp") @Nullable Long timestamp,
-                              @JsonProperty("author") @Nullable Author author,
-                              @JsonProperty("writable") boolean writable) {
+    public UpdateServerStatusCommand(@JsonProperty("timestamp") @Nullable Long timestamp,
+                                     @JsonProperty("author") @Nullable Author author,
+                                     @JsonProperty("writable") @Nullable Boolean writable,
+                                     @JsonProperty("replicating") @Nullable Boolean replicating) {
```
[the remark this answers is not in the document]
ikhoon: +1 Originally, I designed to selectively update one of them. However, I agree that we don't need such flexibility.

server/src/main/java/com/linecorp/centraldogma/server/command/AbstractCommandExecutor.java
```diff
@@ -61,6 +63,7 @@ protected AbstractCommandExecutor(@Nullable Consumer<CommandExecutor> onTakeLead
                                       @Nullable Consumer<CommandExecutor> onReleaseLeadership) {
         this.onTakeLeadership = onTakeLeadership;
         this.onReleaseLeadership = onReleaseLeadership;
+        statusManager = new CommandExecutorStatusManager(this);
```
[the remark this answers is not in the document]
ikhoon: I also considered that, but the actual actions of start and stop occur in the `CommandExecutor`s, and even if `CommandExecutorStatusManager` has the fields, `CommandExecutorStatusManager` should update them indirectly through the `CommandExecutor`.

So I prefer making the role of `CommandExecutorStatusManager` help trigger the actions.

server/src/main/java/com/linecorp/centraldogma/server/command/UpdateServerStatusCommand.java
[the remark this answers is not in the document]
ikhoon: > How about also making writable and replicating not nullable?

I remembered why it was treated as nullable.

Let's say both `replicating` and `writable` are false due to a failure. We may want to restart the cluster with `REPLICATING_ONLY` mode. If w set `writable` to false via `/status` API. The desired status will be written to `server-status.properties`. 
If we don't allow nullable values, the current status of `replicating` will be stored and used when restarting.

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
ikhoon: The command at 734L updates `StandaloneCommandExecutor` but this line updates `ZookeeperCommandExecutor`. We need to stop both executors because a command is executed by `StandaloneCommandExecutor` first and written to zk logs.
https://github.com/line/centraldogma/blob/27718d65b6555240d752c5e6909b43fd276b7f1c/server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java#L1126

server/src/main/java/com/linecorp/centraldogma/server/management/ServerStatus.java
```diff
@@ -0,0 +1,101 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.base.MoreObjects;
+
+/**
+ * The status of the server.
+ */
+@JsonInclude(Include.NON_NULL)
+public final class ServerStatus {
+    private final boolean writable;
+    private final boolean replicating;
```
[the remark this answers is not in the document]
ikhoon: Good idea. It should be an optional value that should be configured by administrators.

If an instant enters read-only mode, others tend to enter together. Some people may want to use CD in read-only mode while they are recovering the broken data or status.

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -718,12 +720,16 @@ private synchronized void replayLogs(long targetRevision) {
...
             ReplicationLog<?> l = null;
             try {
                 final Optional<ReplicationLog<?>> log = loadLog(nextRevision, true);
+                Command<?> command = null;
                 if (log.isPresent()) {
                     l = log.get();
-                    final Command<?> command = l.command();
+                    command = l.command();
                     final Object expectedResult = l.result();
                     final Object actualResult = delegate.execute(command).get();
```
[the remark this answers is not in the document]
minwoox: Oops, I missed that. Thanks for the explanation.

server/src/main/java/com/linecorp/centraldogma/server/command/CommandExecutorStatusManager.java
```diff
@@ -0,0 +1,141 @@
...
+    /**
+     * Updates the status of the executor with the specified {@link UpdateServerStatusCommand}.
+     *
+     * <p>This method could take a long time if the executor is not in the desired state yet.
+     * So it should be not called from an event loop thread.
+     */
+    public synchronized void updateStatus(UpdateServerStatusCommand command) {
+        final ServerStatus serverStatus = command.serverStatus();
+        setWritable(serverStatus.writable());
+        setReplicating(serverStatus.replicating());
```
minwoox, line 73: How about setting replicating first if it changes from false to true?
```suggestion
        final ServerStatus newserverStatus = command.serverStatus();
        if (!replicating() && newserverStatus.replicating()) {
            setReplicating(newserverStatus.replicating());
            setWritable(newserverStatus.writable());
        } else {
            setWritable(newserverStatus.writable());
            setReplicating(newserverStatus.replicating());
        }
```

server/src/main/java/com/linecorp/centraldogma/server/internal/api/AdministrativeService.java
```diff
@@ -53,111 +51,47 @@ public AdministrativeService(CommandExecutor executor) {
...
+     * <p>If {@link UpdateServerStatusRequest#scope()} is omitted, defaults to {@link Scope#ALL}.
+     * If the scope is {@link Scope#ALL}, the new status is propagated to all cluster servers.
+     * If the scope is {@link Scope#LOCAL}, the new status is only applied to the current server.
      */
-    @Patch("/status")
-    @Consumes("application/json-patch+json")
+    @Put("/status")
+    @Consumes("application/json")
     @RequiresAdministrator
     public CompletableFuture<ServerStatus> updateStatus(ServiceRequestContext ctx,
```
minwoox, line 69: nit: ctx isn't used.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/AdministrativeService.java
```diff
@@ -53,111 +51,47 @@ public AdministrativeService(CommandExecutor executor) {
...
-            }
-            return executor().stop().handle((unused, cause) -> {
-                if (cause != null) {
-                    logger.warn("Failed to stop the command executor:", cause);
-                } else {
-                    logger.info("Disabled replication");
-                }
+            return CompletableFuture.supplyAsync(() -> {
+                executor().statusManager().setWritable(newStatus.writable());
+                executor().statusManager().setReplicating(newStatus.replicating());
```
minwoox, line 84: How about adding an API that uses serverStatus?
```java
executor().statusManager().updateStatus(newStatus);
```

server/src/main/java/com/linecorp/centraldogma/server/internal/api/AdministrativeService.java
```diff
@@ -53,111 +51,47 @@ public AdministrativeService(CommandExecutor executor) {
...
+                statusManager.updateStatus(newStatus);
```
minwoox, line 85: In administrativeService we do:
```java
serverStatusManager.updateStatus(c.serverStatus());
statusManager().updateStatus(c);
```
However, we update `CommandExecutorStatusManager` first. Could you apply them in the same order?

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -772,6 +782,15 @@ private synchronized void replayLogs(long targetRevision) {
         }
     }
 
+    private void updateZkCommandStatusLater(UpdateServerStatusCommand command) {
+        canReplicate = command.serverStatus().replicating();
+        // Use a separate executor since executorStatusManager.updateStatus() may stop the executor that calls
+        // this method.
+        ForkJoinPool.commonPool().execute(() -> {
```
minwoox, line 789: Can't we do this only when `if (!canReplicate) {...}`?

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -1070,6 +1089,11 @@ private static String pathFromRevision(long revision) {
     @Override
     protected <T> CompletableFuture<T> doExecute(Command<T> command) throws Exception {
         final CompletableFuture<T> future = new CompletableFuture<>();
+        ExecutorService executor = this.executor;
+        if (command instanceof UpdateServerStatusCommand) {
+            // Use a separate executor because `this.executor()` could be stopped while executing the command.
+            executor = ForkJoinPool.commonPool();
```
minwoox, line 1095: ditto. only when the replicating is false?

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -1070,6 +1089,11 @@ private static String pathFromRevision(long revision) {
     @Override
     protected <T> CompletableFuture<T> doExecute(Command<T> command) throws Exception {
         final CompletableFuture<T> future = new CompletableFuture<>();
+        ExecutorService executor = this.executor;
+        if (command instanceof UpdateServerStatusCommand) {
```
minwoox, line 1093: `command.type() == CommandType.UPDATE_SERVER_STATUS`

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -742,6 +749,9 @@ private synchronized void replayLogs(long targetRevision) {
 
                 updateLastReplayedRevision(nextRevision);
                 info.lastReplayedRevision = nextRevision;
+                if (command instanceof UpdateServerStatusCommand) {
```
minwoox, line 752: `command.type() == CommandType.UPDATE_SERVER_STATUS`

minwoox (commented):
Left a few minor comments.
It looks really great! 👍 👍 👍

server/src/main/java/com/linecorp/centraldogma/server/command/CommandExecutorStatusManager.java
ikhoon: It makes sense. Omitting `!replicating()` condition seems to have the same effect though.

server/src/main/java/com/linecorp/centraldogma/server/command/UpdateServerStatusCommand.java
```diff
@@ -21,32 +21,39 @@
...
+    /**
+     * Creates a new instance with the specified properties.
+     */
     @JsonCreator
-    UpdateServerStatusCommand(@JsonProperty("timestamp") @Nullable Long timestamp,
-                              @JsonProperty("author") @Nullable Author author,
-                              @JsonProperty("writable") boolean writable) {
+    public UpdateServerStatusCommand(@JsonProperty("timestamp") @Nullable Long timestamp,
+                                     @JsonProperty("author") @Nullable Author author,
+                                     @JsonProperty("serverStatus") ServerStatus serverStatus) {
```
trustin, line 46: Is there any backward compatibility / gradual rollout concern?

testing-internal/src/main/java/com/linecorp/centraldogma/testing/internal/CentralDogmaReplicationExtension.java
```diff
@@ -0,0 +1,218 @@
...
+import com.linecorp.centraldogma.testing.internal.auth.TestAuthMessageUtil;
+import com.linecorp.centraldogma.testing.internal.auth.TestAuthProviderFactory;
+import com.linecorp.centraldogma.testing.junit.AbstractAllOrEachExtension;
+
+import io.netty.util.NetUtil;
+
+/**
+ * A JUnit {@link Extension} that starts an embedded Central Dogma cluster based on ZooKeeper.
+ */
+public class CentralDogmaReplicationExtension extends AbstractAllOrEachExtension {
```
trustin, line 63: Can we merge `com.linecorp.centraldogma.server.internal.replication.Cluster` and `Replica` into this, perhaps in a follow-up PR?

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
ikhoon: The `command` may update `writable` status.

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
ikhoon: `instanceof` would be better since:
- `command` is casted to `UpdateServerStatusCommand`
- command can be null. `instanceof` can remove an additional null check.

server/src/main/java/com/linecorp/centraldogma/server/command/UpdateServerStatusCommand.java
ikhoon: The command isn't used yet internally. Are you concerned that some users might have used the command already in a plugin?

testing-internal/src/main/java/com/linecorp/centraldogma/testing/internal/CentralDogmaReplicationExtension.java
ikhoon: Agreed. I think `CentralDogmaReplicationExtension` can incorporate with the features of `Cluster` and `Replica`.

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
ikhoon: Replied at https://github.com/line/centraldogma/pull/921#discussion_r1571865798

server/src/main/java/com/linecorp/centraldogma/server/internal/api/AdministrativeService.java
minwoox: Sorry, my previous comment was unclear. 😓 
We do in `StandaloneCommandExecutor`:
```java
return CompletableFuture.supplyAsync(() -> {
    serverStatusManager.updateStatus(c.serverStatus());
    statusManager().updateStatus(c);
```
However, we do it here:
```java
return CompletableFuture.supplyAsync(() -> {
    executor().statusManager().updateStatus(newStatus);
    serverStatusManager.updateStatus(newStatus);
```
Shouldn't we do the same order?

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
minwoox: > The command may update writable status.

Isn't is ok to use the Zookeeper executor because it only sets the `writable` value to `true`? (i.e. it doesn't do anything with the Zookeeper executor.)

ikhoon: I misunderstood your comment before. It is possible to use the Zookeeper executor.

server/src/main/java/com/linecorp/centraldogma/server/command/UpdateServerStatusCommand.java
trustin: Nope. I guess we're the only user at the moment :sweat_smile:

trustin (approved):

minwoox:
https://github.com/line/centraldogma/actions/runs/8780555404/job/24090707203?pr=921#step:6:482
Could you check the test failure? 😉

ikhoon:
Oops... 😅

minwoox (approved):
Looks excellent! Thanks! 👍 👍 👍

minwoox:
Let's ship it 👍 👍 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -718,12 +721,16 @@ private synchronized void replayLogs(long targetRevision) {
 
         long nextRevision = info.lastReplayedRevision + 1;
         for (;;) {
+            if (!canReplicate) {
```
jrhee17, line 724: Question) Assuming a replica has been set with `canReplicate=false` will `canReplicate = true` be set for running servers without restarting? It seems like even if a `UpdateServerStatusCommand` is issued, `replayLogs` will never update this value due to the `break` statement for replicas

Issue #902, opened 2023-12-07
Directory history displays irrelevant changes


minwoox:
I've identified an issue where the history of a directory includes irrelevant changes when the modified files or directories share the same name as the target directory. For instance, the directory `/a` displays changes for `/ab/a.txt`, `/a.json`, and other unrelated items in its history.

<img width="591" alt="image" src="https://github.com/line/centraldogma/assets/25103250/f1339f8b-8a6f-4b6d-841e-16406f7c9ab1">

thachlp:
Hi @minwoox, as I understand, your `a` repository only has `/ab/a.txt` and `/a.json`, but the history includes `/a_b/a.txt`, right?
Maybe I misunderstood because I can not reproduce on my local.

minwoox:
> your a repository 

It's `a` directory and the history should show the history of the `a` directory only.
However, it also displays the history of other files (e.g. `/abc/a.txt`) that are not related to the `a` directory.
Please, let me know if you can't reproduce it. I will probably provide a repro for this. 😉

thachlp:
Thanks for explaining to me.
So the expected is that, even `/ab/a.txt` is a child of `a`, the history of `a` should only include files, not repository, right?

thachlp:
```
/a
├── a.txt
├── a.json
└── /ab
    ├── a.txt
    └── a.json
```
I mean here, the history of `/a` should only contain the history of `/a/a.txt` and `/a/a.json`?

Please help confirm to make sure I understand correctly 😁. If we require that, I will try to fix this.

minwoox:
Sorry about the ambiguity in my comments. 😓 
```
|---- foo (directory)
|     |---- bar.txt
|     |---- baz.json
|---- foobao (directory)
|     |---- bar1.txt
|     |---- baz1.json
```
When two different directories exist at the same level (e.g., `foo` and `foobao`), each directory should display its own history. However, if the name of another directory (`foobao`) starts with the directory name (`foo`), the history of the former directory is included in the latter's history.

thachlp:
Thanks for clarifying, I can reproduce it now 👍
I will take this.

minwoox:
@thachlp hey~ Any progress so far? 😄

thachlp:
Not much, I have backed from 3 weeks vacation. I will continue this on this weekend 🚀🚀

thachlp:
I tested on local with my update:
### AS BE
`/dogma`
**Revision	Summary	Author	Timestamp**
[...]

PR #922, opened 2024-02-24
Fix history displays irrelevant changes

thachlp:
Motivate: 
The history of a directory includes changes in directories share the name. For instance: the history of `/foo` displays changes for `/foo2`, `/foo3`, `/foobar` ...

Modifications:
Add a slash `(/)` before `**` in the path `(path + "/**")`. The pattern `foo/**` matches `/foo/anything`, `/foo/subdir/file`, but it does not match anything that isn't a subpath of `/foo`.

Result
Fixes https://github.com/line/centraldogma/issues/902

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/admin/service/RepositoryService.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/admin/service/RepositoryService.java
@@ -177,10 +177,11 @@ public class RepositoryService extends AbstractService {
                                                        @Param String path,
                                                        @Param @Default("-1") String from,
                                                        @Param @Default("1") String to) {
+        final String pathPattern = "/".equals(path) ? "/**" : path + "/**";
         return projectApiManager.getProject(projectName).repos().get(repoName)
                                 .history(new Revision(from),
                                          new Revision(to),
-                                         path + "**")
+                                        pathPattern)
                                 .thenApply(commits -> commits.stream()
                                                              .map(DtoConverter::convert)
                                                              .collect(toList()));
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/admin/service/RepositoryService.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/admin/service/RepositoryService.java
@@ -181,7 +181,7 @@ public class RepositoryService extends AbstractService {
         return projectApiManager.getProject(projectName).repos().get(repoName)
                                 .history(new Revision(from),
                                          new Revision(to),
-                                        pathPattern)
+                                         pathPattern)
                                 .thenApply(commits -> commits.stream()
                                                              .map(DtoConverter::convert)
                                                              .collect(toList()));
--- a/server/src/test/java/com/linecorp/centraldogma/server/internal/admin/service/RepositoryServiceTest.java
+++ b/server/src/test/java/com/linecorp/centraldogma/server/internal/admin/service/RepositoryServiceTest.java
@@ -72,33 +72,44 @@ class RepositoryServiceTest {
     void getHistory() throws IOException {
         final WebClient client = dogma.httpClient();
         final String repoName = "myRepo";
-        final AggregatedHttpResponse repoResponse = createRepository(client, repoName);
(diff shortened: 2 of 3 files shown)


minwoox (commented):
Thanks! Could you add some tests, please? 😉

minwoox:
Could you also check if `ContentServiceV1` correctly works, please?

thachlp:
> Could you also check if `ContentServiceV1` correctly works, please?

Ok, let me check

thachlp:
> Thanks! Could you add some tests, please? 😉

Sure, let me try

server/src/test/java/com/linecorp/centraldogma/server/internal/admin/service/RepositoryServiceTest.java
```diff
@@ -0,0 +1,36 @@
...
+        }
+    };
+
+    @Test
+    void getUsersInfo() throws IOException {
+        final WebClient client = dogma.httpClient();
+        final AggregatedHttpResponse userInfo = client.get("/api/v0/users/me").aggregate().join();
+        final JsonNode jsonNode = Jackson.readTree(userInfo.contentUtf8());
+        assertThat(jsonNode.get("login").asText()).isEqualTo("admin@localhost.localdomain");
+    }
```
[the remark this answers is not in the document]
minwoox: It's activated only when the `webAppEnabled` is `true`.
https://github.com/line/centraldogma/blob/c7450c50020205ced4926360c7c20021bea77a9a/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java#L782
https://github.com/line/centraldogma/blob/7152873093d64150637789bb2be871c4e458e6b1/server-auth/saml/src/test/java/com/linecorp/centraldogma/server/auth/saml/SamlAuthTest.java#L69

By the way, could you tell me why you need to retrieve the user information?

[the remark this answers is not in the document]
thachlp: Thanks @minwoox, I can call `/api/v0` now.
I just do the simplest test with `api/v0` to explain to you my problem 🤩

server/src/main/java/com/linecorp/centraldogma/server/internal/admin/service/RepositoryService.java
```diff
@@ -177,10 +177,11 @@ public CompletionStage<List<CommitDto>> getHistory(@Param String projectName,
                                                        @Param String path,
                                                        @Param @Default("-1") String from,
                                                        @Param @Default("1") String to) {
+        final String pathPattern = "/".equals(path) ? "/**" : path + "/**";
```
jrhee17, line 180: The other APIs we have (e.g. watch*, diff*) don't add a `slash` at the end.
For consistency, I think it would be better that all APIs that accept a `pathPattern` have the same behavior.
Also, I think it is perfectly valid to watch `/a**` depending on the use-case.

Question) What do you think of fixing this issue from the client side instead?

jrhee17 (commented):
Thanks for the initiative! Left a high-level review before going into details 🙇

server/src/main/java/com/linecorp/centraldogma/server/internal/admin/service/RepositoryService.java
thachlp: Hi @jrhee17, about fixing this issue from the client side, do you mean we receive the list result from the client and filter?
From my side, it is not good to do logic on the client side. 
@minwoox How do you think?

minwoox: Sorry about the late reply. Yeah, I also thought we should fix the UI (which is the client) to call the path correctly when I first created the issue.
We should let users retrieve whatever history they want so I think we should touch it on the server-side.

thachlp: Sound good @minwoox, could you provide me the screenshot to explain expected from client? I am happy to finish this 🚀

PR #923, opened 2024-02-25
Update the default config as document

thachlp:
**Motivation**
When I try to learn the **_centraldogma_** via document, I see that the default config of the source code is missing two properties from the document https://line.github.io/centraldogma/setup-configuration.html
So I create this pr.

### DIFF
--- a/dist/src/conf/dogma.json
+++ b/dist/src/conf/dogma.json
@@ -41,5 +41,6 @@
     "timeWindowSeconds": 1
   },
   "accessLogFormat": "common",
-  "authentication": null
+  "authentication": null,
+  "cors": null
 }
--- a/dist/src/conf/dogma.json
+++ b/dist/src/conf/dogma.json
@@ -28,7 +28,7 @@
   },
   "webAppEnabled": true,
   "webAppTitle": null,
-  "mirroringEnabled": null,
+  "mirroringEnabled": true,
   "numMirroringThreads": null,
   "maxNumFilesPerMirror": null,
   "maxNumBytesPerMirror": null,
@@ -36,6 +36,10 @@
     "method": "NONE"
   },
   "csrfTokenRequiredForThrift": null,
+  "writeQuotaPerRepository": {
+    "requestQuota": 5,
+    "timeWindowSeconds": 1
+  },
   "accessLogFormat": "common",
   "authentication": null
 }
--- a/dist/src/conf/dogma.json
+++ b/dist/src/conf/dogma.json
@@ -35,10 +35,7 @@
   "replication": {
     "method": "NONE"
   },
-  "writeQuotaPerRepository": {
-    "requestQuota": 5,
-    "timeWindowSeconds": 1
-  },
+  "writeQuotaPerRepository": null,
   "accessLogFormat": "common",
   "authentication": null,
   "cors": null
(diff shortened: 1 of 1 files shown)


dist/src/conf/dogma.json
```diff
@@ -28,14 +28,18 @@
...
   "webAppTitle": null,
-  "mirroringEnabled": null,
+  "mirroringEnabled": true,
   "numMirroringThreads": null,
   "maxNumFilesPerMirror": null,
   "maxNumBytesPerMirror": null,
   "replication": {
     "method": "NONE"
   },
   "csrfTokenRequiredForThrift": null,
```
hmhuan, line 38: I see this field `csrfTokenRequiredForThrift ` in default config but not in documentation. Also the default configuration missing `cors` field

thachlp: Thanks. I added the `cors` field.

dist/src/conf/dogma.json
```diff
@@ -28,14 +28,19 @@
   },
   "webAppEnabled": true,
   "webAppTitle": null,
-  "mirroringEnabled": null,
+  "mirroringEnabled": true,
```
minwoox, line 31: Could you please explain the rationale behind this change?

thachlp: I see in the documentation:
image

minwoox: Oops, I didn't notice that. 😓 Thanks!

minwoox (approved):
Great! Thanks a lot, @thachlp!

dist/src/conf/dogma.json
```diff
@@ -28,14 +28,18 @@
...
   "maxNumFilesPerMirror": null,
   "maxNumBytesPerMirror": null,
   "replication": {
     "method": "NONE"
   },
-  "csrfTokenRequiredForThrift": null,
+  "writeQuotaPerRepository": {
+    "requestQuota": 5,
+    "timeWindowSeconds": 1
+  },
```
ikhoon, line 41: I don't think this `writeQuotaPerRepository` is a sensible default.
Instead, what do you think of removing the settings in https://github.com/line/centraldogma/blob/a5eb4129dc13a6001c93ac02f70a621a51aa978a/site/src/sphinx/setup-configuration.rst#L50-L53
; set null?

thachlp: Hi @ikhoon, from my view as a user, I think it is better to keep the documentation because it is easy to understand to set specific quotas. I should update this config to null.
Is it good?

hmhuan (approved):

jrhee17 (approved):
Nice! Thanks for the cleanup @thachlp 🙇 👍 🙇

dist/src/conf/dogma.json
minwoox: @ikhoon Any thought? 🤔

ikhoon: We haven't fully verified the feature yet. Let's keep it null.

minwoox: Then, @thachlp Please set this to null as @ikhoon  suggested. 🙇 
https://github.com/line/centraldogma/blob/a5eb4129dc13a6001c93ac02f70a621a51aa978a/site/src/sphinx/setup-configuration.rst?plain=1#L50-L53

minwoox:
@thachlp 👍 👍 👍

Issue #928, opened 2024-03-12
`500` is returned when `CentralDogma` server is closed


jrhee17:
We have received reports that `500` is returned when `CentralDogma` is closing.
While we can't do much about requests that ignore health checks, we can try to reduce the count of such requests.
One hypothesis I have is that watch requests return a `500` before returning.

----

When `ProjectManager#close` is called, watch requests are shut down with a `ShuttingDownException` at the following location:

https://github.com/line/centraldogma/blob/54205bfcee1167f84bc1ab1a69e08fa0c3e3f2f3/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java#L374

https://github.com/line/centraldogma/blob/54205bfcee1167f84bc1ab1a69e08fa0c3e3f2f3/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java#L935

It might be worth shutting down the server first before `ProjectManager#close` is called so that HTTP2 connections are shut down with a `GOAWAY`, and close the `ProjectManager` afterwards
https://github.com/line/centraldogma/blob/54205bfcee1167f84bc1ab1a69e08fa0c3e3f2f3/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java#L990

jrhee17:
On second thought, returning `ShuttingDownException` probably propagates the server status to clients so it might be better to keep the current shutdown order.

Instead, we could return a 304 `NOT_MODIFIED` instead of a `500` when translating `ShuttingDownException`.

Additionally, we may optionally also want to let users configure a graceful shutdown period after setting the health check to false.

https://github.com/line/centraldogma/blob/54205bfcee1167f84bc1ab1a69e08fa0c3e3f2f3/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java#L328

PR #929, opened 2024-03-12
Gracefully shut down watches when server is shutting down

jrhee17:
**Motivation**

When `CentralDogma` is shutting down, clients watching the server inevitably receive a 500 HTTP response code.

While for most requests I think this behavior makes sense, watch requests by nature are expected to continuously watch the server and are probably not interested in shut down notifications.

To minimize noise while monitoring CentralDogma servers, I propose that we return a 304 status code for watch requests when the server is shutting down

**Modifications**
- Return a 304 for watch requests when `ShuttingDownException` is bubbled up

**Result**
- Central Dogma Server maintainers can monitor with less noise
- Closes https://github.com/line/centraldogma/issues/928

### DIFF
--- a/client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java
+++ b/client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java
@@ -594,7 +594,8 @@ public interface CentralDogma {
      * {@link CompletableFuture} will be completed with {@code null}.
      *
      * @return the latest known {@link Revision} which contains the changes for the matched files.
-     *         {@code null} if the files were not changed for 1 minute since the invocation of this method.
+     *         {@code null} if the files were not changed for 1 minute since the invocation of this method,
+     *         or the server is shut down during the watch.
      *
      * @deprecated Use {@link WatchFilesRequest#start(Revision)} via
      *             {@link CentralDogmaRepository#watch(PathPattern)}.
@@ -615,7 +616,7 @@ public interface CentralDogma {
      *
      * @return the latest known {@link Revision} which contains the changes for the matched files.
      *         {@code null} if the files were not changed for {@code timeoutMillis} milliseconds
-     *         since the invocation of this method.
+     *         since the invocation of this method, or the server is shut down during the watch.
      *
      * @deprecated Use {@link CentralDogmaRepository#watch(PathPattern)} and
      *             {@link WatchFilesRequest#start(Revision)}.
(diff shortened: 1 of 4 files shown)


thachlp (commented):
@jrhee17 I think `503 Service Unavailable` is more meaning in this case.
How do you think?

jrhee17:
> @jrhee17 I think 503 Service Unavailable is more meaning in this case.

One thing that concerns me is that `SERVICE_UNAVAILABLE` is already being returned in cases where a request times out before reaching the storage layer.

ref: https://github.com/line/centraldogma/blob/54205bfcee1167f84bc1ab1a69e08fa0c3e3f2f3/server/src/main/java/com/linecorp/centraldogma/server/internal/api/HttpApiExceptionHandler.java#L132

This means we can't blindly ignore 503 status codes from our monitoring target since it is possible that the server is really unhealthy, not just from shutting down.


My initial thought was health check failures probably indicate the server is going down anyways. If we were to explicitly notify to clients that the server is going down, I think returning a 304 NOT MODIFIED with a header is also an option. I'm yet to find a specification which indicates server health though.

thachlp:
> > @jrhee17 I think 503 Service Unavailable is more meaning in this case.
> 
> One thing that concerns me is that `SERVICE_UNAVAILABLE` is already being returned in cases where a request times out before reaching the storage layer.
> 
> ref:
> 
> https://github.com/line/centraldogma/blob/54205bfcee1167f84bc1ab1a69e08fa0c3e3f2f3/server/src/main/java/com/linecorp/centraldogma/server/internal/api/HttpApiExceptionHandler.java#L132
> 
> This means we can't blindly ignore 503 status codes from our monitoring target since it is possible that the server is really unhealthy, not just from shutting down.
> 
> My initial thought was health check failures probably indicate the server is going down anyways. If we were to explicitly notify to clients that the server is going down, I think returning a 304 NOT MODIFIED with a header is also an option. I'm yet to find a specification which indicates server health though.

Thanks for explaining to me 🙇

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java
```diff
@@ -615,7 +616,7 @@ default CompletableFuture<Revision> watchRepository(String projectName, String r
      *
      * @return the latest known {@link Revision} which contains the changes for the matched files.
      *         {@code null} if the files were not changed for {@code timeoutMillis} milliseconds
-     *         since the invocation of this method.
+     *         since the invocation of this method, or the server is shut down during the watch.
```
ikhoon, line 619: Instead of saying that the server is shutting down, why don't we focus on that a watch can be finished before the timeout? For example:
```
Even before the timeout, the watch may return null earlier due to issues such as server restart.
```

ikhoon (approved):
Thanks for handling this issue. 🙇‍♂️

minwoox (approved):
Great. Thanks!

PR #938, opened 2024-03-18
Wait for quiet period after health check

jrhee17:
**Motivation**

https://github.com/line/centraldogma/issues/928

When `CentralDogma` server is shutting down, we set the health check to false and then initiate the shutdown process.
However, we do this immediately so clients likely don't have time to react to the health check signal.
I propose that we wait for `quietPeriodMillis` after setting the health check for more graceful shutdown behavior.

Note that I've left the `ServerBuilder#gracefulShutdownTimeout` call as-is since if requests are correctly drained `gracefulShutdownTimeout`, won't have an effect.

**Modifications**

- It is possible that a user provides a negative `quietPeriodMillis`, so added a >= 0 validation
- Added a sleep right after setting the health check to false

**Result**
- Graceful shutdown behavior

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
@@ -326,6 +326,16 @@ public class CentralDogma implements AutoCloseable {
      */
     public CompletableFuture<Void> stop() {
         serverHealth.setHealthy(false);
+
+        final Optional<GracefulShutdownTimeout> gracefulTimeoutOpt = cfg.gracefulShutdownTimeout();
+        if (gracefulTimeoutOpt.isPresent()) {
+            try {
+                Thread.sleep(gracefulTimeoutOpt.get().quietPeriodMillis());
+            } catch (InterruptedException e) {
+                logger.debug("Interrupted while waiting for quiet period", e);
+            }
+        }
+
         numPendingStopRequests.incrementAndGet();
         return startStop.stop().thenRun(numPendingStopRequests::decrementAndGet);
     }
--- a/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
@@ -334,6 +334,7 @@ public class CentralDogma implements AutoCloseable {
                 Thread.sleep(gracefulTimeoutOpt.get().quietPeriodMillis());
             } catch (InterruptedException e) {
                 logger.debug("Interrupted while waiting for quiet period", e);
+                Thread.currentThread().interrupt();
             }
         }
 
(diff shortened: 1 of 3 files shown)


server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -326,6 +326,16 @@ public CompletableFuture<Void> start() {
      */
     public CompletableFuture<Void> stop() {
         serverHealth.setHealthy(false);
+
+        final Optional<GracefulShutdownTimeout> gracefulTimeoutOpt = cfg.gracefulShutdownTimeout();
+        if (gracefulTimeoutOpt.isPresent()) {
+            try {
+                Thread.sleep(gracefulTimeoutOpt.get().quietPeriodMillis());
```
minwoox, line 333: Let's add a comment on why we sleep here. 😉

minwoox (approved):
Thanks!

thachlp (approved):
It is an interesting behaviour, could you guide me how to test on UI or with log?

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -326,6 +326,17 @@ public CompletableFuture<Void> start() {
...
     public CompletableFuture<Void> stop() {
         serverHealth.setHealthy(false);
+
+        final Optional<GracefulShutdownTimeout> gracefulTimeoutOpt = cfg.gracefulShutdownTimeout();
+        if (gracefulTimeoutOpt.isPresent()) {
+            try {
+                // Sleep so that clients have some time to redirect traffic according to the health status
+                Thread.sleep(gracefulTimeoutOpt.get().quietPeriodMillis());
+            } catch (InterruptedException e) {
+                logger.debug("Interrupted while waiting for quiet period", e);
```
ikhoon, line 336: Should we restore the interruption status so that the code in the up-stack can see that an interrupt was issued?
```suggestion
                logger.debug("Interrupted while waiting for quiet period", e);
                Thread.currentThread().interrupt();
```

ikhoon (approved):
Looks good!

jrhee17:
> It is an interesting behaviour, could you guide me how to test on UI or with log?

Once the central dogma server shutdown is initiated, the health check service will be switched off (`/monitor/l7check`)
https://github.com/line/centraldogma/blob/2994712c79f041941d5fb5160368823b59832e29/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java#L553-L555

after waiting a certain period, the project manager will be shut down
https://github.com/line/centraldogma/blob/2994712c79f041941d5fb5160368823b59832e29/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java#L935

The main point is that there are no requests that throw a `ShuttingDownException` while the health check endpoint returns healthy.

PR #941, opened 2024-04-10
Retry the ZooKeeper lock operation to avoid entering read-only mode unwantedly

trustin:
Motivation:

Although not frequent, ZooKeeper seems to fail when `ZooKeeperCommandExecutor.safeLock()` attempts to acquire a lock when the system is under load, leading into read-only mode.

Modifications:

- Retry up to 1 minute to acquire a lock.
- Log what command is causing the problem.
- Miscellaneous:
  - Fix the problem where `ZooKeeperCommandExecutorTest.lockTimeout()` doesn't terminate.

Result:

- Central Dogma does not enter read-only mode unexpectedly under load.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
@@ -90,7 +90,6 @@ import com.google.common.escape.Escaper;
 import com.google.common.escape.Escapers;
 import com.google.common.util.concurrent.MoreExecutors;
 
-import com.linecorp.armeria.common.util.Exceptions;
 import com.linecorp.armeria.common.util.SafeCloseable;
 import com.linecorp.centraldogma.common.CentralDogmaException;
 import com.linecorp.centraldogma.common.Revision;
@@ -172,6 +171,9 @@ public final class ZooKeeperCommandExecutor
     private final QuotaConfig writeQuota;
 
     private MetadataService metadataService;
+
+    // Failing to acquire a lock is a critical problem, so we wait as much as we can.
+    private long lockTimeoutMillis = TimeUnit.MINUTES.toMillis(1);
 
     private volatile EmbeddedZooKeeper quorumPeer;
     private volatile CuratorFramework curator;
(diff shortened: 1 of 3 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -173,6 +172,9 @@ public final class ZooKeeperCommandExecutor
 
     private MetadataService metadataService;
 
+    // Failing to acquire a lock is a critical problem, so we wait as much as we can.
+    private long lockTimeoutMillis = TimeUnit.MINUTES.toMillis(1);
```
ikhoon, line 176: Q) Would it be better to increase the default timeout for the server and client?

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -790,32 +792,70 @@ public void childEvent(CuratorFramework unused, PathChildrenCacheEvent event) th
...
+                    Thread.sleep(100);
+                } catch (InterruptedException ignored) {
+                    // Ignore
+                }
+
+                remainingTimeNanos = deadlineNanos - System.nanoTime();
+                if (remainingTimeNanos <= 0) {
+                    // Timed out.
+                    break;
+                }
```
ikhoon, line 831: nit) What do you think of sleeping after checking the remaining timeout?
```suggestion
                remainingTimeNanos = deadlineNanos - System.nanoTime();
                if (remainingTimeNanos <= 0) {
                    // Timed out.
                    break;
                }

                // Sleep a little bit to avoid high CPU usage.
                try {
                    Thread.sleep(100);
                } catch (InterruptedException ignored) {
                    // Ignore
                }
```

ikhoon (approved):
Thanks for the quick fix! 🙇‍♂️🙇‍♂️

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
trustin: If 'the server and client' means Central Dogma API server and client, then I guess not. Failing to acquire a lock in time is a critical error that shouldn't occur frequently and increasing API timeout because this situation is an overkill in my opinion.

jrhee17 (approved):
Understood the changes as: "wait and retry for up to 1 minute if an NPE occurs unexpectly".
I'm slightly concerned that this might create locks that are never deleted, but should be better than just falling into read-only mode 👍

PR #946, opened 2024-04-22
Use LinkedBlockingQueue instead of LinkedTransferQueue for ThreadPoolExecutor

minwoox:
Motivation:
The use of `LinkedTransferQueue` in JDK 17 has a bug causing a thread to consume all of a CPU's usage: https://bugs.openjdk.org/browse/JDK-8301341. Although this bug has been fixed, the patch has not been backported to JDK 17. Since copying the code from JDK 11 poses licensing issues (https://github.com/openhab/openhab-core/pull/3756), we need a different solution.

Switching to `LinkedBlockingQueue` offers a feasible workaround. While `LinkedTransferQueue` is typically efficient under high multi-thread contention due to its use of CAS, our current usage does not exhibit significant contention as the queue tasks' size remains close to 0.

Modifications:
- Replaced the use of `LinkedTransferQueue` with `LinkedBlockingQueue` in `ThreadPoolExecutor`.

Result:
- You no longer see an excessive CPU usage by a single thread.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
@@ -45,7 +45,7 @@ import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ForkJoinPool;
-import java.util.concurrent.LinkedTransferQueue;
+import java.util.concurrent.LinkedBlockingQueue;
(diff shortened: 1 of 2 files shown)


trustin (approved):
:+1::+1::+1:

minwoox:
JCTools also doesn't implement `java.util.concurrent.BlockingQueue`:
https://github.com/JCTools/JCTools/wiki/Getting-Started-With-JCTools#we-have-the-qs

ikhoon (approved):
Thanks, @minwoox!

PR #943, opened 2024-04-11
Add `project` and `repository` tags to `api` metrics

ikhoon:
Motivation:

It is difficult to check how many commits have been pushed to a specific repository without looking at log files.

Modifications:

- Add `project` and `repository` tags to `api` metrics.
  - `projectName` and `repoName` path params are used to extract the values.

Result:

Central Dogma metrics for REST API now provide project and repository information as tags.

PR #950, opened 2024-04-25
Revert "Add `project` and `repository` tags to `api` metrics (#943)"

minwoox:
This reverts commit bacab639dbbe323ece76ce82adb3feb793c6ee1e.

We noticed an increase in CPU and memory usage. Upon capturing a heap dump, we found a large number of Metric-related objects, leading us to suspect that this pull request might be the cause. Let's revert this change to determine if it is indeed responsible.


trustin (approved):

ikhoon (approved):
LGTM!

jrhee17:
I feel like these metrics are relatively basic and probably still useful.

Perhaps another way to handle this could have been to use a `MeterFilter` to filter out projects that are of little interest to us.

ikhoon:
Agreed. The number of repositories created by batch jobs may be large, but the number of commits ​​will be constant and the metrics for them are less important.

PR #952, opened 2024-05-02
Restrict prefix value in `CentralDogmaConfig`

minwoox:
Motivation:
The current implementation of `CentralDogmaConfig` strips the prefix separated by a colon from the configuration value. This causes issues when the configuration value, without a prefix, contains a colon. For instance, the following configuration value:
```
-----BEGIN EC PRIVATE KEY-----
Proc-Type: 4,ENCRYPTED
DEK-Info: ...
...
```
In this case, `CentralDogmaConfig` throws an exception because there is no `ConfigValueConverter` for converting the value after the colon. This can be avoided by restricting the prefix value to only valid characters.

Modifications:
- Restrict the prefix value to the regular expression pattern `^[a-z0-9_]+$`.
- Use the configuration value as is if `CentralDogmaConfig` fails to find a proper `ConfigValueConverter` for the prefix.

Result:
- `CentralDogmaConfig` will no longer fail due to configuration values with colons.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaConfig.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaConfig.java
@@ -87,7 +87,7 @@ public final class CentralDogmaConfig {
 
     private static final Logger logger = LoggerFactory.getLogger(CentralDogmaConfig.class);
 
-    private static final Pattern PREFIX_PATTERN = Pattern.compile("^[a-z0-9_]+$");
+    private static final Pattern PREFIX_PATTERN = Pattern.compile("^[a-z0-9_-]+$");
 
     private static final Map<String, ConfigValueConverter> CONFIG_VALUE_CONVERTERS;
 
--- a/server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaConfig.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaConfig.java
@@ -43,6 +43,7 @@ import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.ServiceLoader;
 import java.util.function.Predicate;
+import java.util.regex.Pattern;
 
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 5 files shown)


trustin (approved):

server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaConfig.java
```diff
@@ -86,6 +87,8 @@ public final class CentralDogmaConfig {
 
     private static final Logger logger = LoggerFactory.getLogger(CentralDogmaConfig.class);
 
+    private static final Pattern PREFIX_PATTERN = Pattern.compile("^[a-z0-9_]+$");
```
jrhee17, line 90: Optional) Maybe just me, but I find it common to also include dashes in property names

```suggestion
    private static final Pattern PREFIX_PATTERN = Pattern.compile("^[a-z0-9_-]+$");
```

server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaConfig.java
```diff
@@ -127,14 +145,20 @@ public static String convertValue(@Nullable String value, String propertyName) {
...
+
         final String rest = value.substring(index + 1);
 
         final ConfigValueConverter converter = CONFIG_VALUE_CONVERTERS.get(prefix);
         if (converter != null) {
             return converter.convert(prefix, rest);
         }
-
-        throw new IllegalArgumentException("failed to convert " + propertyName + ". value: " + value);
+        logger.warn("No {} found for {}. prefix: {}",
```
jrhee17, line 159: I wasn't able to find an explanation for this change (logging instead of throwing).
Do you mind adding a comment so that I know why this change was necessary?

jrhee17 (approved):
👍 👍 👍

server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaConfig.java
minwoox: The private key starts with `-`, so I intentionally didn't put it. But maybe it was too strict. Let me put it back. 😆

ikhoon (approved):
👍👍

PR #953, opened 2024-05-02
Add a metric for tracking the number of pushes

ikhoon:
Motivation:

It isn't easy to check how many commits have been pushed to a specific repository without looking at log files.

This PR is the orignal commit of #943 that tracks the number of pushes instead of adding `project` and `repo` tags to all `api` metrics. #943 has been reverted because it increased in CPU and memory usage.

Modifications:

- Add a `Counter` that is increased when a commit is pushed to a repository.
  - `project` and `repository` are added as tags.

Result:

You can monitor the number of pushes to a repository using the `commits.push` metric.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
@@ -194,7 +194,9 @@ public class ContentServiceV1 extends AbstractService {
             CommitMessageDto commitMessage,
             @RequestConverter(ChangesRequestConverter.class) Iterable<Change<?>> changes) {
         checkPush(repository.name(), changes);
-        meterRegistry.counter("commits.push", "project", repository.parent().name(), "repository", repository.name())
+        meterRegistry.counter("commits.push",
+                              "project", repository.parent().name(),
+                              "repository", repository.name())
                      .increment();
 
         final long commitTimeMillis = System.currentTimeMillis();
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
@@ -90,6 +90,8 @@ import com.linecorp.centraldogma.server.storage.repository.FindOption;
 import com.linecorp.centraldogma.server.storage.repository.FindOptions;
 import com.linecorp.centraldogma.server.storage.repository.Repository;
 
+import io.micrometer.core.instrument.MeterRegistry;
+
 /**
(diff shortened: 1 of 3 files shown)


jrhee17 (approved):
> https://github.com/line/centraldogma/pull/943 has been reverted because it increased in CPU and memory usage.

This is unfortunate as I was actually really looking forward to the metrics in that PR 😅 Maybe next time.

server/src/test/java/com/linecorp/centraldogma/server/internal/api/converter/PushMetricTest.java
```diff
@@ -0,0 +1,62 @@
...
+class PushMetricTest {
+
+    @RegisterExtension
+    static final CentralDogmaExtension dogma = new CentralDogmaExtension();
+
+    @Test
+    void shouldIncreasePushCounter() {
+        final String projectName = "myPro";
+        final String repoName = "myRepo";
+        final MeterRegistry meterRegistry = dogma.dogma().meterRegistry().get();
```
jrhee17, line 41: nit; It's probably worth creating a separate `SimpleMeterRegistry` since the default `MeterRegistry` is static/globally shared.

jrhee17: Nevermind, realized that it is set per-extension

ikhoon:
> > #943 has been reverted because it increased in CPU and memory usage.
> 
> This is unfortunate as I was actually really looking forward to the metrics in that PR 😅 Maybe next time.

Otherwise, we may add a filter to tag repositories selectively. The filter could be used to exclude less interesting repositories generated mechanically by a plugin or bot.
```java
interface RepositoryMetricFilter {
   boolean shouldTagRepository(String projectName, String repoName);
}

new CentralDogmaBuilder()
   .repositoryMetricFilter((project, repo) -> {
      if (project.equals(registryProject) {
         return false;
      } else {
         return true;
      }
   });
```

minwoox (approved):
Thanks!

server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
```diff
@@ -190,6 +194,10 @@ public CompletableFuture<PushResultDto> push(
             CommitMessageDto commitMessage,
             @RequestConverter(ChangesRequestConverter.class) Iterable<Change<?>> changes) {
         checkPush(repository.name(), changes);
+        meterRegistry.counter("commits.push",
```
minwoox, line 197: Note: Didn't add this counter to ThriftService because it's going to be deprecated.

minwoox:
@ikhoon 👍 👍 👍

Issue #543, opened 2020-11-02
Support Git protocols (HTTP, SSH and Git)


minwoox:
https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols
@Eisig said:
Many GitOps tools(e.g. Fluxcd) use Git. If they could use Centraldogma as a config repo, it is easier to setup a ha git cluster.

trustin:
Yeah, would be nice to have this feature!

PR #954, opened 2024-05-08
Support Git clone over HTTP in Central Dogma.

minwoox:
Motivation:
Adding support for Git clone over HTTP in Central Dogma server would enhance its capabilities and make it more user-friendly. This PR partially addressed #543.
Please note that I didn't impelent Git over HTTP fully because:
- We have APIs for commit separately.
- The underlying Git repository could be changed to another format instead of Git.

Modifications:
- Added support for Git clone over HTTP to Central Dogma.

Result:
- You can now clone your repositories via Git clone over HTTP.
  ```
  git clone -c http.extraHeader="Authorization: Bearer your-token" https://your-dogma.com/foo/bar.git
  ```

References:
- https://git-scm.com/docs/protocol-v2/
- https://git-scm.com/docs/protocol-capabilities/
- https://www.git-scm.com/docs/http-protocol
- https://www.git-scm.com/docs/git-upload-pack
- https://git-scm.com/docs/protocol-common
- https://git-scm.com/docs/pack-format/

The CLI command that allows you to view the packets being sent and received:
```
GIT_TRACE=2 GIT_CURL_VERBOSE=2 GIT_TRACE_PERFORMANCE=2 GIT_TRACE_PACK_ACCESS=2 GIT_TRACE_PACKET=2 GIT_TRACE_PACKFILE=2 GIT_TRACE_SETUP=2 GIT_TRACE_SHALLOW=2  git clone --depth=1 https://github.com/line/armeria.git
```

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/api/GitHttpService.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/api/GitHttpService.java
@@ -24,6 +24,7 @@ import static org.eclipse.jgit.transport.GitProtocolConstants.VERSION_2_REQUEST;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 
 import org.eclipse.jgit.lib.Repository;
 import org.eclipse.jgit.transport.UploadPack;
@@ -192,7 +193,7 @@ public final class GitHttpService {
 
         // https://git-scm.com/docs/protocol-common#_pkt_line_format
         void put(String line) {
-            lineLength(sb, line.length() + 5);
+            lineLength(sb, line.getBytes(StandardCharsets.UTF_8).length + 5);
             sb.append(line).append('\n');
         }
 
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/api/GitHttpService.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/api/GitHttpService.java
@@ -120,7 +120,7 @@ public final class GitHttpService {
     }
 
     private static String maybeRemoveGitSuffix(String repoName) {
-        if (repoName.length() > 5 && repoName.endsWith(".git")) {
+        if (repoName.length() >= 5 && repoName.endsWith(".git")) {
             repoName = repoName.substring(0, repoName.length() - 4);
         }
         return repoName;
@@ -211,7 +211,7 @@ public final class GitHttpService {
         }
 
         void flush() {
-            https: //git-scm.com/docs/protocol-v2/2.31.0#_packet_line_framing
+            // https: //git-scm.com/docs/protocol-v2/2.31.0#_packet_line_framing
             sb.append("0000");
         }
 
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/api/GitHttpService.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/api/GitHttpService.java
@@ -72,35 +72,18 @@ public final class GitHttpService {
 
     // https://git-scm.com/docs/protocol-capabilities/
     private static String capabilityAdvertisement() {
-        final StringBuilder sb = new StringBuilder();
-        sb.append("001e# service=git-upload-pack\n");
-        pktFlush(sb);
-        pktLine(sb, "version 2");
-        pktLine(sb, COMMAND_LS_REFS);
+        final PacketLineFraming packetLineFraming = new PacketLineFraming();
+        packetLineFraming.putWithoutPktLine("001e# service=git-upload-pack\n");
+        packetLineFraming.flush();
+        packetLineFraming.put("version 2");
+        packetLineFraming.put(COMMAND_LS_REFS);
         // Support limited options for now due to the unique characteristics of Git repositories in
         // Central Dogma, such as having only a master branch and no tags, among other specifics.
-        pktLine(sb, COMMAND_FETCH + '=' + OPTION_WAIT_FOR_DONE + ' ' + OPTION_SHALLOW);
+        packetLineFraming.put(COMMAND_FETCH + '=' + OPTION_WAIT_FOR_DONE + ' ' + OPTION_SHALLOW);
         // TODO(minwoox): Migrate hash function https://git-scm.com/docs/hash-function-transition
-        pktLine(sb, "object-format=sha1");
-        pktFlush(sb);
-        return sb.toString();
-    }
-
-    // https://git-scm.com/docs/protocol-common#_pkt_line_format
-    static void pktLine(StringBuilder sb, String line) {
-        lineLength(sb, line.getBytes(StandardCharsets.UTF_8).length + 5);
-        sb.append(line).append('\n');
-    }
-
-    static void pktFlush(StringBuilder sb) {
-        https://git-scm.com/docs/protocol-v2/2.31.0#_packet_line_framing
-        sb.append("0000");
-    }
-
-    private static void lineLength(StringBuilder sb, int length) {
-        for (int i = 3; i >= 0; i--) {
-            sb.append(HEX_DIGITS[(length >>> (4 * i)) & 0xf]);
-        }
+        packetLineFraming.put("object-format=sha1");
+        packetLineFraming.flush();
+        return packetLineFraming.toString();
     }
 
     private final ProjectApiManager projectApiManager;
@@ -119,7 +102,7 @@ public final class GitHttpService {
             return HttpResponse.of(HttpStatus.FORBIDDEN, MediaType.PLAIN_TEXT_UTF_8,
                                    "Unsupported service: " + service);
         }
-        repoName = maybeRemoveGitSuffix(repoName);
+
         if (gitProtocol == null || !gitProtocol.contains(VERSION_2_REQUEST)) {
             return HttpResponse.of(HttpStatus.BAD_REQUEST, MediaType.PLAIN_TEXT_UTF_8,
                                    "Unsupported git-protocol: " + gitProtocol);
@@ -130,8 +113,12 @@ public final class GitHttpService {
                                    "Project not found: " + projectName);
         }
         if (!projectApiManager.getProject(projectName).repos().exists(repoName)) {
-            return HttpResponse.of(HttpStatus.NOT_FOUND, MediaType.PLAIN_TEXT_UTF_8,
-                                   "Repository not found: " + repoName);
+            final String maybeGitSuffixRemoved = maybeRemoveGitSuffix(repoName);
+            if (maybeGitSuffixRemoved == repoName || // We can just check the reference.
+                !projectApiManager.getProject(projectName).repos().exists(maybeGitSuffixRemoved)) {
+                return HttpResponse.of(HttpStatus.NOT_FOUND, MediaType.PLAIN_TEXT_UTF_8,
+                                       "Repository not found: " + repoName);
+            }
         }
         return CAPABILITY_ADVERTISEMENT_RESPONSE.toHttpResponse();
     }
@@ -147,7 +134,6 @@ public final class GitHttpService {
     @Post("/{projectName}/{repoName}/git-upload-pack")
     public HttpResponse gitUploadPack(AggregatedHttpRequest req,
                                       @Param String projectName, @Param String repoName) {
-        repoName = maybeRemoveGitSuffix(repoName);
         final String gitProtocol = req.headers().get("git-protocol");
         if (gitProtocol == null || !gitProtocol.contains(VERSION_2_REQUEST)) {
             return HttpResponse.of(HttpStatus.BAD_REQUEST, MediaType.PLAIN_TEXT_UTF_8,
@@ -165,8 +151,13 @@ public final class GitHttpService {
                                    "Project not found: " + projectName);
         }
         if (!projectApiManager.getProject(projectName).repos().exists(repoName)) {
-            return HttpResponse.of(HttpStatus.NOT_FOUND, MediaType.PLAIN_TEXT_UTF_8,
-                                   "Repository not found: " + repoName);
+            final String maybeGitSuffixRemoved = maybeRemoveGitSuffix(repoName);
+            if (maybeGitSuffixRemoved == repoName || // We can just check the reference.
+                !projectApiManager.getProject(projectName).repos().exists(maybeGitSuffixRemoved)) {
+                return HttpResponse.of(HttpStatus.NOT_FOUND, MediaType.PLAIN_TEXT_UTF_8,
+                                       "Repository not found: " + repoName);
+            }
+            repoName = maybeGitSuffixRemoved;
         }
 
         final Repository jGitRepository =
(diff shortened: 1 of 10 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/api/GitHttpService.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    private final ProjectApiManager projectApiManager;
+
+    public GitHttpService(ProjectApiManager projectApiManager) {
+        this.projectApiManager = requireNonNull(projectApiManager, "projectApiManager");
+    }
+
+    // https://www.git-scm.com/docs/gitprotocol-http#_smart_clients
+    @Get("/{projectName}/{repoName}/info/refs")
```
trustin, line 106: What do you think about using the path similar to what a user sees in their browser location field? i.e. `/projects/{projectName}/repos/{repoName}` ?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/GitHttpService.java
```diff
@@ -0,0 +1,187 @@
...
+        final StringBuilder sb = new StringBuilder();
+        sb.append("001e# service=git-upload-pack\n0000");
+        pktLine(sb, "version 2");
+        pktLine(sb, COMMAND_LS_REFS);
+        // Support limited options for now due to the unique characteristics of Git repositories in
+        // Central Dogma, such as having only a master branch and no tags, among other specifics.
+        pktLine(sb, COMMAND_FETCH + '=' + OPTION_WAIT_FOR_DONE + ' ' + OPTION_SHALLOW);
+        // TODO(minwoox): Migrate hash function https://git-scm.com/docs/hash-function-transition
+        pktLine(sb, "object-format=sha1");
+        sb.append("0000"); // flush-pkt https://git-scm.com/docs/protocol-v2/2.31.0#_packet_line_framing
```
trustin, line 83: Maybe extract this into `pktFlush(sb)`?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/GitHttpService.java
```diff
@@ -0,0 +1,187 @@
...
+            ResponseHeaders.builder(200)
+                           .add(HttpHeaderNames.CONTENT_TYPE, "application/x-git-upload-pack-advertisement")
+                           .add(HttpHeaderNames.CACHE_CONTROL, ServerCacheControl.REVALIDATED.asHeaderValue())
+                           .build(),
+            HttpData.ofUtf8(capabilityAdvertisement()));
+
+    // https://git-scm.com/docs/protocol-capabilities/
+    private static String capabilityAdvertisement() {
+        final StringBuilder sb = new StringBuilder();
+        sb.append("001e# service=git-upload-pack\n0000");
```
trustin, line 75: Could be split into:
```
pktLine(sb, "# service=git-upload-pack")
pktFlush(sb)
```

server/src/main/java/com/linecorp/centraldogma/server/internal/api/GitHttpService.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    // https://www.git-scm.com/docs/gitprotocol-http#_smart_clients
+    @Get("/{projectName}/{repoName}/info/refs")
+    public HttpResponse advertiseCapability(@Header("git-protocol") @Nullable String gitProtocol,
+                                            @Param String service,
+                                            @Param String projectName, @Param String repoName) {
+        if (!"git-upload-pack".equals(service)) {
+            // Return 403 https://www.git-scm.com/docs/http-protocol#_smart_server_response
+            return HttpResponse.of(403);
```
trustin, line 112: Maybe add some content? `Unsupported service: %s`

server/src/main/java/com/linecorp/centraldogma/server/internal/api/GitHttpService.java
```diff
@@ -0,0 +1,187 @@
...
+        }
+        repoName = maybeRemoveGitSuffix(repoName);
+        if (gitProtocol == null || !gitProtocol.contains(VERSION_2_REQUEST)) {
+            return HttpResponse.of(HttpStatus.BAD_REQUEST, MediaType.PLAIN_TEXT_UTF_8,
+                                   "Unsupported git-protocol: " + gitProtocol);
+        }
+
+        if (!projectApiManager.getProject(projectName).repos().exists(repoName)) {
+            return HttpResponse.of(HttpStatus.NOT_FOUND);
```
trustin, line 121: Ditto - could we add content to all error responses for easier troubleshooting?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/GitHttpService.java
minwoox: If so, users need to do `git clone https://dogma.com/projects/foo/repos/repo.git` instead of `git clone https://dogma.com/foo/repo.git` because we cannot control the request URI of Git command. Do you still prefer this approach?

minwoox:
> As a follow-up, could you also add an integration test that makes sure if this implementation works with Spring Cloud Config Server?

Thanks! Created. 😉 
https://github.com/line/centraldogma/issues/955

minwoox:
@trustin, thanks! All addressed. 😉

server/src/main/java/com/linecorp/centraldogma/server/internal/api/GitHttpService.java
```diff
@@ -0,0 +1,204 @@
...
+            uploadPack.setTimeout(0); // Disable timeout because Armeria server will handle it.
+            // HTTP does not use bidirectional pipe.
+            uploadPack.setBiDirectionalPipe(false);
+            uploadPack.setExtraParameters(ImmutableList.of(VERSION_2_REQUEST));
+            try {
+                uploadPack.upload(inputStream, os, null);
+            } catch (IOException e) {
+                // Log until https://github.com/line/centraldogma/pull/719 is implemented.
+                logger.debug("Failed to respond git-upload-pack-request: {}", req, e);
+                throw new RuntimeException("failed to respond git-upload-pack-request: " + req, e);
```
jrhee17, line 188: We seem to be using `verboseResponses`. I guess this could include the authorization header in the response which I'm not sure is a good idea.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/project/ProjectApiManager.java
```diff
@@ -124,4 +124,11 @@ public Project getProject(String projectName) {
         }
         return projectManager.get(projectName);
     }
+
+    public boolean exists(String projectName) {
+        if (INTERNAL_PROJECT_DOGMA.equals(projectName) && !isAdmin()) {
```
jrhee17, line 129: 👍

jrhee17 (approved):
Looks good as a starting point to me 👍  Thanks @minwoox 👍 🙇 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/api/GitHttpService.java
minwoox: Modified to use just body. Thanks!

server/src/main/java/com/linecorp/centraldogma/server/internal/api/GitHttpService.java
```diff
@@ -0,0 +1,204 @@
...
+        pktLine(sb, COMMAND_FETCH + '=' + OPTION_WAIT_FOR_DONE + ' ' + OPTION_SHALLOW);
+        // TODO(minwoox): Migrate hash function https://git-scm.com/docs/hash-function-transition
+        pktLine(sb, "object-format=sha1");
+        pktFlush(sb);
+        return sb.toString();
+    }
+
+    // https://git-scm.com/docs/protocol-common#_pkt_line_format
+    static void pktLine(StringBuilder sb, String line) {
+        lineLength(sb, line.length() + 5);
```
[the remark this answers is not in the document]
minwoox: Yeah, we can do that. 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/api/GitHttpService.java
```diff
@@ -0,0 +1,206 @@
...
+        }
+        if (!projectApiManager.getProject(projectName).repos().exists(repoName)) {
+            return HttpResponse.of(HttpStatus.NOT_FOUND, MediaType.PLAIN_TEXT_UTF_8,
+                                   "Repository not found: " + repoName);
+        }
+
+        final Repository jGitRepository =
+                projectApiManager.getProject(projectName).repos().get(repoName).jGitRepository();
+
+        final ByteStreamMessage body = StreamMessage.fromOutputStream(os -> {
```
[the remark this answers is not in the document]
minwoox: You mean just adding `@Blocking` to this method or is there something else?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/GitHttpService.java
```diff
@@ -0,0 +1,206 @@
...
+        }
+        if (!projectApiManager.getProject(projectName).repos().exists(repoName)) {
+            return HttpResponse.of(HttpStatus.NOT_FOUND, MediaType.PLAIN_TEXT_UTF_8,
+                                   "Repository not found: " + repoName);
+        }
+        return CAPABILITY_ADVERTISEMENT_RESPONSE.toHttpResponse();
+    }
+
+    private static String maybeRemoveGitSuffix(String repoName) {
+        if (repoName.length() > 5 && repoName.endsWith(".git")) {
```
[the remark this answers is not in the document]
minwoox: I didn't realize it. Thanks!

server/src/main/java/com/linecorp/centraldogma/server/internal/api/GitHttpService.java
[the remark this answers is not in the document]
ikhoon: I thought of adding `@Blocking` to the service method.

ikhoon (approved):
Fantastic feature! 🤩✨

server/src/main/java/com/linecorp/centraldogma/server/internal/api/GitHttpService.java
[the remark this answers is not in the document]
ikhoon: Had a chat with @minwoox and I found out that he used `StreamMessage.fromOutputStream()` to support backpressure in case the returned git objects are huge.

I agree with his opinion and using `StreamMessage.fromOutputStream()` makes more sense.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/GitHttpService.java
```diff
@@ -0,0 +1,231 @@
...
+            return HttpResponse.of(HttpStatus.BAD_REQUEST, MediaType.PLAIN_TEXT_UTF_8,
+                                   "Unsupported content-type: " + contentType);
+        }
+
+        if (!projectApiManager.exists(projectName)) {
+            return HttpResponse.of(HttpStatus.NOT_FOUND, MediaType.PLAIN_TEXT_UTF_8,
+                                   "Project not found: " + projectName);
+        }
+        if (!projectApiManager.getProject(projectName).repos().exists(repoName)) {
+            final String maybeGitSuffixRemoved = maybeRemoveGitSuffix(repoName);
```
ikhoon, line 154: I don't see permission for `xxx.git` is correctly checked with `@RequiresReadPermission`. If the repo name in Central Dogma is `xxx`, `RequiresPermissionDecorator` may fail to find proper permission for the repo.
https://github.com/line/centraldogma/blob/c376db1eaf2ef155bcdf6cd277116e48323fb719/server/src/main/java/com/linecorp/centraldogma/server/internal/api/auth/RequiresPermissionDecorator.java#L88-L89

minwoox: Discussed this with @ikhoon and we've decided remove ".git" suffix if it's specified when creating a repository.

https://github.com/line/centraldogma/issues/956

server/src/main/java/com/linecorp/centraldogma/server/internal/api/GitHttpService.java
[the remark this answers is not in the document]
minwoox: Let me revert this because we are using it for non-binary data. We need to use the length of the string rather than the length of the bytes. Although we are only using 1-byte characters, the outcomes may not differ; however, I want to avoid confusing others who read this code.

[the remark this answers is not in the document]
minwoox: Here are the references for non-binary data.
https://git-scm.com/docs/protocol-common#_pkt_line_format
https://lore.kernel.org/all/20210428222219.GA982@localhost/T/
```
As a general rule of thumb, if the pkt-line is a pack or object, it's
binary, and otherwise (a ref advertisement, a negotiation, etc.) it's
text.  Almost any case where GIT_TRACE_PACKET prints the output is going
to be text, and almost any case where it does not is going to be binary.
```

server/src/main/java/com/linecorp/centraldogma/server/internal/api/auth/RequiresPermissionDecorator.java
```diff
@@ -80,6 +81,17 @@ private static HttpResponse throwForbiddenResponse(ServiceRequestContext ctx, St
                                          projectName, repoName, adminOrOwner);
     }
 
+    /**
+     * Removes the trailing ".git" suffix from the repository name if it exists. This is added for
+     * GitHttpService. See {@link GitHttpService}.
+     */
+    private static String maybeRemoveGitSuffix(String repoName) {
+        if (repoName.length() > 5 && repoName.endsWith(".git")) {
```
ikhoon, line 89: Ditto.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/GitHttpService.java
```diff
@@ -0,0 +1,224 @@
...
+        void putWithoutPktLine(String line) {
+            sb.append(line);
+        }
+
+        void delim() {
+            sb.append("0001");
+        }
+
+        void flush() {
+            https: //git-scm.com/docs/protocol-v2/2.31.0#_packet_line_framing
```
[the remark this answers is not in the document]
minwoox: Oops. 😓

server/src/main/java/com/linecorp/centraldogma/server/internal/api/GitHttpService.java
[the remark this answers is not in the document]
jrhee17: I see, the reasoning I originally suggested this was:
1) I didn't think any protocols will use utf-16 code units for the length
2) The jGit library seems to use utf-8 as well
Having said this, I think it's fine if <=127 characters are used only.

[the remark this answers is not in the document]
minwoox: Oops, I was totally wrong about this. Let me revert the revert again. 😅

server/src/main/java/com/linecorp/centraldogma/server/internal/api/GitHttpService.java
trustin: I'm not very opinionated on this, so let's go with `/<proj>/<repo>.git` and change if we really have to.

trustin (approved):
Thanks!

PR #965, opened 2024-07-01
Change xDS resource directory structure

minwoox:
Motivation:
To improve the management and authorization of xDS resources in Central Dogma, the directory structure has been modified. Additionally, a `DiffResultType` is added to allow fetching the new content directly instead of a patch when calling `Repository.diff()`.

Modifications:
- Changed the xDS resource directory structure.
  - All resources are now under the `xds` Central Dogma project. (This will be changed to `_xds` in a follow-up PR.)
  - Users can create an xDS project (which is a Central Dogma repository) within the `xds` project.
    - Users can manage their own resources under Central Dogma's authorization system.
- Added `DiffResultType` for fetching the new content instead of a patch when calling `Repository.diff()`.

Result:
- The xDS resource management is now more organized and secure, leveraging Central Dogma's authorization system.


xds/src/main/java/com/linecorp/centraldogma/xds/internal/ControlPlanePlugin.java
```diff
@@ -67,219 +68,236 @@ public final class ControlPlanePlugin extends AllReplicasPlugin {
...
+                            centralDogmaXdsResources.removeListener(repoName, path);
+                        } else if (path.startsWith(ROUTES_DIRECTORY)) {
+                            centralDogmaXdsResources.removeRoute(repoName, path);
+                        }
+                        break;
+                    default:
+                        // Ignore other types of changes.
+                        // No APPLY_JSON_PATCH because the diff option.
+                        // No RENAME because the resource name in the content always have to be
+                        // changed if the file is renamed.
```
jrhee17, line 293: Question) Is it possible to at least log with warning level here since we can't guarantee this in compile time?

server/src/main/java/com/linecorp/centraldogma/server/storage/project/InternalProjectInitializer.java
```diff
@@ -56,12 +57,26 @@ public InternalProjectInitializer(CommandExecutor executor) {
...
+     * Creates the specified internal project and its internal repositories.
+     */
+    public void initialize(String projectName) {
+        requireNonNull(projectName, "projectName");
+        checkArgument(!INTERNAL_PROJECT_DOGMA.equals(projectName),
+                      "Use initialize() to create %s", projectName);
+        initialize(projectName, false);
+    }
+
+    private void initialize(String projectName, boolean createTokenFile) {
```
jrhee17, line 76: Optional) I'm aware this is internal, but instead of a flag that doesn't really have a meaning would just defining a `initializeProject` method make more sense?

```java
    public void initialize() {
        initializeProject(INTERNAL_PROJECT_DOGMA);
        initializeTokens();
    }

    public void initializeProject(String projectName) {
        ...
```

jrhee17 (commented):
Went through the core related changes, will take a look at the xds related changes next week.

server/src/main/java/com/linecorp/centraldogma/server/storage/repository/DiffOption.java
```diff
@@ -0,0 +1,36 @@
...
+ * under the License.
+ */
+package com.linecorp.centraldogma.server.storage.repository;
+
+import com.linecorp.centraldogma.common.Change;
+
+/**
+ * The options for diffing a repository.
+ */
+public enum DiffOption {
```
ikhoon, line 23: Suggestion) `DiffOption` sounds like how to compare two revisions such as `ignore white spaces`. However, this option is used to specify the result style of changed files.
```suggestion
public enum DiffResultType {

   PATCH,
   
   RAW_DATA
```

minwoox: I like the names. 👍

ikhoon: I realized that the usage is different. I'll think about it more.

xds/src/main/java/com/linecorp/centraldogma/xds/internal/ControlPlanePlugin.java
minwoox: That's a good idea. 👍

server/src/main/java/com/linecorp/centraldogma/server/storage/repository/DiffOption.java
ikhoon: I'm okay to leave it as it is because it is only used at the server side which we can easily change.

ikhoon (approved):
Thanks, @minwoox! 🙇‍♂️

server/src/main/java/com/linecorp/centraldogma/server/storage/repository/DiffOption.java
minwoox: Alright, let me use `DiffResultType` though. 😉

xds/src/main/java/com/linecorp/centraldogma/xds/internal/CentralDogmaSnapshotResources.java
```diff
@@ -27,55 +26,62 @@
...
+            value.values().forEach(versionedResource -> {
+                final String resourceName = Resources.getResourceName(versionedResource.resource());
+                versionedResourcesMap.put(resourceName, versionedResource);
+                resourcesMap.put(resourceName, versionedResource.resource());
+            });
         }
-        return new CentralDogmaSnapshotResources<>(versionedResourcesMap.build(), resourcesMap.build(),
-                                                   Integer.toString(revision.major()));
+        return new CentralDogmaSnapshotResources<>(
+                versionedResourcesMap.build(), resourcesMap.build(), resourceType);
```
jrhee17, line 51: Question) I understood this map contains all resources for all resources under the xds project.
I think this approach is fine since there is probably a low possibility of hash collision - however, I'm wondering if it may be simpler + less resources will be used if a `CentralDogmaXdsResources` is created per repository instead.

I just wanted to understand the reasoning behind the current design before moving forward

xds/src/main/java/com/linecorp/centraldogma/xds/internal/CentralDogmaSnapshotResources.java
```diff
@@ -84,14 +90,29 @@ private CentralDogmaSnapshotResources(
...
+        }
+        throw new IllegalArgumentException("Requesting all resource isn't allowed for " + resourceType);
+    }
+
+    private String allResourceVersion() {
+        if (allResourceVersion != null) {
+            return allResourceVersion;
+        }
+        return allResourceVersion = Hashing.sha256()
+                                           .hashInt(versionedResources.values().hashCode())
```
jrhee17, line 112: Note: Checked that the returned `values` for `ImmutableMap` will 1) consistently return the same hashcode if called on the same `versionedResources` 2) will return a hash of all elements which guarantees some degree of randomness.

jrhee17 (approved):
👍 👍 👍

xds/src/main/java/com/linecorp/centraldogma/xds/internal/CentralDogmaSnapshotResources.java
minwoox: `CentralDogmaSnapshotResources`, which `CentralDogmaXdsResources` creates, has to have all resources so we can't create multiple `CentralDogmaXdsResources`.
However, when we support node groups later, we will create multiple `CentralDogmaXdsResources`s https://github.com/line/centraldogma/blob/70aa9a65938fdfbb1bb2a9e279873f526047288d/xds/src/main/java/com/linecorp/centraldogma/xds/internal/ControlPlanePlugin.java#L91

xds/src/main/java/com/linecorp/centraldogma/xds/internal/CentralDogmaSnapshotResources.java
minwoox: > 2) will return a hash of all elements which guarantees some degree of randomness.

Yeah, I also believe the chances of collision are extremely low given the xDS resources cardinality.
https://stackoverflow.com/a/4014407

jrhee17: I don't doubt the randomness of SHA 256 😄    It's just that I've never thought about the hashcode of `ImmutableMap#values` before

minwoox: So did I before this PR. 😆

minwoox:
I will work on implementing `XdsKubernetesService` that fetches endpoint information from k8s control plane based on this PR. Thanks a lot for reviewing. 😉

PR #967, opened 2024-07-03
Wait for internal projects to be created before starting the mirror migration

ikhoon:
Motivation:

Plugins are started asynchronously when a command executor starts. A mirror migration job needs to access `dogma/dogma` project to check if a migration job was performed before.

I worked on a similar logic before. After not working on it for a long time, I deleted the code when merging the main branch. I was confused and I thought it was redundant.

Modifications:

- Add `whenInitialized()` to `InternalProjectInitializer`
  - Wait for the future to complete before starting a mirror migration job

Result:

The bug will not affect users as the new mirror format hasn't been released.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
@@ -16,49 +16,6 @@
 
 package com.linecorp.centraldogma.server;
 
-import static com.google.common.base.MoreObjects.firstNonNull;
-import static com.google.common.base.Preconditions.checkState;
-import static com.google.common.base.Strings.isNullOrEmpty;
-import static com.linecorp.centraldogma.internal.api.v1.HttpApiV1Constants.API_V0_PATH_PREFIX;
-import static com.linecorp.centraldogma.internal.api.v1.HttpApiV1Constants.API_V1_PATH_PREFIX;
-import static com.linecorp.centraldogma.internal.api.v1.HttpApiV1Constants.HEALTH_CHECK_PATH;
-import static com.linecorp.centraldogma.internal.api.v1.HttpApiV1Constants.METRICS_PATH;
-import static com.linecorp.centraldogma.server.auth.AuthProvider.BUILTIN_WEB_BASE_PATH;
-import static com.linecorp.centraldogma.server.auth.AuthProvider.LOGIN_API_ROUTES;
-import static com.linecorp.centraldogma.server.auth.AuthProvider.LOGIN_PATH;
-import static com.linecorp.centraldogma.server.auth.AuthProvider.LOGOUT_API_ROUTES;
-import static com.linecorp.centraldogma.server.auth.AuthProvider.LOGOUT_PATH;
-import static java.util.Objects.requireNonNull;
-
-import java.io.File;
-import java.io.IOException;
-import java.io.InputStream;
-import java.net.InetSocketAddress;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionStage;
-import java.util.concurrent.Executor;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.ForkJoinPool;
-import java.util.concurrent.LinkedBlockingQueue;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.ThreadPoolExecutor;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
-import java.util.concurrent.atomic.AtomicInteger;
-import java.util.function.BiFunction;
-import java.util.function.Consumer;
-import java.util.function.Function;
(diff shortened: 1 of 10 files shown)


minwoox:
Could you also address this comment in this PR?
https://github.com/line/centraldogma/pull/880/files#r1645377689
Sorry I forgot to check it before merging. 😓

ikhoon:
The current code is not type-safe and easy to make mistakes. Let me try to refactor to use `DepedencyInjector` so that we can annotate `@RequestConverter` in the abstract class.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirroringServicePlugin.java
```diff
@@ -50,6 +50,7 @@ public synchronized CompletionStage<Void> start(PluginContext context) {
             mirroringService = new DefaultMirroringService(new File(cfg.dataDir(), "_mirrors"),
                                                            context.projectManager(),
                                                            context.meterRegistry(),
+                                                           context.internalProjectInitializer(),
```
minwoox, line 53: Can't we just wait here instead of passing the initializer?
The thread that invokes `internalProjectInitializer.whenInitialized().get();` in `migrate()` isn't changed.

ikhoon: I wanted to perform the logic in the `MirroringMigrationService` since `MirroringMigrationService` will be removed in the future.

Anyway, either is fine.

minwoox: Or, how about calling `internalProjectInitializer.whenInitialized().get()` in  `PluginGroupStartStop.doStart()` to prevent the situation when another plugin is added?

ikhoon: +1 Good idea.

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
jrhee17, line 1: Question) Maybe I'm not understanding the situation correctly, but would calling `projectInitializer.initialize()` before the call to `startCommandExecutor` help?

https://github.com/line/centraldogma/blob/531ffc35c6d4d5244647e23408b79b8ee1074b80/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java#L415-L420

The fact that we're passing the `projectInitializer` to plugins implies that `projectInitializer` needs to be in a ready state to be written to. I feel like creating the `dogma` project should be done before plugins are started in any case.

jrhee17 (approved):
Left a question, but approving based on implementation correctness 👍 👍 👍

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
ikhoon: > but would calling projectInitializer.initialize() before the call to startCommandExecutor help?

If a CommandExecutor hasn't started, `projectInitializer.initialize()` will fail.

> dogma project should be done before plugins are started in any case.

Agreed. I will remove `public` from `whenInitialized()` and call `whenInitialized().get()` before the plugins start.

ikhoon: The packages were different so it was not easy to remove `public`. Let me keep it. 😅

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -448,6 +449,13 @@ private CommandExecutor startCommandExecutor(
         final Consumer<CommandExecutor> onTakeLeadership = exec -> {
             if (pluginsForLeaderOnly != null) {
                 logger.info("Starting plugins on the leader replica ..");
+                try {
+                    // Wait until the internal project is initialized.
+                    projectInitializer.whenInitialized().get();
```
minwoox, line 454: Isn't this only called for the leader?
Shouldn't we do this for all plugins?

ikhoon: Others are explicitly invoked after `projectInitializer.initialize()`.
The process happened in the same thread. 
https://github.com/line/centraldogma/blob/e0439173f3f81bc8b3562483bcd9d133f1e40a49/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java#L1093-L1101

minwoox: Ah I missed that. 😓

minwoox (approved):
Looks good, thanks!

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
ikhoon: `onTakeLeadership` is invoked synchronously which is a deadlock. I changed it to be invoked in `PluginGroupStartStop.doStart()`.

minwoox:
👍 👍 👍

PR #973, opened 2024-07-08
Allow adding `@` prefix for internal projects

minwoox:
Motivation:
The xDS module requires its own Central Dogma project for managing xDS resources.
While it's possible to create a project with an arbitrary name that is not currently is use,
adding a `@` prefix to the project name has several benefits:
- It allows adding other internal projects later without worrying about duplicate names
- It makes it easy to distinguish internal projects from non-internals.

Modifications:
- Split valid name pattern to `USER_INPUT_PROJECT_AND_REPO_NAME_PATTERN` and `PROJECT_AND_REPO_NAME_PATTERN`
  - `USER_INPUT_PROJECT_AND_REPO_NAME_PATTERN` is used to validate the names entered by users.
  - `PROJECT_AND_REPO_NAME_PATTERN` is used internally.

Result:
- You can now create an internal project whose name starts with `@`.

### DIFF
--- a/common/src/main/java/com/linecorp/centraldogma/internal/Util.java
+++ b/common/src/main/java/com/linecorp/centraldogma/internal/Util.java
@@ -55,9 +55,20 @@ public final class Util {
      * Start with an alphanumeric character.
      * An alphanumeric character, minus, plus, underscore and dot are allowed in the middle.
      * End with an alphanumeric character.
+     * Use this pattern for project and repository names that are entered by users.
      */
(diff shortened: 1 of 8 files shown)


minwoox:
This version has to be deployed before #965 is deployed.

ikhoon (approved):
Thanks! 👍👍

jrhee17 (approved):
👍 👍 👍

ikhoon:
Still LGTM

trustin (approved):
:+1:

PR #981, opened 2024-07-15
Add xDS group service. 

minwoox:
Motivation:
A user might want to manage her own xDS resources with a group of people who have permission to create and update these resources. By creating an xDS group (a Central Dogma repository), she can easily manage these xDS resources and grant write permissions to specific people, allowing them to manage the resources under Central Dogma's authorization system.

Modifications:
- Added `XdsGroupService` to create a group, which is a Central Dogma repository.
  - With this group, a user can easily manage privileged people under the Central Dogma authorization system.
  - `xds_group.proto` file is also added.
    - All proto files might be published in a separate module later for other clients such as LEGY.
- Removed `MetadataServiceInjector`.
  - `MetadataService` is injected when a decorator is created using a `DependencyInjector`.

Result:
- You can now create xDS groups to manage your xDS resources under the Central Dogma authorization system.


xds/src/main/java/com/linecorp/centraldogma/xds/application/v1/XdsApplicationService.java
```diff
@@ -0,0 +1,117 @@
...
+        final String name = removePrefix("applications/", applicationName);
+        if (projectManager.get(XDS_CENTRAL_DOGMA_PROJECT).repos().exists(name)) {
+            throwAlreadyExists(name);
+        }
+        createRepository(commandExecutor, mds, currentAuthor(), XDS_CENTRAL_DOGMA_PROJECT, name)
+                .handle((unused, cause) -> {
+                    if (cause != null) {
+                        if (cause instanceof RepositoryExistsException) {
+                            throwAlreadyExists(name);
+                        }
```
ikhoon, line 67: Is this exception propagated somewhere?

xds/src/main/java/com/linecorp/centraldogma/xds/application/v1/XdsApplicationService.java
```diff
@@ -0,0 +1,117 @@
...
+                                          .asRuntimeException();
+        }
+
+        // TODO(minwoox): Check the permission.
+        removeRepository(commandExecutor, mds, currentAuthor(), XDS_CENTRAL_DOGMA_PROJECT, name)
+                .handle((unused, cause1) -> {
+                    if (cause1 != null) {
+                        responseObserver.onError(
+                                Status.INTERNAL.withDescription("Failed to delete " + applicationName)
+                                               .withCause(cause1).asRuntimeException());
```
ikhoon, line 110: Add `return` in the end of this block?
```suggestion
                                               .withCause(cause1).asRuntimeException());
                             return;
```

xds/src/main/proto/centraldogma/xds/application/v1/xds_application.proto
```diff
@@ -0,0 +1,62 @@
...
+    option (google.api.http) = {
+      post: "/api/v1/xds/applications"
+      body: "application"
+    };
+  }
+
+  // Deletes an application.
+  rpc DeleteApplication(DeleteApplicationRequest) returns (google.protobuf.Empty) {
+    option (google.api.http) = {
+      delete: "/api/v1/xds/{name=applications/*}"
```
ikhoon, line 41: For simplicity.
```suggestion
      delete: "/api/v1/xds/applications/{name}"
```

xds/src/main/proto/centraldogma/xds/application/v1/xds_application.proto
```diff
@@ -0,0 +1,62 @@
...
+  // Format: applications/{application}
+  string name = 1 [(google.api.field_behavior) = REQUIRED];
+
+  // If set to true, any xds Resources from this application will also be deleted.
+  // (Otherwise, the request will only work if the application has no xDS resources.)
+  // bool force = 2;
+}
+
+message Application {
+  // Format: applications/{application}
```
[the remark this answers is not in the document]
minwoox: https://github.com/aip-dev/google.aip.dev/blob/master/aip/general/0135.md#request-message
I'm trying to adhere to the AIP guidelines as much as possible. 😉

xds/src/main/proto/centraldogma/xds/application/v1/xds_application.proto
minwoox: ditto. I'm trying to adhere to the AIP.
https://github.com/aip-dev/google.aip.dev/blob/master/aip/general/0135.md#guidance

server/src/main/java/com/linecorp/centraldogma/server/internal/api/RepositoryServiceUtil.java
```diff
@@ -0,0 +1,43 @@
...
+import com.linecorp.centraldogma.common.Revision;
+import com.linecorp.centraldogma.server.command.Command;
+import com.linecorp.centraldogma.server.command.CommandExecutor;
+import com.linecorp.centraldogma.server.metadata.MetadataService;
+
+public final class RepositoryServiceUtil {
+
+    public static CompletableFuture<Revision> createRepository(
+            CommandExecutor commandExecutor, MetadataService mds,
+            Author author, String projectName, String repoName) {
```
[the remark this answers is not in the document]
minwoox: These methods are called from `RepositoryServiceV1` and `XdsApplicationService` and I think just leaving this class as a utility class is enough. If you don't mind, let's leave it as is. 😉

server/src/main/java/com/linecorp/centraldogma/server/plugin/PluginInitContext.java
```diff
@@ -54,4 +60,11 @@ public PluginInitContext(CentralDogmaConfig config,
     public ServerBuilder serverBuilder() {
         return serverBuilder;
     }
+
+    /**
+     * Returns the {@link AuthService} of the Central Dogma server.
+     */
+    public Function<? super HttpService, AuthService> authService() {
```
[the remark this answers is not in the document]
minwoox: I have thought about generalizing it but I couldn't imagine another decorator to share except the `AuthService`. Could we do this when we really need it? 😉

xds/src/main/java/com/linecorp/centraldogma/xds/application/v1/XdsApplicationService.java
minwoox: Oops. 😓

xds/src/main/java/com/linecorp/centraldogma/xds/group/v1/XdsGroupService.java
```diff
@@ -0,0 +1,119 @@
...
+        this.projectManager = projectManager;
+        this.commandExecutor = commandExecutor;
+        mds = new MetadataService(projectManager, commandExecutor);
+    }
+
+    @Override
+    public void createGroup(CreateGroupRequest request,
+                            StreamObserver<Group> responseObserver) {
+        final String groupName = request.getGroup().getName();
+        final String name = removePrefix("groups/", groupName);
```
jrhee17, line 58: Taking another look at this PR, it seems like we're going need to extra logic to remove/add `groups/` in all user-facing locations. i.e. API paths, DTOs, UI, metrics, etc.. I'm not sure if it's realistic and worth it to ensure this.

Do you mind explaining why we need this prefix again?

xds/src/main/java/com/linecorp/centraldogma/xds/internal/ControlPlanePlugin.java
```diff
@@ -148,6 +149,12 @@ private void init0(PluginInitContext pluginInitContext) {
                                                    .useBlockingTaskExecutor(true)
                                                    .build();
         sb.route().build(grpcService);
+        final GrpcService xdsApplicationService =
+                GrpcService.builder()
+                           .addService(new XdsGroupService(
+                                   projectManager, pluginInitContext.commandExecutor()))
+                           .enableHttpJsonTranscoding(true).build();
+        sb.service(xdsApplicationService, pluginInitContext.authService());
```
jrhee17, line 157: Question) Just so I understand the current state, IIUC only admins have access to internal projects. Hence, am I understanding correctly that only admins will be able to use `XdsGroupService` for now?

xds/src/main/java/com/linecorp/centraldogma/xds/group/v1/XdsGroupService.java
```diff
@@ -0,0 +1,119 @@
...
+            throw alreadyExistsException(name);
+        }
+        createRepository(commandExecutor, mds, currentAuthor(), XDS_CENTRAL_DOGMA_PROJECT, name)
+                .handle((unused, cause) -> {
+                    if (cause != null) {
+                        if (cause instanceof RepositoryExistsException) {
+                            responseObserver.onError(alreadyExistsException(name));
+                        } else {
+                            responseObserver.onError(
+                                    Status.INTERNAL.withCause(cause).asRuntimeException());
```
jrhee17, line 69: Note: I'm curious on how we'll deal with handling exceptions later on as sometimes we perform special operations like in `ReplicationLagTolerantCentralDogma` for `RevisionNotFoundException`.

For now, I assume creating a repository and getting the repository immediately won't necessary succeed if the get is from a different replica.

Anyways, I guess the current approach is fine for now

xds/src/main/proto/centraldogma/xds/group/v1/xds_group.proto
```diff
@@ -0,0 +1,62 @@
...
+  // Creates a new group.
+  rpc CreateGroup(CreateGroupRequest) returns (Group) {
+    option (google.api.http) = {
+      post: "/api/v1/xds/groups"
+      body: "group"
+    };
+  }
+
+  // Deletes an group.
+  rpc DeleteGroup(DeleteGroupRequest) returns (google.protobuf.Empty) {
```
jrhee17, line 39: Note: This is a red flag personally since I think there is always the possibility that additional fields will need to be added. Having said this, I guess it's fine to try adhering to the AIP like you proposed for now.

jrhee17 (commented):
Looks good overall, the `/groups` prefix is the only issue bothering me. The rest looks good 👍

xds/src/main/proto/centraldogma/xds/group/v1/xds_group.proto
minwoox: > This is a red flag personally since I think there is always the possibility that additional fields will need to be added. 

Would you mind explaining which fields are likely to be added later?
If that's reasonable, I will define a message and return it instead of the `Empty`.

xds/src/main/java/com/linecorp/centraldogma/xds/group/v1/XdsGroupService.java
minwoox: That's a good question. I will add the gRPC exception handler for the xDS services after I all set up the services.

xds/src/main/java/com/linecorp/centraldogma/xds/internal/ControlPlanePlugin.java
minwoox: Anyone can use the `XdsGroupService` for now as I talked before.
https://github.com/line/centraldogma/pull/981/files#diff-6530b293347a15e8133c26cfb4221f0b565c32f0a6cccb8c1d44c0d78108c8bfR62

After modifying the authorization system similar to GitHub, anyone can create a group, and the one who created the group can add anyone or tokens to the group who will have the write permission for the resources in the group.

xds/src/main/java/com/linecorp/centraldogma/xds/group/v1/XdsGroupService.java
minwoox: Like I said before, I'm trying to adhere to the AIP as much as possible:
https://google.aip.dev/122
It recommends in this way:
```
publishers/123/books/les-miserables
users/vhugo1802

Resource name components should usually alternate between collection identifiers (example: publishers, books, users) and resource IDs (example: 123, les-miserables, vhugo1802).
```

>  we're going need to extra logic to remove/add groups/ in all user-facing locations.

Yeah, that is true. But the logic isn't so many, and we still need that logic for validating and removing group name from the cluster name.
I thought the format (`groups/{group}/clusters/{cluster}`), which follows the guideline, was much better than the original format (`{group}/{cluster}`).

server/src/main/java/com/linecorp/centraldogma/server/plugin/PluginInitContext.java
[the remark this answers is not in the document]
minwoox: I think it's not easy to create `AuthProvider`, which is used to create an `AuthService`,  in CentralDogmaConfig with the current design.

xds/src/main/java/com/linecorp/centraldogma/xds/internal/ControlPlanePlugin.java
jrhee17: I see, I must have glanced through the diffset. I understood that for now, anyone can create/delete anyone's group 👍

xds/src/main/java/com/linecorp/centraldogma/xds/group/v1/XdsGroupService.java
jrhee17: If you're strong on following the AIP, then sure. Ideally I prefer that there is a cross-cutting solution to apply this, but I guess we'll be doing this manually then.

e.g. If a new API is added where groupName is the parameter, we should be conscious to add/remove the prefix.

xds/src/main/proto/centraldogma/xds/group/v1/xds_group.proto
jrhee17: Let's just follow the AIP for now since the point is to ease these types of discussions.

jrhee17 (approved):
👍 👍 👍

server/src/main/java/com/linecorp/centraldogma/server/plugin/PluginInitContext.java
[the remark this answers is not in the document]
ikhoon: I see. Let's keep the code now and consider it later if changes are needed.

xds/src/main/java/com/linecorp/centraldogma/xds/group/v1/XdsGroupService.java
```diff
@@ -0,0 +1,119 @@
...
+     */
+    public XdsGroupService(ProjectManager projectManager, CommandExecutor commandExecutor) {
+        this.projectManager = projectManager;
+        this.commandExecutor = commandExecutor;
+        mds = new MetadataService(projectManager, commandExecutor);
+    }
+
+    @Override
+    public void createGroup(CreateGroupRequest request,
+                            StreamObserver<Group> responseObserver) {
```
ikhoon, line 56: Optional) What do you think of using Reactive-gRPC instead of the vanilla gRPC? It is only used for server modules so we are relatively free of dependencies. `Reactive-gRPC` could remove boilerplate code such as `.onNext() + .onCopmlete()`.

ikhoon (approved):
Thanks!

xds/src/main/java/com/linecorp/centraldogma/xds/group/v1/XdsGroupService.java
minwoox: >  Ideally I prefer that there is a cross-cutting solution to apply this

I also prefer applying the cross-cutting solution, which is why I hesitated to follow the AIP at first. 😕 However, I've decided to try using gRPC for this API, and it’s not easy to apply AOP even if I don’t follow AIP because it needs parameters in the message. So, I've decided to just follow the AIP.

xds/src/main/java/com/linecorp/centraldogma/xds/group/v1/XdsGroupService.java
minwoox: Thanks for the suggestion. 👍 
Well, honestly, I didn't want to use any other gRPC libraries because the logic is relatively simple (All methods are unary) and it might take more time (maybe just for me) when troubleshooting.

However, that doesn't mean that I will never use grpc-reactor. I will definitely use it when I think it gives more advantages or when I want to use it. 🤣

PR #984, opened 2024-07-18
Introduce `XdsCentralDogmaBuilder`

jrhee17:
**Motivation**

CentralDogma is used deployed in multiple regions/zones. Given that we will soon be supporting zone-aware load balancing from Armeria's side, we can use CentralDogma to take advantage of zone-aware load balancing.

For instance:
- we may route only reads to the CentralDogma in the same zone, whereas write requests will be done to a single pre-designated zone.
- in DR situations, we can quickly route traffic between zones dynamically

**Modifications**
- Introduced `XdsCentralDogmaBuilder`, which builds an `ArmeriaCentralDogma` based on parameters
  - Most methods in `AbstractArmeriaCentralDogmaBuilder` didn't make sense for xDS, so `AbstractCentralDogmaBuilder` is extended directly
- Moved `ArmeriaCentralDogma` from the client package to the internal package so that `XdsCentralDogmaBuilder` can also build an `ArmeriaCentralDogma`.
- The `CentralDogma` created by `XdsCentralDogmaBuilder` creates an `XdsBootstrap` internally that should be closed. For this reason, `CentralDogma` now inherits `AutoCloseable` which closes underlying resources such as `XdsBootstrap` or `EndpointGroup`.
- Misc) Modified `CentralDogmaRuleDelegate` since configuring a `CentralDogmaExtension` with `server#port` threw an exception since the ipv6 scope id wasn't understood.

**Result**
- `java-armeria-xds` client is introduced


ikhoon (approved):
Looks good! 👍👍

minwoox (approved):
Looks awesome. 👍 👍 👍

PR #988, opened 2024-07-18
Add xDS resource services.

minwoox:
Motivation:
We need APIs for creating, updating and deleting xDS resources.

Modifications:
- Added `xds_{resource}.proto` files that defines APIs for xDS resources.
  - The API provides both gRPC and HTTP API.
  - The API follows the [API Improvement Proposals by Google](https://google.aip.dev/) so it's recommnded to read the documentation especially
    [Create](https://google.aip.dev/133), [Update](https://google.aip.dev/134) and [Delete](https://google.aip.dev/135).
- Added `XdsClusterService`, `XdsEndpointService`, `XdsListenerService` and `XdsRouteService`.
  - They implemented the `Xds{Resource}ServiceImplBase` generated from the proto files.
  - Each create API request has `parent` (`groups/{group}`) and `{resource}_id`.
    - The resource name is automatically composed using those fields: `groups/{group}/{resources}/{resource}_id`. e.g `groups/foo/clusters/echo-cluster`
    - Thus, the resource name in the resource is ignored: https://github.com/aip-dev/google.aip.dev/blob/master/aip/general/0133.md#user-specified-ids
    - An endpoint whose type is `ClusterLoadAssignment` does not have name field.
      - Thus, the endpoint name is converted to a cluster name and it's used for retreiving.
      - e.g. `groups/foo/endpoints/echo-cluster` -> `groups/foo/clusters/echo-cluster`
  - Each update HTTP API request has `{resource}.name` for identifying the resource.
    - Thus, the resource name in the resource is ignored. However, update gRPC API uses the resource name.
    - The endpoint name is also converted to a cluster name.

Result:
- You can now create, update and delete xDS resource through gRPC and HTTP APIs.


ikhoon (approved):
Looks nice! 👍👍

xds/src/main/java/com/linecorp/centraldogma/xds/endpoint/v1/XdsEndpointService.java
```diff
@@ -0,0 +1,124 @@
...
+                                                      .toBuilder()
+                                                      .setClusterName(clusterName)
+                                                      .build();
+        xdsResourceManager.push(responseObserver, group, fileName(endpointId),
+                                "Create endpoint: " + clusterName, endpoint, currentAuthor());
+    }
+
+    private static String clusterName(String parent, String endpointId) {
+        // Use /clusters/ instead of /endpoints/ for the cluster name.
+        // /endpoints/ will be used for the file name.
```
jrhee17, line 82: Note: Understood the peculiarities of `XdsEndpointService`

xds/src/main/java/com/linecorp/centraldogma/xds/cluster/v1/XdsClusterService.java
```diff
@@ -0,0 +1,100 @@
...
+     */
+    public XdsClusterService(XdsResourceManager xdsResourceManager) {
+        this.xdsResourceManager = xdsResourceManager;
+    }
+
+    @Override
+    public void createCluster(CreateClusterRequest request, StreamObserver<Cluster> responseObserver) {
+        final String parent = request.getParent();
+        final String group = removePrefix("groups/", parent);
+        xdsResourceManager.checkGroup(group);
```
jrhee17, line 57: Note: Calling `createGroup` and immediately calling `createCluster` might fail if the created group is not propagated across replicas yet

xds/src/main/proto/centraldogma/xds/endpoint/v1/xds_endpoint.proto
```diff
@@ -0,0 +1,102 @@
...
+
+  // Deletes a endpoint.
+  rpc DeleteEndpoint(DeleteEndpointRequest) returns (google.protobuf.Empty) {
+    option (google.api.http) = {
+      delete: "/api/v1/xds/{name=groups/*/endpoints/**}"
+    };
+  }
+
+  // TODO(minwoox): Implement the following methods.
+  //rpc addLocalityLbEndpoint(addLocalityLbEndpointRequest) returns
```
jrhee17, line 57: So do we consider `LocalityLbEndpoint` as a `resource`?  or do we use `UpdateEndpointRequest.update_mask`?

jrhee17 (approved):
👍 👍 👍

xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsResourceManager.java
```diff
@@ -0,0 +1,187 @@
...
+                responseObserver.onError(Status.INTERNAL.withCause(cause).asRuntimeException());
+                return null;
+            }
+            if (entries.isEmpty()) {
+                // TODO(minwoox): implement allowMissing.
+                responseObserver.onError(Status.NOT_FOUND.withDescription("Resource not found: " + resourceName)
+                                                         .asRuntimeException());
+                return null;
+            }
+            task.run();
```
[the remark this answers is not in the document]
minwoox: That is correct. 👍

xds/src/main/java/com/linecorp/centraldogma/xds/cluster/v1/XdsClusterService.java
minwoox: That is correct. 👍 It's like creating a repository and immediately pushing a file might fail.

xds/src/main/proto/centraldogma/xds/endpoint/v1/xds_endpoint.proto
minwoox: This API will add a `LocalityLbEndpoint` to an existing `ClusterLoadAssignment`. I thought this would be useful for situations where each instance registers its information dynamically. The intended flow is:
- A user creates a `ClusterLoadAssignment` with empty endpoints.
- Each instance will (de)register its information via this API to the `ClusterLoadAssignment`.

xds/src/main/java/com/linecorp/centraldogma/xds/cluster/v1/XdsClusterService.java
jrhee17: Understood, also noted that the revision isn't returned in the response.

xds/src/main/proto/centraldogma/xds/endpoint/v1/xds_endpoint.proto
jrhee17: I see, I think we can worry about this API later then. I was wondering how this API fits into your idea of following the google AIP.

As a side-note, I also think an interesting point we will have to think about is how to guarantee consistency for concurrent `[add|remove]LocalityLbEndpoint` calls

minwoox: >  I was wondering how this API fits into your idea of following the google AIP.

I was trying to use a custom verb so we don't have to follow the create or update AIP
e.g. `patch: /api/v1/xds/{endpoint_name=groups/*/endpoints/**}:register"`

> we will have to think about is how to guarantee consistency for concurrent [add|remove]LocalityLbEndpoint calls

I was going to introduce a sort of CAS commit with a windowing. Will share the idea once it's concretized.

PR #990, opened 2024-07-22
Syntax highlighting for well known configuration files

ikhoon:
Motivation:

Only JSON was the only supported language to syntax highlight. If we support other files, readability could be increased.

Modifications:

- Map file extension to Monaco's language ID.

Result:

Syntax highlighting is supported for YAML and XML.


minwoox (approved):
👍 👍 👍

PR #991, opened 2024-07-22
Correctly log elapsed milliseconds in `CentralDogmaBeanFactory`

KarboniteKream:
Motivation:

When the elapsed time in `CentralDogmaBeanFactory#get()` is logged, it prints nanoseconds with `ms` unit.

Modifications:

- Convert the elapsed time to milliseconds before logging.

Result:

The logs correctly display the elapsed time of obtaining the initial value.


jrhee17 (approved):
Thanks! Separately, we probably need to fix our CI 🙇

minwoox (approved):
Thanks! 👍 👍 👍

client/java/src/main/java/com/linecorp/centraldogma/client/updater/CentralDogmaBeanFactory.java
```diff
@@ -328,7 +328,7 @@ public <T> T get(T defaultValue, Class<T> beanType, Consumer<T> changeListener,
                     throw ex;
                 }
             }
-            final long elapsedMillis = System.nanoTime() - t0;
+            final long elapsedMillis = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - t0);
```
ikhoon, line 331: 😱

ikhoon (approved):
Nice catch! 👍👍

minwoox:
@KarboniteKream 👍 👍 👍

PR #993, opened 2024-07-23
Provide a way to configure each plugin.

minwoox:
Motivation:
Configuring user-defined plugins is currently difficult because there is no way to configure their properties via `dogma.json`.

Modifications:
- Added `PluginConfig` interface that represents a configuration of a plugin.
- Added `pluginConfigs` to `CentralDogmaConfig` that accepts a list of `PluginConfig`.
  - Users can now configure each plugin using `PluginConfig`.
- Moved mirroring plugin properties from `CentralDogmaConfig` to `MirroringServicePluginConfig`.
  - Compatibility was not maintained as few users are running the Central Dogma server, and they can be guided directly. Let me know if a different approach is preferred.

Result:
- Users can now specify plugin configurations in the `plugins` property.
  ```json
  {
    "dataDir": "./data",
    ...
    "pluginConfigs": [
      {
        "type": "com.example.MyPlugin",
        "enabled": true, // enabled by default
        "myPluginTimeout": 10
      },
      // Other plugins
    ]
  }
  ```
- (Breaking changes) `mirroringEnabled`, `numMirroringThreads`, `maxNumFilesPerMirror`, and `maxNumBytesPerMirror` have been removed from `CentralDogmaBuilder` and `CentralDogmaConfig`.
  - These properties should now be specified in the plugins property:
    ```json
    {
      "dataDir": "./data",
      ...
      "pluginConfigs": [
        {
          "type": "com.linecorp.centraldogma.server.mirror.MirroringServicePluginConfig",
          "enabled": true,
          "numMirroringThreads": 16,
          "maxNumFilesPerMirror": 8192
        }  
      ]
    }
    ```

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaBuilder.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaBuilder.java
@@ -80,6 +80,22 @@ public final class CentralDogmaBuilder {
             "maximumWeight=134217728," + // Cache up to apx. 128-megachars.
             "expireAfterAccess=5m";      // Expire on 5 minutes of inactivity.
 
+    static final boolean XDS_CONTROL_PLANE_FOUND;
+
+    static {
+        XDS_CONTROL_PLANE_FOUND =
+                exists("com.linecorp.centraldogma.xds.internal.ControlPlanePlugin");
+    }
+
+    private static boolean exists(String className) {
+        try {
+            Class.forName(className, false, CentralDogmaBuilder.class.getClassLoader());
+            return true;
+        } catch (Throwable t) {
+            return false;
+        }
+    }
+
     // Armeria properties
     // Note that we use nullable types here for optional properties.
     // When a property is null, the default value will be used implicitly.
@@ -107,9 +123,13 @@ public final class CentralDogmaBuilder {
     private int numMirroringThreads = DEFAULT_NUM_MIRRORING_THREADS;
     private int maxNumFilesPerMirror = DEFAULT_MAX_NUM_FILES_PER_MIRROR;
     private long maxNumBytesPerMirror = DEFAULT_MAX_NUM_BYTES_PER_MIRROR;
+
+    private boolean xdsControlPlaneEnabled = XDS_CONTROL_PLANE_FOUND;
+
     @Nullable
     private GracefulShutdownTimeout gracefulShutdownTimeout;
     private ReplicationConfig replicationConfig = ReplicationConfig.NONE;
+    @Nullable
     private String accessLogFormat;
 
     // AuthConfig properties
@@ -407,6 +427,15 @@ public final class CentralDogmaBuilder {
     }
 
     /**
+     * Sets whether xDS control plane is enabled or not.
+     * If unspecified, xDS control plane is enabled.
+     */
+    public CentralDogmaBuilder xdsControlPlaneEnabled(boolean xdsControlPlaneEnabled) {
+        this.xdsControlPlaneEnabled = xdsControlPlaneEnabled;
+        return this;
+    }
+
+    /**
      * Sets the graceful shutdown timeout. If unspecified, graceful shutdown is disabled.
      */
     public CentralDogmaBuilder gracefulShutdownTimeout(GracefulShutdownTimeout gracefulShutdownTimeout) {
@@ -584,8 +613,8 @@ public final class CentralDogmaBuilder {
                                       numRepositoryWorkers, repositoryCacheSpec,
                                       maxRemovedRepositoryAgeMillis, gracefulShutdownTimeout,
                                       webAppEnabled, webAppTitle, mirroringEnabled, numMirroringThreads,
-                                      maxNumFilesPerMirror, maxNumBytesPerMirror, replicationConfig,
-                                      null, accessLogFormat, authCfg, quotaConfig,
+                                      maxNumFilesPerMirror, maxNumBytesPerMirror, xdsControlPlaneEnabled,
+                                      replicationConfig, null, accessLogFormat, authCfg, quotaConfig,
                                       corsConfig);
     }
 }
--- a/server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaBuilder.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaBuilder.java
@@ -49,6 +49,7 @@ import com.linecorp.centraldogma.server.auth.AuthConfig;
 import com.linecorp.centraldogma.server.auth.AuthProvider;
 import com.linecorp.centraldogma.server.auth.AuthProviderFactory;
 import com.linecorp.centraldogma.server.auth.Session;
+import com.linecorp.centraldogma.server.plugin.PluginConfig;
 import com.linecorp.centraldogma.server.storage.repository.Repository;
 
 import io.micrometer.core.instrument.MeterRegistry;
--- a/server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaBuilder.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaBuilder.java
@@ -71,42 +71,29 @@ public final class CentralDogmaBuilder {
     private static final ServerPort DEFAULT_PORT = new ServerPort(36462, SessionProtocol.HTTP);
 
     static final int DEFAULT_NUM_REPOSITORY_WORKERS = 16;
-    static final int DEFAULT_NUM_MIRRORING_THREADS = 16;
-    static final int DEFAULT_MAX_NUM_FILES_PER_MIRROR = 8192;
-    static final long DEFAULT_MAX_NUM_BYTES_PER_MIRROR = 32 * 1048576; // 32 MiB
     static final long DEFAULT_MAX_REMOVED_REPOSITORY_AGE_MILLIS = 604_800_000;  // 7 days
 
     static final String DEFAULT_REPOSITORY_CACHE_SPEC =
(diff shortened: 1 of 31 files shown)


server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaBuilder.java
```diff
@@ -406,6 +426,15 @@ public CentralDogmaBuilder maxNumBytesPerMirror(long maxNumBytesPerMirror) {
         return this;
     }
 
+    /**
+     * Sets whether xDS control plane is enabled or not.
+     * If unspecified, xDS control plane is enabled.
+     */
+    public CentralDogmaBuilder xdsControlPlaneEnabled(boolean xdsControlPlaneEnabled) {
```
jrhee17, line 433: Note) This PR starts leaking xDS logic into the core implementation.

jrhee17 (approved):
Looks good in terms of correctness 👍 👍 👍

server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaBuilder.java
ikhoon: It would not make sense `xdsControlPlaneEnabled` is set to `true` when `XDS_CONTROL_PLANE_FOUND == false`

ikhoon: Optional) Otherwise, we may generalize this to plug-in plugins. 
```
interface Plugin {
   // A unique name for this plugin
   String name();
}
```
```json5
{
  # All plugins are enabled by default.
  "plugins": {
    "include": [ "mirror" ]
    "exclude": [ "xds" ]
  }
}
```

minwoox: > xdsControlPlaneEnabled is set to true when XDS_CONTROL_PLANE_FOUND == false

So, I've added a logic for that:
https://github.com/line/centraldogma/pull/993/files#diff-170293b368ccad4c74d5f040277fa7439a76a671becd4db7dfbc59576bac95a0R301

> Otherwise, we may generalize this to plug-in plugins.

That's a good suggestion. Let me apply for that.

minwoox:
I haven't fixed the documentation yet. I will update it once the changes are accepted.

server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaBuilder.java
minwoox: @jrhee17, @ikhoon Changed to configure each plugin. PTAL. 🙇

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirroringServicePluginConfig.java
```diff
@@ -0,0 +1,82 @@
...
+
+import static com.google.common.base.MoreObjects.firstNonNull;
+import static com.google.common.base.Preconditions.checkArgument;
+
+import javax.annotation.Nullable;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.base.MoreObjects;
+
+final class MirroringServicePluginConfig {
```
ikhoon, line 26: What do you think of making `PluginConfig` an interface and extending `PluginConfig` to provide mirror configuration as we did in `MirrorCredential`?

With this approach, users could set plugin configuration type-safely when using builder methods. For example:
```java
new CentralDogmaBuilder(...)
   . pluginConfigs(new MirrorPluginConfig(...), 
                   ControlPlanePluginConfig.disabled())
```

minwoox: Addressed. 👍

ikhoon (approved):
Looks better! 👍👍

server/src/main/java/com/linecorp/centraldogma/server/plugin/PluginConfigDeserializer.java
```diff
@@ -0,0 +1,75 @@
...
+    /**
+     * Creates a new instance.
+     */
+    public PluginConfigDeserializer() {
+        super(PluginConfig.class);
+    }
+
+    @Override
+    public PluginConfig deserialize(JsonParser jp, DeserializationContext ctxt)
+            throws IOException, JsonProcessingException {
```
jrhee17, line 45:
```suggestion
            throws IOException {
```

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirroringServicePlugin.java
```diff
@@ -47,12 +50,30 @@ public synchronized CompletionStage<Void> start(PluginContext context) {
         DefaultMirroringService mirroringService = this.mirroringService;
         if (mirroringService == null) {
             final CentralDogmaConfig cfg = context.config();
+            final Optional<PluginConfig> optional =
+                    cfg.pluginConfigs().stream().filter(plugin -> configType().isInstance(plugin))
```
jrhee17, line 54: nit; I think it may be worth just constructing a map `Class -> PluginConfig` in `CentralDogmaConfig` so that we don't have to do this every time. I think it's fine not to do this in this PR though.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirroringServicePlugin.java
```diff
@@ -47,12 +50,30 @@ public synchronized CompletionStage<Void> start(PluginContext context) {
...
+            final long maxNumBytesPerMirror;
+
+            final MirroringServicePluginConfig mirroringServicePluginConfig =
+                    (MirroringServicePluginConfig) optional.orElse(null);
+            if (mirroringServicePluginConfig != null) {
+                numThreads = mirroringServicePluginConfig.numMirroringThreads();
+                maxNumFilesPerMirror = mirroringServicePluginConfig.maxNumFilesPerMirror();
+                maxNumBytesPerMirror = mirroringServicePluginConfig.maxNumBytesPerMirror();
+            } else {
+                numThreads = MirroringServicePluginConfig.INSTANCE.numMirroringThreads();
```
jrhee17, line 67: Question) If `cfg.pluginConfigs()` doesn't have a `MirroringServicePluginConfig`, I thought this plugin wouldn't be added in the first place since it wouldn't be in the following list.

https://github.com/line/centraldogma/blob/326b5099ba973b937add2a99ae82a21cbcef4c3c/server/src/main/java/com/linecorp/centraldogma/server/PluginGroup.java#L86-L90

I'm probably missing something, but do you mind explaining when this scenario can occur?

jrhee17 (approved):
Left some questions, but still looks good 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirroringServicePlugin.java
minwoox: > If cfg.pluginConfigs() doesn't have a MirroringServicePluginConfig, I thought this plugin wouldn't be added in the first place since it wouldn't be in the following list.

Had a chat about this with @ikhoon and we've decided to enable the plugin if there's no configuration specified in the `plugins` property.

Do you think we shouldn't add the plugin if the `PluginConfig` is missing? 🤔

jrhee17: I think I just missed the following line. No objection to enabling plugins without a config specified 👍 

https://github.com/line/centraldogma/blob/326b5099ba973b937add2a99ae82a21cbcef4c3c/server/src/main/java/com/linecorp/centraldogma/server/plugin/Plugin.java#L58

minwoox: Do you think it will be less confusing if we use `pluginConfigs` instead of `plugins`? 🤔 
```
{
  "dataDir": "./data",
  ...
  "pluginConfigs": [
    {
      "type": "com.linecorp.centraldogma.server.mirror.MirroringServicePluginConfig",
      "enabled": true,
      "numMirroringThreads": 16,
      "maxNumFilesPerMirror": 8192
    }  
  ]
}
```

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirroringServicePlugin.java
minwoox: That's a good suggestion. Addressed. 👍 
Note that I've added the additional `pluginConfigMap` method and left the `pluginConfigs()` as is for JSON serialization.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirroringServicePlugin.java
jrhee17: Good point 👍  I think so (I assume you are referring to the naming exposed publicly)

minwoox: Yeah, the name in the configuration file. 😉 
I thought the name `plugins` might have caused confusion because it's not plugins themselves, but rather plugin configurations, which can be omitted.

Thanks, fixed. 😉

minwoox:
Thanks a lot for reviewing. 😉

PR #994, opened 2024-07-23
Do not use a global cache for Node.js

ikhoon:
Motivation:

We have seen a weird error message in CI builds.
```
Couldn't follow symbolic link '/home/runner/.gradle/caches/nodejs/server/node-v16.14.0-linux-x64/bin/npx'.
```
We used to use a custom gradle cache key in the CI build workflow. But in the end, we removed it and delegated a cache logic to `gradle/gradle-build-action`. So I think we will use the cached node.js without referring to the glocal cache location.

Modification:

- Remove the custom cache location for Node.js

Result:

Stable CI builds

### DIFF
--- a/server/build.gradle
+++ b/server/build.gradle
@@ -7,12 +7,6 @@ node {
     npmVersion = '10.8.1'
     download = true
     npmInstallCommand = "ci"
-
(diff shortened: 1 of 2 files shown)


minwoox (approved):
👍 👍 👍

PR #996, opened 2024-07-26
Add JsonProperty to ControlPlanePluginConfig constructor parameter

minwoox:
Modification:
- Add `JsonProperty` to `ControlPlanePluginConfig` constructor parameter

Result:
- You can now configure `ControlPlanePluginConfig` via configuration file properly.

### DIFF
--- a/xds/src/main/java/com/linecorp/centraldogma/xds/internal/ControlPlanePluginConfig.java
+++ b/xds/src/main/java/com/linecorp/centraldogma/xds/internal/ControlPlanePluginConfig.java
@@ -17,6 +17,8 @@ package com.linecorp.centraldogma.xds.internal;
 
 import javax.annotation.Nullable;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
+
 import com.linecorp.centraldogma.server.plugin.AbstractPluginConfig;
 
 /**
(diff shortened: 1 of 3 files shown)


jrhee17 (approved):
👍 👍 👍

xds/src/main/java/com/linecorp/centraldogma/xds/internal/ControlPlanePluginConfig.java
```diff
@@ -26,7 +28,7 @@ public final class ControlPlanePluginConfig extends AbstractPluginConfig {
     /**
      * Creates a new instance.
      */
-    public ControlPlanePluginConfig(@Nullable Boolean enabled) {
+    public ControlPlanePluginConfig(@JsonProperty("enabled") @Nullable Boolean enabled) {
```
ikhoon, line 31: Don't we need `@JsonCreator`?

minwoox: It's not needed when there's only one constructor. Let me add it though for clarity.

trustin (approved):
:+1:

ikhoon (approved):
Thanks! 🙇‍♂️🙇‍♂️

PR #999, opened 2024-07-28
Add @ConsumseJson to contents API.

### LINKED ISSUE #987: Add `@Consume(content-type)` to all REST services class or methods
All Central Dogma APIs use JSON (Patch) to exchange data. A Central Dogma server can't handle other types such as XML or YAML.

Currently, the request `content-type` is not specified in some API methods.
https://github.com/line/centraldogma/blob/24603eb149d3945fde2cc1ced71a9efdcbc231fe/server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java#L187-L191
That results in generating a converter not found error message if `content-type` is missing.
For example, a commit is sent without `content-type`, `No suitable request converter` is returned from which users don't know what is wrong.
```sh
$ curl -XPOST "http://127.0.0.1:36462/api/v1/projects/foo/repos/bar/contents" \
 -H "Authorization: Bearer appToken-***" \
 -d '{
    "commitMessage" : {
        "summary": "hello",
        "detail": "a",
        "markup": "MARKDOWN"
    },
    "changes" : [
        {
            "path" : "/foo0.json",
            "type" : "UPSERT_JSON",
            "content" : {"a": "bar2"}
        }
    ]
 }'
{"exception":"java.lang.IllegalArgumentException","message":"No suitable request converter found for a @RequestObject 'CommitMessageDto'"}%
```

If `@ConsumeJson` is added to the method, `405 Method Not Allowed` is returned which would make it easier to identify the cause.

chickenchickenlove:
### Motivation:
- When a client sends a request with a JSON body but does not configure the `content-type` in header, Central Dogma returns the response`: {"exception":"java.lang.IllegalArgumentException","message":"No suitable request converter found for a @RequestObject 'CommitMessageDto'"}`. This response is quite ambiguous, making it difficult for the client to understand the problem.

### Modifications:
- Add `@ConsumesJson` to each method that uses `ChangesRequestConverter.class`. 

### FYI
1. If `@RequestConverter` is declared on method signatures, Armeria will add an object Resolver. ([link](https://github.com/line/armeria/blob/e2b298dd2f54243e8801540f13e90166f75578b5/core/src/main/java/com/linecorp/armeria/internal/server/annotation/AnnotatedValueResolver.java#L501-L508))
2. Then, when Armeria receives a request from client, Armeria will try to resolve the request.  ([link](https://github.com/line/armeria/blob/e2b298dd2f54243e8801540f13e90166f75578b5/core/src/main/java/com/linecorp/armeria/internal/server/annotation/AnnotatedValueResolver.java#L829-L862))
3. `ChangesRequestConverter` tries to parse body from request. Actually, `ChangeRequestConverter` delegates to `JacksonRequestConvertFunction`.
4. `JacksonRequestConverter` validates contents type. ([here](https://github.com/line/armeria/blob/e2b298dd2f54243e8801540f13e90166f75578b5/core/src/main/java/com/linecorp/armeria/server/annotation/JacksonRequestConverterFunction.java#L98-L106))
5. If there is no `content-type` or `content-type` is not `application/json`, `JacksonRequestConverter` will call `RequestConverterFunction.fallthrough()`. 

With this flow, client will receive response `: {"exception":"java.lang.IllegalArgumentException","message":"No suitable request converter found for a @RequestObject 'CommitMessageDto'"}`.

If `@ConsumsJson` is delcared to each method having `ChangesRequestConverter.class` on their method signature, Armeria will not try to resolve request without header `Content-Type: application/json`.

### Result:
- Client will receive 415 response.
```sh
$ curl -X POST localhost:8080/my-post \
-d '{"name": "John Doe", "email": "john.doe@example.com"}'
>>
Status: 415
Description: Unsupported Media Type
```
- Closes #987.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
@@ -51,6 +51,7 @@ import com.linecorp.armeria.common.HttpResponse;
 import com.linecorp.armeria.common.HttpStatus;
 import com.linecorp.armeria.common.util.Exceptions;
 import com.linecorp.armeria.server.ServiceRequestContext;
+import com.linecorp.armeria.server.annotation.ConsumesJson;
 import com.linecorp.armeria.server.annotation.Default;
 import com.linecorp.armeria.server.annotation.Get;
 import com.linecorp.armeria.server.annotation.Param;
@@ -187,6 +188,7 @@ public class ContentServiceV1 extends AbstractService {
      * <p>Pushes a commit.
      */
     @Post("/projects/{projectName}/repos/{repoName}/contents")
+    @ConsumesJson
     @RequiresWritePermission
     public CompletableFuture<PushResultDto> push(
             ServiceRequestContext ctx,
@@ -226,6 +228,7 @@ public class ContentServiceV1 extends AbstractService {
      * <p>Previews the actual changes which will be resulted by the given changes.
      */
     @Post("/projects/{projectName}/repos/{repoName}/preview")
+    @ConsumesJson
     public CompletableFuture<Iterable<ChangeDto<?>>> preview(
             ServiceRequestContext ctx,
             @Param @Default("-1") String revision,
--- a/server/src/test/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1Test.java
+++ b/server/src/test/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1Test.java
@@ -417,7 +417,7 @@ class ContentServiceV1Test {
                 "application/xml, 415",
                 "application/x-www-form-urlencoded, 415"
         })
-        void getFileWithoutSuitableContentType(String mediaType, int expectedStatusCode) {
+        void getFileWithVariusContentType(String mediaType, int expectedStatusCode) {
             // Given:
             final WebClient client = dogma.httpClient();
             final String body =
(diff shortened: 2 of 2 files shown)


ikhoon:
It looks good. 

> Client will receive 415 response.

Would you add a test that verifies the status? It is a simple test but will be helpful to detect regressions in the future.

chickenchickenlove:
@ikhoon nim, i make a new commit to apply your comments. 

I discovered something in the process.
When the `Content-Type` is not configured for each client, the default `Content-Type` may differ.

the [values](https://github.com/line/armeria/blob/0960d091bfc7f350c17e68f57cc627de584b9705/core/src/main/java/com/linecorp/armeria/server/DefaultRoute.java#L111) may differ! 
`dogma.client` don't configure `Content-Type`. thus `value` is `null`.
`curl` configure `application/x-form-urlencoded` as default. thus `value` is `application/x-form/urlencoded`. 
`aiohttp.ClientSession()` in python configure `text/plain` as default. thus `value` is `text/plain`. 


Since the `Dogma client` defaults to a `null` Content-Type, it always will receive 400 response. 
Therefore, to properly test this, I used parameterized tests to vary the `Content-Type`.

ikhoon (approved):
Thanks, @chickenchickenlove! 👍👍

jrhee17 (approved):
👍 👍

minwoox (approved):
Thanks, @chickenchickenlove ❤️

PR #1000, opened 2024-07-29
Fix to fetch the file content with revision

ikhoon:
Motiviation:

A revision number should be specified to get old data.

Modifications:

- Use the revision number specified in the path variable to fetch file content

Result:

Old content is now correctly rendered


trustin (approved):
:+1:

minwoox (approved):
👍 👍 👍

jrhee17 (approved):
👍 👍 👍

PR #1007, opened 2024-08-13
Ignore invalid configurations when migrating legacy mirrors and credentials

ikhoon:
Motivation:

If an exception is raised while performing a mirror migration job, it tries to roll back the migration.
https://github.com/line/centraldogma/blob/5912bf7dd52bda8c414ba1e25c9a1155f981b42d/server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirroringMigrationService.java#L135-L139 
However, the rollback also could fail if there is no migration on a project. `entries` may be empty which results in raising `RedundantChangeException`.
https://github.com/line/centraldogma/blob/5912bf7dd52bda8c414ba1e25c9a1155f981b42d/server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirroringMigrationService.java#L266-L267

Initially, I thought it was better to stop the whole migration job immediately to make the migration status atomic if an exception was raised. However, that was a bad choice. If it fails, it remains in `REPLICATION_ONLY` mode, which is difficult to fix manually.

Modification:

- Always recover the server status to `WRITABLE` even when a mirror migration job completes exceptionally.
- Warn invalid mirror and credential configurations and continue to migrate the remaining projects.
- Try to roll back if the new migration directory is not empty.

Result:

- The server status is correctly reset to `WRITABLE` when a mirror migration job fails.

### DIFF
--- a/server-mirror-git/src/test/java/com/linecorp/centraldogma/server/internal/mirror/MirroringMigrationServiceTest.java
+++ b/server-mirror-git/src/test/java/com/linecorp/centraldogma/server/internal/mirror/MirroringMigrationServiceTest.java
@@ -120,6 +120,22 @@ class MirroringMigrationServiceTest {
             "  ]\n" +
             '}';
 
+    // An invalid mirror config that has no localRepo.
+    static final String INVALID_MIRROR =
+            "{\n" +
+            "  \"type\": \"single\",\n" +
+            "  \"enabled\": true,\n" +
+            "  \"schedule\": \"0 * * * * ?\",\n" +
+            "  \"direction\": \"REMOTE_TO_LOCAL\",\n" +
+            "  \"localPath\": \"/\",\n" +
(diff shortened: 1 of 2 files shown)


minwoox:
https://github.com/line/centraldogma/blob/c8967c93bdd883ee4041a4e24ef781a3b64c4da0/server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirroringMigrationService.java#L414
You will fix this in a different PR?

ikhoon:
A mirror configuration that does not have `localRepo` will not perform correctly. So I let the configuration be ignored from migration. The exception is caught by https://github.com/line/centraldogma/blob/65b410f017bc94a58a3a961866324281d9552e93/server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirroringMigrationService.java#L210-L214

minwoox (approved):
Thanks! 👍 👍 👍

PR #1010, opened 2024-08-13
Check if the internal project and repos exist before creating them

minwoox:
Motivation:
Central Dogma server currently attempts to create internal projects and repositories during starting up, even if they already exist. This is unnecessary for most scenarios, except when setting up a new cluster.

Modifications:
- Added checks to verify if the internal project and repositories exist before attempting to create them.
- Propagated the read-only exception to the caller if it's raised while creating the internal project.
  - It doesn't make sense to continue to run the cluster because the internal project must exist when the cluster gets back to non read-only mode.

Result:
- Central Dogma servers now checks for the existence of internal projects and repos before attempting to create them.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/command/Command.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/command/Command.java
@@ -375,8 +375,11 @@ public interface Command<T> {
      */
     static <T> Command<T> forcePush(Command<T> delegate) {
         requireNonNull(delegate, "delegate");
-        checkArgument(delegate.type() == CommandType.NORMALIZING_PUSH || delegate.type() == CommandType.PUSH,
-                      "delegate: %s (expected: NORMALIZING_PUSH or PUSH)", delegate);
+        checkArgument(delegate.type() == CommandType.CREATE_PROJECT ||
+                      delegate.type() == CommandType.CREATE_REPOSITORY ||
+                      delegate.type() == CommandType.NORMALIZING_PUSH || delegate.type() == CommandType.PUSH,
+                      "delegate: %s (expected: CREATE_PROJECT, CREATE_REPOSITORY, NORMALIZING_PUSH or PUSH)",
+                      delegate);
         return new ForcePushCommand<>(delegate);
     }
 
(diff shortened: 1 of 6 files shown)


server/src/main/java/com/linecorp/centraldogma/server/storage/project/InternalProjectInitializer.java
```diff
@@ -86,17 +91,15 @@ public void initialize(String projectName) {
...
-            final Throwable peeled = Exceptions.peel(cause);
-            if (peeled instanceof ReadOnlyException) {
-                // The executor has stopped right after starting up.
-                return;
-            }
-            if (!(peeled instanceof ProjectExistsException)) {
-                throw new Error("failed to initialize an internal project: " + projectName, peeled);
+        if (!projectManager.exists(projectName)) {
+            try {
+                executor.execute(createProject(creationTimeMillis, Author.SYSTEM, projectName))
```
ikhoon, line 96: Suggestion: Could you wrap `createProject()` and `createRepository()` with `forcePush()` command to make it work in `REPLICATION_ONLY` mode?

minwoox: Good suggestion. Applied. 👍

server/src/main/java/com/linecorp/centraldogma/server/command/Command.java
```diff
@@ -375,8 +375,13 @@ static Command<Void> updateServerStatus(ServerStatus serverStatus) {
...
         requireNonNull(delegate, "delegate");
-        checkArgument(delegate.type() == CommandType.NORMALIZING_PUSH || delegate.type() == CommandType.PUSH,
-                      "delegate: %s (expected: NORMALIZING_PUSH or PUSH)", delegate);
+        checkArgument(delegate.type() == CommandType.CREATE_PROJECT ||
+                      delegate.type() == CommandType.CREATE_REPOSITORY ||
+                      delegate.type() == CommandType.NORMALIZING_PUSH || delegate.type() == CommandType.PUSH,
+                      "delegate: %s (expected: CREATE_PROJECT, CREATE_REPOSITORY, NORMALIZING_PUSH or PUSH)",
+                      delegate);
+        checkArgument(delegate.author().equals(Author.SYSTEM), "delegate.author: %s (expected: SYSTEM)",
+                      delegate.author());
```
ikhoon, line 384: Should we also allow administrators to run force push?

ikhoon: Nevermind. There was no association between `Author` and `User`.

ikhoon:
The test failure in CI will be fixed by https://github.com/line/centraldogma/pull/1016

ikhoon (approved):
Thanks, @minwoox! 👍👍

PR #1011, opened 2024-08-14
Log `TokenNotFoundException` at the debug level

ikhoon:
Motivation:

`TokenNotFoundException` may be useful to debug a request failure but the warning log would be noisy to the maintainers of a Central Dogma cluster.

Modifications:

- Use the `DEBUG` level to log `TokenNotFoundException`.
  - `IllegalArgumentException` is also logged at `DEBUG`.

Result:

`TokenNotFoundException` is now logged at the `DEBUG` level in `ApplicationTokenAuthorizer`.


minwoox (approved):
👍

jrhee17 (approved):
👍 👍 👍

Issue #700, opened 2022-07-12
Provide a way to trigger Git mirroring immediately


minwoox:
Users might want to do that when the repo has been modified manually on CD UI.

PR #1012, opened 2024-08-14
Reject invalid mirroring URI

ikhoon:
Motivation:

An expected remote URI for mirroring is `<domain.com>/<org>/<repo>.git`. However, both the mirror UI and REST API didn't validate the input.

For example, a SSH URL, `github.com:line/centraldogma.git` is not valid in Central Dogma but can be committed. Exceptions are raised later when it is used.

Additionally, I also fixed a bug where settings tabs are not visible to admin that I found while fixing this bug.

Modifications:

- Improved the regular expression for `remoteUri` to strictly check the input.
- Validate a remote URI in the mirror REST API.
- Always set the project role of adminstrator to `OWNER`

Result:

Invalid remote URIs are now rejected by the mirror UI and API.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/MirrorConfig.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/MirrorConfig.java
@@ -99,6 +99,9 @@ public final class MirrorConfig {
         this.localRepo = requireNonNull(localRepo, "localRepo");
         this.localPath = firstNonNull(localPath, "/");
         this.remoteUri = requireNonNull(remoteUri, "remoteUri");
+        // Validate the remote URI.
+        RepositoryUri.parse(remoteUri, direction == MirrorDirection.REMOTE_TO_LOCAL ? "git" : "dogma");
+
         if (gitignore != null) {
             if (gitignore instanceof Iterable &&
                 Streams.stream((Iterable<?>) gitignore).allMatch(String.class::isInstance)) {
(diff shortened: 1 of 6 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/MirrorConfig.java
```diff
@@ -99,6 +100,11 @@ public MirrorConfig(@JsonProperty("id") String id,
         this.localRepo = requireNonNull(localRepo, "localRepo");
         this.localPath = firstNonNull(localPath, "/");
         this.remoteUri = requireNonNull(remoteUri, "remoteUri");
+
+        // Validate the remote URI.
+        final String suffix = remoteUri.getScheme().equals(SCHEME_DOGMA) ? "dogma" : "git";
```
minwoox, line 105: Shouldn't we do this in the `MirrorDto` or `MirroringServiceV1`?
How about also checking if the local repo exists?

minwoox:
How about trying mirroring before pushing so that an invalid configuration isn't committed at all?
I've implemented the k8s service in that way.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/MirrorConfig.java
ikhoon: > Shouldn't we do this in the MirrorDto or MirroringServiceV1?

I chose this place because the segments of remote URI in `MirrorDto` need to be assembled.
https://github.com/line/centraldogma/blob/5fec9f111c862f47d24350a1d16f8c36b8acf0a8/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/DefaultMetaRepository.java#L301-L306

> How about also checking if the local repo exists?

That would be covered by Mirror UI which provides a select box to list local repos.
However, validation on the server side is a also good idea. I will do it in the f/u work. 😉

minwoox (approved):
👍 👍 👍

ikhoon:
> How about trying mirroring before pushing so that an invalid configuration isn't committed at all?

This PR focuses on validating a remote URI. I will do it in the next milestone while implementing https://github.com/line/centraldogma/issues/700

PR #1013, opened 2024-08-14
Remove a trailing slash in the webapp paths

ikhoon:
Motivation:

`/app/projects` is more concise and simple than `/app/projects/`

Modifications:

- Set `trailingSlash` to false and revert workarounds to support a trailing slash.
- Miscellanous) Generated a Central Dogma favicon from `central_dogma.png` and added to `public/` directory
- Bump Armeria version to 1.30.0

Result:

Paths in the webapp no longer contains a trailing slash.


dependencies.toml
```diff
@@ -3,7 +3,7 @@
 #     If its classes are exposed in Javadoc, update offline links as well.
 #
 [versions]
-armeria = "1.29.4"
+armeria = "1.30.0"
```
minwoox, line 6: Do you want to update dependencies in a separate PR?

minwoox (approved):
👍 👍 👍

dependencies.toml
ikhoon: It was updated to use the new API in Armeria. The dependency upgrade PR also includes Armeria updates.

PR #1017, opened 2024-08-16
Access JAR file only once to find Envoy extension classes

minwoox:
In #1008, I used reflection to find Envoy extension classes in the Jar file.
However, I've noticed that accessing the JAR file for the second time takes significantly longer, though I haven't been able to determine the cause.


ikhoon (approved):
Thanks, @minwoox!

jrhee17 (approved):

PR #1018, opened 2024-08-20
Fix bugs in UI pagination

ikhoon:
Motivation:

- When a filter value is set, the rows are filtered by the value but the original page index is not updated correctly.
- `New repository` button wasn't shown up in the anonymous mode.

Modifications:

- Fixed `Filter` to set the page index to 0 when a filter value is updated.
- Fixed `WithProjectRole` component to show the child components in the anonymous mode.

Result:

- Pagination works properly even when filters are applied.
- The new repository button is now correctly exposed in anonymous mode.


jrhee17 (approved):

minwoox (approved):
Thanks! 👍 👍 👍

PR #1020, opened 2024-08-20
Skip the directory entry when creating xDS resources at server startup

minwoox:
Motivation:
`Repository.find(...)` fetches not just files but for directories. We should skip the directories when creating xDS resources.

### DIFF
--- a/xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsResourceWatchingService.java
+++ b/xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsResourceWatchingService.java
@@ -29,6 +29,7 @@ import org.slf4j.LoggerFactory;
 
 import com.linecorp.centraldogma.common.Change;
 import com.linecorp.centraldogma.common.Entry;
+import com.linecorp.centraldogma.common.EntryType;
 import com.linecorp.centraldogma.common.RepositoryNotFoundException;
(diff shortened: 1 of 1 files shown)


xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsResourceWatchingService.java
```diff
@@ -85,6 +86,9 @@ protected void init() {
             logger.info("Creating xDS resources from {} at revision: {}", groupName, normalizedRevision);
             final Map<String, Entry<?>> entries = repository.find(normalizedRevision, pathPattern()).join();
             for (Entry<?> entry : entries.values()) {
+                if (entry.type() == EntryType.DIRECTORY || !entry.hasContent()) {
```
jrhee17, line 89: Question) Considering only a JSON type is supported, would it make more sense to whitelist instead so that unexpected json deserialization failures could be avoided?

```suggestion
                if (entry.type() != EntryType.JSON || !entry.hasContent()) {
```

jrhee17 (approved):
👍 👍 👍

xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsResourceWatchingService.java
minwoox: That's a good suggestion. 👍

PR #1022, opened 2024-08-22
Avoid Using Future.join() During Control Plane Plugin Startup

minwoox:
Motivation:
The use of `repository.find(...).join()` during the control plane plugin startup introduces unnecessary delays.

Modifications:
- Removed the use of `future.join()` during the control plane plugin startup to avoid blocking the control plane executor.

Result:
- Reduced startup time.

### DIFF
--- a/xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsResourceWatchingService.java
+++ b/xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsResourceWatchingService.java
@@ -79,6 +79,7 @@ public abstract class XdsResourceWatchingService {
      * Must be executed by {@link #executor()}.
      */
     protected void init() {
+        logger.info("Initializing {}...", getClass().getSimpleName());
         final Builder<CompletableFuture<Void>> futures = ImmutableList.builder();
         for (Repository repository : xdsProject.repos().list().values()) {
             final String groupName = repository.name();
@@ -117,6 +118,7 @@ public abstract class XdsResourceWatchingService {
         CompletableFuture.allOf(futures.build().toArray(new CompletableFuture[0])).join();
         // Watch dogma repository to add newly created xDS projects.
         watchDogmaRepository();
+        logger.info("{} initialized.", getClass().getSimpleName());
     }
 
     private void watchDogmaRepository() {
--- a/xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsResourceWatchingService.java
+++ b/xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsResourceWatchingService.java
@@ -94,9 +94,8 @@ public abstract class XdsResourceWatchingService {
             futures.add(findFuture.handleAsync((entries, cause) -> {
                 // Executed by the repository worker thread.
                 if (cause != null) {
-                    logger.warn("Unexpected exception while finding {} at revision: {}",
-                                groupName, normalizedRevision, cause);
-                    return null;
+                    throw new RuntimeException("Unexpected exception while finding " + groupName +
+                                               " at revision: " + normalizedRevision, cause);
                 }
                 for (Entry<?> entry : entries.values()) {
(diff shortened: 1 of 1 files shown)


xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsResourceWatchingService.java
```diff
@@ -84,24 +88,33 @@ protected void init() {
...
-                    throw new RuntimeException("Unexpected exception while building an xDS resource from " +
-                                               groupName + path, t);
+                for (Entry<?> entry : entries.values()) {
+                    if (entry.type() != EntryType.JSON || !entry.hasContent()) {
+                        continue;
+                    }
+                    final String path = entry.path();
+                    final String contentAsText = entry.contentAsText();
+                    try {
+                        handleXdsResource(path, contentAsText, groupName);
```
minwoox, line 107: In `handleXdsResource`, there is a heavy computation: serializing the text to a protobuf message.
Will revisit this later, if this PR doesn't make any changes.

xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsResourceWatchingService.java
```diff
@@ -84,26 +89,36 @@ protected void init() {
...
-            final Map<String, Entry<?>> entries = repository.find(normalizedRevision, pathPattern()).join();
-            for (Entry<?> entry : entries.values()) {
-                if (entry.type() != EntryType.JSON || !entry.hasContent()) {
-                    continue;
+            final CompletableFuture<Map<String, Entry<?>>> findFuture =
+                    repository.find(normalizedRevision, pathPattern());
+            futures.add(findFuture.handleAsync((entries, cause) -> {
+                // Executed by the repository worker thread.
+                if (cause != null) {
+                    logger.warn("Unexpected exception while finding {} at revision: {}",
```
jrhee17, line 97: Question) It seems like this behavior has changed from failing the startup to just logging a warning.
When restarting, are we expected to check for this log? or would just explicitly throwing a `RuntimeException` make more sense?

xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsResourceWatchingService.java
```diff
@@ -84,26 +89,36 @@ protected void init() {
...
+                for (Entry<?> entry : entries.values()) {
+                    if (entry.type() != EntryType.JSON || !entry.hasContent()) {
+                        continue;
+                    }
+                    final String path = entry.path();
+                    final String contentAsText = entry.contentAsText();
+                    try {
+                        handleXdsResource(path, contentAsText, groupName);
+                    } catch (Throwable t) {
+                        logger.warn("Unexpected exception while building an xDS resource from {}.",
```
jrhee17, line 110: ditto

Question) It seems like this behavior has changed from failing the startup to just logging a warning.
When restarting, are we expected to check for this log? or would just explicitly throwing a `RuntimeException` make more sense?

jrhee17 (approved):
👍 👍

xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsResourceWatchingService.java
minwoox: I thought it was okay to skip the repository but it was probably not a good idea.
Let me re-throw the exception in this case. 😉

xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsResourceWatchingService.java
minwoox: Let me just leave a warning log in this case.
This could happen when an admin made a commit via `ContentServiceV1` and I think it doesn't make sense to stop Central Dogma because of just one file. We can just ignore it. 😉

ikhoon (approved):
👍👍

PR #1023, opened 2024-08-22
`XdsCentralDogmaBuilder` accepts a `localClusterName` instead of `serviceCluster`

jrhee17:
**Motivation**

While verifying whether zone-aware routing works correctly, I found that the `serviceCluster` options wasn't being respected.

> (string) Name of the local cluster (i.e., the cluster that owns the Envoy running this configuration). In order to enable zone aware routing this option must be set. If local_cluster_name is defined then clusters must be defined in the Bootstrap static cluster resources. This is unrelated to the --service-cluster option which do
es not affect zone aware routing.

ref: https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/bootstrap/v3/bootstrap.proto#envoy-v3-api-field-config-bootstrap-v3-clustermanager-local-cluster-name

I'm not sure why I thought using `serviceCluster` was a good idea, but it is probably a better idea to accept `localClusterName` if we want to take advantage of zone-aware routing

**Modifications**
- Renamed `serviceCluster` to `localClusterName`
- Set `localClusterName` to `Bootstrap#clusterManager#localClusterName`

**Result**
- `XdsCentralDogmaBuilder` supports zone-aware routing correctly


minwoox (approved):
👍 👍 👍

ikhoon (approved):
🙇‍♂️🙇‍♂️

PR #1024, opened 2024-08-22
Fix Deadlock Issue in Control Plane service

minwoox:
Modification:
A mistake in #1022 caused a deadlock in the control plane executor. The issue arose because the thread was waiting on a future to complete, but it needed to execute and complete that future itself, leading to a deadlock. https://github.com/line/centraldogma/pull/1022/files#diff-f99a1af4ac33cff7d57aea503c56b19cbadb765e6b1a48067de36c1998b367acR117 https://github.com/line/centraldogma/pull/1022/files#diff-f99a1af4ac33cff7d57aea503c56b19cbadb765e6b1a48067de36c1998b367acR115

Modifications:
- Changed to wait on the startup thread instead of the control plane executor.

Result:
- The deadlock issue during the control plane service startup has been resovled.

### DIFF
--- a/xds/src/main/java/com/linecorp/centraldogma/xds/internal/ControlPlanePlugin.java
+++ b/xds/src/main/java/com/linecorp/centraldogma/xds/internal/ControlPlanePlugin.java
@@ -17,7 +17,6 @@
 package com.linecorp.centraldogma.xds.internal;
 
 import java.util.concurrent.CompletionStage;
-import java.util.concurrent.TimeUnit;
 
 import com.google.common.base.MoreObjects;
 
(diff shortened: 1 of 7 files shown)


jrhee17 (approved):
👍 👍

jrhee17:
Question) Was there no way to test this issue? I'm surprised this wasn't caught in the CI in this repo, but it was caught in our internal repo

minwoox:
> Was there no way to test this issue? I'm surprised this wasn't caught in the CI in this repo, but it was caught in our internal repo

Let me add a test case for this. I need another plugin that creates an xds repository before startup.

ikhoon (approved):
👍👍

PR #1025, opened 2024-08-22
Ensure Compliance with AIP Rules for Resource Creation and Update

minwoox:
Motivation:
Identified violations of two AIP (API Improvement Proposals) rules:
- CREATE: If a user tries to create a resource with an ID that would result in a duplicate resource name, the service must return an ALREADY_EXISTS error.
- UPDATE: If the method call is on a resource that already exists and all fields match, the existing resource should be returned unchanged.

Modifications:
- Implemented the aforementioned AIP rules during resource creation and update.
- Added `ControlPlaneExceptionHandlerFunction``, which converts exceptions raised within Central Dogma to appropriate gRPC status codes.

Result:
- The service now correctly returns an ALREADY_EXISTS error for duplicate resource creation attempts.
- Resource updates that do not change any fields will now correctly return the existing resource without modification.


xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsResourceManager.java
```diff
@@ -134,11 +137,16 @@ public void checkGroup(String group) {
...
-            StreamObserver<T> responseObserver, String group, String fileName,
-            String summary, T resource, Author author) {
+            StreamObserver<T> responseObserver, String group, String resourceName, String fileName,
+            String summary, T resource, Author author, boolean create) {
         final Change<JsonNode> change;
         try {
-            change = Change.ofJsonUpsert(fileName, JSON_MESSAGE_MARSHALLER.writeValueAsString(resource));
+            final String jsonText = JSON_MESSAGE_MARSHALLER.writeValueAsString(resource);
+            if (create) {
+                change = Change.ofJsonPatch(fileName, null, jsonText);
```
jrhee17, line 146: Question) I'm wondering if there could be a race condition of two create requests are called concurrently, then the result would be a merged resource instead of the resource from either request.

e.g.
```
request1 -> addOperation("field1")
request2 -> addOperation("field2")

-> the resulting stored value has both "field1" and "field2"
```

1) Is this scenario possible? 2) I'm curious why it is necessary to use `ofJsonPatch`

jrhee17 (approved):
👍 👍

xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsResourceManager.java
minwoox: > Is this scenario possible?

It's not possible.

>  I'm curious why it is necessary to use ofJsonPatch

`Change.ofJsonPatch(fileName, null, jsonText);` means that the commit must be made only when the `fileName` doesn't exist. You can check it here how it works:
- https://github.com/line/centraldogma/blob/main/common/src/main/java/com/linecorp/centraldogma/common/Change.java#L223
- https://github.com/line/centraldogma/blob/main/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java#L1107-L1116

Without `ofJsonPatch`, when a user creates multiple resources with the same ID, the last commit will override the previous one even though it should fail.

jrhee17: I see, I missed the following point: https://github.com/line/centraldogma/blob/7add9c25e6c6e421a9c28ea17e5ee6c2b7e3b150/common/src/main/java/com/linecorp/centraldogma/internal/jsonpatch/SafeReplaceOperation.java#L55-L56

Sounds good to me 👍

PR #1028, opened 2024-08-27
Paginate the commit history page with query params

ikhoon:
Motivation:

It could paginate the commit history page but a page index or a page size is not exposed in query parameters. Because these values were managed only as internal states of Javascript, sharing specific history as a URL was impossible.

Modifications:

- Update `Breadcrumbs` to specify query parameters
- Increase the width of `CompareButton`
- Add an option to render a custom component when data is empty in `DynamicDataTable`
- Add an option to hide a go-to button in `PaginationBar`.
- Expose a browse button when if the path is a directory in `HistoryList`
- Refactored `HistoryListPage` to use query parameters to specify revision ranges.
- Use a form tag to submit the revision range with an enter key in `ChangesViewPage`
- Link to a history page instead of a commit page in a `FileEditor`
  - Commits for a specific file must be retrieved sequentially from the `HEAD` revision, the Central Dogma server has a limit of 1000 if `maxCommits` is not specified. This means a commit for a specific file is not indexed and may not be found in the recent 1000 commits.
  - As we can't scan the entire history, it would be better to show the history and guide the user to find a commit rather than failing to find a commit for the file.

Result:

You can now share a specific range of commit history as a URL.

### DIFF
--- a/webapp/src/dogma/common/components/Breadcrumbs.tsx
+++ b/webapp/src/dogma/common/components/Breadcrumbs.tsx
@@ -39,8 +39,10 @@ export const Breadcrumbs = ({
         if (omitIndexList.includes(i)) {
           return null;
         }
-        let query0 = '';
-        if (!omitQueryList.includes(i)) {
+        let query0;
+        if (omitQueryList.includes(i) || omitQueryList.includes(i - asPathNestedRoutes.length)) {
+          query0 = '';
+        } else {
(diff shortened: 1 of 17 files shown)


ikhoon:
image

jrhee17 (approved):

minwoox (approved):
👍 👍 👍

PR #1029, opened 2024-08-28
Fix regex for validating mirror remoteUrl

minwoox:
Motivation:
The regex for validating mirror remoteUrl is wrong. `[.-\/]` is interpreted as `.` to `/`. However, it should be `[.\-\/]` which allows `.`, `-` and `/` characters.

Modifications:
- Fixed regex for validating mirror remoteUrl

Result:
- You can now correctly setup a mirror url when it contains `-` in its name.

### DIFF
--- a/webapp/src/dogma/features/project/settings/mirrors/MirrorForm.tsx
+++ b/webapp/src/dogma/features/project/settings/mirrors/MirrorForm.tsx
@@ -300,7 +300,7 @@ const MirrorForm = ({ projectName, defaultValue, onSubmit, isWaitingResponse }: 
                 type="text"
                 defaultValue={defaultValue.remoteUrl}
                 placeholder="my.git.com/org/myrepo.git"
-                {...register('remoteUrl', { required: true, pattern: /^[\w.-]+(:[0-9]+)?\/[\w.-\/]+.git$/ })}
+                {...register('remoteUrl', { required: true, pattern: /^[\w.-]+(:[0-9]+)?\/[\w.\-\/]+.git$/ })}
               />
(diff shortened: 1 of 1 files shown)


jrhee17 (approved):
👍 👍 👍

webapp/src/dogma/features/project/settings/mirrors/MirrorForm.tsx
```diff
@@ -300,7 +300,7 @@ const MirrorForm = ({ projectName, defaultValue, onSubmit, isWaitingResponse }:
                 type="text"
                 defaultValue={defaultValue.remoteUrl}
                 placeholder="my.git.com/org/myrepo.git"
-                {...register('remoteUrl', { required: true, pattern: /^[\w.-]+(:[0-9]+)?\/[\w.-\/]+.git$/ })}
+                {...register('remoteUrl', { required: true, pattern: /^[\w.-]+(:[0-9]+)?\/[\w.\-\/]+.git$/ })}
```
ikhoon, line 303: 😅

ikhoon (approved):
Thanks!

webapp/src/dogma/features/project/settings/mirrors/MirrorForm.tsx
ikhoon: By the way, do we have to escape the first `-`?

minwoox: We don't have to because `-` is in the last position in `[]`. However, let me add it for clarification.

PR #1030, opened 2024-08-29
Remove hostnamePatterns from credential files

minwoox:
Motivation:
We decided to repurpose the `MirrorCredential` to manage all repository credentials, not just those specific to mirroring. The hostnamePatterns property, which was specific to mirroring, is no longer necessary. With the introduction of a dedicated mirroring API, hostnamePatterns can be safely removed.

Modifications:
- Added a job that removes the `hostnamePatterns` property from existing credential files. This prepares the system for the complete removal of this property in the next PR.
- Remove `MirroringMigrationService` to avoid conflict while executing the job.

Result:
- The `hostnamePatterns` property is now deprecated and will be removed entirely in the subsequent PR.


server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/RemovingHostnamePatternsService.java
```diff
@@ -0,0 +1,119 @@
...
+                                                        ((ObjectNode) content).without("hostnamePatterns")));
+                    }
+
+                    if (!changes.isEmpty()) {
+                        numMigratedProjects++;
+                        logger.info("hostnamePatterns in credentials are removed in the project: {}",
+                                    project.name());
+                    }
+
+                    commandExecutor.execute(Command.forcePush(Command.push(
```
ikhoon, line 101: Should we skip pushing this commit if `changes` is empty?

ikhoon (approved):
Thanks! 🙏

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirroringService.java
```diff
@@ -131,12 +131,11 @@ public synchronized void start(CommandExecutor commandExecutor) {
                     }
                 }));
 
-        // Migrate the old mirrors.json to the new format if exists.
         try {
-            new MirroringMigrationService(projectManager, commandExecutor).migrate();
```
jrhee17, line 136: Note: I understood that it will be mentioned in the release notes clearly that https://github.com/line/centraldogma/pull/880 should be released before this version (0.67.0)

---

Note: I think other parties hosting their own CentralDogma servers may have difficulty upgrading with the recent versions. Ideally, we may want to consider in the future:

1) Add warning logs to check if `hostnamePatterns` is indeed not being used, and guide users to migrate (possibly in the following location)
https://github.com/line/centraldogma/blob/534f6bf8e6e9918d2fcf878612518bd4f25aeee7/server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/credential/AbstractMirrorCredential.java#L92
2) After a couple of releases, remove the `hostnamePatterns` field and data

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/RemovingHostnamePatternsService.java
```diff
@@ -0,0 +1,121 @@
...
+
+    void start() throws Exception {
+        // Enter read-only mode.
+        commandExecutor.execute(Command.updateServerStatus(ServerStatus.REPLICATION_ONLY))
+                       .get(1, TimeUnit.MINUTES);
+        logger.info("Start removing hostnamePatterns in credential ...");
+        if (commandExecutor instanceof ZooKeeperCommandExecutor) {
+            logger.debug("Waiting for 30 seconds to make sure that all cluster have been notified of the " +
+                         "read-only mode ...");
+            Thread.sleep(30000);
```
jrhee17, line 67: Note: For the record, I don't think removing hostnames is a critical operation, and can be considered more of a data cleanup.

The current approach requires that 1) writes are blocked (which requires us to announce operations) and 2) maintainers also need to clear time for monitoring and possible operations.

Given that mirror configs are rarely modified, perhaps just pushing with a specific revision/retrying is safer.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/RemovingHostnamePatternsService.java
```diff
@@ -0,0 +1,121 @@
...
+    private final CommandExecutor commandExecutor;
+
+    RemovingHostnamePatternsService(ProjectManager projectManager, CommandExecutor commandExecutor) {
+        this.projectManager = projectManager;
+        this.commandExecutor = commandExecutor;
+    }
+
+    void start() throws Exception {
+        // Enter read-only mode.
+        commandExecutor.execute(Command.updateServerStatus(ServerStatus.REPLICATION_ONLY))
```
jrhee17, line 61: Question) Will this code path be executed for every leader change?

Given that we already had several issues with readonly mode, I prefer that the command isn't issued when we're not aware.

e.g. servers randomly go through leader election during the weekend, and servers go into readonly mode without us being aware.

jrhee17 (commented):
Left a minor question 🙇

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirroringService.java
minwoox: > I understood that it will be mentioned in the release notes clearly that https://github.com/line/centraldogma/pull/880 (0.67.0) should be released before this version

That is correct. 👍 

> Ideally, we may want to consider in the future:

That's a good idea. 👍 
Since the PR hasn't been merged yet, we can still proceed with your suggestion. However, I opted for this approach because:
- This change isn't related to a widely-used client feature.
- We can provide guidance to the server operators, who are fewer in number.
- (Personally) I'd like to wrap up the mirror-related tasks as quickly as possible."

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/RemovingHostnamePatternsService.java
minwoox: > For the record, I don't think removing hostnames is a critical operation, and can be considered more of a data cleanup.

That is correct. 👍
I've implemented it similarly to how `MirroringMigrationService` is done, but it might be a bit excessive.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/RemovingHostnamePatternsService.java
minwoox: > Question) Will this code path be executed for every leader change?

It shouldn't. 😓 Let me add a file to prevent it.

jrhee17 (approved):
👍 👍 👍

PR #1031, opened 2024-08-29
Rename `MirroringCredential` to `Credential`

minwoox:
Motivation:
We decided to repurpose the `MirrorCredential` to manage all repository credentials, not just those specific to mirroring. To reflect this role, we should remove `Mirror` prefix from the class name.

Caveat: This commit must be deployed to central dogma replicas after applying the changes from #1030.

Modifications:
- Renamed `MirroringCredential` to `Credential` and moved its package.
- Removed `hostnamePatterns` property in `Credential`.

Result:
- The renamed `Credential` class can now be used for managing various types of repository credentials, beyond just mirroring.


server/src/main/java/com/linecorp/centraldogma/server/credential/Credential.java
```diff
@@ -0,0 +1,61 @@
...
+import com.fasterxml.jackson.annotation.JsonSubTypes.Type;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+
+import com.linecorp.centraldogma.server.internal.credential.AccessTokenCredential;
+import com.linecorp.centraldogma.server.internal.credential.NoneCredential;
+import com.linecorp.centraldogma.server.internal.credential.PasswordCredential;
+import com.linecorp.centraldogma.server.internal.credential.PublicKeyCredential;
+
+/**
+ * The authentication credentials which are required when accessing the Git repositories.
```
ikhoon, line 31: Should we update Javadoc to reflect on the current usage?

server/src/main/java/com/linecorp/centraldogma/server/credential/Credential.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.EXISTING_PROPERTY, property = "type")
+@JsonSubTypes({
+        @Type(value = NoneCredential.class, name = "none"),
+        @Type(value = PasswordCredential.class, name = "password"),
+        @Type(value = PublicKeyCredential.class, name = "public_key"),
+        @Type(value = AccessTokenCredential.class, name = "access_token")
+})
+@JsonInclude(JsonInclude.Include.NON_NULL)
```
ikhoon, line 40: Should we ignore unknown fields in case the migration didn't remove `hostnamePatterns` somehow?
```suggestion
@JsonInclude(JsonInclude.Include.NON_NULL)
@JsonIgnoreProperties(ignoreUnknown = true)
```

jrhee17 (approved):
Understood that this is a simple rename, removal of hostnamePatterns

server/src/main/java/com/linecorp/centraldogma/server/credential/Credential.java
minwoox: Add addressed. 😉

ikhoon (approved):
👍👍

PR #1032, opened 2024-08-29
Ignore known exceptions from logging

ikhoon:
Motivation:

Central Dogma does not use Armeria's `LoggingClient`, so known exceptions handled by `HttpApiExceptionHandler` are logged in `DefaultUnloggedExceptionsReporter`.

Moditications:

- Set `ctx.setShouldReportUnloggedExceptions(false)` to ignore known exceptions they don't need to be logged.

Result:

You no longer see warnings about ignorable exceptions.


minwoox (approved):
👍 👍 👍

jrhee17 (approved):

PR #1033, opened 2024-08-30
Fix that the created group is not watched by XdsResourceWatchingService

minwoox:
Motivation:
The newly created group is not watched by `XdsResourceWatchingService` due to a bug.

Modification:
- Fixed to watch the created group
- Added more debugging logs and metric for watching.groups

Result:
- The newly created group is now correctly watched by `XdsResourceWatchingService`.


xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsResourceWatchingService.java
```diff
@@ -134,8 +140,9 @@ private void watchDogmaRepository() {
                     final boolean added = watchingGroups.add(groupName);
                     if (!added) {
                         // Already watching.
-                        return;
+                        continue;
```
ikhoon, line 143: 😱

jrhee17 (approved):
👍 👍

PR #1038, opened 2024-09-26
Support OAuth Token for Kubernetes Authentication via Credential Service

minwoox:
Motivation:
The current approach stores the Kubernetes oauthToken in the Kubernetes configuration, which is not ideal for sensitive information. To improve security, we need a way to specify and manage the oauthToken through the credential service, allowing it to be securely retrieved and used by the Kubernetes service.

Modifications:
- Added functionality to use the `oauthToken` for Kubernetes authentication via the credential service if the token is specified with `credential:`.

Result:
- Kubernetes OAuth tokens can now be securely managed and retrieved through the credential service.


jrhee17:
Let me wait until this is rebased over https://github.com/line/centraldogma/pull/1031 since it is difficult to review the changes related to OAuth Token.

minwoox:
> Let me wait until this is rebased over https://github.com/line/centraldogma/pull/1031 since it is difficult to review the changes related to OAuth Token.

Yeah, let me ping you when it's ready. 😉

minwoox:
@jrhee17, @ikhoon this is now ready. 😉

xds/src/main/java/com/linecorp/centraldogma/xds/k8s/v1/XdsKubernetesEndpointFetchingService.java
```diff
@@ -134,15 +136,24 @@ protected void handleXdsResource(String path, String contentAsText, String group
...
         final KubernetesEndpointsUpdater updater =
-                new KubernetesEndpointsUpdater(commandExecutor, kubernetesEndpointGroup, executorService,
+                new KubernetesEndpointsUpdater(commandExecutor, future, executorService,
                                                groupName, watcherName, endpointWatcher.getClusterName());
         updaters.put(watcherName, updater);
-        kubernetesEndpointGroup.addListener(endpoints -> {
-            if (endpoints.isEmpty()) {
-                return;
+        future.handle((kubernetesEndpointGroup, cause) -> {
+            if (cause != null) {
```
ikhoon, line 143: Would you leave a warning log for `cause`? I guess https://github.com/line/centraldogma/pull/1038/files#diff-131baeee793bf759dddfe5493e4e204e04f40869576f3d654124581c973114beR202-R205 exception is neither propagated nor logged.

xds/src/main/java/com/linecorp/centraldogma/xds/k8s/v1/XdsKubernetesEndpointFetchingService.java
```diff
@@ -134,15 +136,24 @@ protected void handleXdsResource(String path, String contentAsText, String group
...
+                updater.close();
+                return null;
             }
-            executorService.execute(updater::maybeSchedule);
-        }, true);
+            kubernetesEndpointGroup.addListener(endpoints -> {
+                if (endpoints.isEmpty()) {
+                    return;
+                }
+                executorService.execute(updater::maybeSchedule);
```
ikhoon, line 153: How about passing `kubernetesEndpointGroup` as an argument to `maybeSchedule` to avoid calling `kubernetesEndpointGroupFuture.join()`?

xds/src/main/java/com/linecorp/centraldogma/xds/k8s/v1/XdsKubernetesService.java
```diff
@@ -96,43 +102,57 @@ public void createServiceEndpointWatcher(CreateServiceEndpointWatcherRequest req
...
-                return null;
-            }
-            scheduledFuture.cancel(false);
-            kubernetesEndpointGroup.closeAsync();
+        final ContextAwareBlockingTaskExecutor taskExecutor =
+                ServiceRequestContext.current().blockingTaskExecutor();
+        final CompletableFuture<KubernetesEndpointGroup> future =
+                createKubernetesEndpointGroup(watcher, xdsResourceManager.xdsProject().metaRepo(),
+                                              taskExecutor);
+        future.handle((kubernetesEndpointGroup, cause) -> {
```
ikhoon, line 116: Should we safely call `future.join()` because the services are marked as `@Blocking`?

xds/src/main/java/com/linecorp/centraldogma/xds/k8s/v1/XdsKubernetesService.java
```diff
@@ -96,43 +102,57 @@ public void createServiceEndpointWatcher(CreateServiceEndpointWatcherRequest req
...
+                    return;
+                }
+                kubernetesEndpointGroup.closeAsync();
+                responseObserver.onError(
+                        Status.INTERNAL.withDescription(
+                                "Failed to retrieve k8s endpoints within 5 seconds. watcherName: " +
+                                watcher.getName()).asRuntimeException());
+            }, 5, TimeUnit.SECONDS);
+
+            whenReady.handle((endpoints, cause1) -> {
```
ikhoon, line 141: Ditto.

xds/src/main/java/com/linecorp/centraldogma/xds/k8s/v1/XdsKubernetesEndpointFetchingService.java
minwoox: That's a good suggestion. 👍 I'll add a log for that.
Initially, I didn’t take any action because the KubernetesEndpointGroup is already created in XdsClusterService without any issues, so I assumed the exception was a rare occurrence. However, it's still important to log it. 😉

xds/src/main/java/com/linecorp/centraldogma/xds/k8s/v1/XdsKubernetesEndpointFetchingService.java
minwoox: That's a good suggestion. 👍

xds/src/main/java/com/linecorp/centraldogma/xds/k8s/v1/XdsKubernetesService.java
minwoox: Yeah, we can do that, but I wanted to avoid blocking the thread as much as possible.
Please let me know if you still think we should still use `future.join`.

xds/src/main/java/com/linecorp/centraldogma/xds/k8s/v1/XdsKubernetesService.java
```diff
@@ -142,32 +162,51 @@ private static void validateWatcherAndPush(
...
 
-        if (!isNullOrEmpty(kubernetesConfig.getOauthToken())) {
-            configBuilder.withOauthToken(kubernetesConfig.getOauthToken());
+        final String oauthToken = kubernetesConfig.getOauthToken();
+        if (isNullOrEmpty(oauthToken)) {
+            return CompletableFuture.completedFuture(configBuilder.build());
         }
 
-        return configBuilder.build();
+        if (!oauthToken.startsWith("credential:")) {
```
jrhee17, line 196: Question) Would it make more sense to enforce users to use credentials?

Since we're still free to make breaking changes, we could just rename `oauthToken` to `credentialId` or `oauthCredentialId` while we're at it.

jrhee17 (approved):

xds/src/main/java/com/linecorp/centraldogma/xds/k8s/v1/XdsKubernetesService.java
minwoox: Yeah, that's a good idea. 👍

xds/src/main/java/com/linecorp/centraldogma/xds/k8s/v1/XdsKubernetesService.java
```diff
@@ -96,43 +102,57 @@ public void createServiceEndpointWatcher(CreateServiceEndpointWatcherRequest req
...
+            final ScheduledFuture<?> scheduledFuture = taskExecutor.schedule(() -> {
+                if (!completed.compareAndSet(false, true)) {
+                    return;
+                }
+                kubernetesEndpointGroup.closeAsync();
+                responseObserver.onError(
+                        Status.INTERNAL.withDescription(
+                                "Failed to retrieve k8s endpoints within 5 seconds. watcherName: " +
+                                watcher.getName()).asRuntimeException());
+            }, 5, TimeUnit.SECONDS);
```
ikhoon, line 139: Q) 5 seconds seems short compared to request timeouts. Why did you set 5 seconds as the deadline?

minwoox: I chose a value of less than 10 seconds, which is the request timeout.
This API has two phases: calling k8s and committing.
I estimated that the API would need at least 5 seconds for the committing phase so I allocated 5 seconds for the k8s.
I could potentially increase the default timeout and assign the larger value but I felt it wasn't necessary at this point. 
Do you have any recommendations?

xds/src/main/java/com/linecorp/centraldogma/xds/k8s/v1/XdsKubernetesService.java
minwoox: Addressed. 😉

minwoox:
Let me merge this to work on supporting merging multiple k8s clusters into one. Thanks!

PR #1040, opened 2024-09-27
Improve the robustness of old ZooKeeper log removal

trustin:
Motivation:

`OldLogRemover` in `ZooKeeperCommandExecutor` currently catches a `Throwable` when deleting an old log or its log blocks. However, it has two issues doing so:

- It doesn't handle an exception that's raised when reading the metadata of the old log.
- `Throwable` is way too wide exception to catch. Catching a `KeeperException` whose code is `NONODE` will be enough.
  - Note that the failure will only transfer the leadership to other replica, rather than stopping the whole replication process.

Modifications:

- `OldLogRemover` now catches `KeeperException` whose code is `NONODE` only.
- An attempt to read a missing log node's metadata is now handled properly.
- Added more detail to the log messages about missing nodes
  - Split `deleteLog()` into `deleteLog()` and `deleteLogBlock()`

Result:

- The leadership is not transferred anymore when `OldLogRemover` attempts to retrieve a missing log node's metadata, which is not really a critical issue.
  - Instead, the leadership will be transferred when an exception occurs not because of a missing node.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
@@ -57,7 +57,6 @@ import javax.annotation.Nullable;
 import org.apache.curator.RetryPolicy;
 import org.apache.curator.framework.CuratorFramework;
 import org.apache.curator.framework.CuratorFrameworkFactory;
-import org.apache.curator.framework.api.transaction.CuratorOp;
 import org.apache.curator.framework.imps.CuratorFrameworkState;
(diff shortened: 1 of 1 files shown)


ikhoon (approved):
Looks nice. Thanks, @trustin! 🙇‍♂️🙇‍♂️

jrhee17 (approved):
Interesting, I'm not sure how a null `LogMeta` is possible, but agree with the isolation/additional logging 👍

minwoox:
> Throwable is way too wide exception to catch. Catching a KeeperException whose code is NONODE will be enough.
Note that the failure will only transfer the leadership to other replica, rather than stopping the whole replication process.

Question: Is there any chance that the replica, which receives the leadership, raises the same exception?

minwoox (approved):
Thanks!

PR #1041, opened 2024-09-30
Provide a way to trigger Git mirroring manually

### LINKED ISSUE #236: Manual mirroring schedule
In some cases, a user might want to trigger mirroring manually, rather than on automated schedule, sometimes via the web UI and sometimes via HTTP API.

### LINKED ISSUE #700: Provide a way to trigger Git mirroring immediately

ikhoon:
Motivation:

- There is no easy way to test Git mirroring configurations. 
- Some users may want to run Git mirroring manually instead of using the periodic scheduler.

Modifications:

- Add `POST /projects/{projectName}/mirrors/{mirrorId}/run` REST API to `MirroringServiceV1` so as to trigger a mirroring task.
- Add `verboseResponses` to `CentralDogmaConfig` so as to contain full stack traces in error responses.
  - The cause of mirroring failures will be displayed in the UI for administrators or `verboseResponses` is enabled.
- Change `AbstractMirror` to return `MirrorResult` which is used to notify the result in the UI.
- Make `MirrorDto.schedule` nullable.
  - Exclude a mirror configuration from scheduling `schedule` it is null.
- Move `MirrorException` to `common` and expose it in the error responses.
- Add `cronstrue` dependency to `package.json` to display a human readable description on the mirror form UI.
- Increase `maxWidth` of `NotificationWrapper` to render stack traces well.

Result:

- You can now trigger a mirroring task in the mirror UI.
- Fixes #700
- Fixes #236


ikhoon:
[trigger-mirror.webm](https://github.com/user-attachments/assets/3dc05873-2e74-4bcf-b90d-d023acd196bc)

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirrorRunner.java
```diff
@@ -0,0 +1,149 @@
...
+                0, mirrorConfig.numMirroringThreads(),
+                // TODO(minwoox): Use LinkedTransferQueue when we upgrade to JDK 21.
+                60, TimeUnit.SECONDS, new LinkedBlockingQueue<>(),
+                new DefaultThreadFactory("mirror-api-worker", true));
+        threadPoolExecutor.allowCoreThreadTimeOut(true);
+        worker = ExecutorServiceMetrics.monitor(meterRegistry, threadPoolExecutor,
+                                                "mirrorApiWorker");
+    }
+
+    public CompletableFuture<MirrorResult> run(String projectName, String mirrorId) {
```
jrhee17, line 79: Optional) For better UX, if a `mirrorRunId` and the `triggeredTime` is returned I think it may be more clear to users whether the mirror result is from an in-flight mirror or a new attempt.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/MirroringServiceV1.java
```diff
@@ -125,11 +133,21 @@ private CompletableFuture<PushResultDto> createOrUpdate(String projectName,
         });
     }
 
+    // Mirroring may be a long-running task, so we need to increase the timeout.
+    @RequestTimeout(value = 5, unit = TimeUnit.MINUTES)
+    @Post("/projects/{projectName}/mirrors/{mirrorId}/run")
+    public CompletableFuture<MirrorResult> runMirror(@Param String projectName, @Param String mirrorId)
```
jrhee17, line 139: Question) Is there no need to at least limit the people who can run mirrors to those who own the project?

```suggestion
    @RequiresWritePermission
    public CompletableFuture<MirrorResult> runMirror(@Param String projectName, @Param String mirrorId)
```

site/src/sphinx/setup-configuration.rst
```diff
@@ -262,6 +263,11 @@ Core properties
 
     - the path of the management service. If not specified, the management service is mounted at ``/internal/management``.
 
+- ``verboseResponses`` (boolean)
```
jrhee17, line 266: Question) I'm not confident exceptions don't contain sensitive information (e.g. like credentials). If we do add this property, are you planning on enabling it in our in-house clusters by default?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/auth/ApplicationTokenAuthorizer.java
```diff
@@ -53,16 +54,20 @@ public class ApplicationTokenAuthorizer implements Authorizer<HttpRequest> {
...
         this.tokenLookupFunc = requireNonNull(tokenLookupFunc, "tokenLookupFunc");
+        this.verboseResponses = verboseResponses;
     }
 
     @Override
     public CompletionStage<Boolean> authorize(ServiceRequestContext ctx, HttpRequest data) {
         final OAuth2Token token = AuthTokenExtractors.oAuth2().apply(data.headers());
         if (token != null && token.accessToken().equals(CsrfToken.ANONYMOUS)) {
             AuthUtil.setCurrentUser(ctx, User.ANONYMOUS);
+            HttpApiUtil.setVerboseResponses(ctx, User.ANONYMOUS, verboseResponses);
```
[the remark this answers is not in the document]
ikhoon: Good point. I missed that. 😅

site/src/sphinx/setup-configuration.rst
ikhoon: Right. I intended to add this feature for use in our in-house systems.

- The sensitive information of credentials is marked when it is serialized.
- The permission manager will check if the caller is granted to access the information. 

Even though sensitive information is exposed to the caller, I think, the caller will be a user who already has the appropriate permissions.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/MirroringServiceV1.java
ikhoon: We need it.

client/java-armeria/src/main/java/com/linecorp/centraldogma/internal/client/armeria/ArmeriaCentralDogma.java
```diff
@@ -133,6 +134,7 @@ public final class ArmeriaCentralDogma extends AbstractCentralDogma {
                         .put(RepositoryExistsException.class.getName(), RepositoryExistsException::new)
                         .put(InvalidPushException.class.getName(), InvalidPushException::new)
                         .put(ReadOnlyException.class.getName(), ReadOnlyException::new)
+                        .put(MirrorException.class.getName(), MirrorException::new)
```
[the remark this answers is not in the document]
ikhoon: I added it mechanically. We may add mirror API in `CentralDogma` client later.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirrorRunner.java
ikhoon: It is difficult to create a unique/incremental ID without an external system. `triggeredTime` can be easily added.

site/src/sphinx/setup-configuration.rst
jrhee17: I see, if you are confident then would it make more sense to just enable by default instead of adding a flag?

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirrorRunner.java
jrhee17: I was imagining a simple UUID inside `MirrorKey`, but I guess just returning `triggeredTime` is fine too

jrhee17 (approved):

site/src/sphinx/setup-configuration.rst
minwoox: I feel it's too verbose if we return the stack trace for all HTTP APIs.
What do you think about only returning the stack trace for specific APIs (e.g. mirroring) when accessed by non-admin users?

minwoox (commented):
Looks great all in all. Left small suggestions. 😉

site/src/sphinx/setup-configuration.rst
ikhoon: Later, the exception could be used to generate the client-side stack trace as we did in Armeria gRPC implementation.
https://github.com/line/armeria/blob/main/grpc/src/main/java/com/linecorp/armeria/internal/server/grpc/AbstractServerCall.java#L571-L575

Fine-tuning is a good idea, but for now I think this option is enough for our use case.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/MirroringServiceV1.java
```diff
@@ -125,11 +133,27 @@ private CompletableFuture<PushResultDto> createOrUpdate(String projectName,
         });
     }
 
+    /**
+     * POST /projects/{projectName}/mirrors/{mirrorId}/run
```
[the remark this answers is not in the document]
ikhoon: Unfortunately, we can't use a path pattern for a custom verb but should use regular expressions. I didn't want to use regex for this sample path pattern.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/AbstractMirror.java
```diff
@@ -61,16 +62,18 @@ public abstract class AbstractMirror implements Mirror {
     private final String remoteBranch;
     @Nullable
     private final String gitignore;
+    @Nullable
```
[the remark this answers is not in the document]
ikhoon: Does that mean we check mirroring settings but do not push the mirrored data to the local repository when `enabled = false`?

I thought users could test a new mirror with `enabled = true & scheduled = disabled` settings.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/MirroringServiceV1.java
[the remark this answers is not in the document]
minwoox: Then, it's fine as is. I might give it a try to get it working one day.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirrorRunner.java
```diff
@@ -0,0 +1,154 @@
...
+                        if (!mirror.enabled()) {
+                            throw new MirrorException("The mirror is disabled: " + mirrorKey);
+                        }
+
+                        return mirror.mirror(workDir, commandExecutor,
+                                             mirrorConfig.maxNumFilesPerMirror(),
+                                             mirrorConfig.maxNumBytesPerMirror());
+                    }, worker);
+            // Remove the inflight request when the mirror task is done.
+            future.handleAsync((unused0, unused1) -> inflightRequests.remove(mirrorKey));
```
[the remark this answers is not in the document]
ikhoon: I didn't use `handle` intentionally because `inflightRequests.remove` could be called recursively inside `inflightRequests.computeIfAbsent()`. I wasn't confident that all supported JDK versions were safe for the operation.


Let me specify `worker` for the job.

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -567,7 +570,6 @@ private Server startServer(ProjectManager pm, CommandExecutor executor,
                                @Nullable SessionManager sessionManager,
                                InternalProjectInitializer projectInitializer) {
         final ServerBuilder sb = Server.builder();
-        sb.verboseResponses(true);
```
ikhoon, line 570: I found out that `verboseResponses()` was always enabled with `ServerBuilder` which may not be recommended because we came to a conclusion that stack traces should not be exposed to non-administrator users.

minwoox:
Is this ready?

ikhoon:
Sorry, I missed the test failure. Fixed it. PTAL.

xds/src/test/java/com/linecorp/centraldogma/xds/k8s/v1/XdsKubernetesServiceTest.java
```diff
@@ -179,7 +180,9 @@ void invalidProperty() throws IOException {
         watcher = watcher(watcherId, "nginx-service", "invalid-credential-id");
         response = createWatcher(watcher, watcherId);
         assertThat(response.status()).isSameAs(HttpStatus.BAD_REQUEST);
-        assertThat(response.contentUtf8()).contains("failed to find credential 'invalid-credential-id'");
+        assertThatJson(response.contentUtf8())
+                .node("grpc-code").isEqualTo("INVALID_ARGUMENT")
+                .node("message").isEqualTo("failed to find credential 'invalid-credential-id' in @xds/meta");
```
ikhoon, line 185: This test failed because `Flags.verborseResponses()` is now disabled by default.
Changed `XdsKubernetesService` to set exception messages as gRPC status descriptions.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/AbstractMirror.java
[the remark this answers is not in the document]
minwoox: Yeah, let's leave it as is and revisit this later if there's any demand for this.

minwoox (approved):
Excellent work! 🎉 🎉 🎉

PR #1045, opened 2024-10-23
Remove the legacy webapp

ikhoon:
Motivation:

The new UI becomes stablized and we don't get addtional bug reports for them. So I think it is safe to remove the legacy webapp to improve build speed and reduce the artifact size.

Modifications:

- Remove `server/src/main/resource` folder
- Remove build scripts for the legacy webapp in `build.gradle`
- Remove `/legacy-webapp` route path from the server.

Result:

The legacy webapp is no longer supported.


ikhoon:
If there are any concerns about deleting the legacy app, let's keep this PR open until we are ready.

minwoox (approved):
I think we are good to go. 👍

PR #1047, opened 2024-10-31
Fix `:dist:copyLib` Task not to copy duplicate libraries

minwoox:
Motivation:
The `:dist:copyLib` task currently copies dependencies into the lib directory from multiple modules. Since each module resolves dependencies independently, it can result in duplicate libraries with differing versions.

Modifications:
- Updated the task to use the dependencies section for centralized dependency resolution.
  - @ikhoon suggested this. 😉 

Result:
- The libraries in the lib directory are now deduplicated.


ikhoon (approved):
🙇‍♂️🙇‍♂️

PR #1049, opened 2024-11-01
Fix a bug where `Created by me` filter fails to render

ikhoon:
Motivation:

If a project is removed, `ProjectDto.creator` is null. `TypeError: Cannot read properties of undefined (reading 'email')` will be raised for the removed project.

Modifications:

- Use null-safe operator (?) to access `projectDto.creator.email`

Result:

`Created by me` filter correctly excludes removed projects.


minwoox (approved):
👍 👍 👍

PR #1050, opened 2024-11-01
Copy JCommander library to distribution directory

minwoox:
I made a mistake in #1047.

### DIFF
--- a/dist/build.gradle
+++ b/dist/build.gradle
@@ -35,6 +35,8 @@ dependencies {
     implementation project(':server-auth:shiro')
     implementation project(':server-mirror-git')
     implementation project(':xds')
+    // Add the optionalImplementation to copy the JAR into the distribution directory.
+    implementation libs.jcommander
 
     // Logging
     runtimeOnly libs.logback12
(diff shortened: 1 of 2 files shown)


ikhoon (approved):

dist/build.gradle
```diff
@@ -35,6 +35,8 @@ dependencies {
     implementation project(':server-auth:shiro')
     implementation project(':server-mirror-git')
     implementation project(':xds')
+    // Add the optionalImplementation to copy the JAR into the distribution directory.
+    implementation libs.jcommander
```
ikhoon, line 39: Because other `optionalImplementation` uses it.
```suggestion
    runtimeOnly libs.jcommander
```

minwoox: Isn't it okay because this is declared in dist module?

ikhoon: It should be okay. I meant `libs.slf4j1.jul.to.slf4j` whose dependency is `optionalImplementation ` in `server` module is declared as `runtimeOnly`.
 https://github.com/line/centraldogma/blob/461fc63efd657d625eb0760742fc54240ebc3d7c/server/build.gradle#L55-L56

PR #1053, opened 2024-11-05
Expose internal projects to users who have `MEMBER` permission

ikhoon:
Motivation:

Currently non-admin users can't access `@xds` project so they need to access the resources only through API. The absence of UI will be quite inconvenient for users. We plan to create UI for xDS, but we don't know when it will be ready.

As a workaround, it might be better to view the xDS confiugrations by browsing UI first and switch to the dedicated UI later. For that I propose to open internal projects to `MEMBER` users.

Modifications:

- Fixed `ProjectApiManager` to contain internal projects if the login user is a member of the internal project.
- Fixed REST API services to inject the current user to `ProjectApiManager` when listing projects.
- As an exception, Thrift API does not allow access to internal projects.

Result:

Administrators now allow users to access the `@xds` project by registering them as members.


minwoox (approved):
Looks great, thanks!

common/src/main/java/com/linecorp/centraldogma/internal/Util.java
```diff
@@ -164,6 +164,29 @@ public static String validateProjectName(String projectName, String paramName) {
...
+    public static boolean isValidProjectName(String projectName, boolean allowInternal) {
+        requireNonNull(projectName, "projectName");
+        if (allowInternal) {
+            return PROJECT_AND_REPO_NAME_PATTERN.matcher(projectName).matches();
+        } else {
+            return USER_INPUT_PROJECT_AND_REPO_NAME_PATTERN.matcher(projectName).matches();
+        }
+    }
+
     public static boolean isValidProjectName(String projectName) {
```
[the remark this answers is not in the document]
ikhoon: I wanted to minimize the changes. Cleaned it up.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/ProjectServiceV1.java
```diff
@@ -189,10 +189,11 @@ public CompletableFuture<Void> purgeProject(@Param String projectName, Author au
     @Consumes("application/json-patch+json")
     @Patch("/projects/{projectName}")
     @RequiresAdministrator
-    public CompletableFuture<ProjectDto> patchProject(@Param String projectName, JsonNode node, Author author) {
+    public CompletableFuture<ProjectDto> patchProject(@Param String projectName, JsonNode node, Author author,
```
jrhee17, line 192: Question) This method updates the metadata and immediately returns the project via `projectApiManager.getProject`.

However, `projectApiManager.getProject` relies on `Project#metadata` which is updated asynchronously.
Were you able to confirm this API will work correctly in a distributed setting? It seems like if a thread reschedule is done in the following line the most recent metadata may not be accessed.

https://github.com/line/centraldogma/blob/4d4af7e350226986d1e3dc2ee3566e44b6f48428/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java#L1525

server/src/main/java/com/linecorp/centraldogma/server/internal/api/GitHttpService.java
```diff
@@ -113,10 +118,10 @@ public HttpResponse advertiseCapability(@Header("git-protocol") @Nullable String
             return HttpResponse.of(HttpStatus.NOT_FOUND, MediaType.PLAIN_TEXT_UTF_8,
                                    "Project not found: " + projectName);
         }
-    if (!projectApiManager.getProject(projectName).repos().exists(repoName)) {
+        if (!projectApiManager.getProject(projectName, user).repos().exists(repoName)) {
```
jrhee17, line 121: Question) Is it possible to just use the user in the thread when possible?  The current context/user seems like a cross cutting concern, and I'm not sure if it is practical to change signatures whenever we need to access a user from a new location.

e.g.
```suggestion
        if (!projectApiManager.getProject(projectName).repos().exists(repoName)) {
```

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/project/ProjectApiManager.java
```diff
@@ -116,14 +129,34 @@ public CompletableFuture<Revision> unremoveProject(String projectName, Author au
...
             throw new IllegalArgumentException("Cannot access " + projectName);
         }
-        return projectManager.get(projectName);
+        if (user.isAdmin()) {
+            return project;
+        }
+        final ProjectMetadata metadata = project.metadata();
+        if (metadata != null) {
+            // Only show internal projects to the members of the project.
+            if (metadata.memberOrDefault(user.id(), null) != null) {
```
jrhee17, line 155: For the record, do you mind explaining your reasoning on why you chose to check this here instead of at `RequiresPermissionDecorator`? Permission related logic is scattered across the code base, and I would like to organize in my mind which checks go where.

jrhee17 (approved):
👍 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/api/GitHttpService.java
ikhoon: I wasn't sure all executors that Central Dogma uses are context-aware. Initially, I was thinking of using thread context, but I also thought there might be different opinions about this. When considering the two patterns, I concluded the explicit one had more advantages.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/ProjectServiceV1.java
ikhoon: Good point. I'll improve `RepositoryListener` to be notified synchronously while a commit is performed in the F/U work.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/project/ProjectApiManager.java
ikhoon: Not all methods are decorated with `RequiresPermissionDecorator`. 
https://github.com/line/centraldogma/blob/c615ba5d4f0216b33e35f5141da54c068fb8b743/server/src/main/java/com/linecorp/centraldogma/server/internal/api/RepositoryServiceV1.java#L82-L84

It depends on `ApiRequestConverter` to get `Project`.
https://github.com/line/centraldogma/blob/53885c45e5ddf4e56eefa7162ea5ca743e711e85/server/src/main/java/com/linecorp/centraldogma/server/internal/api/converter/HttpApiRequestConverter.java#L61-L62

 I agree with your opinion about using `RequiresPermissionDecorator`. Currently `RequiresPermissionDecorator` is used for checking the permission for repositories but not for the members of projects. We may need some refactoring to use `RequiresPermissionDecorator` for project permission. As we will continue to update permissions, I will keep this as a point of improvement.

ikhoon: Please understand that I cannot make changes to the PR right away because I need to deploy this feature to the internal cluster.

Issue #1048, opened 2024-11-01
Refactor Authorization and Project Configurations in Central Dogma


minwoox:
## Summary

The current authorization system in Central Dogma has several limitations and issues regarding security, role permissions, and project management. This proposal addresses these issues by refining roles and permissions, adding visibility options, and improving configuration management.

## Problems and Limitations

1. **Insecure Anonymous Role**  
   - **Current State**: Central Dogma includes an `Anonymous` role that allows repository access without authentication. 
   - **Issue**: Exposing repositories to unauthenticated users is a security risk.
   - **Solution**: Remove the `Anonymous` role entirely to restrict access to logged-in users only.

2. **Ineffective Guest Role and Repository Visibility**  
   - **Current State**: The `Guest` role can have read or write permissions, which is not ideal for permissions management. It can make sensitive information vulnerable.
   - **Issue**: The existence of a `Guest` role with write permissions is unusual, and it limits options for repository visibility.
   - **Solution**: Eliminate the `Guest` role. Introduce a visibility option (`public` or `private`) at the repository level:
     - **Private repositories**: Only accessible to authenticated users with explicit permissions.
     - **Public repositories**: Available to all authenticated users for read access.

3. **Limited Project-Level Permissions**  
   - **Current State**: A project in Central Dogma has multiple repositories, but project owners cannot easily assign permissions to individual repositories within the project. Currently, a user must be registered as a project member to access specific repositories.
   - **Issue**: This constraint makes it challenging to manage permissions across multiple repositories in a single project.
   - **Solution**: Implement fine-grained repository-level permissions that do not require project-level membership. This will allow project owners to control permissions for each repository independently.

4. **Lack of Team Structure**  
   - **Current State**: All permissions must be set individually, as Central Dogma lacks team-based access control.
   - **Issue**: Managing access at an individual level can be tedious and error-prone, especially for large teams.
   - **Solution**: Introduce team-based access control, allowing permissions to be assigned to teams instead of individual users.

5. **Terminology Alignment with Industry Standards**  
   - **Current State**: Central Dogma uses `Admin` to describe the Central Dogma cluster owner, but GitHub and similar platforms use `Admin` for repository ownership, which may cause confusion.
   - **Issue**: Users may confuse Central Dogma cluster-level administration with repository-level ownership.
   - **Solution**: Rename `Admin` (the term used for cluster owners) to something clearer and more intuitive. **Proposed name**: `ClusterAdmin`.

[...]

PR #1054, opened 2024-11-06
Rename Admin Role to SystemAdmin for Clarity

minwoox:
Motivation:
The term "Admin" does not clearly convey its purpose as a system-wide administrator role, as discussed in [issue #1048](https://github.com/line/centraldogma/issues/1048). Renaming this role improves clarity for users and developers.

Modifications:
- Renamed all "Admin" to "SystemAdmin" across the codebase.
- Updated `MetadataService.updateTokenLevel()` to avoid directly accessing the `admin` property.

Result:
- The role name "Admin" has been replaced with "SystemAdmin".
- (Breaking)
  - `authentication.administrators` in dogma.json is now `authentication.systemAdministrators`.
- (Deprecation)
  - Use `systemAdmin` property instead of `admin` when creating a token via REST API:
    ```json
    {"appId": "foo", "isSystemAdmin": true, ...}
     ```
  - Use `SYSTEMADMIN` instead of `admin` when changing the token level:
    ```json
    {"level":"SYSTEMADMIN"}
    ```

### DIFF
--- a/common/src/main/java/com/linecorp/centraldogma/common/RedundantChangeException.java
+++ b/common/src/main/java/com/linecorp/centraldogma/common/RedundantChangeException.java
@@ -16,37 +16,33 @@
 
 package com.linecorp.centraldogma.common;
 
+import static java.util.Objects.requireNonNull;
+
+import javax.annotation.Nullable;
+
 /**
  * A {@link CentralDogmaException} that is raised when attempted to push a commit without effective changes.
  */
 public class RedundantChangeException extends CentralDogmaException {
 
     private static final long serialVersionUID = 8739464985038079688L;
-
-    /**
-     * Creates a new instance.
-     */
-    public RedundantChangeException() {}
+    @Nullable
+    private final Revision headRevision;
 
     /**
      * Creates a new instance.
      */
     public RedundantChangeException(String message) {
         super(message);
+        headRevision = null;
     }
 
     /**
      * Creates a new instance.
      */
-    public RedundantChangeException(Throwable cause) {
-        super(cause);
-    }
-
-    /**
-     * Creates a new instance.
-     */
-    public RedundantChangeException(String message, Throwable cause) {
-        super(message, cause);
+    public RedundantChangeException(Revision headRevision, String message) {
+        super(message);
+        this.headRevision = requireNonNull(headRevision, "headRevision");
     }
 
     /**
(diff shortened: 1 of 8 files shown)


server/src/main/java/com/linecorp/centraldogma/server/command/AbstractCommandExecutor.java
```diff
@@ -126,7 +126,7 @@ public final <T> CompletableFuture<T> execute(Command<T> command) {
         if (!isStarted()) {
             throw new ReadOnlyException("running in read-only mode. command: " + command);
         }
-        if (!writable && !(command instanceof AdministrativeCommand)) {
+        if (!writable && !(command instanceof SystemAdministrativeCommand)) {
```
jrhee17, line 129: Double checked that the superclass `[System]AdministrativeCommand` isn't exposed in the JSON in any way

server/src/main/java/com/linecorp/centraldogma/server/metadata/Token.java
```diff
@@ -122,11 +125,11 @@ public String secret() {
     }
 
     /**
-     * Returns whether this token has administrative privileges.
+     * Returns whether this token has system administrative privileges.
      */
     @JsonProperty
-    public boolean isAdmin() {
-        return isAdmin;
+    public boolean isSystemAdmin() {
```
jrhee17, line 131: Note: 
1. While releasing, if a user tries to access a token via UI served by old servers to a new server he/she may encounter errors.
2. If a new token is registered while releasing, old servers may not be able to access it
3. Regarding rollback plan: if a new token is registered we need to clean up the token before performing rollback. Having said this, I'm not sure if there is an easy way to detect this at the moment.

jrhee17 (approved):
👍 👍

server/src/main/java/com/linecorp/centraldogma/server/metadata/Token.java
minwoox: > While releasing, if a user tries to access a token via UI served by old servers to a new server he/she may encounter errors.

That's correct. 👍 We need to let them refresh the page after finishing the rolling restart.

> If a new token is registered while releasing, old servers may not be able to access it

That's correct. 👍 I thought it should be okay, but it probably shouldn't be. 😓 
Let me create a separate commit to solve this issue. 

> Regarding rollback plan:

I'm implementing a client that migrates `admin` to `systemAdmin`.
Let me also implement the client the vise-versa.

minwoox: @jrhee17 Here it is. 😉 
https://github.com/line/centraldogma/pull/1056

ikhoon (approved):
👍

PR #1055, opened 2024-11-07
Enable `respect_dns_ttl` by default for xDS cluster

ikhoon:
Motivation:

Envoy refreshes DNS every 5 seconds regardless of DNS TTL. The frequent refresh consumes additional CPU and memory resources and may put a heavy load on DNS servers.

`respect_dns_ttl` is an option to let Envoy use the TTL of DNS respones. https://github.com/envoyproxy/envoy/issues/6876
Respecting DNS TTL seems more sensible behavior for Armeria xDS modules.

Modifications:

- Set `respect_dns_ttl` to true when creating an xDS cluster.
  - Due to the limation of protobuf, `respect_dns_ttl` in a request is ignored but the value still can be modified via xDS update API.

Result:

`respect_dns_ttl` is now enabled by default for xDS cluster.

### DIFF
--- a/xds/src/main/java/com/linecorp/centraldogma/xds/cluster/v1/XdsClusterService.java
+++ b/xds/src/main/java/com/linecorp/centraldogma/xds/cluster/v1/XdsClusterService.java
@@ -64,10 +64,20 @@ public final class XdsClusterService extends XdsClusterServiceImplBase {
         }
 
         final String clusterName = parent + CLUSTERS_DIRECTORY + clusterId;
-        // Ignore the specified name in the cluster and set the name
-        // with the format of "groups/{group}/clusters/{cluster}".
-        // https://github.com/aip-dev/google.aip.dev/blob/master/aip/general/0133.md#user-specified-ids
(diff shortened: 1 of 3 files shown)


minwoox:
I’m not sure if modifying a user-created cluster without notice is a good approach. Could we guide users to specify the value when creating the cluster instead?

jrhee17 (approved):

ikhoon:
> Getting record ttl is difficult. Envoy uses c-ares to query DNS. In current repository, ares_gethostbyname is used for resolving name. Unfortunately, ares_gethostbyname doesn't provide record ttl information to callback function.

https://github.com/envoyproxy/envoy/issues/6876#issuecomment-492507307

I thought it was a defect of Envoy which couldn't acquire DNS TTL in its implementation. Armeria DNS resolver respects DNS TTL which is an industrial standard.
Although it is the current Envoy's behavior I didn't see we should follow the non-standard specification.

However, I am not strong in this direction because I can also guide users in setting up the property.

minwoox:
That makes sense. Thanks for the explanation. 👍

minwoox (approved):
👍 👍 👍

PR #1056, opened 2024-11-08
Prepare Token Updates for #1054 Compatibility

minwoox:
Motivation
To ensure smooth compatibility with #1054, the `Token` structure needs an update to support changes in `admin` property.

Modifications:
- Enhanced the Token deserializer to accept both `admin` and `systemAdmin` properties, ensuring backward compatibility.
- Updated `MetadataService.updateTokenLevel()` to avoid direct access to the `admin` property.

Result:
- The `Token` structure now supports the upcoming changes, allowing seamless updates with #1054.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java
@@ -57,6 +57,7 @@ import com.linecorp.centraldogma.internal.jsonpatch.ReplaceOperation;
 import com.linecorp.centraldogma.internal.jsonpatch.TestAbsenceOperation;
 import com.linecorp.centraldogma.server.QuotaConfig;
 import com.linecorp.centraldogma.server.command.CommandExecutor;
+import com.linecorp.centraldogma.server.internal.admin.service.TokenNotFoundException;
 import com.linecorp.centraldogma.server.storage.project.Project;
 import com.linecorp.centraldogma.server.storage.project.ProjectManager;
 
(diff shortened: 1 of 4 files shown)


server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java
```diff
@@ -1011,14 +1012,25 @@ public CompletableFuture<Revision> updateTokenLevel(Author author, String appId,
...
+                                                 final Tokens tokens0 = tokens.object();
+                                                 final Token token = tokens0.get(appId);
+                                                 if (token == null) {
+                                                     throw new TokenNotFoundException("App ID: " + appId);
+                                                 }
+                                                 if (toBeAdmin == token.isAdmin()) {
+                                                     throw new IllegalArgumentException(
+                                                             "The token is already " +
+                                                             (toBeAdmin ? "admin" : "user"));
+                                                 }
```
ikhoon, line 1024: Question) The serialized format of `tokens.json` seems the same. Did you change to validate this? I am asking because this could override the previous state for multiple updates. For example, one deprecates a token, and another makes this token admin. This may not happen in the real world as the permission is only granted to admins.

minwoox: > The serialized format of tokens.json seems the same.

That is true except the token now deserializes from both: `{"appId": "..", "admin": true, ...}` and `{"appId": "..", "systemAdmin": true, ...}`. This is the preliminary work for #1054.
Let's imagine that we rolling-update with #1054. When a new structured token (with `systemAdmin`) is committed, the old servers will fail to read the token because it cannot recognize `systemAdmin` property.

ikhoon: I understood that the JSON pointer to `/appIds/{appId}/admin` should be `/appIds/{appId}/systemAdmin` when #1054 is merged.

ikhoon (approved):

server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java
minwoox: After this PR gets merged, the JSON pointer does not point `/appIds/{appId}/admin` anymore. It replaces the whole `Token` at `/appIds/{appId}`. 😉

jrhee17 (approved):
👍 👍 👍

PR #1057, opened 2024-11-11
Provide a way to listen to the events of mirroring

ikhoon:
Motivation:

Mirroring failure is only recorded as a warning, but there is a need to handle it in a different way. For example, it can be recorded in metrics or end-users can be notified immediately.

`MirrorListener` is provided as an extension point to utilize various events occurring in the mirror.

Modifications:

- Introduce `MirrorListener` whose implementations can be loaded dynamically via Java SPI.
- `onStart()`, `onComplete()` and `onError()` events are added.
- The default behavior is preserved in `DefaultMirrorListener` which is only used when no custom `MirrorListener` is configured.

Result:

You can now use `MirrorListener` to listen to `Mirror` events.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirrorSchedulingService.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirrorSchedulingService.java
@@ -22,6 +22,7 @@ import static java.util.Objects.requireNonNull;
 
 import java.io.File;
 import java.time.Duration;
+import java.time.Instant;
 import java.time.ZonedDateTime;
 import java.util.List;
 import java.util.ServiceLoader;
@@ -52,10 +53,11 @@ import com.google.common.util.concurrent.MoreExecutors;
 import com.linecorp.centraldogma.common.MirrorException;
 import com.linecorp.centraldogma.server.MirroringService;
 import com.linecorp.centraldogma.server.command.CommandExecutor;
+import com.linecorp.centraldogma.server.metadata.User;
 import com.linecorp.centraldogma.server.mirror.Mirror;
 import com.linecorp.centraldogma.server.mirror.MirrorListener;
 import com.linecorp.centraldogma.server.mirror.MirrorResult;
-import com.linecorp.centraldogma.server.storage.project.Project;
+import com.linecorp.centraldogma.server.mirror.MirrorTask;
 import com.linecorp.centraldogma.server.storage.project.ProjectManager;
 
 import io.micrometer.core.instrument.MeterRegistry;
@@ -66,7 +68,7 @@ public final class MirrorSchedulingService implements MirroringService {
 
     private static final Logger logger = LoggerFactory.getLogger(MirrorSchedulingService.class);
 
-    private static final MirrorListener mirrorListener;
+    static final MirrorListener mirrorListener;
 
     static {
         final List<MirrorListener> listeners =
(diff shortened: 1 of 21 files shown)


minwoox (approved):
Looks great! 👍 👍 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirrorSchedulingService.java
```diff
@@ -62,6 +66,19 @@ public final class MirrorSchedulingService implements MirroringService {
 
     private static final Logger logger = LoggerFactory.getLogger(MirrorSchedulingService.class);
 
+    private static final MirrorListener mirrorListener;
```
jrhee17, line 69: Note) I understood that `MirrorListener` isn't invoked for manually triggered mirrors from `MirrorServiceV1`.

While this may make sense in terms of our purpose (monitoring), it may be confusing since a `Mirror` has been done yet `MirrorListener` isn't invoked.

server/src/main/java/com/linecorp/centraldogma/server/mirror/MirrorListener.java
```diff
@@ -0,0 +1,50 @@
...
+ *     <li>{@link #onError(Mirror, Throwable)}: Logs a warning message for the error of {@link Mirror}.</li>
+ * </ul>
+ */
+@Nullable
+public interface MirrorListener {
+
+    /**
+     * Invoked when the {@link Mirror} operation is started.
+     */
+    void onStart(Mirror mirror);
```
jrhee17, line 39: Optional; If a `MirrorTask` is passed at `onStart` which contains a `mirrorId`, users may be able to differentiate the flow of a single `Mirror` across multiple callbacks via `mirrorId`. This may be useful since multiple mirrors may be run in a single leader with the introduction of zone-aware mirroring.

jrhee17 (approved):

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirrorSchedulingService.java
ikhoon: Good point. Let me `scheduled` flag to methods of `MirrorListener`.

server/src/main/java/com/linecorp/centraldogma/server/mirror/MirrorListener.java
ikhoon: `mirrorId` means the ID for a single mirroring task, not `Mirror#id()`, right?

> This may be useful since multiple mirrors may be run in a single leader with the introduction of zone-aware mirroring.

Sorry, I didn't understand this. Did you mean multiple mirrors can be run concurrently? Could you elaborate on this, please?

server/src/main/java/com/linecorp/centraldogma/server/mirror/MirrorListener.java
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * Invoked when the {@link Mirror} operation is completed.
+     */
+    void onComplete(Mirror mirror, MirrorResult result);
+
+    /**
+     * Invoked when the {@link Mirror} operation is failed.
+     */
+    void onError(Mirror mirror, Throwable cause);
```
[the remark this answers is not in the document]
ikhoon: Fixed to pass `MirrorTask` which contains those properties.

PR #1059, opened 2024-11-12
Move `Delete Project` button to `Danger zone` tab

ikhoon:
Motivation:

We got two feedbacks about UX:

- `Delete Project`'s location was inappropriate.

  > After I click on Project Settings, the button is immediately replaced
  > with Delete Project, while I'm still hovering over it.

- Automatically focus to the project seach box on the welcome page.

  > would it be possible to add something like auto-focus to the
  > "Search project..." field when opening the web UI?
  > Perhaps also supporting the / key would help, for consistency with
  > the Projects page

Modifications:

- Create `Danger zone` tab and move `Delete Project` to it.
- Miscellaneous)
  - Remove `repos` segment from all `Breadcrumb`
    - I felt it was unnecessary to include `repos` since it is assumed to be a repo without additional explanation.
  - Fix to auto-focus to the search box on the welcome page.
  - Fix broken APIs in the mock server.

Result:

Improved usability of the UI.


jrhee17 (approved):

minwoox (approved):
👍 👍 👍

PR #1060, opened 2024-11-14
Introduce Role-Based Authentication for Repository Management

minwoox:
Motivation:
To address the migration of mirroring configurations from projects to repositories and to simplify access management,
we decided to:
- Hide the meta repository.
- Replace permission-based authentication with a role-based system using `RepositoryRole` (`READ`, `WRITE`, `ADMIN`). The `ADMIN` role has access to the configurations.

This change resolves the issue where users with `WRITE` permission for the meta repository were only allowed to creating mirroring configurations.

With `RepositoryRole`, access becomes more structured and extensible for future enhancements (e.g., introducing custom roles). While permission-based authentication is replaced here, this does not preclude the future coexistence of role-based and permission-based systems.

Modifications:
- Added `RepositoryRole` with hierarchical roles (`READ`, `WRITE`, `ADMIN`).
- Replaced `RequiresPermission` annotations with `RequiresRepositoryRole`.
- Renamed `RequiresRole` to `RequiresProjectRole`.
  - Updated it to accept a single `ProjectRole`, utilizing the hierarchical model to avoid redundant role specifications.
- Removed APIs for managing permissions, replacing them with role management APIs.

Result:
- Role-based access simplifies management and aligns with repository-specific mirroring configurations.
- (Breaking) APIs for managing permissions are removed.

To-do:
- Update the documentation to reflect the new systme, including updated screenshots

Migration plan:
- Deploy [PR](https://github.com/line/centraldogma/pull/1061), which supports deserialization of both legacy and new metadata format.
- Deploy intermediate commit supporting both permission and role APIs, ensuring metadata is stored in the new format.
  - The commit also migrate the legacy format to new format.
- Deploy this commit, which exclusively supports role-based APIs.

PR #1061, opened 2024-11-14
Prepare RepositoryMetadata role Updates for #1060 Compatibility

minwoox:
Motivation
To ensure smooth compatibility with #1060, the `RepositoryMetadata` structure needs an update to support changes.
AS-IS:
```json
{
  "name": "minu-test",
  "perRolePermissions": {
    "owner": [
      "READ",
      "WRITE"
    ],
    "member": ["READ"],
    "guest": []
  },
  "perUserPermissions": {
    "foo@dogma.com": [
      "READ"
    ],
    "bar@dogma.com": [
      "READ",
      "WRITE"
    ]
  },
  "perTokenPermissions": {
    "goodman": [
      "READ"
    ]
  },
  "creation": {
    "user": "minu.song@dogma.com",
    "timestamp": "2024-08-19T02:47:23.370762417Z"
  }
}
```

TO-BE:
```json
{
  "name": "minu-test",
  "roles": {
    "projects": {
      "member": "READ",
      "guest": null
    }
    "users": {
      "foo@dogma.com": "READ",
      "bar@dogma.com": "WRITE"
    },
    "tokens": {
      "goodman": "READ"
    }
  },
  "creation": {
    "user": "minu.song@dogma.com",
    "timestamp": "2024-08-19T02:47:23.370762417Z"
  }
}
```

Modifications:
- Enhanced the `RepositoryMetadata` deserializers to accept both legacy and new formats.
- Added `RepositoryRole`.
  - It will be used for repository instead of `Permission`.
- Removed `MetadataApiService.updateSpecificUserPermission` and `updateSpecificTokenPermission`.
  - They are not used in the UI and we don't even have test cases for that.
  - Will add those APIs later with better API design when we need it.

Result:
- The `RepositoryMetadata` structure now supports the upcoming changes in #1060.


server/src/main/java/com/linecorp/centraldogma/server/metadata/RepositoryMetadataDeserializer.java
```diff
@@ -0,0 +1,78 @@
...
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.deser.std.StdDeserializer;
+import com.google.common.collect.ImmutableMap;
+
+import com.linecorp.centraldogma.internal.Jackson;
+import com.linecorp.centraldogma.server.QuotaConfig;
+
+final class RepositoryMetadataDeserializer extends StdDeserializer<RepositoryMetadata> {
```
jrhee17, line 33: I understood these custom deserializers will be removed once migration is complete

jrhee17 (approved):
updateSpecificUserPermissionupdateSpecificUserPermission

server/src/main/java/com/linecorp/centraldogma/server/metadata/RepositoryMetadataDeserializer.java
minwoox: That is correct. 👍 👍

ikhoon (approved):
👍👍

server/src/main/java/com/linecorp/centraldogma/server/metadata/Roles.java
```diff
@@ -0,0 +1,83 @@
...
+        this.projectRoles = requireNonNull(projectRoles, "projectRoles");
+        this.users = requireNonNull(users, "users");
+        this.tokens = requireNonNull(tokens, "tokens");
+    }
+
+    /**
+     * Returns the {@link ProjectRoles} of the repository.
+     */
+    @JsonProperty("projects")
+    public ProjectRoles projectRoles() {
```
jrhee17, line 55: I understood this property as: the  default `RepositoryRole` that will be assigned to project members that aren't explicitly set by `users` and `tokens`.

server/src/main/java/com/linecorp/centraldogma/server/metadata/RepositoryMetadataDeserializer.java
```diff
@@ -0,0 +1,106 @@
...
+    RepositoryMetadataDeserializer() {
+        super(RepositoryMetadata.class);
+    }
+
+    @Override
+    public RepositoryMetadata deserialize(JsonParser p, DeserializationContext ctxt)
+            throws IOException {
+        final JsonNode jsonNode = p.readValueAsTree();
+        final String name = jsonNode.get("name").textValue();
+        final JsonNode perRolePermissionsNode = jsonNode.get("perRolePermissions");
```
jrhee17, line 55: Note: I understood that if both `perRolePermissions` and `roles` is in the json node, then `perRolePermissions` (the legacy format) has higher priority.

common/src/main/java/com/linecorp/centraldogma/common/RepositoryRole.java
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * Able to read a file from a repository.
+     */
+    READ,
+    /**
+     * Able to write a file to a repository.
+     */
+    WRITE,
+    /**
+     * Able to manage a repository.
```
jrhee17, line 33: Compared with the current permission scheme, I understood the difference is:
1) the ability to set ADMIN for a repository
2) repository settings can't change project admin roles.

jrhee17 (approved):
Understood the new changes

server/src/main/java/com/linecorp/centraldogma/server/metadata/Roles.java
minwoox: > the default RepositoryRole that will be assigned to project members

That's correct. 👍 

> that aren't explicitly set by users and tokens.

That is not correct. A `RepositoryRole` can still be assigned to a user. The `RepositoryRole`, which operates at a higher level, will be applied in such situation. https://github.com/line/centraldogma/pull/1060/files#diff-dc8a031ec8b91f16d76b42e152e9c3be8dd07b6f77bea922030074acc78ce78eR800-R823

server/src/main/java/com/linecorp/centraldogma/server/metadata/RepositoryMetadataDeserializer.java
minwoox: >  I understood that if both perRolePermissions and roles is in the json node

They cannot be. Let me ass an assertion logic. 😉

minwoox:
This PR will be ready after https://github.com/line/centraldogma/pull/1069 is merged.

server/src/main/java/com/linecorp/centraldogma/server/metadata/RepositorySupport.java
```diff
@@ -116,16 +114,7 @@ CompletableFuture<Revision> push(String projectName, String repoName,
 
         return executor.execute(Command.transform(null, author, projectName, repoName, Revision.HEAD,
                                                   commitSummary, "", Markup.PLAINTEXT, transformer))
-                       .thenApply(CommitResult::revision)
-                       .exceptionally(cause -> {
-                           final Throwable peeled = Exceptions.peel(cause);
-                           if (peeled instanceof RedundantChangeException) {
```
minwoox, line 122: I've realized that we don't need this logic anymore after introducing the transformer in #1069 because it will raise an IllegalArgumentException if there's no change.

server/src/main/java/com/linecorp/centraldogma/server/metadata/ProjectRoles.java
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * Represents the roles assigned to project members and guests for a specific repository.
+ */
+public final class ProjectRoles {
+
+    /**
+     * Returns a new {@link ProjectRoles} with the specified {@link RepositoryRole}s.
+     */
+    public static ProjectRoles of(@Nullable RepositoryRole member, @Nullable RepositoryRole guest) {
+        return new ProjectRoles(member, guest);
```
ikhoon, line 34: nit: Create `EMTPY_PROJECT_ROLES` and reuse if both `member` and `guest` are null?

server/src/main/java/com/linecorp/centraldogma/server/metadata/RepositorySupport.java
ikhoon: I'm not sure `IllegalArgumentException` is a proper exception to understand the context.

server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java
```diff
@@ -471,16 +471,48 @@ public CompletableFuture<Revision> updatePerRolePermissions(Author author,
...
-                                                     "/perRolePermissions");
-        final Change<JsonNode> change =
-                Change.ofJsonPatch(METADATA_JSON,
-                                   new ReplaceOperation(path, Jackson.valueToTree(perRolePermissions))
-                                           .toJsonNode());
         final String commitSummary =
                 "Update the role permission of the '" + repoName + "' in the project '" + projectName + '\'';
-        return metadataRepo.push(projectName, Project.REPO_DOGMA, author, commitSummary, change);
+        final ContentTransformer<JsonNode> transformer = new ContentTransformer<>(
+                METADATA_JSON, EntryType.JSON, node -> {
```
ikhoon, line 477: Optional) I could see some code similar to this in other places. How about creating `RepositoryMetadataTransformer` and duplicating them?
```java
private static class RepositoryMetadataTransformer extends ContentTransformer {
    
   RepositoryMetadataTransformer(Function<RepositoryMetadata, RepositoryMetadata>) {
       ...
   }
}
```

server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java
```diff
@@ -471,16 +471,48 @@ public CompletableFuture<Revision> updatePerRolePermissions(Author author,
...
+            final ProjectMetadata projectMetadata = projectMetadata(node);
+            final RepositoryMetadata repositoryMetadata = projectMetadata.repo(repoName);
+            assert repositoryMetadata.name().equals(repoName);
+            if (repositoryMetadata.perRolePermissions().equals(perRolePermissions)) {
+                throw new IllegalArgumentException(
```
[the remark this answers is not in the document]
minwoox: That's a good suggestion! How about the following approach:  
- Introduce `UserNotFoundException`
  - This exception would be used exclusively on the server side, similar to `TokenNotFoundException`.  
- Raise `UserNotFoundException` and `TokenNotFoundException`
  - These exceptions would be triggered when attempting to update an entity that doesn't exist.  
- Raise `RedundantChangeException`
  - This exception would be raised when an update is attempted with the same content as the existing one.  
  - The `RedundantChangeException` would be swallowed, and the system would simply return the head revision. In other words, users wouldn’t encounter this exception."

server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java
minwoox: That's a good suggestion. 👍 
`RepositoryMetadataTransformer`, `ProjectMetadataTransformer` and `TokensTransformer ` are added. 😉

server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java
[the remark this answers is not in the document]
minwoox: Updated, PTAL. 🙇

server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java
```diff
@@ -968,35 +1067,38 @@ public CompletableFuture<Revision> destroyToken(Author author, String appId) {
...
+    private static Map<String, Token> newAppIds(Tokens tokens, String appId, Token newToken) {
+        final ImmutableMap.Builder<String, Token> appIdsBuilder =
+                ImmutableMap.builderWithExpectedSize(tokens.appIds().size());
+        for (Entry<String, Token> entry : tokens.appIds().entrySet()) {
+            if (!entry.getKey().equals(appId)) {
+                appIdsBuilder.put(entry);
+            } else {
+                appIdsBuilder.put(appId, newToken);
+            }
+        }
```
ikhoon, line 1098: Side note: It seems useful if we have an `ImmutableMap` utility that helps to update or remove.
```java
class ImmutableMapUtil {
    Map<K, V> remove(Map<K, V> old, K key) { 
      // create a new map and remove `key`.
      ... 
    }

    Map<K, V> add(Map<K, V> old, K key, V value) { ... }
}
```

ikhoon (approved):
Thanks! 👍

server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java
minwoox: That looks useful. 👍 Let me try that later. 😉

PR #1062, opened 2024-11-18
Provide a way to pin a mirror to a zone

ikhoon:
Motivation:

I propose running the mirror in a zone close to the git server or only in zones that are allowed access.

Modifications:

- Add `zone` as an optional property to mirror configurations.
- Add `GET /api/v1/mirror/config` API to provide zone-related configurations.
  - The API is used to select a pinned zone on the mirror form.
- Add `zonePinned` flag to `MirroringServicePluginConfig` to enable zone-pinned mirroring.
  - The option is disabled by default.
  - `DefaultMirroringServicePlugin.target()` returns `PluginTarget.ZONE_LEADER_ONLY` if `zonePinned == true`
  - Fixed `MirrorSchedulingService` to only run a pinned zone if `zonePinned == true`
- Add `ZoneConfig` to replace `zone: string` configuration.
  - `allZones` is newly added to specify the list of zone names.
- Breaking) Make `Plugin.target()` take `CentralDogmaConfig` as an argument.

Result:

You can now specify a zone where you want to perform mirroring.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/PluginGroup.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/PluginGroup.java
@@ -108,8 +108,8 @@ final class PluginGroup {
                          .collect(toImmutableMap(Entry::getKey, e -> {
                              final PluginTarget target = e.getKey();
                              final List<Plugin> targetPlugins = e.getValue();
-                             final String poolName = "plugins-for-" + target.name().toLowerCase().replace("_",
-                                                                                                          "-");
+                             final String poolName =
+                                     "plugins-for-" + target.name().toLowerCase().replace("_", "-");
                              return new PluginGroup(targetPlugins,
                                                     Executors.newSingleThreadExecutor(
                                                             new DefaultThreadFactory(poolName, true)));
--- a/server/src/main/java/com/linecorp/centraldogma/server/PluginGroup.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/PluginGroup.java
@@ -16,10 +16,12 @@
 package com.linecorp.centraldogma.server;
 
 import static com.google.common.collect.ImmutableList.toImmutableList;
+import static com.google.common.collect.ImmutableMap.toImmutableMap;
 import static java.util.Objects.requireNonNull;
 
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.ServiceLoader;
 import java.util.concurrent.CompletableFuture;
@@ -27,12 +29,14 @@ import java.util.concurrent.CompletionStage;
 import java.util.concurrent.Executor;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.stream.Collectors;
 
 import javax.annotation.Nullable;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Iterables;
@@ -63,30 +67,28 @@ final class PluginGroup {
      *
      * @param target the {@link PluginTarget} which would be loaded
      */
+    @VisibleForTesting
     @Nullable
     static PluginGroup loadPlugins(PluginTarget target, CentralDogmaConfig config) {
-        return loadPlugins(PluginGroup.class.getClassLoader(), target, config, ImmutableList.of());
+        return loadPlugins(PluginGroup.class.getClassLoader(), config, ImmutableList.of()).get(target);
     }
 
     /**
      * Returns a new {@link PluginGroup} which holds the {@link Plugin}s loaded from the classpath.
-     * {@code null} is returned if there is no {@link Plugin} whose target equals to the specified
+     * An empty map is returned if there is no {@link Plugin} whose target equals to the specified
      * {@code target}.
      *
      * @param classLoader which is used to load the {@link Plugin}s
-     * @param target the {@link PluginTarget} which would be loaded
      */
-    @Nullable
-    static PluginGroup loadPlugins(ClassLoader classLoader, PluginTarget target, CentralDogmaConfig config,
-                                   List<Plugin> plugins) {
+    static Map<PluginTarget, PluginGroup> loadPlugins(ClassLoader classLoader, CentralDogmaConfig config,
+                                                      List<Plugin> plugins) {
         requireNonNull(classLoader, "classLoader");
-        requireNonNull(target, "target");
         requireNonNull(config, "config");
 
         final ServiceLoader<Plugin> loader = ServiceLoader.load(Plugin.class, classLoader);
         final ImmutableMap.Builder<Class<?>, Plugin> allPlugins = new ImmutableMap.Builder<>();
         for (Plugin plugin : Iterables.concat(plugins, loader)) {
-            if (target == plugin.target() && plugin.isEnabled(config)) {
+            if (plugin.isEnabled(config)) {
                 allPlugins.put(plugin.configType(), plugin);
             }
         }
@@ -94,11 +96,31 @@ final class PluginGroup {
         // IllegalArgumentException is thrown if there are duplicate keys.
         final Map<Class<?>, Plugin> pluginMap = allPlugins.build();
         if (pluginMap.isEmpty()) {
-            return null;
+            return ImmutableMap.of();
         }
 
-        return new PluginGroup(pluginMap.values(), Executors.newSingleThreadExecutor(new DefaultThreadFactory(
-                "plugins-for-" + target.name().toLowerCase().replace("_", "-"), true)));
+        final Map<PluginTarget, PluginGroup> pluginGroups =
+                pluginMap.values()
+                         .stream()
+                         .collect(Collectors.groupingBy(plugin -> plugin.target(config)))
+                         .entrySet()
+                         .stream()
+                         .collect(toImmutableMap(Entry::getKey, e -> {
+                             final PluginTarget target = e.getKey();
+                             final List<Plugin> targetPlugins = e.getValue();
+                             final String poolName = "plugins-for-" + target.name().toLowerCase().replace("_",
+                                                                                                          "-");
+                             return new PluginGroup(targetPlugins,
+                                                    Executors.newSingleThreadExecutor(
+                                                            new DefaultThreadFactory(poolName, true)));
+                         }));
+
+        pluginGroups.forEach((target, group) -> {
+            logger.debug("Loaded plugins for target {}: {}", target,
+                         group.plugins().stream().map(plugin -> plugin.getClass().getName())
+                              .collect(Collectors.toList()));
+        });
+        return pluginGroups;
     }
 
     private final List<Plugin> plugins;
@@ -119,9 +141,10 @@ final class PluginGroup {
     /**
      * Returns the first {@link Plugin} of the specified {@code clazz} as wrapped by an {@link Optional}.
      */
-    <T extends Plugin> Optional<T> findFirstPlugin(Class<T> clazz) {
+    @Nullable
+    <T extends Plugin> T findFirstPlugin(Class<T> clazz) {
         requireNonNull(clazz, "clazz");
-        return plugins.stream().filter(clazz::isInstance).map(clazz::cast).findFirst();
+        return plugins.stream().filter(clazz::isInstance).map(clazz::cast).findFirst().orElse(null);
     }
 
     /**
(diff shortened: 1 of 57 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirrorSchedulingService.java
```diff
@@ -209,9 +221,21 @@ private void schedulePendingMirrors() {
...
                                   continue;
                               }
+                              if (zoneConfig != null) {
+                                  String pinnedZone = m.zone();
+                                  if (pinnedZone == null) {
+                                      // Use the first zone if the mirror does not specify a zone.
+                                      pinnedZone = zoneConfig.allZones().get(0);
+                                  }
+                                  if (!pinnedZone.equals(currentZone)) {
+                                      // Skip the mirror if the zone does not match.
```
jrhee17, line 231: Question) Would it be worth checking if the `pinnedZone` is contained in `ZoneConfig#allZones` to detect ill-configured mirrors?

server/src/main/java/com/linecorp/centraldogma/server/PluginGroup.java
```diff
@@ -63,42 +67,60 @@ final class PluginGroup {
...
+     * An empty map is returned if there is no {@link Plugin} whose target equals to the specified
      * {@code target}.
      *
      * @param classLoader which is used to load the {@link Plugin}s
-     * @param target the {@link PluginTarget} which would be loaded
      */
-    @Nullable
-    static PluginGroup loadPlugins(ClassLoader classLoader, PluginTarget target, CentralDogmaConfig config,
-                                   List<Plugin> plugins) {
+    static Map<PluginTarget, PluginGroup> loadPlugins(ClassLoader classLoader, CentralDogmaConfig config,
```
jrhee17, line 83: Note: Understood that with this change, we don't allow the same plugin to be added with different `PluginTarget`s (i.e. `PluginA+LEADER_ONLY`, `PluginA+ZONE_LEADER_ONLY`)

jrhee17 (approved):
Left only nits but looks good to me 👍 👍

xds/src/main/java/com/linecorp/centraldogma/xds/internal/ControlPlanePlugin.java
```diff
@@ -67,8 +68,8 @@ public Class<?> configType() {
     @Override
     public String toString() {
         return MoreObjects.toStringHelper(this)
-                          .add("configType", configType())
-                          .add("target", target())
+                          .add("configType", configType().getName())
+                          .add("target", PluginTarget.LEADER_ONLY)
```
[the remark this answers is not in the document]
ikhoon: Oops... 😱

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirroringServicePlugin.java
```diff
@@ -97,8 +127,9 @@ public MirrorSchedulingService mirroringService() {
     @Override
     public String toString() {
         return MoreObjects.toStringHelper(this)
-                          .add("configType", configType())
-                          .add("target", target())
+                          .omitNullValues()
+                          .add("configType", configType().getName())
+                          .add("target", pluginTarget)
```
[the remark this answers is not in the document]
ikhoon: It was intentional and I thought it was safe because `target(CentralDogmaConfig)` is called right after the plugin is initialized. `pluginTarget` may be not null for most of the object's lifecycle.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirrorRunner.java
```diff
@@ -91,10 +100,16 @@ private CompletableFuture<MirrorResult> run(MirrorKey mirrorKey, User user) {
...
+                                                      mirrorKey.projectName + '/' + mirrorKey.mirrorId);
                         }
 
-                        final MirrorTask mirrorTask = new MirrorTask(mirror, user, Instant.now(), true);
+                        final String zone = mirror.zone();
+                        if (zone != null && !zone.equals(currentZone)) {
+                            throw new MirrorException("The mirror is not in the current zone: " + currentZone);
+                        }
+                        final MirrorTask mirrorTask = new MirrorTask(mirror, user, Instant.now(),
+                                                                     currentZone, true);
```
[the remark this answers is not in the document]
ikhoon: It was my mistake. It should be false.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirrorSchedulingService.java
ikhoon: Most users may choose the `pinnedZone` from the UI's select box, but it also makes sense to validate on the server side.

jrhee17: Sorry, let me rephrase my concern.

For instance, if servers are deployed in JP1, JP2, JP3 and we decide to migrate (JP3 -> JP4), is there no need for us to be aware of failing mirror jobs? I think moving servers across zones is not such a rare occasion.

server/src/main/java/com/linecorp/centraldogma/server/mirror/MirroringServicePluginConfig.java
```diff
@@ -68,38 +70,48 @@ public MirroringServicePluginConfig(
...
         checkArgument(this.maxNumBytesPerMirror > 0,
                       "maxNumBytesPerMirror: %s (expected: > 0)", this.maxNumBytesPerMirror);
+        this.zonePinned = zonePinned;
     }
 
     /**
      * Returns the number of mirroring threads.
      */
-    @JsonProperty
+    @JsonProperty("numMirroringThreads")
```
minwoox, line 79: revert?

site/src/sphinx/mirroring.rst
```diff
@@ -117,6 +118,15 @@ Setting up a mirroring task
     of strings where each line represents a single pattern. The file pattern expressed in gitignore is relative to the
     path of ``remoteUri``.
 
+- ``zone`` (string, optional)
+
+   - the zone where the mirroring task is executed.
+
+   - If unspecified:
+
+     - a mirroring task is executed in the first zone of ``zone.zones`` configuration.
```
minwoox, line 127: allZones?

minwoox (commented):
Left some nits. Looks great. 👍

server/src/main/java/com/linecorp/centraldogma/server/ZoneConfig.java
```diff
@@ -0,0 +1,94 @@
...
+public final class ZoneConfig {
+
+    private final String currentZone;
+    private final List<String> allZones;
+
+    /**
+     * Creates a new instance.
+     */
+    @JsonCreator
+    public ZoneConfig(@JsonProperty("currentZone") String currentZone,
```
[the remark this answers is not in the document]
ikhoon: It's possible, but I don't think `all` conveys the exact meaning, so I'd like to keep the current expression.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirroringServicePlugin.java
```diff
@@ -48,27 +68,37 @@ public synchronized CompletionStage<Void> start(PluginContext context) {
...
+            final ZoneConfig zoneConfig;
 
             if (mirroringServicePluginConfig != null) {
                 numThreads = mirroringServicePluginConfig.numMirroringThreads();
                 maxNumFilesPerMirror = mirroringServicePluginConfig.maxNumFilesPerMirror();
                 maxNumBytesPerMirror = mirroringServicePluginConfig.maxNumBytesPerMirror();
+                if (mirroringServicePluginConfig.zonePinned()) {
+                    zoneConfig = cfg.zone();
+                    if (zoneConfig == null) {
+                        throw new IllegalStateException(
```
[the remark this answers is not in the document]
ikhoon: `zoneConfig` won't be null because it was checked in
https://github.com/line/centraldogma/blob/5f0a9a1eb7ed2db53db195cd5c8577d9bce8b459/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java#L267-L270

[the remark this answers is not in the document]
ikhoon: I just added this logic for double checking.

server/src/main/java/com/linecorp/centraldogma/server/mirror/MirroringServicePluginConfig.java
ikhoon: It looks like a boilerplate but I prefer this pattern because it is more deterministic. 
We also use a standard getter for boolean values for which `is` is dropped. I wasn't sure if that was intended or not.
https://github.com/line/centraldogma/blob/5f0a9a1eb7ed2db53db195cd5c8577d9bce8b459/server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaConfig.java#L500-L503

Anyway, it is unrelated to this PR so I will revert it.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirrorSchedulingService.java
ikhoon: It makes more sense. Let me report a failure to `MirrorListener`.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirroringServicePlugin.java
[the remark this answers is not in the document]
minwoox: Then, how about using `assert`? We use `assert` for this purpose. 😉

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirrorSchedulingService.java
minwoox: Note that I also want to generalize the mirror listener so that it also reports the when a k8s aggregator fails.

server/src/main/java/com/linecorp/centraldogma/server/mirror/MirroringServicePluginConfig.java
minwoox: Yes, we're currently using it in a mixed manner, and it might be a good time to define a standard. 🤔

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirrorSchedulingService.java
ikhoon: Are you planning to design a general listener to receive arbitrary events?

minwoox: Yes. 😄

ikhoon: It sounds nice. I'm looking forward to the Central Dogma event bus. 🚌

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirroringServicePlugin.java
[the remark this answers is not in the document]
ikhoon: Fixed.

server/src/main/java/com/linecorp/centraldogma/server/ZoneConfig.java
```diff
@@ -0,0 +1,94 @@
...
+                      @JsonProperty("allZones") List<String> allZones) {
+        requireNonNull(currentZone, "currentZone");
+        requireNonNull(allZones, "allZones");
+        this.currentZone = convertValue(currentZone, "zone.currentZone");
+        this.allZones = allZones;
+        if (!allZones.contains(currentZone)) {
+            throw new IllegalArgumentException("The current zone must be one of the all zones.");
+        }
```
minwoox, line 50: nit:
```suggestion
        checkArgument(allZones.contains(currentZone), "The current zone: %s, (expected: one of %s)",
                      currentZone, allZones);
```

minwoox (approved):
👍 👍 👍

minwoox (approved):
👍 👍 👍

server/src/main/java/com/linecorp/centraldogma/server/ZoneConfig.java
ikhoon: I was lazy. I used the code generated by GitHub Copilot as is. 😅

PR #1066, opened 2024-11-27
Clear Filter Value When Clicking on a Directory in the File List View

minwoox:
In the file list view, users can filter directories and files by a path String. However, the filtered directory is navigated by a click, the children content is not displayed automatically because the filter value is not cleared.

### DIFF
--- a/webapp/build.gradle
+++ b/webapp/build.gradle
@@ -74,7 +74,7 @@ if (!rootProject.hasProperty('noLint')) {
     task prettier(type: NpmTask) {
         dependsOn tasks.eslint
 
-        args = ['run', 'format']
+        args = ['run', 'format:fix']
 
         inputs.dir('src')
         inputs.file('package.json')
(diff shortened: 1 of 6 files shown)


ikhoon:
We may reset the stats of `DynamicDataTable` by specifying `key`.
```diff
--- webapp/src/dogma/features/file/FileList.tsx
+++ webapp/src/dogma/features/file/FileList.tsx
@@ -125,7 +125,7 @@ const FileList = <Data extends object>({
   );
   return (
     <Box>
-      <DynamicDataTable columns={columns as ColumnDef<Data>[]} data={data} />
+      <DynamicDataTable key={slug} columns={columns as ColumnDef<Data>[]} data={data} />
```

I haven't fully tested but it seems to work.

minwoox:
> I haven't fully tested but it seems to work.

Checked it works perfectly. Thanks!

jrhee17 (approved):

ikhoon (approved):

PR #1070, opened 2024-12-05
Implement `LogMeta#toString` to easily debug zookeeper log removal warnings

jrhee17:
**Motivation**

With the improvement introduced by https://github.com/line/centraldogma/pull/1040, we have been observing logs where CentralDogma attempts to delete logs that don't exist.

In order to 1) verify which nodes are being prematurely deleted and 2) to observe trends as to why this may occur, I propose that `LogMeta#toString` be implemented.


minwoox (approved):
👍 👍 👍

PR #1071, opened 2024-12-05
Disable sphinx for aarch_64

jrhee17:
Until we actually upload the binary since:

1) A simple `assemble` task fails for osx users
2) Now the CI is failing as well


minwoox (approved):
👍 👍 👍

PR #1073, opened 2024-12-11
Expose a port number of `remoteUrl` in the mirror UI

ikhoon:
Motivation:

I found that the port number of a remote URI was missing and wasn't shown in the mirror UI.

Modifications:

- Use `URI.getAuthority()` to include the port number of a remote URI.

Result:

The custom port number of a remote URI is correctly displayed in the mirror UI.


minwoox (approved):
👍

PR #1076, opened 2024-12-24
Fix where memeber or token with null permission

minwoox:
I discovered a case that a member with null permission is added to the metadata.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/metadata/RepositoryMetadataDeserializer.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/metadata/RepositoryMetadataDeserializer.java
@@ -15,8 +15,6 @@
  */
 package com.linecorp.centraldogma.server.metadata;
 
-import static com.google.common.collect.ImmutableMap.toImmutableMap;
-
 import java.io.IOException;
 import java.util.Collection;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


jrhee17 (approved):

server/src/main/java/com/linecorp/centraldogma/server/metadata/RepositoryMetadataDeserializer.java
```diff
@@ -94,7 +94,16 @@ private static RepositoryRole repositoryRole(Collection<Permission> permissions)
     }
 
     private static Map<String, RepositoryRole> convert(Map<String, Collection<Permission>> permissions) {
-        return permissions.entrySet().stream()
-                          .collect(toImmutableMap(Entry::getKey, entry -> repositoryRole(entry.getValue())));
+        final Builder<String, RepositoryRole> builder = ImmutableMap.builder();
+        for (Entry<String, Collection<Permission>> entry : permissions.entrySet()) {
+            final Collection<Permission> value = entry.getValue();
+            if (!value.isEmpty()) {
+                final RepositoryRole repositoryRole = repositoryRole(value);
```
ikhoon, line 101: `repositoryRole` already returns `null` if `value` is empty. How about setting `repositoryRole` if it is not null to avoid double-checking.
```java
final RepositoryRole repositoryRole = repositoryRole(value);
if (!repositoryRole) {
    builder.put(entry.getKey(), repositoryRole);
}
```

Otherwise, add a filter condition to the existing stream operations?

```java
permissions.entrySet().stream()
           .filter(sce -> !sce.getValue().isEmpty())
           .collect(toImmutableMap(Entry::getKey, entry -> repositoryRole(entry.getValue())));
```

ikhoon (approved):

PR #1079, opened 2024-12-29
Spring Boot Centraldogma Loader

flex-gyeongil:
Disclaimer
This PR contains multiple conceptual things which is on-going development under flex-team.

We affected huge conflict between upstream. so want to resolve some discussion point about our current approach.

Motivation 1 
To make centraldogma k8s native

boot-loader, and jib is related

Motivation 2
To support SAML group authorization


build.gradle
```diff
@@ -78,28 +81,11 @@ configure(projectsWithFlags('java')) {
...
-            testImplementation libs.slf4j1.jul.to.slf4j
-            testImplementation libs.slf4j1.jcl.over.slf4j
-            testImplementation libs.slf4j1.log4j.over.slf4j
-            testRuntimeOnly libs.logback12
-        }
+        implementation libs.slf4j1.api
+        testImplementation libs.slf4j1.jul.to.slf4j
+        testImplementation libs.slf4j1.jcl.over.slf4j
+        testImplementation libs.slf4j1.log4j.over.slf4j
+        testRuntimeOnly libs.logback12
```
flex-gyeongil, line 88: To make separated concern for other configurations.
I had some dependency issues by this.

cloud-native/boot-loader/src/main/java/com/linecorp/armeria/server/saml/SamlEndpoint.java
```diff
@@ -0,0 +1,156 @@
...
+
+        // FIXME Should be applied to armeria upstream!!
+        final int port = uri.getPort() > 0 ? uri.getPort() : defaultPort;
+        if (!(port == 80 && "http".equals(scheme) || port == 443 && "https".equals(scheme))) {
+            builder.port(uri.getPort() > 0 ? uri.getPort() : defaultPort);
+        }
+        return builder
+                .path(uri.getPath())
+                .toUriString();
+    }
```
flex-gyeongil, line 123: Temporal handling due to known port erasing.

cloud-native/boot-loader/src/main/kotlin/com/linecorp/centraldogma/config/AuthConfig.kt
```diff
@@ -0,0 +1,135 @@
...
+    override fun quorumPort(): Int = ZooKeeperServerConfigSpec.validatePort(quorumPort, "quorumPort")
+
+    override fun electionPort(): Int = ZooKeeperServerConfigSpec.validatePort(electionPort, "electionPort")
+
+    override fun clientPort(): Int = clientPort
+
+    override fun groupId(): Int? = groupId
+
+    override fun weight(): Int = weight
+}
```
flex-gyeongil, line 62: Discussion point 1.

Due to hard coupling between configs and jackson deserialization, I wanna suggest to split configuration spec and it's implementation.

This approach can separate dependencies between jackson and others.

cloud-native/boot-loader/src/main/kotlin/com/linecorp/centraldogma/spring/loader/CentralDogmaSpringLoader.kt
```diff
@@ -0,0 +1,108 @@
...
+        val centralDogmaConfig =
+            object : CentralDogmaConfigSpec by properties {
+                override fun toString(): String = "Contains sensitive values. Please use actuator instead."
+            }
+
+        return CentralDogmaManager(
+            CentralDogma.forConfig(centralDogmaConfig, meterRegistry)
+                .authProviderCreator(authProviderCreator)
+                .tokenAuthorizerCreator(tokenAuthorizerCreator),
+        )
```
flex-gyeongil, line 102: Discussion Point 2.

configurable token authorizer and auth provider creator.

current implementation has strong coupling between json config file.

to resolve this coupling, I'd extract it

cloud-native/boot-loader/src/test/resources/full-config.yaml
```diff
@@ -0,0 +1,104 @@
...
+        W0zPsF4W+n+09pbG5JYOgVFkM1gVc56PmTUwFa+4D6Wqq/LyRowkXvjyWPoAE6JI
+        jbFhu2UTMKpL6kA9j/48b/ixULOblfV6ahB+kA9Pw9kjMAgcXUTm6hRepmu+gTcC
+        AwEAAaMhMB8wHQYDVR0OBBYEFDXaz1aatlHvzZi355hb+PHMtlK4MA0GCSqGSIb3
+        DQEBBQUAA4IBAQCXcCcwTpj2bWMzrxJOv1YScPT/KzRNlj3Bu6bKh4UW3//wwfO4
+        1uUhiYL9ont+EFCX6qtxbrWz16fahklaHKIc4AlKvwplwAqw2FUlNKhztmDDC0UM
+        as9o/TUs9rIthOofrfho8FeLULsiKQbZjwSp2o81xooIElBxilkXin+5GjyTQIlL
+        COMJngk/mFdOP5oCvAXVI3YsISgUsDFmR1A1vO0oiUKZY2evruMb7O+Pk2JieyJZ
+        AYI29mmfoDtaBH9pb1OpqEs1dpkTT1EkM51qOQv0wDGV7cksPOHqBNnJHzc9oyuW
+        tCaFCVVxUeL48EUfHU3efqCMw9ba++1LlYAp
+        -----END CERTIFICATE-----
```
flex-gyeongil, line 92: In k8s, or container environment, certificates can injected by environment variable and spring boot can handle it as configuration properties

common/src/main/java/com/linecorp/centraldogma/internal/Jackson.java
```diff
@@ -79,7 +80,8 @@ public final class Jackson {
         prettyMapper.enable(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS);
 
         registerModules(new SimpleModule().addSerializer(Instant.class, InstantSerializer.INSTANCE)
-                                          .addDeserializer(Instant.class, InstantDeserializer.INSTANT));
+                                          .addDeserializer(Instant.class, InstantDeserializer.INSTANT),
+                        new KotlinModule.Builder().build());
```
flex-gyeongil, line 84: for... some kotlin stuff.

it/server/src/test/java/com/linecorp/centraldogma/it/TestAllReplicasPlugin.java
```diff
@@ -43,8 +45,8 @@ public CompletionStage<Void> stop(PluginContext context) {
     }
 
     @Override
-    public Class<?> configType() {
+    public Class<? extends PluginConfig> configType() {
         // Return the plugin class itself because it does not have a configuration.
-        return TestAllReplicasPlugin.class;
+        return NoopPluginConfig.class;
```
flex-gyeongil, line 50: Discussion point 3.

From some reason. Plugins have specification for duplication check in plugin group.

It seems weird to check duplication by plugin config class because for that check, some plugins which has no configurations returns itself.

Plugin config may be resuable, then duplication check key should be plugin class, not plugin config.

HDYT?

server/src/main/java/com/linecorp/centraldogma/server/PluginGroup.java
```diff
@@ -80,27 +81,33 @@ static PluginGroup loadPlugins(PluginTarget target, CentralDogmaConfig config) {
...
-        // IllegalArgumentException is thrown if there are duplicate keys.
-        final Map<Class<?>, Plugin> pluginMap = allPlugins.build();
-        if (pluginMap.isEmpty()) {
+        if (allPlugins.isEmpty()) {
             return ImmutableMap.of();
         }
 
         final Map<PluginTarget, PluginGroup> pluginGroups =
-                pluginMap.values()
+                allPlugins
```
flex-gyeongil, line 110: related with Discution point 3.

duplication check by plugin class count.

settings.gradle
```diff
@@ -12,8 +12,9 @@ includeWithFlags ':client:java-armeria-legacy',             'java', 'publish', '
 includeWithFlags ':client:java-armeria-xds',                'java', 'publish', 'relocate'
 includeWithFlags ':client:java-spring-boot2-autoconfigure', 'java', 'publish', 'relocate', 'no_aggregation'
 includeWithFlags ':client:java-spring-boot2-starter',       'java', 'publish', 'relocate', 'no_aggregation'
-includeWithFlags ':client:java-spring-boot3-autoconfigure', 'java17', 'publish', 'relocate'
-includeWithFlags ':client:java-spring-boot3-starter',       'java17', 'publish', 'relocate'
+includeWithFlags ':client:java-spring-boot3-autoconfigure', 'java17', 'spring-boot3', 'publish', 'relocate'
+includeWithFlags ':client:java-spring-boot3-starter',       'java17', 'spring-boot3', 'publish', 'relocate'
+includeWithFlags ':cloud-native:boot-loader',               'kotlin', 'spring-boot3'
```
jrhee17, line 17: Question) Have you tried specifying `java17`? I think it could help avoid manually setting the supported minimum java version in gradle scripts

```suggestion
includeWithFlags ':cloud-native:boot-loader',               'java17', 'kotlin', 'spring-boot3'
```

jrhee17 (commented):
I haven't discussed this with the other maintainers, but let me leave my thoughts for now:

> Due to hard coupling between configs and jackson deserialization, I wanna suggest to split configuration spec and it's implementation.

I'm neutral about this proposition.

I do think it's possible to add ways to create a CentralDogmaConfig (e.g. `CentralDogmaConfig.fromYaml()`).

Having said this, in terms of 1) keeping the config in-sync with the `cloud-native` module 2) deduplication I think the proposal makes sense

> configurable token authorizer and auth provider creator.

Ideally, I think I prefer that the `CentralDogma.java` class provides ways to be configured and other applications `cloud-native` can consume this.

What do you think of exposing an SPI provider which is analogous to `AuthProviderCreator`, `TokenAuthorizerCreator` instead?

> From some reason. Plugins have specification for duplication check in plugin group.

Sounds like a bug to me, but it's probably better to check with the original authors @minwoox @ikhoon

cloud-native/boot-loader/src/main/kotlin/com/linecorp/centraldogma/config/AuthConfig.kt
minwoox: That's a good suggestion but I prefer dererring the separation until we use a configuration format that JAckson does not support, as Jackson already supports the YAML format:
```java
public static CentralDogmaConfig load(File configFile) throws JsonMappingException, JsonParseException {
    requireNonNull(configFile, "configFile");
    try {
        if (configFile.getPath().endsWith(".yaml")) {
            return yamlMapper.readValue(configFile, CentralDogmaConfig.class);        
        } else {
            return objectMapper.readValue(configFile, CentralDogmaConfig.class);
        }
    } catch (JsonParseException | JsonMappingException e) {
        throw e;
    } catch (IOException e) {
        throw new IOError(e);
    }
}
```

I also tried a simple PoC to see if it works:
```
@Test
void foo() throws JsonProcessingException {
    ObjectMapper jsonMapper = new ObjectMapper();
    ObjectMapper yamlMapper = new ObjectMapper(new YAMLFactory());
    final Foo foo = jsonMapper.readValue("{\"a\":\"foo\",\"bar\":{\"b\":\"c\"}}", Foo.class);
    System.err.println(foo);
    final Foo fooYaml = yamlMapper.readValue(
            "a: foo\n" +
            "bar:\n" +
            "  b: c", Foo.class);
    System.err.println(fooYaml);
}

public static class Foo {

    private final String a;
    private final Bar bar;

    @JsonCreator
    public Foo(@JsonProperty("a") String a, @JsonProperty("bar") Bar bar) {
        this.a = a;
        this.bar = bar;
    }

    @Override
    public String toString() {
                          .add("a", a)
                          .add("bar", bar)
                          .toString();
    }
}

public static class Bar {

    private final String b;

    @JsonCreator
    public Bar(@JsonProperty("b") String b) {
        this.b = b;
    }

    @Override
    public String toString() {
                          .add("b", b)
                          .toString();
    }
}
```

build.gradle
minwoox: Not sure which problem you've gone through but make sure not to include the slf4j1 for spring boot3 module.

cloud-native/boot-loader/src/main/java/com/linecorp/armeria/server/saml/SamlEndpoint.java
minwoox: Oops what's the bug btw?

cloud-native/boot-loader/src/test/resources/full-config.yaml
minwoox: `ConfigValueConverter` also supports that:
https://github.com/line/centraldogma/blob/8ca35bf184d5532292456f81c239d728c8b43601/server/src/main/java/com/linecorp/centraldogma/server/DefaultConfigValueConverter.java#L56-L57
But it's added recently so it might not be available when you started this implementation. 😓

it/server/src/test/java/com/linecorp/centraldogma/it/TestAllReplicasPlugin.java
minwoox: That makes sense. 👍

minwoox:
> What do you think of exposing an SPI provider which is analogous to AuthProviderCreator, TokenAuthorizerCreator instead?

I like this approach. 👍

PR #1081, opened 2024-12-30
Migrate Legacy Metadata Format to new Role-Based Format

minwoox:
Motivation:
In #1060, role-based authentication was introduced, changing the structure of metadata to support roles (READ, WRITE, ADMIN).

Modifications:
- Implemented a migration plugin to convert legacy metadata into the new role-based format.

Result:
- The legacy metadata format is migrated to new role-based format.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/MetadataRolesMigrationPlugin.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/MetadataRolesMigrationPlugin.java
@@ -1,0 +1,127 @@
+/*
+ * Copyright 2024 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.linecorp.centraldogma.server.internal;
+
+import java.time.Instant;
+import java.util.concurrent.CompletionStage;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Stopwatch;
+import com.google.common.collect.ImmutableMap;
+
+import com.linecorp.armeria.common.util.UnmodifiableFuture;
+import com.linecorp.centraldogma.common.Author;
+import com.linecorp.centraldogma.common.Change;
+import com.linecorp.centraldogma.common.Entry;
+import com.linecorp.centraldogma.common.Markup;
+import com.linecorp.centraldogma.common.Query;
+import com.linecorp.centraldogma.common.Revision;
+import com.linecorp.centraldogma.internal.Jackson;
+import com.linecorp.centraldogma.server.CentralDogmaConfig;
+import com.linecorp.centraldogma.server.command.Command;
+import com.linecorp.centraldogma.server.command.CommandExecutor;
+import com.linecorp.centraldogma.server.command.CommitResult;
+import com.linecorp.centraldogma.server.metadata.MetadataService;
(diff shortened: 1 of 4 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/MetadataRolesMigrationPlugin.java
```diff
@@ -0,0 +1,127 @@
...
+                                                                    context.commandExecutor());
+        final Stopwatch stopwatch = Stopwatch.createStarted();
+        for (Project project : context.projectManager().list().values()) {
+            final String name = project.name();
+            try {
+                metadataService.migrateMetadata(name).join();
+            } catch (Exception ex) {
+                // No need to rollback because the migration is not committed. Just log the error.
+                logger.warn("Failed to migrate metadata roles of {}", name, ex);
+                return UnmodifiableFuture.completedFuture(null);
```
jrhee17, line 81: Question)

1. If a metadata is already migrated, will a `RedundantChangeException` be thrown?
2. If an exception is thrown, would it be a better idea to log and continue instead of returning? Whether the metadata was migrated or not doesn't seem to affect the behavior.

jrhee17 (approved):

server/src/main/java/com/linecorp/centraldogma/server/internal/MetadataRolesMigrationPlugin.java
minwoox: > If a metadata is already migrated, will a RedundantChangeException be thrown?

Just the head revision is returned.
https://github.com/line/centraldogma/blob/a813dcd7de41670b2303681808be512897fa930a/server/src/main/java/com/linecorp/centraldogma/server/metadata/RepositorySupport.java#L128-L132

> If an exception is thrown, would it be a better idea to log and continue instead of returning? Whether the metadata was migrated or not doesn't seem to affect the behavior.

Yeah, we can probably do that. Let me make a change. 😉

server/src/main/java/com/linecorp/centraldogma/server/internal/MetadataRolesMigrationPlugin.java
```diff
@@ -0,0 +1,127 @@
...
+        }
+        logger.info("Starting metadata roles migration ...");
+
+        final MetadataService metadataService = new MetadataService(context.projectManager(),
+                                                                    context.commandExecutor());
+        final Stopwatch stopwatch = Stopwatch.createStarted();
+        for (Project project : context.projectManager().list().values()) {
+            final String name = project.name();
+            try {
+                metadataService.migrateMetadata(name).join();
```
ikhoon, line 77: Q) The blocking code will prevent other plugins from executing. Is that intentional?
https://github.com/line/centraldogma/blob/91ad720df74534b21a86b06d3fb9faee0d3e526b/server/src/main/java/com/linecorp/centraldogma/server/PluginGroup.java#L186-L187

server/src/main/java/com/linecorp/centraldogma/server/internal/MetadataRolesMigrationPlugin.java
```diff
@@ -0,0 +1,127 @@
...
+            logger.debug("Metadata roles have already been migrated. Skipping auto migration...");
+            return UnmodifiableFuture.completedFuture(null);
+        }
+        logger.info("Starting metadata roles migration ...");
+
+        final MetadataService metadataService = new MetadataService(context.projectManager(),
+                                                                    context.commandExecutor());
+        final Stopwatch stopwatch = Stopwatch.createStarted();
+        for (Project project : context.projectManager().list().values()) {
+            final String name = project.name();
```
ikhoon, line 75: Should we add a migration log for each project so we can identify the progress and investegate when something goes wrong?

ikhoon (approved):
👍👍

server/src/main/java/com/linecorp/centraldogma/server/internal/MetadataRolesMigrationPlugin.java
minwoox: > Is that intentional?

That is true. This approach is inspired by the mirroring migration job:
https://github.com/line/centraldogma/pull/880/files#diff-52b26e4b2cfaf28da0494ec3a92464b4b609594aeccb0a6f3c74d49a7454cf5cR135

server/src/main/java/com/linecorp/centraldogma/server/internal/MetadataRolesMigrationPlugin.java
minwoox: It's a relatively simple migration so I thought just leaving the warning log was enough. 😄

trustin (approved):

minwoox:
https://github.com/line/centraldogma/pull/1115
Will do the migration so let me close this PR.
Thanks for the review!

PR #1085, opened 2025-01-03
Add mirror access control

ikhoon:
Motivation:

Sometimes you need to restrict access to specific repositories without allowing mirroring for all repositories. These repositories may contain sensitive information or code that is only available to a select group of members.

Modifications:

- Add `MirrorAccessController` that can allow or disallow access to the remote repositories for mirroring.
  - It is used to check if a scheduled mirroring task or a task triggered by UI is allowed to access.
  - `MirrorListener.onCreate()` and `MirrorListener.onUpdate()` events are newly added. The events could be used as an extension point to detect a new remote URL pattern and integrate external systems such as a workflow or an email notification.
- Add CRUD REST API for administrators
  - GET `/api/v1/mirror/access` to retrieve all mirror access controls.
  - POST `/api/v1/mirror/access` to create a mirror access controls.
  - and so on...
- In addition to the REST API, a mirror access control can be added with `MirrorAccessController.allow(...)` API which is exposed by `PluginContext.mirrorAccessController()`.
- Add `CrudRepository` abstraction to easily create and update a collection of entities in a directory.
  - This API would be useful especially when we implement UI-based CRUD operations.
- Add mirror access control UI for administrators
  - CRUD operations are supported.

Result:

- Administrators can restrict access to remote repositories when mirroring.
- You can receive notifications when a new mirror is created or an existing mirror is updated through `MirrorListener`.


ikhoon:
Screenshot 2025-01-03 at 5 20 51 PM
Screenshot 2025-01-03 at 5 22 03 PM

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirrorAccessController.java
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    public CompletableFuture<List<MirrorAccessControl>> list() {
+        return repository().findAll().thenApply(list -> list.stream()
+                                                            .map(HasRevision::object)
+                                                            .collect(toImmutableList()));
+    }
+
+    @Override
+    public CompletableFuture<Boolean> allow(String targetPattern, String reason, int order) {
```
minwoox, line 88: Is there any reason that you encourage using this via `PluginContext.mirrorAccessController()` instead of the REST API?

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirrorAccessController.java
```diff
@@ -0,0 +1,182 @@
...
+        final Author author = Author.SYSTEM;
+        final MirrorAccessControl accessControl =
+                new MirrorAccessControl(idGenerator.generateId().toString(), targetPattern, true,
+                                        reason, 0, UserAndTimestamp.of(author));
+        logger.info("Allowing the target pattern: {}", accessControl);
+        // If there is a duplicate target pattern, the order will be considered first.
+        // If the order is the same, the latest one will be considered first.
+        return repository().save(accessControl, author).thenApply(unused -> true);
```
minwoox, line 96: How can we delete this access control later? Do we just need to disallow it?

minwoox (commented):
Basically looks great! Will also review the UI part. 👍

server/src/main/java/com/linecorp/centraldogma/server/mirror/MirrorListener.java
```diff
@@ -34,6 +34,16 @@
 @Nullable
 public interface MirrorListener {
 
+    /**
+     * Invoked when a new {@link Mirror} is created.
+     */
+    void onCreate(Mirror mirror, MirrorAccessController accessController);
```
jrhee17, line 40: Question) What do you think of only passing the mirrorId here and instead creating a `MirrorContext` similar to how `PluginContext` is defined which contains `ProjectManager`, `CommandExecutor`, etc..

`CentralDogma` could maintain an internal `CompositeMirrorListener` which holds this context, and pass it on to user-defined child `MirrorListener`s.

This way
1) we can actually handle the failure from the extra lookup in when creating/updating a mirror
2) there is more freedom in `MirrorListener` to modify/read from the repository

jrhee17 (approved):

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/HasRevision.java
```diff
@@ -0,0 +1,46 @@
...
+package com.linecorp.centraldogma.server.internal.storage.repository;
+
+import static java.util.Objects.requireNonNull;
+
+import com.linecorp.centraldogma.common.Revision;
+
+/**
+ * An interface that provides a {@link Revision} with an object.
+ */
+public interface HasRevision<T> {
```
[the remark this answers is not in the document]
ikhoon: The difference is that it was designed as an interface. As it is not a public API, I prefer keeping it as is instead of merging it. I will look at the use cases and consider a unified API later.
If `HasRevision` is unuseful, I also consider removing it.

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -570,6 +585,11 @@ private CommandExecutor startCommandExecutor(
             // Trigger the exception if any.
             startFuture.get();
             projectInitializer.initialize();
+            final CrudRepository<MirrorAccessControl> accessControlRepository =
+                    new GitCrudRepository<>(MirrorAccessControl.class, executor, pm,
+                                            INTERNAL_PROJECT_DOGMA, Project.REPO_DOGMA,
+                                            MIRROR_ACCESS_CONTROL_PATH);
+            mirrorAccessController.setRepository(accessControlRepository);
```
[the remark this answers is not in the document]
ikhoon: It was the initial design. But I changed to set `GitCrudRepository` lazily, because `dogma` project and its repositories may not be created if a server is newly started.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitCrudRepository.java
```diff
@@ -0,0 +1,143 @@
...
+import com.linecorp.centraldogma.server.command.CommitResult;
+import com.linecorp.centraldogma.server.internal.storage.repository.CrudRepository;
+import com.linecorp.centraldogma.server.internal.storage.repository.HasRevision;
+import com.linecorp.centraldogma.server.storage.project.ProjectManager;
+import com.linecorp.centraldogma.server.storage.repository.Repository;
+
+/**
+ * A {@link CrudRepository} implementation which stores JSON objects in a Git repository.
+ */
+public final class GitCrudRepository<T> implements CrudRepository<T> {
```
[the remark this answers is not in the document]
ikhoon: It seemed like JSON would continue to be used as a serialization format in the future. But from a performance perspective, I thought we should consider other alternatives to git, so I expressed `Git` as a variant.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirrorAccessController.java
ikhoon: It is possible to use REST API in a Plugin but it can be a hassle.
The plugin needs to manage an access token for each phase.

I was thinking of implementing a plugin to expose a callback as a REST API.
When the callback is invoked, a target pattern will be added to a whitelist or a blacklist via these APIs.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirrorAccessController.java
ikhoon: Currently, it isn't supported by `MirrorAccessControler` API. If necessary, should we use the UI to manage it?

Considering the current usability, I thought the design is sufficient.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirrorAccessController.java
```diff
@@ -0,0 +1,182 @@
...
+        });
+    }
+
+    @Override
+    public CompletableFuture<Map<String, Boolean>> isAllowed(Iterable<String> repoUris) {
+        return repository().findAll().thenApply(acl -> {
+            if (acl.isEmpty()) {
+                // If there is no access control, it is allowed by default.
+                return Streams.stream(repoUris)
+                              .collect(toImmutableMap(uri -> uri, uri -> true));
```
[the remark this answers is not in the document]
ikhoon: Good point. `uri` is not unique.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirrorSchedulingService.java
```diff
@@ -221,6 +229,21 @@ private void scheduleMirrors() {
...
+                                  if (!allowed) {
+                                      logger.debug("The mirroring from {} is not allowed. mirror: {}",
+                                                   m.remoteRepoUri(), m);
+                                      continue;
+                                  }
+                              } catch (Exception e) {
+                                  logger.warn("Failed to check the access control. mirror: {}",
+                                              m, e);
+                                  continue;
+                              }
```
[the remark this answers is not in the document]
ikhoon: `MirrorListener.onDisallowed()` is added to notify if a mirror is disallowed.
I didn't address the warning log for the exception because it will be raised rarely and the error may be a system-level error.

server/src/main/java/com/linecorp/centraldogma/server/mirror/MirrorListener.java
ikhoon: I agree that the extra lookup is useless for the default implementation but the overhead would not be significant. 

However, it is not enough to just give a mirror ID. Since a snapshot of when it was created is needed, a revision should also be provided. Instead of adding more contexts to the API, I think the current style is not bad. 

It may be possible to give more freedom, but I doubt the features are useful at this point.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/HasRevision.java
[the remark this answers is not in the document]
minwoox: > The difference is that it was designed as an interface. 

There's only the default implementation so I thought we could merge it with the `HolderWithRevision`.

>  I will look at the use cases and consider a unified API later.

That also sounds nice. 👍

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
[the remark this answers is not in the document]
minwoox: > because dogma project and its repositories may not be created if a server is newly started.

Alright, then how about removing the constructor parameter, then?

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirrorAccessController.java
minwoox: > It is possible to use REST API in a Plugin but it can be a hassle.

What I meant was exposing this API via `MirrorAccessControlService`. Is there any reason for this?

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirrorAccessController.java
minwoox: I see. I was just curious because the ID is automatically generated.

minwoox:
Sorry for causing the conflict. 😓

ikhoon:
No worries. I expected it in advance.

server/src/main/java/com/linecorp/centraldogma/server/mirror/MirrorListener.java
jrhee17: > we can actually handle the failure from the extra lookup in when creating/updating a mirror

For more context, I imagined that we will notify users when a mirror is created/updated, but not `allow`ed.
My concern wasn't really about overhead, but rather if there is a way to monitor/apply fallback logic if the callback isn't invoked due to failure from the extra IO. 
(I doubt we will be looking at exception logs since we aren't actively monitoring logs at the moment)

I'm fine with the current design, but wanted to make the above point clear

webapp/src/dogma/features/settings/mirror-access/MirrorAccessControlForm.tsx
```diff
@@ -0,0 +1,201 @@
...
+
+          <FormControl isRequired isInvalid={errors.targetPattern != null}>
+            <FormLabel>
+              <LabelledIcon icon={LuRegex} text="Git URI pattern" />
+            </FormLabel>
+            <Input
+              id="targetPattern"
+              name="targetPattern"
+              placeholder=".*://github.com/line/.*"
+              defaultValue={defaultValue.targetPattern}
```
minwoox, line 109: Can we validate the pattern here as well?

minwoox (approved):
Looks great! 👍 👍 👍

webapp/src/dogma/common/components/DeleteConfirmationModal.tsx
```diff
@@ -31,14 +47,20 @@ export const DeleteConfirmationModal = ({
   handleDelete,
   isLoading,
 }: DeleteConfirmationModalProps): JSX.Element => {
+  let from;
+  if (repoName) {
+    from = ` from ${repoName}`;
+  } else if (projectName) {
+    from = ` from ${projectName}`;
+  }
```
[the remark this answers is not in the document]
ikhoon: Yes, `null` or `undefined` is rendered as an empty string.

webapp/src/dogma/features/settings/mirror-access/MirrorAccessControlForm.tsx
ikhoon: `targetPattern` format is not fixed. It could be a plain text or a regular expression.
Do you intend to check if `targetPattern` is a valid regex?

minwoox: NVM. I missed the point that it can be plaintext.

server/src/main/java/com/linecorp/centraldogma/server/mirror/MirrorListener.java
ikhoon: I agree that the async processing is not good from a transaction perspective. If we only pass the mirror ID, there is an advantage that `MirrorListener` is always called when the commit is pushed.

Initially, I wished to design a simple API. I will consider improving this point in the future.

minwoox:
👍 👍 👍

PR #1086, opened 2025-01-03
Move Mirroring Configuration to Individual Repositories

minwoox:
Motivation:
Mirroring configurations should belong to the repository where the mirroring is executed to ensure proper ownership and management alignment.

Modifications:
- Implemented migration job that relocated mirroring configurations to their respective repositories.
  - Updated Mirroring REST API to reflect these changes.	
- Updated credential IDs in mirroring configurations to use a full resource ID format:
  - Example for a project credential: `projects/foo/credentials/foo-credential`.
  - Example for a repository credential: `projects/foo/repos/bar/credentials/bar-credential`.
- Adjusted role requirements for credential management:
  - Project OWNER: Can create, update, and delete credentials.
  - Project MEMBER: Can read credentials.
  - Removed dependency on Meta repository access roles.

Result:
- Mirroring configurations are now repository-specific.
- Dependency on Meta repository access roles has been eliminated.


minwoox:
Will continue after https://github.com/line/centraldogma/pull/1085 is merged. 😉

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MigratingMirrorToRepositoryService.java
```diff
@@ -0,0 +1,196 @@
...
+
+        // Enter read-only mode.
+        commandExecutor.execute(Command.updateServerStatus(ServerStatus.REPLICATION_ONLY))
+                       .get(1, TimeUnit.MINUTES);
+        logger.info("Starting Mirrors migration ...");
+        if (commandExecutor instanceof ZooKeeperCommandExecutor) {
+            logger.debug("Waiting for 10 seconds to make sure that all cluster have been notified of the " +
+                         "read-only mode ...");
+            Thread.sleep(10000);
+        }
```
minwoox, line 87: I prefer migrating without entering read-only mode because
- It's not critical if mirroring is temporarily unavailable for a minute.
- Any issues can be promptly addressed if they arise.

If this is acceptable, I will remove these lines of code. Please let me know your opinion.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/DefaultMetaRepository.java
```diff
@@ -125,57 +152,124 @@ public CompletableFuture<Mirror> mirror(String id, Revision revision) {
...
+                       });
     }
 
-    private List<Mirror> parseMirrors(Map<String, Entry<?>> entries, List<Credential> credentials)
-            throws JsonProcessingException {
+    private <T> CompletableFuture<T> convert(
+            List<String> credentialRepoNames,
+            boolean hasProjectCredential,
+            BiFunction<Map<String, List<Credential>>, List<Credential>, T> convertFunction) {
+        final ImmutableList<CompletableFuture<List<Credential>>> repoCredentialFutures;
```
jrhee17, line 226: Personally, I prefer the extra IO instead of all of this complexity. Having said this, I guess the current code is fine as well.

ikhoon:
While reviewing this PR, I found out two bugs in `DefaultMirrorAccessControler`. Fixed at [`5cae78a` (#1086)](https://github.com/line/centraldogma/pull/1086/commits/5cae78a716b405f666f5e38073a80810aaf1fbbd)

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirrorRunner.java
```diff
@@ -95,50 +95,54 @@ public MirrorRunner(ProjectApiManager projectApiManager, CommandExecutor command
...
+                                    throw new MirrorException("The mirror is disabled: " +
+                                                              mirrorKey.repoName + '/' + mirrorKey.mirrorId);
+                                }
+
+                                return mirrorAccessController.isAllowed(mirror).thenApplyAsync(allowed -> {
+                                    if (!allowed) {
+                                        throw new MirrorAccessException(
+                                                "The mirroring from " + mirror.remoteRepoUri() +
+                                                " is not allowed: " +
+                                                mirrorKey.projectName + '/' + mirrorKey.mirrorId);
```
ikhoon, line 120: To apply the same approach as above here
```suggestion
                                                mirrorKey.repoName + '/' + mirrorKey.mirrorId);
```

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/DefaultMetaRepository.java
```diff
@@ -125,57 +152,124 @@ public CompletableFuture<Mirror> mirror(String id, Revision revision) {
                 throw new RepositoryMetadataException("failed to load the mirror configuration", e);
             }
 
-            final CompletableFuture<List<Credential>> credentials;
             if (Strings.isNullOrEmpty(c.credentialId())) {
```
ikhoon, line 155: nit:
```suggestion
            if (c.credentialId().isEmpty()) {
```

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/MirrorConfig.java
```diff
@@ -147,12 +192,22 @@ id, enabled, schedule, direction, findCredential(credentials, credentialId),
         throw new IllegalArgumentException("could not find a mirror provider for " + mirrorContext);
     }
 
-    public static Credential findCredential(Iterable<Credential> credentials,
-                                            @Nullable String credentialId) {
-        if (credentialId != null) {
-            for (Credential c : credentials) {
+    private Credential findCredential(Map<String, List<Credential>> repoCredentials,
+                                      List<Credential> projectCredentials) {
+        if (repoCredential) {
```
ikhoon, line 197: Suggestion) This logic seems complicated to be included in a data class. What do you think of extracting the methods to `MirrorConverter`?

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MigratingMirrorToRepositoryService.java
```diff
@@ -0,0 +1,195 @@
...
+            }
+            try {
+                final MirrorConfig mirrorConfig = Jackson.treeToValue(content, MirrorConfig.class);
+                final String repoName = mirrorConfig.localRepo();
+                if (repoName == null) {
+                    warnInvalidMirrorConfig(entry, content);
+                    continue;
+                }
+                final MirrorConfig newMirrorConfig = mirrorConfig.withCredentialId(
+                        projectMirrorCredentialId(repository.parent().name(), mirrorConfig.credentialId()));
```
ikhoon, line 160: Should we skip migration for a `MirrorConfig` whose the credential ID starts with `projects/`? Because we don't rollback if the migration process is halt somehow. This migration could be partially applied.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/DefaultMetaRepository.java
```diff
@@ -125,57 +152,124 @@ public CompletableFuture<Mirror> mirror(String id, Revision revision) {
...
-                credentials = projectCredentials();
+                if (!parent().repos().exists(repoName)) {
+                    throw mirrorNotFound(revision, mirrorFile);
+                }
+                return CompletableFuture.completedFuture(c.toMirror(parent(), Credential.FALLBACK));
+            }
+
+            final List<String> credentialRepoNames;
+            final boolean hasProjectCredential;
+            if (c.repoCredential()) {
```
ikhoon, line 164: Suggestion) What do you think of adding a new resource repository layer to find a credential from a given ID? This logic could be simplified as we don't have to care if the ID is a repo credential.
```java
interface MetaResourceRepository {
    T find(String resourceName, Class<T> clazz);
}

Credential credential = repositoryRepository.find(mirror.credentialId(), Credential.class);
```

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MigratingMirrorToRepositoryService.java
ikhoon: Agreed.

ikhoon (commented):
I couldn't fully review this PR. I will do another round tomorrow.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/CredentialServiceV1.java
```diff
@@ -66,7 +67,7 @@ public CredentialServiceV1(ProjectApiManager projectApiManager, CommandExecutor
      *
      * <p>Returns the list of the credentials in the project.
      */
-    @RequiresRepositoryRole(value = RepositoryRole.READ, repository = Project.REPO_META)
+    @RequiresProjectRole(ProjectRole.MEMBER)
```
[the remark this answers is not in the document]
minwoox: We need to allow the admin of each repository to read the credentials. Since it's read, I thought we could give the read permission to all members.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/MirroringServiceV1.java
```diff
@@ -118,100 +120,116 @@ private static Map<String, Object> mirrorZoneConfig(CentralDogmaConfig config) {
      *
      * <p>Returns the list of the mirrors in the project.
      */
-    @RequiresRepositoryRole(value = RepositoryRole.READ, repository = Project.REPO_META)
+    @RequiresProjectRole(ProjectRole.OWNER)
```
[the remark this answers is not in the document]
minwoox: In this case, I think we need to allow only owners to read the whole repository mirroring configurations. Each repository admin can do crud in its repository.

server/src/main/java/com/linecorp/centraldogma/server/mirror/Mirror.java
```diff
@@ -63,6 +63,13 @@ public interface Mirror {
      */
     Credential credential();
 
+    /**
+     * Returns the ID of the credential which is used to access the Git repositories.
+     * It is in the form of {@code "projects/<project>/credentials/<credential>"} or
+     * {@code "projects/<project>/repos/<repo>/credentials/<credential>"}.
+     */
+    String mirrorCredentialId();
```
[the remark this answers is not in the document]
minwoox: Alright, let me try that. 😉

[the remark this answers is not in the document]
minwoox: @ikhoon 
I tried that approach, but it seems to make the logic and codebase more complicated.
What do you think of just removing `credential.id()` and using `crendential.name()` as we do for representing an xDS resource?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/MirroringServiceV1.java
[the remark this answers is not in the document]
ikhoon: I see. I understood that a project credential could be read by the repo admin but a project mirror couldn't.

server/src/main/java/com/linecorp/centraldogma/server/mirror/Mirror.java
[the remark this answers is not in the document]
minwoox: Never mind. Let me just add the resourceName field to the classes. 😉

server/src/main/java/com/linecorp/centraldogma/server/internal/api/MirroringServiceV1.java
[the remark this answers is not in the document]
minwoox: That is correct. 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/DefaultMetaRepository.java
minwoox: I've used this logic not to fetch the resource one by one when multiple credentials are needed to be found. Let me encapsulate the logic into another class then.

webapp/src/dogma/common/components/ChakraLink.tsx
```diff
@@ -1,12 +1,24 @@
...
   disabled?: boolean;
+  href: string;
 };
-export const ChakraLink = (props: ChakraLinkProps) => {
-  if (props.disabled) {
-    return <Box {...props}></Box>;
-  }
-  return <Link as={NextLink} {...props}></Link>;
-};
+export const ChakraLink = React.forwardRef<HTMLAnchorElement, ChakraLinkProps>(
```
minwoox, line 9: This is for solving the error:
```
hydration-error-info.js:63 Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?
```

server/src/main/java/com/linecorp/centraldogma/server/credential/Credential.java
```diff
@@ -26,35 +26,42 @@
...
 @JsonIgnoreProperties(ignoreUnknown = true)
 public interface Credential {
 
-    Credential FALLBACK = new NoneCredential("", true);
+    Credential FALLBACK = new NoneCredential("");
 
     /**
      * Returns the ID of the credential.
      */
-    @JsonProperty("id")
```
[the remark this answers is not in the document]
minwoox: The ID is in the resource name and we can easily extract the ID from it.
We also use the same approach for storing an xDS resources.

it/xds-member-permission/src/test/java/com/linecorp/centraldogma/server/test/XdsMemberPermissionTest.java
```diff
@@ -111,7 +112,7 @@ void shouldAllowMembersToAccessInternalProjects() throws Exception {
         final AggregatedHttpResponse credentialResponse =
                 adminWebClient.prepare()
                               .post("/api/v1/projects/@xds/credentials")
-                              .contentJson(new NoneCredential("test", true))
+                              .contentJson(new NoneCredential(credentialName("@xds", "test")))
```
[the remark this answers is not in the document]
minwoox: It's not possible to know the repository name with the ID so we always specify the resource name when creating a credential. The resource name already contains the ID.

server/src/main/java/com/linecorp/centraldogma/server/credential/LegacyCredential.java
```diff
@@ -0,0 +1,69 @@
...
+@JsonSubTypes({
+        @Type(value = NoneLegacyCredential.class, name = "none"),
+        @Type(value = PasswordLegacyCredential.class, name = "password"),
+        @Type(value = PublicKeyLegacyCredential.class, name = "public_key"),
+        @Type(value = AccessTokenLegacyCredential.class, name = "access_token")
+})
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonIgnoreProperties(ignoreUnknown = true)
+@Deprecated
+public interface LegacyCredential {
```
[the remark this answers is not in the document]
minwoox: I ended up with this implementation for easier migration. The new Credential doesn't have to validate the old legacy format.

server/src/main/java/com/linecorp/centraldogma/server/credential/Credential.java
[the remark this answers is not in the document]
minwoox: Discussed with @ikhoon and we have decided:
- Not to store `id` in `Credential` such as we do for xDS resources.
- To introduce `CreateCredentialRequest` which contains `credentialId` so that the resource name is automatically generated on the service.
- To use the same approach, which uses the resource name for indentifying it, for all other resources such as mirror.

common/src/main/java/com/linecorp/centraldogma/internal/CredentialUtil.java
```diff
@@ -0,0 +1,128 @@
...
+    // TODO(minwoox): remove ._ from the ID.
+    public static final Pattern REPO_CREDENTIAL_ID_PATTERN =
+            Pattern.compile(
+                    "^projects/([^/]+)/repos/([^/]+)/credentials/([a-z](?:[a-z0-9-_.]{0,61}[a-z0-9])?)$");
+
+    public static String validateCredentialName(String projectName, String repoName,
+                                                String credentialName) {
+        requireNonNull(credentialName, "credentialName");
+        if (credentialName.isEmpty()) {
+            // Allow an empty credential ID for Credential.FALLBACK.
```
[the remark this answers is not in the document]
minwoox: Renamed it. PTAL. 🙇

server/src/main/java/com/linecorp/centraldogma/server/internal/api/CredentialServiceV1.java
```diff
@@ -111,9 +118,15 @@ public CompletableFuture<Credential> getCredentialById(User loginUser,
...
+    @RequiresProjectRole(ProjectRole.OWNER)
     public CompletableFuture<PushResultDto> createCredential(@Param String projectName,
-                                                             Credential credential, Author author, User user) {
+                                                             CreateCredentialRequest request,
+                                                             Author author, User user) {
+        final String credentialId = request.credentialId();
+        final String credentialName = credentialName(projectName, credentialId);
+        // Ignore the credential name in the request and set the name with the format of
+        // "projects/{project}/credentials/{credentialId}".
+        final Credential credential = request.credential().withName(credentialName);
```
ikhoon, line 129: As the credential name in `Credential` is not used, should we make the field nullable so that the client doesn't have to fill it?

ikhoon (approved):
Thanks, @minwoox!

server/src/main/java/com/linecorp/centraldogma/server/internal/api/CredentialServiceV1.java
minwoox: @ikhoon The `name` is nullable when a credential is created:
https://github.com/line/centraldogma/blob/aec299376aa2baefcbba198baa608fff6b71ae51/server/src/main/java/com/linecorp/centraldogma/server/internal/credential/AccessTokenCredential.java#L42|
Do you mean the front-end side?

ikhoon: I was confused. I thought the client should fill `credential.name` in the request body.
From my understanding, the value is optional so clients could omit the value.

minwoox: That's correct. 👍

minwoox:
Thanks a lot for reviewing. 🙇

PR #1091, opened 2025-01-22
Remove GUEST write permission

ikhoon:
Motivation:

A public repository in GitHub grants READ permission to normal users but does not give WRITE permission to them. Similarly, it does not make sense to grant a `WRITE` permission to `GUEST` role in Central Dogma.

Related: https://github.com/line/centraldogma/issues/1048

Modifications:

- Disallow setting the GUEST WRITE permission in the repo role UI.
- The old WRITE permission of GUEST is automatically converted to READ permission.
  - I will write a migration to clean the GUEST WRITE permission data stored in `metadata.json` later. Because two role formats are fixed in `metadata.json` and I don't think the new format is stabilized. The new format could be changed when a new permission feature is added.
- Breaking) GUEST users no longer write a change to a repository. A MEMBER WRITE role is explicitly necessary.

Result:

The GUEST WRITE permission is removed from the repository roles.


jrhee17 (approved):

minwoox (approved):
👍 👍 👍

PR #1093, opened 2025-02-05
Prohibit application tokens from accessing repositories with GUEST permission

ikhoon:
Motivation:

We are aiming to implement stricter and more explicit access control in Central Dogma.

It has been observed that many tokens are accessing repositories as a GUEST role without being registered with the project. If the token is exposed to the outside, an attacker could gain access to all repositories that allow GUEST access.

To resolve the issue, I propose restricting new tokens from accessing repositories that have GUEST permission. For backward compatibility, the existing tokens are still allowed to access repositories with GUEST permission.

Modifications:

- Add `allowGuestAccess` field to `Token` class to indicate if a token is allowed to access a repository with the GUEST role.
  - If the value is null in `tokens.json`, it defaults to `true` for backward compatibility.
- Fixed `MetadataService` to prohibit tokens whose `allowGuestAccess` is false from accessing repositories with the GUEST role.

Result:

New application tokens no longer have access to repositories with the `GUEST` role.


server/src/main/java/com/linecorp/centraldogma/server/metadata/Token.java
```diff
@@ -80,22 +85,26 @@ public Token(@JsonProperty("appId") String appId,
...
                  @JsonProperty("systemAdmin") @Nullable Boolean isSystemAdmin,
+                 @JsonProperty("allowGuestAccess") @Nullable Boolean allowGuestAccess,
                  @JsonProperty("creation") UserAndTimestamp creation,
                  @JsonProperty("deactivation") @Nullable UserAndTimestamp deactivation,
                  @JsonProperty("deletion") @Nullable UserAndTimestamp deletion) {
         assert isAdmin != null || isSystemAdmin != null;
         this.appId = Util.validateFileName(appId, "appId");
         this.secret = Util.validateFileName(secret, "secret");
         this.isSystemAdmin = isSystemAdmin != null ? isSystemAdmin : isAdmin;
+        // Allow guest access by default for backward compatibility.
```
jrhee17, line 96: Note: Although `allowGuestAccess=true` is allowed via API, it isn't allowed via UI. I understood the reason is this field will eventually be deprecated and guest access won't be allowed entirely.

jrhee17 (approved):

server/src/main/java/com/linecorp/centraldogma/server/metadata/Token.java
ikhoon: Right. Users can't set `allowGuestAccess=true` either in the UI or the REST API.
`allowGuestAccess` is true for old tokens or administrative tokens.

minwoox (approved):
Thanks! Looks great. 👍

minwoox:
👍 👍 👍

PR #1094, opened 2025-02-06
(WIP) Report watcher metrics related to the latest revision

### LINKED ISSUE #1092: A request to expose the latest revision and latest committed at of the watcher as metrics
Hello, I have a feature request for centraldogma.
How about adding the following metrics?

```
centraldogma_latest_commited_at{project="xxx", repository="xxx", watcher="xxx"} timestamp
centraldogma_latest_revision{project="xxx", repository="xxx", watcher="xxx"} revisionNumber
```

**Motivation:**
Having these metrics would allow us to represent changes in centraldogma using Grafana annotations.
Many services use centraldogma as a feature flag system, and being able to visualize changes in Grafana would be highly useful.

be-hase:
close #1092 

I have only implemented a prototype so far.

### DIFF
--- a/client/java-armeria/src/main/java/com/linecorp/centraldogma/internal/client/armeria/ArmeriaCentralDogma.java
+++ b/client/java-armeria/src/main/java/com/linecorp/centraldogma/internal/client/armeria/ArmeriaCentralDogma.java
@@ -110,6 +110,8 @@ import com.linecorp.centraldogma.internal.Jackson;
 import com.linecorp.centraldogma.internal.Util;
 import com.linecorp.centraldogma.internal.api.v1.WatchTimeout;
 
+import io.micrometer.core.instrument.MeterRegistry;
+
 public final class ArmeriaCentralDogma extends AbstractCentralDogma {
 
     private static final MediaType JSON_PATCH_UTF8 = MediaType.JSON_PATCH.withCharset(StandardCharsets.UTF_8);
@@ -144,8 +146,9 @@ public final class ArmeriaCentralDogma extends AbstractCentralDogma {
     private final SafeCloseable safeCloseable;
 
     public ArmeriaCentralDogma(ScheduledExecutorService blockingTaskExecutor,
-                               WebClient client, String accessToken, SafeCloseable safeCloseable) {
-        super(blockingTaskExecutor);
+                               WebClient client, String accessToken, SafeCloseable safeCloseable,
+                               @Nullable MeterRegistry meterRegistry) {
+        super(blockingTaskExecutor, meterRegistry);
         this.client = requireNonNull(client, "client");
         authorization = "Bearer " + requireNonNull(accessToken, "accessToken");
         this.safeCloseable = safeCloseable;
--- a/client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogmaBuilder.java
+++ b/client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogmaBuilder.java
@@ -46,6 +46,8 @@ import com.google.common.net.InetAddresses;
 import com.linecorp.centraldogma.common.RevisionNotFoundException;
 import com.linecorp.centraldogma.internal.CsrfToken;
 
+import io.micrometer.core.instrument.MeterRegistry;
+
 /**
  * Builds a {@link CentralDogma} client.
  */
@@ -70,6 +72,8 @@ public abstract class AbstractCentralDogmaBuilder<B extends AbstractCentralDogma
     private int maxNumRetriesOnReplicationLag = DEFAULT_MAX_NUM_RETRIES_ON_REPLICATION_LAG;
     private long retryIntervalOnReplicationLagMillis =
             TimeUnit.SECONDS.toMillis(DEFAULT_RETRY_INTERVAL_ON_REPLICATION_LAG_SECONDS);
+    @Nullable
+    private MeterRegistry meterRegistry;
 
     /**
      * Returns {@code this}.
@@ -297,7 +301,7 @@ public abstract class AbstractCentralDogmaBuilder<B extends AbstractCentralDogma
     private List<URL> findProfileResources(ClassLoader classLoader) throws IOException {
         final ImmutableList.Builder<URL> urls = ImmutableList.builder();
         for (String p : profileResourcePaths) {
-            for (final Enumeration<URL> e = classLoader.getResources(p); e.hasMoreElements();) {
+            for (final Enumeration<URL> e = classLoader.getResources(p); e.hasMoreElements(); ) {
                 urls.add(e.nextElement());
             }
         }
@@ -414,5 +418,24 @@ public abstract class AbstractCentralDogmaBuilder<B extends AbstractCentralDogma
      */
     protected long retryIntervalOnReplicationLagMillis() {
         return retryIntervalOnReplicationLagMillis;
+    }
+
+    /**
+     * TODO
+     * @param meterRegistry
+     * @return
+     */
+    public B meterRegistry(MeterRegistry meterRegistry) {
+        requireNonNull(meterRegistry, "meterRegistry");
+        this.meterRegistry = meterRegistry;
+        return self();
+    }
+
+    /**
+     * TODO
+     * @return
+     */
+    protected final MeterRegistry meterRegistry() {
+        return meterRegistry;
     }
 }
--- a/client/java/src/main/java/com/linecorp/centraldogma/client/AbstractWatcher.java
+++ b/client/java/src/main/java/com/linecorp/centraldogma/client/AbstractWatcher.java
@@ -49,6 +49,9 @@ import com.linecorp.centraldogma.common.Query;
 import com.linecorp.centraldogma.common.RepositoryNotFoundException;
 import com.linecorp.centraldogma.common.Revision;
 
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tags;
+
 abstract class AbstractWatcher<T> implements Watcher<T> {
 
     private static final Logger logger = LoggerFactory.getLogger(AbstractWatcher.class);
@@ -59,6 +62,7 @@ abstract class AbstractWatcher<T> implements Watcher<T> {
         STOPPED
     }
 
+    private final CentralDogma centralDogma;
     private final ScheduledExecutorService watchScheduler;
     private final String projectName;
     private final String repositoryName;
@@ -69,6 +73,9 @@ abstract class AbstractWatcher<T> implements Watcher<T> {
     private final long maxDelayMillis;
     private final double multiplier;
     private final double jitterRate;
+    @Nullable
+    private final MeterRegistry meterRegistry;
+    private final Tags tags;
 
     private final List<Map.Entry<BiConsumer<? super Revision, ? super T>, Executor>> updateListeners =
             new CopyOnWriteArrayList<>();
@@ -82,9 +89,11 @@ abstract class AbstractWatcher<T> implements Watcher<T> {
     @Nullable
     private volatile CompletableFuture<?> currentWatchFuture;
 
-    AbstractWatcher(ScheduledExecutorService watchScheduler, String projectName, String repositoryName,
-                    String pathPattern, boolean errorOnEntryNotFound, long delayOnSuccessMillis,
-                    long initialDelayMillis, long maxDelayMillis, double multiplier, double jitterRate) {
+    AbstractWatcher(CentralDogma centralDogma, ScheduledExecutorService watchScheduler, String projectName,
+                    String repositoryName, String pathPattern, boolean errorOnEntryNotFound,
+                    long delayOnSuccessMillis, long initialDelayMillis, long maxDelayMillis, double multiplier,
+                    double jitterRate, @Nullable MeterRegistry meterRegistry) {
+        this.centralDogma = centralDogma;
         this.watchScheduler = watchScheduler;
         this.projectName = projectName;
         this.repositoryName = repositoryName;
@@ -95,6 +104,15 @@ abstract class AbstractWatcher<T> implements Watcher<T> {
         this.maxDelayMillis = maxDelayMillis;
         this.multiplier = multiplier;
         this.jitterRate = jitterRate;
+        this.meterRegistry = meterRegistry;
+        tags = Tags.of(
+                "project", projectName,
+                "repository", repositoryName,
+                "pathPattern", pathPattern,
+                // Although it is a rare case, there is a possibility of using the same watcher for the same
+                // project, repo, and pathPattern. Therefore, the watcher’s hash code should be included as a tag.
+                "id", String.valueOf(System.identityHashCode(this))
+        );
     }
 
     @Override
@@ -247,6 +265,7 @@ abstract class AbstractWatcher<T> implements Watcher<T> {
                  logger.debug("watcher noticed updated file {}/{}{}: rev={}",
                               projectName, repositoryName, pathPattern, newLatest.revision());
                  notifyListeners(newLatest);
+                 recordLatestRevision(newLatest.revision());
                  if (!initialValueFuture.isDone()) {
                      initialValueFuture.complete(newLatest);
                  }
@@ -315,6 +334,27 @@ abstract class AbstractWatcher<T> implements Watcher<T> {
                 }
             });
         }
+    }
+
+    private void recordLatestRevision(Revision revision) {
+        if (meterRegistry == null) {
+            return;
+        }
+        meterRegistry.gauge("centraldogma.watcher.latest.revision", tags, revision.major());
+        centralDogma.forRepo(projectName, repositoryName)
+                    .history(PathPattern.of(pathPattern))
+                    .get(revision, revision)
+                    .whenComplete((commits, e) -> {
+                        if (e == null) {
+                            commits.stream().findFirst().ifPresent(commit -> {
+                                meterRegistry.gauge("centraldogma.watcher.latest.commit.time", commit.when());
+                            });
+                        } else {
+                            logger.warn(
+                                    "Failed to retrieve the commit to record metrics. watcher={}/{}{}, rev={}",
+                                    projectName, repositoryName, pathPattern, revision, e);
+                        }
+                    });
     }
 
     @Override
--- a/client/java-armeria-legacy/src/main/java/com/linecorp/centraldogma/client/armeria/legacy/LegacyCentralDogma.java
+++ b/client/java-armeria-legacy/src/main/java/com/linecorp/centraldogma/client/armeria/legacy/LegacyCentralDogma.java
@@ -92,14 +92,16 @@ import com.linecorp.centraldogma.internal.thrift.RevisionConverter;
 import com.linecorp.centraldogma.internal.thrift.WatchFileResult;
 import com.linecorp.centraldogma.internal.thrift.WatchRepositoryResult;
 
+import io.micrometer.core.instrument.MeterRegistry;
+
 final class LegacyCentralDogma extends AbstractCentralDogma {
 
     private final CentralDogmaService.AsyncIface client;
     private final EndpointGroup endpointGroup;
 
     LegacyCentralDogma(ScheduledExecutorService blockingTaskExecutor, CentralDogmaService.AsyncIface client,
-                       EndpointGroup endpointGroup) {
-        super(blockingTaskExecutor);
+                       EndpointGroup endpointGroup, @Nullable MeterRegistry meterRegistry) {
+        super(blockingTaskExecutor, meterRegistry);
         this.client = requireNonNull(client, "client");
         this.endpointGroup = endpointGroup;
     }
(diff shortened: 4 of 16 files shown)


client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogmaBuilder.java
```diff
@@ -70,6 +72,8 @@ public abstract class AbstractCentralDogmaBuilder<B extends AbstractCentralDogma
     private int maxNumRetriesOnReplicationLag = DEFAULT_MAX_NUM_RETRIES_ON_REPLICATION_LAG;
     private long retryIntervalOnReplicationLagMillis =
             TimeUnit.SECONDS.toMillis(DEFAULT_RETRY_INTERVAL_ON_REPLICATION_LAG_SECONDS);
+    @Nullable
+    private MeterRegistry meterRegistry;
```
be-hase, line 76: By default, it might be a good idea to use `Metrics.globalRegistry`.
```
private MeterRegistry meterRegistry = Metrics.globalRegistry;
```

client/java-armeria/src/main/java/com/linecorp/centraldogma/internal/client/armeria/ArmeriaCentralDogma.java
```diff
@@ -144,8 +146,9 @@ public final class ArmeriaCentralDogma extends AbstractCentralDogma {
     private final SafeCloseable safeCloseable;
 
     public ArmeriaCentralDogma(ScheduledExecutorService blockingTaskExecutor,
-                               WebClient client, String accessToken, SafeCloseable safeCloseable) {
-        super(blockingTaskExecutor);
+                               WebClient client, String accessToken, SafeCloseable safeCloseable,
```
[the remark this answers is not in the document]
ikhoon: We don't consider certain changes in `internal` packages as breaking changes. Don't worry about that.

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogmaBuilder.java
```diff
@@ -415,4 +419,20 @@ public final B retryIntervalOnReplicationLagMillis(long retryIntervalOnReplicati
...
+    public B meterRegistry(MeterRegistry meterRegistry) {
+        requireNonNull(meterRegistry, "meterRegistry");
+        this.meterRegistry = meterRegistry;
+        return self();
+    }
+
+    /**
+     * Returns the {@link MeterRegistry}.
+     */
+    protected final MeterRegistry meterRegistry() {
```
ikhoon, line 435:
```suggestion
    @Nullable
    protected final MeterRegistry meterRegistry() {
```

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogmaBuilder.java
ikhoon: I prefer the default to be null because some people may not be interested in the metrics.

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractWatcher.java
```diff
@@ -43,22 +43,31 @@
...
+import io.micrometer.core.instrument.Meter.Id;
+import io.micrometer.core.instrument.Meter.Type;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tags;
+
 abstract class AbstractWatcher<T> implements Watcher<T> {
 
     private static final Logger logger = LoggerFactory.getLogger(AbstractWatcher.class);
+    private static final String LATEST_REVISION_METER_NAME = "centraldogma.watcher.latest.revision";
+    private static final String LATEST_COMMIT_TIME_METER_NAME = "centraldogma.watcher.latest.commit.time";
```
ikhoon, line 62:
```suggestion
    private static final String LATEST_REVISION_METER_NAME = "centraldogma.client.watcher.latest.revision";
    private static final String LATEST_COMMIT_TIME_METER_NAME = "centraldogma.client.watcher.latest.commit.time";
```

Some people may want to use a different prefix and add custom tags for their metrics. 
What do you think of adding a customization point for that? 

I realized that the `centraldogma-client` module has no dependency on Armeria. Actually, we have decided to add Armeria dependency to `centraldogma-client` but didn't put it in action.

To conclude, let's add Armeria dependency and use `MeterIdPrefix`, which I proposed in the issue.

be-hase: I also wrote this in the README, but since this is a relatively unique metric, I couldn’t think of many use cases for customization at this point.

I considered using this, but copying and pasting it didn’t seem ideal, so I decided not to for now.
(That said, it's also not a good idea for centraldogma-client to depend on Armeria.)
[MeterIdPrefix in Armeria](https://github.com/line/armeria/blob/main/core/src/main/java/com/linecorp/armeria/common/metric/MeterIdPrefix.java)

And if we want to rename or retag metrics, we can transform them using a meter filter.
[Micrometer Meter Filters Documentation](https://github.com/micrometer-metrics/micrometer-docs/blob/main/src/docs/concepts/meter-filters.adoc#transforming-metrics)

...How about?

be-hase:
( Hmm... For some reason, the CI is failing on Windows... Why...?

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractWatcher.java
```diff
@@ -124,6 +150,13 @@ void start() {
         if (state.compareAndSet(State.INIT, State.STARTED)) {
             scheduleWatch(0);
         }
+        if (meterRegistry != null) {
+            // emit metrics once the values are ready
+            initialValueFuture.thenAccept(latest -> {
```
jrhee17, line 155: Question) What do you think of just watching `this.latest.revision()` while setting -1 if not set?

I think the current approach introduces the additional complexity of maintaining an additional state.

e.g. Assuming:
```java
val watcher = dogma.watcher()...
watcher.start()
watcher.close()
// `initialValueFuture` is completed after the close and a gauge is registered
```

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractWatcher.java
```diff
@@ -95,6 +112,15 @@ private enum State {
...
         this.multiplier = multiplier;
         this.jitterRate = jitterRate;
+        this.meterRegistry = meterRegistry;
+        tags = Tags.of(
+                "project", projectName,
+                "repository", repositoryName,
+                "pathPattern", pathPattern,
+                // There is a possibility of using the same watcher for the same project, repo, and pathPattern.
+                // Therefore, the watcher’s hash code should be included as a tag.
+                "watcher_hash", String.valueOf(System.identityHashCode(this))
```
jrhee17, line 122: Optional) What do you think of assigning an ID instead of a hashcode on the off-chance there is a conflict? It can also help users check how many watchers there are in a JVM easily.

e.g.

```
private static final AtomicLong WATCHER_ID = new AtomicLong();
...
"watcher_id", String.valueOf(WATCHER_ID.incrementAndGet())
```

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractWatcher.java
```diff
@@ -317,6 +356,32 @@ private void notifyListeners(Latest<T> latest) {
...
+        centralDogma.forRepo(projectName, repositoryName)
+                    .history()
+                    .get(revision, revision)
+                    .whenComplete((commits, e) -> {
+                        if (e == null) {
+                            commits.stream().findFirst().ifPresent(commit -> {
+                                // noinspection ConstantValue
+                                if (latestCommit.getAndSet(commit) == null) {
+                                    // emit metrics once the values are ready
+                                    meterRegistry.gauge(LATEST_COMMIT_TIME_METER_NAME, tags, latestCommit,
```
jrhee17, line 373: Question) I understand the usefulness of last commit time since it may be more convenient to check when a file was last committed as opposed to checking the revision numbers.

Having said this, this commit time may not accurate represent the current watched object since it is done in a separate IO.

What do you think of handling commit time separately by embedding commit time information in the watch response? This will involve server-side changes as well, so it may be better to handle in a separate PR.

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractWatcher.java
be-hase: Ah, that sounds good too. If others prefer this approach, I’ll consider changing it.

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractWatcher.java
be-hase: I also struggled with this.  

Since I judged that it wouldn't become too complex, I wanted to provide as accurate metrics as possible, which is why I implemented it this way.  

However, if the **Central Dogma team** determines that it would make maintenance more complex, I think it's fine to set it to **-1**.

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractWatcher.java
be-hase: > What do you think of handling commit time separately by embedding commit time information in the watch response? This will involve server-side changes as well, so it may be better to handle in a separate PR.

Yes, I actually wanted this as well. However, since the API interface on the **Central Dogma server** side will also change, there seem to be many things that need to be discussed.

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractWatcher.java
ikhoon: I see. How about just adding an infix `.client`? I want to distinguish it from the server metric. We may add additional metrics under `centraldogma.client` namespace.

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractWatcher.java
ikhoon: How about allowing users to set the name of a `Watcher`? If it is absent, we may create a name automatically.

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractWatcher.java
ikhoon: Isn’t it more important when the client receives a watch for a revision? 
If you think the commit time recorded in the server is meaningful, I prefer @jrhee17's approach.

be-hase: Ahhh, now that you mention it, that makes sense.

It seems better to simply record the value at the time the watcher receives it, without calling the get history API.
I'll make that change. It should be much simpler.

e.g. `centraldogma.client.watcher.latest.received.time`

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractWatcher.java
be-hase: Ah, I overlooked the client infix remark. You're absolutely right. I'll fix it!

be-hase:
I have fixed it.

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractWatcher.java
```diff
@@ -317,6 +362,17 @@ private void notifyListeners(Latest<T> latest) {
...
     }
 
+    private void updateLatestReceivedTime() {
+        if (meterRegistry == null) {
+            return;
+        }
+
+        if (latestReceivedTime.getAndSet(Instant.now().getEpochSecond()) == 0) {
+            // emit metrics once the values are ready
+            meterRegistry.gauge(LATEST_RECEIVED_TIME_METER_NAME, tags, latestReceivedTime, AtomicLong::get);
```
jrhee17, line 372: Question) Will this actually register `latestReceivedTime` subsequently after the first registration since the `Meter.Id` is already registered? Would it make more sense to register the gauge once and define the update function to fetch `AbstractWatcher#latestReceivedTime` instead?

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractWatcher.java
jrhee17: Alternatively, `Double.NaN` could also be used

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractWatcher.java
```diff
@@ -82,9 +96,16 @@ private enum State {
     @Nullable
     private volatile CompletableFuture<?> currentWatchFuture;
 
+    // unix epoch seconds
+    // In Prometheus, it is common to handle data with second precision,
+    // so we intentionally use second precision.
+    // ref: https://prometheus.io/docs/prometheus/latest/querying/functions/#time
```
jrhee17, line 102: Optional) I think `TimeGauge` can be used if we want to make this metric implementation-agnostic.

e.g.
```
TimeGauge.builder("...", ..., TimeUnit.MILLISECONDS, ...)
```

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractWatcher.java
ikhoon: If you prefer precise metrics, should we use a lock to register metrics, close the watcher and remove metrics? 
- The lock will solve the problem that jrhee17 mentioned.
- The metric registration could be placed in one method. 

```java
ReentrantLock lock = ...

if (initialValueFuture.complete(newLatest)) {
  updateWatchMetrics();
}
void updateWatchMetrics() {
   lock.lock();

   if (isStopped()) {
      return;
   }

   if (this.latest == null) {
     // register two metrics
   }
   // update the last received time

   lock.unlock();
}

void close() {
    lock.lock();

    ....
    lock.unlock();
}
```

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractWatcher.java
```diff
@@ -95,6 +116,16 @@ private enum State {
         this.maxDelayMillis = maxDelayMillis;
         this.multiplier = multiplier;
         this.jitterRate = jitterRate;
+        this.meterRegistry = meterRegistry;
+        tags = Tags.of(
+                "project", projectName,
+                "repository", repositoryName,
+                "pathPattern", pathPattern,
```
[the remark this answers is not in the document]
be-hase: Ah, `path` looks good to me.

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractWatcher.java
be-hase: ...Thinking about this topic again, the `-1` seems fine.

if we don't care about `-1`, we can simply write `... >= 0` in PromQL.

(I prefer `-1` because NaN has some quirks in handling.)

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractWatcher.java
```diff
@@ -247,6 +296,7 @@ private void doWatch(int numAttemptsSoFar) {
                  logger.debug("watcher noticed updated file {}/{}{}: rev={}",
                               projectName, repositoryName, pathPattern, newLatest.revision());
                  notifyListeners(newLatest);
+                 latestReceivedTimeSeconds.set(Instant.now().getEpochSecond());
```
ikhoon, line 299: nit: Should we use `volatile long` instead? For now, we don't get any advantage from `AtomicLong`.

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractWatcher.java
```diff
@@ -124,6 +157,16 @@ void start() {
         if (state.compareAndSet(State.INIT, State.STARTED)) {
             scheduleWatch(0);
         }
+        if (meterRegistry != null) {
+            meterRegistry.gauge(LATEST_REVISION_METER_NAME, tags, this,
+                                watcher -> Optional.ofNullable(watcher.latest)
+                                                   .map(it -> it.revision().major())
+                                                   .orElse(-1));
```
ikhoon, line 164: Micro optimization) Should we use `if ( != null)` instead because we prefer to minimize object allocations even though it is trivial?

ikhoon (approved):
Thanks! 🙏🚀

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractWatcher.java
be-hase: Ah, that's right.

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractWatcher.java
be-hase: OK, fixed

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractWatcher.java
```diff
@@ -124,6 +156,20 @@ void start() {
         if (state.compareAndSet(State.INIT, State.STARTED)) {
             scheduleWatch(0);
         }
+        if (meterRegistry != null) {
+            meterRegistry.gauge(LATEST_REVISION_METER_NAME, tags, this,
+                                watcher -> {
+                                    if (watcher.latest == null) {
+                                        return -1;
+                                    } else {
+                                        return watcher.latest.revision().major();
```
minwoox, line 165: nit:
```suggestion
                                    final Latest<T> latest = watcher.latest;
                                    if (latest == null) {
                                        return -1;
                                    } else {
                                        return latest.revision().major();
```

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractWatcher.java
```diff
@@ -142,6 +188,12 @@ public void close() {
         if (currentWatchFuture != null && !currentWatchFuture.isDone()) {
             currentWatchFuture.cancel(false);
         }
+
+        if (meterRegistry != null) {
+            meterRegistry.remove(new Id(LATEST_REVISION_METER_NAME, tags, null, null, Type.GAUGE));
```
minwoox, line 193: Not sure if it's a good idea to remove this gauge since it takes long to remove it when there are a lot of meters in the map.
Anyway, how about creating the gauge in `start()` method and remove it here?
```java
this.gauge = Gauge.builder(LATEST_REVISION_METER_NAME, this, watcher -> {
    final Latest<T> latest = watcher.latest;
    if (latest == null) {
        return -1;
    } else {
        return latest.revision().major();
    }
}).tags(tags).register(meterRegistry);
...
meterRegistry.remove(gauge);
```

minwoox (commented):
Looks great. 👍 Left small suggestions. 😉

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractWatcher.java
ikhoon: > remove this gauge since it takes long to remove

The performance regression has been resolved, so removal operations will no longer take a long time. https://github.com/micrometer-metrics/micrometer/pull/5750#issuecomment-2550808550

be-hase: Just to be sure, I checked the implementation, and it seems that using `remove(Meter meter)` results in the same behavior.
https://github.com/micrometer-metrics/micrometer/blob/9b1d4042579ec0691c7b235478e7fe50ce5d8eab/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java#L748

minwoox: > The performance regression has been resolved, so removal operations will no longer take a long time. https://github.com/micrometer-metrics/micrometer/pull/5750#issuecomment-2550808550

Thanks for the link. I didn't know they fixed it. 😉 

> it seems that using remove(Meter meter) results in the same behavior.

Yeah, it simply delegates the call to the `remove` method that takes a meter.
I left this comment while considering a scenario where `watcher.close()` is called without calling `watcher.start()`.
In that case, it would attempt to remove unregistered meters.
I understand this is unlikely in most cases, but I wanted to clarify it since modifying the code wouldn’t be difficult. 😉

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractWatcher.java
minwoox: Yeah, it also doen't access the volatile field twice. 😉

minwoox (approved):
Thanks a lot for adding this nice metric, @be-hase! 😉

be-hase:
If there are no issues, would you mind merging it...?

minwoox:
@be-hase, Thanks again! 🙇

be-hase:
thanks :)

PR #1096, opened 2025-02-10
Provide a way to get last file revisions

ikhoon:
Motivation:

Central Dogma's content API does not return a file revision when the file is modified.
Alternatively, users can use the history API to find the file revision. One downside of the history API is that we couldn't use it to find the last file revisions of multiple files at once, requiring N queries instead.

If the content API can directly return the last file revision, it would eliminate the need to use the history API and execute N queries. Additionally, since the request is simple, caching would be beneficial from the server’s perspective. If multiple servers in the cluster perform the same operation due to a specific event, the cache hit rate is expected to be high.

Modifications:

- REST API)
  - Add `includeLastFileRevision (int)` as a query parameter to `listFiles` and `getFiles` API. If the value is greater than 1, `EntryDto` may include the modified file revision.
- Java Client API) 
  - `FilesWithRevisionRequest` is added to fluent specify `includeLastFileRevision` in the get files or list files API.
- Internal)
  - Add `blockingFindLastFileRevisions()` to search for file revision within a range of revisions. If a file is not found in the revisions, `Revision.INIT (1)` is filled instead.

Result:

You can now retrieve the last file revisions using the list files API and the get files API.
```java
CentralDogmaRepository repo = client.forRepo("myProj", "myRepo");

Map<String, Entry<?>> listFiles = 
  repo.file(PathPattern.all())
      // Specify the maximum number of revisions to search for the last file revision.
      // If the last file revision is not found from the specified range of revisions, 
      // `Revision.INIT` is returned instead.
      .includeLastFileRevision(100)
      .list()
      .join();
```


server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/cache/CacheableQueryCall.java
```diff
@@ -36,14 +36,16 @@ final class CacheableQueryCall extends CacheableCall<Entry<?>> {
...
     final Query<?> query;
+    final int includeLastFileRevision;
     final int hashCode;
 
-    CacheableQueryCall(Repository repo, Revision revision, Query<?> query) {
+    CacheableQueryCall(Repository repo, Revision revision, Query<?> query, int includeLastFileRevision) {
         super(repo);
         this.revision = requireNonNull(revision, "revision");
         this.query = requireNonNull(query, "query");
+        this.includeLastFileRevision = includeLastFileRevision <= 1 ? -1 : includeLastFileRevision;
```
jrhee17, line 46: Do we not need to check equality based on `includeLastFileRevision` as well?

jrhee17 (approved):

client/java-armeria/src/main/java/com/linecorp/centraldogma/internal/client/armeria/ArmeriaCentralDogma.java
```diff
@@ -408,29 +408,68 @@ private static Revision normalizeRevision(AggregatedHttpResponse res) {
     @Override
     public CompletableFuture<Map<String, EntryType>> listFiles(String projectName, String repositoryName,
                                                                Revision revision, PathPattern pathPattern) {
+        return listFiles0(projectName, repositoryName, revision, pathPattern, -1,
```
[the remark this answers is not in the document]
ikhoon: Good idea.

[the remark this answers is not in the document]
ikhoon: > 0 represents "check backwards history for 0 revisions"

If so, should 0 and 1 return the same result?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/cache/CacheableQueryCall.java
ikhoon: That's my mistake.

client/java-armeria/src/main/java/com/linecorp/centraldogma/internal/client/armeria/ArmeriaCentralDogma.java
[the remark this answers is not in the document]
jrhee17: > Personally, I think it makes more sense that:
> 0 represents "check backwards history for 0 revisions"
validate on includeLastFileRevision<0

I meant it feels more intuitive to me that the parameter represents the number of revisions to go backwards. (i.e. `Revision#backward(int)`).

I think the current implementation is fine as long as 1) the default value is well defined with validation 2) and the javadocs are clear. It might be helpful if the parameter name better represents the meaning of "check x number of revision including the head revision" as well.

common/src/main/java/com/linecorp/centraldogma/common/Entry.java
```diff
@@ -140,6 +158,14 @@ public Revision revision() {
         return revision;
     }
 
+    /**
+     * Returns a new {@link Entry} with the specified {@link Revision}.
+     */
+    public Entry<T> withRevision(Revision revision) {
+        requireNonNull(revision, "revision");
+        return new Entry<>(revision, path, type, content, true);
```
minwoox, line 166: Probably cache the `allowNullContent` and reuse it here for better accuracy?

common/src/main/java/com/linecorp/centraldogma/common/Entry.java
```diff
@@ -93,7 +93,20 @@ public static Entry<String> ofText(Revision revision, String path, String conten
...
+    /**
+     * Returns a newly-created {@link Entry} whose content is nullable.
+     *
+     * @param revision the revision of the {@link Entry}
+     * @param path the path of the {@link Entry}
+     * @param content the content of the {@link Entry}
+     * @param type the type of the {@link Entry}
+     * @param <T> the content type. {@link JsonNode} if JSON. {@link String} if text.
+     */
+    public static <T> Entry<T> ofNullable(Revision revision, String path, EntryType type, @Nullable T content) {
```
minwoox, line 108: What do you think of removing the `content` from the parameters if you want to make a nullable content entry?
Otherwise, I think we can use the above method which takes a `@Nullable T content`.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
```diff
@@ -537,6 +543,28 @@ private Map<String, Entry<?>> blockingFind(
                 result.put(path, entry);
             }
 
+            if (maxCommitsForFileRevision > 1 && !result.isEmpty()) {
+                // Fetch the last revisions of the files if the FETCH_LAST_FILE_REVISION option is enabled.
+                final List<String> files = result.values().stream()
+                                                 .filter(entry -> entry.type().type() != Void.class)
+                                                 // Remove the heading `/`
+                                                 .map(entry -> entry.path().substring(1))
+                                                 .collect(Collectors.toList());
```
minwoox, line 552: Immutable?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
```diff
@@ -549,6 +577,90 @@ private Map<String, Entry<?>> blockingFind(
...
+    //               without retrieving the history.
+    private Map<String, Revision> blockingFindLastFileRevisions(Revision from, Revision to,
+                                                                List<String> paths) {
+        final RevisionRange range = normalizeNow(from, to);
+        final RevisionRange descendingRange = range.toDescending();
+        final Set<String> remainingFiles = new HashSet<>(paths);
+        final Map<String, Revision> lastRevisionMap = new HashMap<>();
+
+        // At this point, we are sure: from.major >= to.major and read lock is acquired.
+        try (ObjectReader reader = jGitRepository.newObjectReader();
```
minwoox, line 590: What do you think of using the caching object reader that is introduced in #1097? 😉

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
```diff
@@ -549,6 +577,90 @@ private Map<String, Entry<?>> blockingFind(
...
+            // Ensure subdirectories are traversed.
+            treeWalk.setRecursive(true);
+
+            int numProcessedCommits = 0;
+            boolean needsToUpdateFilter = true;
+            for (RevCommit revCommit : revWalk) {
+                if (revCommit.getParentCount() == 0) {
+                    break; // Reached the root commit
+                }
+
```
minwoox, line 610: Probably we can add:
```
assert revCommit.getParentCount() == 1;
```

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
```diff
@@ -549,6 +577,90 @@ private Map<String, Entry<?>> blockingFind(
...
+                numProcessedCommits++;
+
+                treeWalk.reset();
+                treeWalk.addTree(revCommit.getTree());
+                treeWalk.addTree(revCommit.getParent(0).getTree());
```
minwoox, line 615: We can do:
```suggestion
                treeWalk.reset(revCommit.getTree().getId(), revCommit.getParent(0).getTree());
```

server/src/main/java/com/linecorp/centraldogma/server/storage/repository/Repository.java
```diff
@@ -157,7 +157,21 @@ default CompletableFuture<Entry<?>> get(Revision revision, String path) {
...
     default <T> CompletableFuture<Entry<T>> get(Revision revision, Query<T> query) {
-        return getOrNull(revision, query).thenApply(res -> {
+        return get(revision, query, -1);
+    }
+
+    /**
+     * Performs the specified {@link Query}. If {@code includeLastFileRevision} is greater than 1, it attempts
+     * to find the last file revision from the last {@code includeLastFileRevision} commits.
+     *
+     *
```
minwoox, line 167:
```suggestion
     *
```

minwoox (commented):
Looks great. 👍 Left some suggestions. 😉

common/src/main/java/com/linecorp/centraldogma/common/Entry.java
ikhoon: Makes sense.

client/java-armeria/src/main/java/com/linecorp/centraldogma/internal/client/armeria/ArmeriaCentralDogma.java
```diff
@@ -1031,9 +1073,11 @@ private static String toString(AggregatedHttpResponse res) {
         return res.content(charset);
     }
 
-    private static <T> Entry<T> toEntry(Revision revision, JsonNode node, QueryType queryType) {
+    private static <T> Entry<T> toEntry(JsonNode node, QueryType queryType) {
         final String entryPath = getField(node, "path").asText();
         final EntryType receivedEntryType = EntryType.valueOf(getField(node, "type").asText());
+        final Revision revision = new Revision(getField(node, "revision").asInt());
```
trustin, line 1079: The client may talk to an older Central Dogma instance. Can you make it handle an old server response as well and add a test case for it?

ikhoon: Good point.

minwoox: It seems like @ikhoon just moved this logic into this method:
https://github.com/line/centraldogma/blob/e072bcc80b0977811a0349a01a12ef5ff88c954d/common/src/main/java/com/linecorp/centraldogma/internal/api/v1/EntryDto.java#L56-L59
https://github.com/line/centraldogma/blob/584f3766d6facd0a77693a65b9e2fd9e49c0b212/client/java-armeria/src/main/java/com/linecorp/centraldogma/internal/client/armeria/ArmeriaCentralDogma.java#L868

ikhoon:
I changed the milestone to 0.76.0 as it is not an urgent issue.

client/java-armeria/src/main/java/com/linecorp/centraldogma/internal/client/armeria/ArmeriaCentralDogma.java
minwoox: The entry node always has the revision so it shouldn't be a problem?

PR #1098, opened 2025-02-12
Cache an entire repository to increase cache hit rate

ikhoon:
Motivation:

The size of the repositories in Central Dogma are relatively small. Therefore, caching the entire repo rather than caching each file separately would result in a higher cache hit rate. Additionally, when checking repository access patterns internally, it was found that a client tends to send multiple queries to fully scan repository files instead of using the all path pattern `/**`.

Modifications:

- Changed `CachingRepository` to cache all files in a repository when finding files with a path pattern. Filtering is performed to the cached entries.
- Doubled the default repository cache spec to approximately 256-megachars.
- Strip `/` if a matching path starts with it to apply `PathPatternFilter` to `Entry.path()`

Result:

It increases cache hits through repo-level cache and allows the server to quickly fill the cache when starting up.


server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/cache/CachingRepository.java
```diff
@@ -113,10 +106,40 @@ public CompletableFuture<Map<String, Entry<?>>> find(Revision revision, String p
...
+            final ImmutableMap.Builder<FindOption<?>, Object> newOptions = ImmutableMap.builder();
+            options.forEach((key, value) -> {
+                if (key != FindOption.MAX_ENTRIES) {
+                    newOptions.put(key, value);
+                }
+            });
+            cacheableOptions = newOptions.build();
+        }
+
+        return cache.get(new CacheableFindCall(repo, normalizedRevision, ALL_PATH, cacheableOptions))
```
jrhee17, line 122: I understood worst case, the cached size can be `(number of revisions) * (size of directory)`.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/cache/CachingRepository.java
```diff
@@ -113,10 +106,40 @@ public CompletableFuture<Map<String, Entry<?>>> find(Revision revision, String p
...
+                            final PathPatternFilter filter = PathPatternFilter.of(pathPattern);
+                            stream = stream.filter(entry -> filter.matches(entry.getKey()));
+                        }
+                        if (maxEntries != null) {
+                            stream = stream.limit(maxEntries);
+                        }
+
+                        // Use LinkedHashMap to 1) keep the order and 2) allow callers to mutate it.
+                        return stream.collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue,
+                                                               (oldV, newV) -> oldV,
```
jrhee17, line 141: Question) Is this just a precaution? or is it actually possible that multiple values (files) exist for a single key (path)?

server/src/main/java/com/linecorp/centraldogma/server/storage/repository/Repository.java
```diff
@@ -198,6 +198,7 @@ default <T> CompletableFuture<Entry<T>> getOrNull(Revision revision, Query<T> qu
             }
 
             @SuppressWarnings("unchecked")
+
```
jrhee17, line 201:
```suggestion
```

jrhee17 (approved):

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/cache/CachingRepository.java
ikhoon: Possible. Based on the user pattern I observed, most clients were only interested in HEAD revision. Old revisions were accessed occasionally. The practical cached size would be `(1 or 2 revisions) * (size of directory)`.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/cache/CachingRepository.java
ikhoon: `(oldV, newV) -> oldV` was just added because of the signature of `Collectors.toMap()`.
https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collectors.html#toMap-java.util.function.Function-java.util.function.Function-java.util.function.BinaryOperator-java.util.function.Supplier-

To use `LinkedHashMap`, `mergeFunction` is necessary.

minwoox: > allow callers to mutate it.

Is this necessary? I'm asking because we used to return the map as is.
Also what do you think of returning the map as-is when `maxEntries` is `null` so that we don't do the additional `stream.collect` call.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/PathPatternFilter.java
```diff
@@ -226,6 +226,11 @@ public boolean matches(String path) {
             return true;
         }
 
+        if (path.charAt(0) == '/') {
+            // Strip the leading '/' since pathPatterns was compiled without it.
+            path = path.substring(1);
```
minwoox, line 231: Was this a bug?

minwoox (commented):
Looks great. 👍 
Left some questions.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/PathPatternFilter.java
ikhoon: No, it isn't. The path of Central Dogma repository starts with `/` whereas the path of jGit does not start with `/`.
Previously, `PathPatternFilter` was designed to match the paths in jGit's tree. I changed it so that we can also use `PathPatternFilter` to filter `Entry.path()`.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/cache/CachingRepository.java
ikhoon: > Is this necessary?

Our current implementation mutates the returned entries.
https://github.com/line/centraldogma/blob/3a261ba4ec9685b27d5287b239503df4bce767e5/server-mirror-git/src/main/java/com/linecorp/centraldogma/server/internal/mirror/AbstractGitMirror.java#L356-L366

ikhoon: >  what do you think of returning the map as-is when maxEntries is null 

Did you mean `pathPattern == ALL_PATH && maxEntries == null`? As the caller mutates it, creating a new collection would be safer than returning the cached one.

Anyway, I believe using immutable data is a good approach. We need to fix the callers to create a new collection to mutate. I will handle it in the following PR.

minwoox: > Anyway, I believe using immutable data is a good approach. We need to fix the callers to create a new collection to mutate. I will handle it in the following PR.

I agree. I didn't expect that the call would modify the map. (I think I implemented it in that way 😓 )

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/PathPatternFilter.java
minwoox: Thanks for the explanation. 👍

minwoox (approved):
👍 👍 👍

PR #1099, opened 2025-02-14
Cache project metadata and tokens

minwoox:
Motivation:
Currently, only the raw contents of `ProjectMetadata` and `Tokens` are cached when fetched from `MetadataService`. However, deserialization takes a significant amount of time, so caching the deserialized instances would improve performance.

Modifications:
- Introduced `CacheableCall` interface.
- Added `Repository.execute(CacheableCall)` to cache calls.
- Cached deserialized instances of `ProjectMetadata` and `Tokens` retrieved by `MetadataService`.
- Remove stripe locks that were used for `CacheableCompareTreesCall`.

Result:
- Improved performance by avoiding redundant deserialization of `ProjectMetadata` and `Tokens`.


jrhee17 (approved):

server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java
```diff
@@ -100,10 +99,8 @@ public class MetadataService {
     public MetadataService(ProjectManager projectManager, CommandExecutor executor) {
         this.projectManager = requireNonNull(projectManager, "projectManager");
         this.executor = requireNonNull(executor, "executor");
-        metadataRepo = new RepositorySupport<>(projectManager, executor,
-                                               entry -> convertWithJackson(entry, ProjectMetadata.class));
-        tokenRepo = new RepositorySupport<>(projectManager, executor,
-                                            entry -> convertWithJackson(entry, Tokens.class));
+        metadataRepo = new RepositorySupport<>(projectManager, executor, ProjectMetadata.class);
```
ikhoon, line 102: In fact, `ProjectMetadata` is already cached in `DefaultProject` and automatically refreshed when the raw data is updated.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/RepositoryCache.java
```diff
@@ -69,71 +70,32 @@ public RepositoryCache(String cacheSpec, MeterRegistry meterRegistry) {
...
-
-    public <T> T load(CacheableCall<T> key, Supplier<T> supplier, boolean logIfMiss) {
-        CompletableFuture<T> existingFuture = getIfPresent(key);
-        if (existingFuture != null) {
-            final T existingValue = existingFuture.getNow(null);
-            if (existingValue != null) {
-                // Cached already.
-                return existingValue;
-            }
+        final CompletableFuture<T> future = new CompletableFuture<>();
```
ikhoon, line 80: Question) Should we check if the cache was populated by a previous call first before creating a new future?

server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java
ikhoon: What do you think of replacing the implementation of `fetchMetadata0()` with `projectManager.get(projectName).metadata()`?

minwoox: I can't replace it completely because fetching is still used here: https://github.com/line/centraldogma/blob/b9cebef958d6100f7da7f16da53c72d474ab888c/server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java#L170
Let me add `getMetdata` instead. 😉

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/RepositoryCache.java
minwoox: Creating a future is cheaper than call `Cache.getIfPresent`.

ikhoon:
Suggestion) It would be worth caching `Tokens` in a strong reference so we don't lose the cache. 
How about caching `Tokens` using `RepositoryListener`? We may remove some costs such as creating `CompletableFuture`, hash lookup, and cache miss.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/RepositoryCache.java
minwoox: This approach is recommended by the team. (Sorry for not giving you enough information.)
https://github.com/ben-manes/caffeine/wiki/Faq#recursive-computations

minwoox:
> Suggestion) It would be worth caching Tokens in a strong reference so we don't lose the cache.

That's a good suggestion. Let me handle it in this PR. 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/RepositoryCache.java
ikhoon: I would like to know what your reasoning is. Could you elaborate on that? Is `.getIfPresent` more expensive cost than `.putIfAbsent`?

server/src/main/java/com/linecorp/centraldogma/server/metadata/Roles.java
```diff
@@ -73,6 +75,24 @@ public Map<String, RepositoryRole> tokens() {
         return tokens;
     }
 
+    @Override
+    public int weight() {
+        int weight = 0;
```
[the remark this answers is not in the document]
ikhoon: Good to know

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/RepositoryCache.java
minwoox: If you want to check if the cache was popolulated, you need to call `getIfPresent` first and you need to call `put` or `putIfAbsent` if it's not created.
But in this approach, we can just call `putIfAbsent` only once. Did I miss something?

minwoox:
> Suggestion) It would be worth caching Tokens in a strong reference so we don't lose the cache.

@ikhoon `MetadataService` is created multiple places so I need some refactoring. Let me create a separate PR for that if you don't mind.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/RepositoryCache.java
ikhoon: I understand your intention. There is a trade-off for that. If the cache hit ratio is high, most calls will be handled by `getIfPresent()`.

As the change is trivial, let's keep it as it. I just wondered the reason. :-)

ikhoon:
> Let me create a separate PR for that if you don't mind.

Let's do that.

ikhoon (approved):
LGTM once the CI builds pass. 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/RepositoryCache.java
minwoox: That is also true.
Creating a future is relatively cheap and the code is shorter so I prefer this way. 😉

server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java
```diff
@@ -91,92 +92,102 @@ public class MetadataService {
...
     private final CommandExecutor executor;
+    private final InternalProjectInitializer projectInitializer;
 
     private final Map<String, CompletableFuture<Revision>> reposInAddingMetadata = new ConcurrentHashMap<>();
 
     /**
      * Creates a new instance.
      */
-    public MetadataService(ProjectManager projectManager, CommandExecutor executor) {
+    public MetadataService(ProjectManager projectManager, CommandExecutor executor,
```
minwoox, line 102: Please reveiew this class with the hiding whitespace mode. 😉

server/src/main/java/com/linecorp/centraldogma/server/internal/api/sysadmin/TokenService.java
```diff
@@ -87,14 +86,11 @@ public TokenService(CommandExecutor executor, MetadataService mds) {
      * <p>Returns the list of the tokens generated before.
      */
     @Get("/tokens")
-    public CompletableFuture<Collection<Token>> listTokens(User loginUser) {
+    public Collection<Token> listTokens(User loginUser) {
         if (loginUser.isSystemAdmin()) {
-            return mds.getTokens()
-                      .thenApply(tokens -> tokens.appIds().values());
+            return mds.getTokens().appIds().values();
```
minwoox, line 91: We can simply use the cached tokens in this case.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/sysadmin/TokenService.java
```diff
@@ -127,12 +123,12 @@ public CompletableFuture<HttpResult<Token>> createToken(@Param String appId,
             tokenFuture = mds.createToken(author, appId, isSystemAdminToken);
         }
         return tokenFuture
-                .thenCompose(unused -> mds.findTokenByAppId(appId))
+                .thenCompose(unused -> fetchTokensByAppId(appId))
```
minwoox, line 126: We have to fetch the tokens because the tokens are just updated.

ikhoon:
> - hopefully the following comment is addressed before RepositoryListener is used more extensively

Agreed. I'm currently working on the issue.

minwoox:
> I didn't check cases where a create directly leads to a get

That was my concern, but it's probably okay because:
- There's a high chance that the metadata will be up-to-date within an RTT before the client sends a request.
- The client can easily retry if it fails to read the latest data.

> hopefully the following comment is addressed before RepositoryListener is used more extensively

Yeah, it's the best solution for this situation. 😉

PR #1101, opened 2025-02-24
Synchronously invoke `RepositoryListener`

ikhoon:
Motivation:

`RepositoryListener` used `Repository.watch()` to continuously invoke the callback. Since the callback is not invoked immediately upon commit completion, the cache cannot guarantee a happen-before relationship with the write operation.
https://github.com/line/centraldogma/pull/1099#pullrequestreview-2632353089

Modifications:

- Fix `CommitWatchers` to reuse `Watch` added for `RepositoryListener`
  - The watch is removed when `WatchListener.onUpdate()` returns true.
- `recurseWatch()` no longer uses a scheduler to submit a new watch task when notified.
  - `find().join()` operation is invoked in the same thread as the commit to ensure that the callback is invoked before the commit's returned future.

Result:

`RepositoryListener` is now immediately invoked when the watching files are modified.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/CommitWatchers.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/CommitWatchers.java
@@ -37,6 +37,7 @@ import com.google.common.annotations.VisibleForTesting;
 
 import com.linecorp.centraldogma.common.CentralDogmaException;
 import com.linecorp.centraldogma.common.Revision;
+import com.linecorp.centraldogma.server.internal.storage.repository.git.Watch.WatchListener;
 
 final class CommitWatchers {
 
@@ -45,8 +46,9 @@ final class CommitWatchers {
     @VisibleForTesting
     final Map<PathPatternFilter, Set<Watch>> watchesMap = new WatcherMap(8192);
 
-    void add(Revision lastKnownRev, String pathPattern, CompletableFuture<Revision> future) {
-        add0(PathPatternFilter.of(pathPattern), new Watch(lastKnownRev, future));
+    void add(Revision lastKnownRev, String pathPattern,
+             @Nullable CompletableFuture<Revision> future, @Nullable WatchListener listener) {
+        add0(PathPatternFilter.of(pathPattern), new Watch(lastKnownRev, future, listener));
     }
 
     private void add0(final PathPatternFilter pathPattern, Watch watch) {
@@ -57,8 +59,12 @@ final class CommitWatchers {
             watches.add(watch);
         }
 
-        watch.future.whenComplete((revision, cause) -> {
-            if (watch.removed) {
+        final CompletableFuture<Revision> future = watch.future();
+        if (future == null) {
+            return;
+        }
+        future.whenComplete((revision, cause) -> {
+            if (watch.wasRemoved()) {
                 return;
             }
 
@@ -113,7 +119,7 @@ final class CommitWatchers {
         // Notify the matching promises found above.
         final int numEligiblePromises = eligibleWatches.size();
         for (int i = 0; i < numEligiblePromises; i++) {
-            eligibleWatches.get(i).future.complete(revision);
+            eligibleWatches.get(i).notify(revision);
         }
     }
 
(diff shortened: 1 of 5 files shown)


server/src/main/java/com/linecorp/centraldogma/server/storage/repository/Repository.java
```diff
@@ -568,9 +568,4 @@ default <T> CompletableFuture<MergedEntry<T>> mergeFiles(Revision revision, Merg
      * {@link RepositoryListener#pathPattern()} are pushed to this {@link Repository}.
      */
     void addListener(RepositoryListener listener);
-
-    /**
-     * Removes the {@link RepositoryListener} from this {@link Repository}.
-     */
-    void removeListener(RepositoryListener listener);
```
ikhoon, line 575: This API was removed to simplify the implementations. Since this API is designed for internal usage, I don’t think the removal will cause any breaking changes.

jrhee17 (approved):

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/CommitWatchers.java
```diff
@@ -123,7 +129,12 @@ void close(Supplier<CentralDogmaException> causeSupplier) {
             for (final Set<Watch> watches : watchesMap.values()) {
                 for (final Iterator<Watch> i = watches.iterator(); i.hasNext();) {
                     final Watch w = i.next();
-                    eligibleWatches = move(eligibleWatches, i, w);
+                    if (w.future() == null) {
```
minwoox, line 132:
```suggestion
                    if (!w.canRemove()) {
```

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/Watch.java
```diff
@@ -0,0 +1,91 @@
...
+        this.listener = listener;
+        canRemove = future != null;
+    }
+
+    void notify(Revision revision) {
+        if (future != null) {
+            future.complete(revision);
+        } else {
+            assert listener != null;
+            listener.onUpdate(revision, null);
```
minwoox, line 50: Shounldn't `lastKnownRevision` be updated?

minwoox (commented):
Looks great. 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/Watch.java
ikhoon: It is unnecessary for the current implementation, so I didn't do it.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/CommitWatchers.java
ikhoon: I thought `future() == null` was more expressive in conveying the current state, but using  `canRemove()` also sounds good.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/Watch.java
```diff
@@ -0,0 +1,91 @@
...
+    void notifyFailure(Throwable cause) {
+        if (future != null) {
+            future.completeExceptionally(cause);
+        } else {
+            assert listener != null;
+            listener.onUpdate(null, cause);
+        }
+    }
+
+    public Revision lastKnownRevision() {
```
minwoox, line 63:
```suggestion
    Revision lastKnownRevision() {
```

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/Watch.java
minwoox: I was just feeling a bit odd that  `lastKnownRevision` was not updated, as it's technically no longer the last known revision. But since updating it isn't necessary, I'm okay with it.

minwoox (approved):
Thanks! 👍 👍 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/Watch.java
ikhoon: I agree. Given that it’s an internal implementation that does not interact with the public API, I aimed to minimize modifications.

PR #1102, opened 2025-02-25
Prohibit pushing to internal project and repository

minwoox:
Motivation:
The internal project `dogma` and the internal repository `dogma` should be protected from unintended modifications. Even a system admin should not be allowed to push changes to them.

Modifications:
- Added validation to prevent write operations to the `dogma` project and repository.

Result:
- Pushing to the `dogma` project and repository from UI are now prohibited.


ikhoon:
> Even a system admin should not be allowed to push changes to them.

I have some doubts about this direction. Generally, even the system admin should not modify the internal files. However, bugs could occur at any time. If we remove the ability, it may not be easy to handle issues caused by incorrectly formatted data or legacy data.

Instead, how about adding a warning at the top of the editor when `dogma` files are open?

jrhee17:
> Instead, how about adding a warning at the top of the editor when dogma files are open?

Agreed, this sounds safer (for us)

jrhee17 (approved):
> However, the real issue is that if the data is incorrectly formatted, the system enters read-only mode—we’ve seen this happen in past outages.
Even if it doesn’t go into read-only mode, fixing files through the UI is still risky.

I thought this doesn't apply only to the `dogma` project since other internal metadata repositories also run the same risk - same goes with modifying the data directly.
Ideally, I guess a separate role could make more sense (dogma administrator: can access all projects/repos vs. server administrator: has permission to write to internal metadata).

Having said this, I don't think I've touched the dogma project directly over the past year. I guess we can always change this behavior later if needed.

ikhoon:
When we internally migrated the old SAML SSO to the new one, I wrote a migration script to replace the login IDs with email addresses in `dogma/dogma/tokens.json`. 

If write access is disallowed, we will have to develop a new API for those temporary issues.

minwoox:
Discussed with the team and we decided to prohibit changing the internal project and repository from UI.

minwoox:
@ikhoon Updated. PTAL. 😉

ikhoon (approved):
👍👍

trustin (approved):
Should we also prevent all projects and repositories that start with `@`? e.g. `@xds`

minwoox:
> Should we also prevent all projects and repositories that start with @? e.g. @xds

Will do this when the xDS APIs are stabilized. 😉
Currently, I'm sometimes committing via UI.

PR #1104, opened 2025-03-05
Improve the server shutdown process

ikhoon:
Motivation:

When a Central Dogma server stops, it first closes  `GitRepository` and other resources internally. The shutdown process does not respect `GracefulShutdownTimeout` but immediately closes them. As a result, inflight requests fail with `ShuttingDownException` even though they were received before the shutdown process started.

To gracefully shut down the server, I propose closing the Armeria server first. During the Armeria shutdown process, it will wait for requests to be finished during the GracefulShutdownTimeout.

Motifications:

- Close the server first and then shut down other resources.
- Configure Armeria `GracefulShutdown` to fail long-running requests with `ShuttingdownException`
- Fix `HttpApiExceptionHandler` to return 304 Not Modified for watch requests after the graceful shutdown timeout.

Result:

The Central Dogma server gracefully terminates requests while respecting `GracefulShutdownTimeout`.

### DIFF
--- a/it/server/src/test/java/com/linecorp/centraldogma/it/GracefulShutdownTest.java
+++ b/it/server/src/test/java/com/linecorp/centraldogma/it/GracefulShutdownTest.java
@@ -23,16 +23,17 @@ import static org.awaitility.Awaitility.await;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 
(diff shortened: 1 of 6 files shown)


jrhee17 (approved):

minwoox (approved):
👍 👍 👍

minwoox:
👍 👍 👍

PR #1109, opened 2025-03-13
Pass the mirror creator and updater to `MirrorListener`

ikhoon:
Motivation:

The user information may be useful if the implementation of `MirrorListener` is integrated with external systems.

Modifications:

- Add `User` to `onCreate()` and `onUpdate()`

Result:

The mirror creator and updater are accessible in `MirrorListener`.


minwoox (approved):
👍 👍 👍

jrhee17 (approved):

PR #1115, opened 2025-03-19
Hide meta repository from users

minwoox:
Motivation:
The `meta` repository contains mirror and credential configurations, which are now accessed via REST APIs. Since authorization for the `meta` repository is no longer required, it can be completely hidden from users.

Modifications:
- Hid `meta` repositories in all projects.
- Added a job to remove the `meta` repository's `RepositoryMetadata` from the `/metadata.json` file.

Result:
- You can no longer see or access the `meta` repository directly.


server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
```diff
@@ -201,8 +198,7 @@ public CompletableFuture<PushResultDto> push(
             Author author,
             CommitMessageDto commitMessage,
             @RequestConverter(ChangesRequestConverter.class) Iterable<Change<?>> changes) {
-        final User user = AuthUtil.currentUser(ctx);
-        checkPush(repository.name(), changes, user.isSystemAdmin());
+        checkMetaRepoPush(repository.name(), changes);
```
ikhoon, line 201: Should we remove this validation as accessing `meta` is prohibited by the decorator?

server/src/main/java/com/linecorp/centraldogma/server/command/StandaloneCommandExecutor.java
```diff
@@ -292,7 +292,8 @@ private CompletableFuture<Void> purgeRepository(PurgeRepositoryCommand c) {
     }
 
     private CompletableFuture<CommitResult> push(RepositoryCommand<?> c, boolean normalizing) {
-        if (c.projectName().equals(INTERNAL_PROJECT_DOGMA) || c.repositoryName().equals(Project.REPO_DOGMA) ||
+        if (c.projectName().equals(INTERNAL_PROJECT_DOGMA) ||
+            Project.internalRepos().contains(c.repositoryName()) ||
```
trustin, line 296: Could you replace all `Project.internalRepos().contains()` with `Project.isInternalRepo()` (a new static method) for brevity? We could introduce some naming rule in the future more easily as well.

trustin (approved):

server/src/main/java/com/linecorp/centraldogma/server/command/StandaloneCommandExecutor.java
minwoox: That's a good point. Addressed. 😉

ikhoon (approved):
👍👍

server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
minwoox: A system admin can still do the push and I left it because of these lines:
https://github.com/line/centraldogma/pull/1115/files#diff-365d4212ade9fe684cd89a04b979999f32b398703bd3ebbfa99d51d10617c3b4R455

jrhee17 (approved):

minwoox:
Thanks for reviewing. 😉

PR #1112, opened 2025-03-14
Ensure Complete Kubernetes Endpoints Update

minwoox:
Motivation:
Currently, k8s aggregator may update endpoints partially before all `KubernetesEndpointGroup`s are initialized.

Modifications:
- Changed the logic to push Kubernetes endpoints only after all `KubernetesEndpointGroup`s are initialized, preventing partial updates.

Result:
- k8s endpoints are not partially updated anymore.

PR #1118, opened 2025-03-27
Restore debouncing logic for Kubernetes endpoint updates

minwoox:
Motivation:
The debouncing logic for updating Kubernetes endpoints was removed in #1112 because `KubernetesEndpointGroup` was modified to retrieve all endpoints at once when started. However, this change caused issues as backend restarts led to frequent endpoint updates. To mitigate this, the debouncing logic needs to be restored.

Modifications:
- Reintroduced the debouncing logic for Kubernetes endpoint updates to prevent excessive updates during backend restarts.

Result:
- Reduced unnecessary updates to Kubernetes endpoints.


jrhee17 (approved):

ikhoon (approved):
👍👍

PR #1122, opened 2025-04-10
Add support for setting individual repository to read-only mode

minwoox:
Motivation:  
Currently, there's no way to enforce read-only mode on a repository.
Introducing this capability enables use cases such as freezing a repository during migration or
preventing updates to a Kubernetes repository from upstream changes.

Modifications:  
- Introduced `RepositoryStatus` to represent the status of a repository.
- Added a REST API to update a repository's `RepositoryStatus`.
- Relaxed required roles for certain RepositoryService APIs.

Result:  
- Repositories can now be set to read-only mode, preventing write operations.


server/src/main/java/com/linecorp/centraldogma/server/command/AbstractCommandExecutor.java
```diff
@@ -125,6 +138,34 @@ public final void setWritable(boolean writable) {
...
+        if (command instanceof NormalizableCommit) {
+            assert command instanceof RepositoryCommand;
+            final RepositoryCommand<?> repositoryCommand = (RepositoryCommand<?>) command;
+            if (InternalProjectInitializer.INTERNAL_PROJECT_DOGMA.equals(repositoryCommand.projectName())) {
+                return;
+            }
+            String repoName = repositoryCommand.repositoryName();
+            if (Project.REPO_META.equals(repoName)) {
+                // Use REPO_DOGMA for the meta repository because the meta repository will be removed.
+                repoName = Project.REPO_DOGMA;
```
jrhee17, line 151: Question) Does it every make sense to mark `REPO_DOGMA` as not writable? If `REPO_DOGMA` becomes unwritable, how would one restore a repository to be writable? Would it make sense to just not allow making `REPO_DOGMA` unwritable instead?

jrhee17 (approved):

server/src/main/java/com/linecorp/centraldogma/server/command/AbstractCommandExecutor.java
minwoox: > how would one restore a repository to be writable? 

So, force push is used. 😉 
https://github.com/line/centraldogma/pull/1122/files#diff-ce9a71ae9a3b82069d8129da07529226ca4ced78c8ce6e6363f8f908e649edbcR129

Anyway, I realized I needed to make a few more changes. I’ll give you a ping once I’m done. 😉

server/src/main/java/com/linecorp/centraldogma/server/internal/api/sysadmin/ServerStatusService.java
```diff
@@ -81,9 +81,13 @@ public CompletableFuture<ServerStatus> updateStatus(UpdateServerStatusRequest st
                 serverStatusManager.updateStatus(newStatus);
                 return status();
             }, serverStatusManager.sequentialExecutor());
-        } else {
-            return execute(Command.updateServerStatus(newStatus))
-                    .thenApply(unused -> status());
         }
+        if (!oldStatus.replicating() && statusRequest.serverStatus().replicating()) {
```
minwoox, line 85: This is not related to the repository status but I've just wanted to add.

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -1165,6 +1167,7 @@ private <T> T blockingExecute(Command<T> command) throws Exception {
                 final long lastRevision = recentRevisions.stream().mapToLong(Long::parseLong).max().getAsLong();
                 replayLogs(lastRevision);
             }
+            throwExceptionIfRepositoryNotWritable(command);
```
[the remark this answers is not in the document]
minwoox: > Is this check necessary since the command will be verified at L1172 delegate.execute(command).get() anyways?

No it isn't. 😉 
I was going to use the replication only field from the `ServerStatus` but I've realized that it isn't needed for a repository. Also reverted the `ServerStatus` change and introduced `RepositoryStatus`. PTAL. 🙇

ikhoon (approved):
👍👍

server/src/main/java/com/linecorp/centraldogma/server/metadata/MetadataService.java
```diff
@@ -1207,4 +1217,71 @@ private static <T> ImmutableMap<String, T> removeFromMap(Map<String, T> map, Str
...
+            transformer = new ProjectMetadataTransformer((headRevision, projectMetadata) -> {
+                final RepositoryMetadata repositoryMetadata = projectMetadata.repos().get(Project.REPO_DOGMA);
+                if (repositoryMetadata != null) {
+                    throwIfRedundant(repositoryStatus, headRevision, repositoryMetadata, Project.REPO_DOGMA);
+                }
+                final RepositoryMetadata newRepositoryMetadata = RepositoryMetadata.ofDogma(repositoryStatus);
+                final Builder<String, RepositoryMetadata> builder = ImmutableMap.builder();
+                builder.put(Project.REPO_DOGMA, newRepositoryMetadata);
+                projectMetadata.repos().forEach((name, metadata) -> {
+                    if (!Project.REPO_DOGMA.equals(name)) {
```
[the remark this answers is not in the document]
minwoox: `meta` repository `RepositoryMetadata` are gone from this PR:
https://github.com/line/centraldogma/pull/1115/files

server/src/main/java/com/linecorp/centraldogma/server/metadata/RepositoryMetadata.java
```diff
@@ -120,7 +139,9 @@ public Roles roles() {
 
     /**
      * Returns who created this repository when.
+     * This returns {@code null} if this repository is {@link Project#REPO_DOGMA}.
```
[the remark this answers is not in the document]
minwoox: Since this returns a `UserAndTimestamp`, it also needs to include the time. 
I think we can just return the project creation time, but this value isn't going to be used so I've chosen retruning null.

server/src/main/java/com/linecorp/centraldogma/server/command/Command.java
```diff
@@ -391,11 +391,12 @@ static <T> Command<T> forcePush(Command<T> delegate) {
         requireNonNull(delegate, "delegate");
         checkArgument(delegate.type() == CommandType.CREATE_PROJECT ||
                       delegate.type() == CommandType.CREATE_REPOSITORY ||
-                      delegate.type() == CommandType.NORMALIZING_PUSH || delegate.type() == CommandType.PUSH,
-                      "delegate: %s (expected: CREATE_PROJECT, CREATE_REPOSITORY, NORMALIZING_PUSH or PUSH)",
+                      delegate.type() == CommandType.NORMALIZING_PUSH ||
+                      delegate.type() == CommandType.TRANSFORM ||
+                      delegate.type() == CommandType.PUSH,
+                      "delegate: %s " +
+                      "(expected: CREATE_PROJECT, CREATE_REPOSITORY, NORMALIZING_PUSH, TRANSFORM or PUSH)",
```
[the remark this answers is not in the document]
minwoox: Let me just remove the Javadoc since similar mistakes are likely to happen again. Thanks!

server/src/main/java/com/linecorp/centraldogma/server/internal/api/RepositoryServiceV1.java
```diff
@@ -197,6 +259,44 @@ public Map<String, Integer> normalizeRevision(ServiceRequestContext ctx,
...
     }
 
+    /**
+     * GET /projects/{projectName}/repos/{repoName}/status
+     *
+     * <p>Returns the repository status.
+     */
+    @Get("/projects/{projectName}/repos/{repoName}/status")
+    @RequiresRepositoryRole(RepositoryRole.ADMIN)
+    public RepositoryDto status(Project project, Repository repository) {
```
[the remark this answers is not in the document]
minwoox: Thanks! Updated. 😉

server/src/main/java/com/linecorp/centraldogma/server/metadata/RepositoryMetadata.java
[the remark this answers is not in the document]
ikhoon: I preferred to avoid null-checking if we could fill in the data. I don't think we should handle it in this PR. Let's consider it as a separate issue.

server/src/main/java/com/linecorp/centraldogma/server/command/AbstractCommandExecutor.java
jrhee17: The reason I said this is due to how replay logs works:

1. Assume that a repository is already marksed as read only
2. We call `MetadataService#updateRepositoryStatus`
3. If there are any outstanding logs for `REPO_DOGMA`, an exception will always occur while trying to replay logs here: https://github.com/line/centraldogma/blob/f1ff2b65639cb47fa049a35d81ec1183eab8f91f/server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java#L1166
4. The `REPO_DOGMA` can't be updated via zookeeper logs - hence we are in a irrecoverable state.

Anyways, for most cases I guess this won't be an issue since I doubt we'll be marking `REPO_DOGMA` as read-only

jrhee17 (approved):

server/src/main/java/com/linecorp/centraldogma/server/command/AbstractCommandExecutor.java
minwoox: > an exception will always occur while trying to replay logs here:

We change to `PushAsIsCommand` before storing log:
https://github.com/line/centraldogma/blob/f1ff2b65639cb47fa049a35d81ec1183eab8f91f/server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java#L1176
Thus, when it's replaying `PushAsIsCommand`s are used.

StandAloneCommandExecutor raises an exception only when the command is instance of a normalize commit:
https://github.com/line/centraldogma/pull/1122/files#diff-1b79753a6234a28cbdd41ef181ff57c2c3ed6d439b4e7fb6092994f35359d7beR116

So exception won't be raised.

minwoox: What I wanted to say is that read-only mode prohibits a new commit but allows replayed commits. 😉

minwoox:
Thanks for reviewing. 😉

server/src/main/java/com/linecorp/centraldogma/server/command/AbstractCommandExecutor.java
jrhee17: > If there are any outstanding logs for REPO_DOGMA (we don't hold a log while checking whether a repo is read-only) , an exception will always occur while trying to replay logs here:

My point was rather that `throwExceptionIfRepositoryNotWritable` is not checked inside the distributed lock, so it's possible that a commit is made after the push from `updateRepositoryStatus` goes through.

The aforementioned commit will prevent additional commits from going through at `replayLogs`, hence an irrecoverable state.
This will only occur for `REPO_DOGMA`

jrhee17: Nevermind this case since I doubt it will actually occur in production 🙇

jrhee17: @minwoox explained offline.

Since the current implementation ignores `NormalizableCommit`s, commands executed by ZK won't be a target for `throwExceptionIfRepositoryNotWritable `.
Hence, 
1) commits that newly try to write to zk will fail due to `throwExceptionIfRepositoryNotWritable`
2) commits that have already been written to zk will always be processed since they aren't `NormalizableCommit`s

minwoox: Thanks for the clarification. 👍

PR #1123, opened 2025-04-15
Prepare migration to unify meta and dogma repositories

minwoox:
Motivation:
We have decided to combine the `dogma` and `meta` repositories into a single `dogma` repository to simplify repository management. This change is a preliminary step toward that migration.

Modifications:
- Updated logic to use the `dogma` repository as the `MetaRepository` once the migration is complete.

Result:
- Prepared migration to unify meta and dogma repositories


server/src/main/java/com/linecorp/centraldogma/server/internal/storage/project/DefaultProject.java
```diff
@@ -132,13 +147,24 @@ public class DefaultProject implements Project {
...
         repos = newRepoManager(rootDir, repositoryWorker, purgeWorker, cache);
 
+        final boolean useDogmaRepoAsMetaRepo;
+        if (dogmaProject == null) {
+            useDogmaRepoAsMetaRepo = false;
+        } else {
+            final Repository dogmaProjectDogmaRepository = dogmaProject.repos().get(REPO_DOGMA);
+            final Entry<JsonNode> entry = dogmaProjectDogmaRepository.getOrNull(
+                    Revision.HEAD, Query.ofJson(META_TO_DOGMA_MIGRATION_JOB)).join();
+            useDogmaRepoAsMetaRepo = entry != null;
```
ikhoon, line 157: If you intended to refresh `MetaRepository`, adding a new command to propagate the event seems like a complex approach. Instead, it might be simpler to add a property such as `{ "metaRepository: <repoName> }` in `metadata.json` to indicate which repository is used for `MetaRepository` and handle `resetMetaRepository()` logic inside `RepositoryListener`.
https://github.com/line/centraldogma/blob/920d7b115b5c27c9ea313eb3b0025606cf4b5a6f/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/project/DefaultProject.java#L262

minwoox: Thanks for the suggestion.
Here's the migration flow I'm planning:
- Set the project's Dogma repository to read-only
- Migrate data from the meta repository to the Dogma repository
- Add a /meta-to-dogma-migrated.json file to the Dogma repository
- Reset the meta repository on all replicas via the command
- Set the project's Dogma repository to active
- Proceed to the next project

The migration is a one-time operation, and the related logic will be removed afterward.
I tried to avoid polluting the existing logic as much as possible, and I felt it would be more intuitive for the leader to handle both the migration and status transition.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/project/DefaultProject.java
```diff
@@ -274,18 +300,31 @@ public Author author() {
...
-        }
+    private MetaRepository setMetaRepository(boolean useDogmaRepoAsMetaRepo) {
+        final String repoName = useDogmaRepoAsMetaRepo ? REPO_DOGMA : REPO_META;
+        final DefaultMetaRepository metaRepo = new DefaultMetaRepository(repos.get(repoName));
+        this.metaRepo = metaRepo;
+        return metaRepo;
+    }
+
+    @Override
+    public MetaRepository metaRepo() {
```
jrhee17, line 324: Note: I'm going to assume the returned value of this method isn't cached anywhere so that a stale version of `MetaRepository` isn't read

server/src/main/java/com/linecorp/centraldogma/server/command/StandaloneCommandExecutor.java
```diff
@@ -106,6 +108,10 @@ protected <T> CompletableFuture<T> doExecute(Command<T> command) throws Exceptio
             return (CompletableFuture<T>) createProject((CreateProjectCommand) command);
         }
 
+        if (command instanceof ResetMetaRepositoryCommand) {
```
jrhee17, line 111: Note) I understood `ResetMetaRepositoryCommand` will be issued as soon as the metadata is migrated.
Otherwise if there are any updates by the users to the old metadata, a stale version may be read when `resetMetaRepository` is called.

jrhee17 (approved):

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/project/DefaultProject.java
minwoox: That is true. Thanks for double-checking. 😉

server/src/main/java/com/linecorp/centraldogma/server/command/StandaloneCommandExecutor.java
minwoox: >  I understood ResetMetaRepositoryCommand will be issued as soon as the metadata is migrated

That is true. Also, I can check whether the repository is migrated correctly or not from this log after the command is executed:
https://github.com/line/centraldogma/pull/1123/files#diff-1b79753a6234a28cbdd41ef181ff57c2c3ed6d439b4e7fb6092994f35359d7beR214

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/project/DefaultProject.java
ikhoon: I see. As the code is only used during migration, I think it’s fine to leave it as is without spending extra effort.

ikhoon (approved):
👍👍

minwoox:
Thanks for reviewing. 😉

PR #1126, opened 2025-04-16
Allow overriding SAML ACS endpoint for proxy support

minwoox:
Motivation:
Currently, SAML endpoints are automatically generated using the default hostname and active port.
This behavior can cause assertion failures when Central Dogma is deployed behind a proxy,
as the active port may differ from the externally visible one.

Modifications:
- Added `SamlAuthProperties.acs` property to allow users to manually specify the ACS endpoint that matches the proxy's exposed URL.

Result:
- SAML authentication now works correctly when Central Dogma is behind a proxy by allowing the ACS endpoint to be explicitly configured.


ikhoon (approved):
👍👍

jrhee17 (approved):

PR #1128, opened 2025-04-17
Migrate meta repository to dogma repository

minwoox:
Motivation:  
Central Dogma has historically used both `dogma` and `meta` repositories to manage metadata.  
The `meta` repository existed mainly to store mirroring configurations and was accessible via the Web file UI.  
Since mirroring is now managed exclusively through a REST API and the `meta` repository is hidden from users,  
maintaining it separately no longer provides value.  
Combining them simplifies management and allows both to be encrypted using the same repository encryption key.

Modifications:  
- Added a migration job that moves contents from the `meta` repository into the `dogma` repository.

Result:  
- Metadata is now consolidated under the `dogma` repository.


server/src/main/java/com/linecorp/centraldogma/server/internal/storage/MigratingMetaToDogmaRepositoryService.java
```diff
@@ -0,0 +1,171 @@
...
+                return;
+            }
+
+            try {
+                commandExecutor.execute(Command.resetMetaRepository(Author.SYSTEM, projectName))
+                               .get(1, TimeUnit.MINUTES);
+            } catch (Throwable t) {
+                logger.warn("Failed to reset meta repository. project: {}", projectName, t);
+                // Do not continue the migration if the migration fails.
+                return;
```
jrhee17, line 121: If a migration fails, would it be safer to just set `RepositoryStatus` back to active?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/MigratingMetaToDogmaRepositoryService.java
```diff
@@ -0,0 +1,171 @@
...
+            logger.info("Dogma repository in the project '{}' is set to read-only.", projectName);
+            try {
+                migrateMetaRepository(project);
+                numMigratedProjects++;
+                logger.info("Meta repository in the project '{}' is migrated to dogma repository.",
+                            projectName);
+            } catch (Throwable t) {
+                logger.warn("Failed to migrate meta repository. project: {}", projectName, t);
+                // Do not continue the migration if the migration fails.
+                return;
```
jrhee17, line 112: If a migration fails, would it be safer to just set `RepositoryStatus` back to active?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/MigratingMetaToDogmaRepositoryService.java
```diff
@@ -0,0 +1,171 @@
...
+            }
+
+            if (REPO_DOGMA.equals(project.metaRepo().name())) {
+                logger.debug("The project '{}' has already been migrated.", projectName);
+                continue;
+            }
+            logger.info("Migrating meta repository to dogma repository in the project: {} ...", projectName);
+
+            metadataService.updateRepositoryStatus(
+                    Author.SYSTEM, projectName, REPO_DOGMA, RepositoryStatus.READ_ONLY).join();
```
jrhee17, line 100: Since existing servers won't understand the new command (and changes related to force push), I understood that the other changes will be released first, and then this PR will be released

jrhee17 (approved):

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/MigratingMetaToDogmaRepositoryService.java
minwoox: That's a good suggestion. 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/MigratingMetaToDogmaRepositoryService.java
minwoox: That is correct. 👍

ikhoon (approved):
🙇‍♂️🙇‍♂️

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/MigratingMetaToDogmaRepositoryService.java
```diff
@@ -0,0 +1,180 @@
...
+        final Map<String, Entry<?>> entryMap =
+                metaRepository.find(Revision.HEAD, "/credentials/*.json,/repos/**.json").get();
+        final ImmutableList.Builder<Change<?>> builder =
+                ImmutableList.builderWithExpectedSize(entryMap.size() + 1);
+        for (Map.Entry<String, Entry<?>> entry : entryMap.entrySet()) {
+            //noinspection unchecked
+            final Entry<JsonNode> value = (Entry<JsonNode>) entry.getValue();
+            builder.add(Change.ofJsonUpsert(entry.getKey(), value.content()));
+        }
+        builder.add(Change.ofJsonUpsert(META_TO_DOGMA_MIGRATED,
```
ikhoon, line 166: If this job fails after migrating some projects, those migrated projects start using the `dogma` repository to read and store metadata anyway. It means the metadata may be updated to the `dogma` repository after the failure. 

My concern is that the new data is overwritten by the old data in the `meta` repository if this job runs again. Should we skip the migration for a project if the project was migrated before?

minwoox: > Should we skip the migration for a project if the project was migrated before?

Yeah, it's already skipped. 😉 
https://github.com/line/centraldogma/pull/1128/files#diff-9fcf4e0be2d26c552e6c780fa3a045a8254cd705cfea370feb64086aff86ba26R106

PR #1129, opened 2025-04-17
Fix missing migration marker file in dogma repository

minwoox:
Motivation:
While implementing #1123, I forgot to add the `/meta-to-dogma-migrated.json` file to the `dogma` repository. This file is necessary to indicate that the migration from `meta` to `dogma` has been completed.

Modifications:
- Added the `/meta-to-dogma-migrated.json` file to the `dogma` repository when a project is created.

Result:
- The `dogma` repository now contains the correct marker file.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/project/DefaultProject.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/project/DefaultProject.java
@@ -22,6 +22,7 @@ import static com.linecorp.centraldogma.server.storage.project.InternalProjectIn
 import static java.util.Objects.requireNonNull;
 
 import java.io.File;
+import java.time.Instant;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
@@ -185,9 +186,20 @@ public class DefaultProject implements Project {
     private void createReservedRepos(long creationTimeMillis, boolean useDogmaRepoAsMetaRepo) {
         if (!repos.exists(REPO_DOGMA)) {
             try {
-                repos.create(REPO_DOGMA, creationTimeMillis, Author.SYSTEM);
+                final Repository dogmaRepository =
+                        repos.create(REPO_DOGMA, creationTimeMillis, Author.SYSTEM);
+                if (useDogmaRepoAsMetaRepo) {
+                    dogmaRepository.commit(
+                            Revision.HEAD, creationTimeMillis, Author.SYSTEM,
+                            "Add " + META_TO_DOGMA_MIGRATED + " file to dogma repository", "", Markup.PLAINTEXT,
+                            Change.ofJsonUpsert(META_TO_DOGMA_MIGRATED,
+                                                Jackson.writeValueAsString(
+                                                        ImmutableMap.of("timestamp", Instant.now())))).join();
+                }
             } catch (RepositoryExistsException ignored) {
                 // Just in case there's a race.
+            } catch (JsonProcessingException e) {
+                // Should never happen because map is used.
             }
         }
         if (!useDogmaRepoAsMetaRepo && !repos.exists(REPO_META)) {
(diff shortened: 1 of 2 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/storage/project/DefaultProject.java
```diff
@@ -185,9 +186,20 @@ private RepositoryManager newRepoManager(File rootDir, Executor repositoryWorker
     private void createReservedRepos(long creationTimeMillis, boolean useDogmaRepoAsMetaRepo) {
         if (!repos.exists(REPO_DOGMA)) {
             try {
-                repos.create(REPO_DOGMA, creationTimeMillis, Author.SYSTEM);
+                final Repository dogmaRepository =
+                        repos.create(REPO_DOGMA, creationTimeMillis, Author.SYSTEM);
+                if (useDogmaRepoAsMetaRepo) {
+                    dogmaRepository.commit(
```
jrhee17, line 192: Note) I understood for existing projects, will `META_TO_DOGMA_MIGRATED` be written manually inside the migration job.  I guess this logic exists in case projects are created before the migration job runs.

Minor question) Is it possible that a project is created with `useDogmaRepoAsMetaRepo = false` even though the migration job completes if a project is created at the same time as the migration job runs?

jrhee17 (approved):

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/project/DefaultProject.java
minwoox: >  I understood for existing projects, will META_TO_DOGMA_MIGRATED be written manually inside the migration job. I guess this logic exists in case projects are created before the migration job runs.

This is for the project that is created after the migration job runs.
- Migration job runs
- dogma/dogma has `"/meta-to-dogma-migration-job.json"` file
- a project is created without meta repository (e.g. `foo`)
- when the project is reopened it checks `"foo/dogma/meta-to-dogma-migrated.json"` and uses the `dogma` repository as the meta repository.
- Previoulsy, I forgot to add `"foo/dogma/meta-to-dogma-migrated.json"` in above case.

> Minor question) Is it possible that a project is created with useDogmaRepoAsMetaRepo = false even though the migration job completes if a project is created at the same time as the migration job runs?

That can happen. Let me run the migration job again to prevent the case. Thanks!

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/project/DefaultProject.java
```diff
@@ -185,9 +186,20 @@ private RepositoryManager newRepoManager(File rootDir, Executor repositoryWorker
...
+                            Revision.HEAD, creationTimeMillis, Author.SYSTEM,
+                            "Add " + META_TO_DOGMA_MIGRATED + " file to dogma repository", "", Markup.PLAINTEXT,
+                            Change.ofJsonUpsert(META_TO_DOGMA_MIGRATED,
+                                                Jackson.writeValueAsString(
+                                                        ImmutableMap.of("timestamp", Instant.now())))).join();
```
ikhoon, line 197: `CreateProjectCommand` is executed in each cluster. Does that mean `Change.ofJsonUpsert(META_TO_DOGMA_MIGRATED)` could raise `RedundantChangeException`?

minwoox: > Change.ofJsonUpsert(META_TO_DOGMA_MIGRATED) could raise RedundantChangeException?

I don't think so. This happens only after the migration job is completed. After the migration job is completed,
each replica creates its own dogma repository with this commit.

ikhoon (approved):
Thanks! 👍👍

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/project/DefaultProject.java
minwoox: It runs again:
https://github.com/line/centraldogma/pull/1128/files#diff-9fcf4e0be2d26c552e6c780fa3a045a8254cd705cfea370feb64086aff86ba26R90-R94

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/project/DefaultProject.java
ikhoon: Understood now. I was confused. The commit was submitted to a command executor.

Issue #421, opened 2019-07-19
Flaky test: `ArmeriaCentralDogmaBuilderTest.buildingWithProfile()`


ikhoon:
Travis-CI [build#869](https://travis-ci.com/line/centraldogma/builds/119732002#L939)  failed intermittently

```java
com.linecorp.centraldogma.client.armeria.ArmeriaCentralDogmaBuilderTest > buildingWithProfile FAILED
    org.awaitility.core.ConditionTimeoutException: Assertion condition defined as a lambda expression 
      in com.linecorp.centraldogma.client.armeria.ArmeriaCentralDogmaBuilderTest that uses com.linecorp.armeria.client.endpoint.EndpointGroup 
    Expecting:
      <[Endpoint{1.2.3.4.xip.io:36462, ipAddr=1.2.3.4, weight=1000}]>
    to contain exactly in any order:
      <[Endpoint{1.2.3.4.xip.io:36462, ipAddr=1.2.3.4, weight=1000},
        Endpoint{5.6.7.8.xip.io:8080, ipAddr=5.6.7.8, weight=1000}]>
    but could not find the following elements:
      <[Endpoint{5.6.7.8.xip.io:8080, ipAddr=5.6.7.8, weight=1000}]>
     within 10 seconds.
        at org.awaitility.core.ConditionAwaiter.await(ConditionAwaiter.java:145)
        at org.awaitility.core.AssertionCondition.await(AssertionCondition.java:122)
        at org.awaitility.core.AssertionCondition.await(AssertionCondition.java:32)
        at org.awaitility.core.ConditionFactory.until(ConditionFactory.java:902)
        at org.awaitility.core.ConditionFactory.untilAsserted(ConditionFactory.java:723)
        at com.linecorp.centraldogma.client.armeria.ArmeriaCentralDogmaBuilderTest.buildingWithProfile(ArmeriaCentralDogmaBuilderTest.java:69)
        Caused by:
        java.lang.AssertionError: 
        Expecting:
          <[Endpoint{1.2.3.4.xip.io:36462, ipAddr=1.2.3.4, weight=1000}]>
        to contain exactly in any order:
          <[Endpoint{1.2.3.4.xip.io:36462, ipAddr=1.2.3.4, weight=1000},
            Endpoint{5.6.7.8.xip.io:8080, ipAddr=5.6.7.8, weight=1000}]>
        but could not find the following elements:
[...]

PR #1137, opened 2025-04-24
Fix uploading artifact in GitHub Actions

minwoox:
I found that the artifact was not uploaded due to a mismatched name.


client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/AbstractArmeriaCentralDogmaBuilder.java
```diff
@@ -179,7 +179,7 @@ private EndpointGroup endpointGroup0() throws UnknownHostException {
         if (groups.size() == 1) {
             group = groups.get(0);
         } else {
-            group = new CompositeEndpointGroup(groups, EndpointSelectionStrategy.roundRobin());
+            group = EndpointGroup.of(EndpointSelectionStrategy.roundRobin(), groups);
```
minwoox, line 182: https://github.com/line/centraldogma/issues/421 still fails.

I found that there's a race condition in how endpoints are updated:
```
2025-04-25 00:54:39.959 [INFO ](c.l.a.c.e.DynamicEndpointGroup) [Test worker] New endpoints have been set: [4.3.2.1:3 (weight: 1000), 8.7.6.5:4 (weight: 1000)]
2025-04-25 00:54:39.960 [DEBUG](c.l.a.c.e.d.DnsAddressEndpointGroup) [armeria-common-worker-epoll-2-8] 5.6.7.8.foo.com Resolved: Endpoint{5.6.7.8.foo.com:2, ipAddr=5.6.7.8, weight=1000} (TTL: 60)
2025-04-25 00:54:39.960 [INFO ](c.l.a.c.e.DynamicEndpointGroup) [armeria-common-worker-epoll-2-8] New endpoints have been set: [5.6.7.8.foo.com/5.6.7.8:2 (weight: 1000)]
2025-04-25 00:54:39.960 [INFO ](c.l.a.c.e.DynamicEndpointGroup) [armeria-common-worker-epoll-2-8] New endpoints have been set: [1.2.3.4.foo.com/1.2.3.4:1 (weight: 1000), 4.3.2.1:3 (weight: 1000), 5.6.7.8.foo.com/5.6.7.8:2 (weight: 1000), 8.7.6.5:4 (weight: 1000)]
2025-04-25 00:54:39.960 [INFO ](c.l.a.c.e.DynamicEndpointGroup) [armeria-common-worker-epoll-2-7] New endpoints have been set: [1.2.3.4.foo.com/1.2.3.4:1 (weight: 1000), 4.3.2.1:3 (weight: 1000), 8.7.6.5:4 (weight: 1000)]
```

In the log above, the four endpoints should be set after the three-endpoint set, but in reality, the three-endpoint list is applied later. This means an outdated list may override a newer one.
It looks like this is caused by the following line, which doesn’t guarantee the order of method calls:
https://github.com/line/centraldogma/blob/main/client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/CompositeEndpointGroup.java#L57

I believe we can resolve this by switching to CompositeEndpointGroup from Armeria directly, which handles update ordering properly.

jrhee17 (approved):

client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/AbstractArmeriaCentralDogmaBuilder.java
minwoox: I also found a bug in `CompositeEndpointGroup` in Armeria. Let me fix that. 😓

ikhoon (approved):

Issue #963, opened 2024-06-19
Provide a easy way to import a directory or files into Central Dogma programmatically


trustin:
A user often needs to import the files in a filesytem into Central Dogma for testing. At the moment, a user need to create a project and a repository, build a commit and push it using the client API. Their life could be simplified a lot by providing the APIs like the following:

```java
public interface CentralDogma {
    // Imports the files in the given directory into Central Dogma.
    // First component in the path becomes the project name.
    // Second component in the path becomes the repository name.
    // e.g. foo/bar/alice.json will create a new project 'foo' and
    //      a new repository 'bar' under the project 'foo', and
    //      add 'alice.json' to the root directory of the repository 'bar'.
    //
    // Placeholder files like '.keep' and '.gitkeep' could be ignored automatically.
    CompletableFuture<Void> importDir(Path dir);
}

public interface CentralDogmaRepository {
    // Imports the files in the given directory into the repository.
    // Placeholder files like '.keep' and '.gitkeep' could be ignored automatically.
    CompletableFuture<PushResult> importDir(Path dir);
}
```

hyunw9:
I’m interested in contributing to Central Dogma and would love to help improve it.
Is it okay if I take this issue?

hyunw9:
Hello, I’m currently implementing ImportResult, and I noticed it’s not yet defined. 
Could you advise on which fields would be appropriate to include in this class?

jrhee17:
Taking another look at this issue

1) I'm unsure if this API will really be useful for the normal client path - the motivation for adding these APIs seems to be mostly for testing purposes only
2) Assuming this API is used for test purposes, I wonder if the API would make more sense if added on the server-side.

e.g.
```java
public class CentralDogmaRuleDelegate {

    @Nullable
    protected Path importDir() {
        return null;
    }
```

@trustin do you mind if commenting on whether you prefer the original approach or not?
I'm fine with either direction, but at the moment I'm not sure if I'm understanding the objective of the issue correctly.

minwoox:
> Assuming this API is used for test purposes, I wonder if the API would make more sense if added on the server-side.

I guess what @trustin meant was testing with a real server.
That’s probably why he specified the `ClassLoader` in his example.

PR #1140, opened 2025-05-10
Provide an easy way to import a directory into Central Dogma programmatically

hyunw9:
### Motivation

- Related Issus : #963 
- Previously, the user had to call each client API to import files ( createProject(), createRepository(), commit(), push())
- With this feature, users can import files or create projects more easily 

### Modification

**Interface Addition**  
  - Added `importDir(Path)`, `importResourceDir(String)`, and `importResourceDir(String, ClassLoader)` methods to the `CentralDogma` interface. 

- **Delegation in AbstractCentralDogma**  
  - Parses the given `Path` or `String` to determine the target `<project>/<repository>` and delegates the actual import logic to the corresponding `CentralDogmaRepository`.
  https://github.com/hyunw9/centraldogma/blob/48a97cdb816cfe6499293b677c580750b8ca762d/client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogma.java#L225-L233

- **Core Implementation in CentralDogmaRepository**  
  - Implements actual [file resolution](https://github.com/hyunw9/centraldogma/blob/48a97cdb816cfe6499293b677c580750b8ca762d/client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaRepository.java#L94-L108), and [change computation](https://github.com/hyunw9/centraldogma/blob/48a97cdb816cfe6499293b677c580750b8ca762d/client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaRepository.java#L78C5-L92C6).
  - Automatically creates the project/repository if they do not exist, using `createProject()` and `createRepository()` with proper exception handling (e.g., `ProjectExistsException`).

- **Business Logic (change building)**  
  - Local file changes are converted via [`toChange()`](https://github.com/hyunw9/centraldogma/blob/48a97cdb816cfe6499293b677c580750b8ca762d/client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaRepository.java#L78-L92), with placeholder files ignored.
  - For `importResourceDir()`, resource paths are resolved using `ClassLoader.getResource()` with appropriate fallbacks.
  - [Exclude predefined placeholder](https://github.com/hyunw9/centraldogma/blob/48a97cdb816cfe6499293b677c580750b8ca762d/client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaRepository.java#L98) files (e.g. .keep) during import

- **Testing**
  - Added unit tests for `CentralDogmaRepository` to verify import logic with mock `CentralDogma`.
  - Added integration tests (`ArmeriaCentralDogmaTest`) to confirm actual repository creation, content import, and correct diff results.

- **Design Note**
  - Considered using `join()` vs `thenCompose()` when handling project/repo creation.
 
<details>
<summary>detail</summary>
<div markdown="1">

There were cases where the `commit()` operation was attempted before the project or repository had been fully created, resulting in `ProjectNotFoundException` or `RepositoryNotFoundException`.

Initially, we handled this by chaining `createProject()` and `createRepository()` with `thenCompose()` and fallback logic to tolerate already-existing resources.  
Here is the original implementation:

```java
private CompletableFuture<Void> ensureProjectAndRepoExist() {
    final Function<Throwable, Throwable> unwrap = (t) -> {
        if (t instanceof CompletionException) {
            return t.getCause();
        }
        return t;
    };

    return centralDogma.createProject(projectName()).handle((r, ex) -> {
        if (ex != null && !(unwrap.apply(ex) instanceof ProjectExistsException)) {
            throw new CompletionException(unwrap.apply(ex));
        }
        return null;
    }).thenCompose(unused ->
        centralDogma.createRepository(projectName(), repositoryName())
                    .handle((r, ex) -> {
                        if (ex != null && !(unwrap.apply(ex) instanceof RepositoryExistsException)) {
                            throw new CompletionException(unwrap.apply(ex));
                        }
                        return null;
                    })
    ).thenApply(unused -> null);
}
```
While this ensures safe creation of the project and repository, it introduces multiple handle() blocks and deep exception wrapping, which may degrade readability and traceability.
</div>
</details>

### Result 

Without manually creating projects or repositories, users can now import files from both local file systems and classpath resources into Central Dogma with a single method call.

### DIFF
--- a/client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogma.java
+++ b/client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogma.java
@@ -235,12 +235,29 @@ public abstract class AbstractCentralDogma implements CentralDogma {
 
     @Override
     public CompletableFuture<ImportResult> importResourceDir(String dir) {
-        return null;
+        final Path path = Paths.get(dir);
+        if (path.getNameCount() < 2) {
+            return exceptionallyCompletedFuture(
+                    new IllegalArgumentException("Path must be <project>/<repo>[/…]: " + dir));
+        }
+        final String project = path.getName(0).toString();
+        final String repo = path.getName(1).toString();
+
+        return forRepo(project, repo).importResourceDir(dir);
     }
 
     @Override
     public CompletableFuture<ImportResult> importResourceDir(String dir, ClassLoader classLoader) {
-        return null;
+
+        final Path path = Paths.get(dir);
+        if (path.getNameCount() < 2) {
+            return exceptionallyCompletedFuture(
+                    new IllegalArgumentException("Path must be <project>/<repo>[/…]: " + dir));
+        }
+        final String project = path.getName(0).toString();
+        final String repo = path.getName(1).toString();
+
+        return forRepo(project, repo).importResourceDir(dir, classLoader);
     }
 
     /**
--- a/client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogma.java
+++ b/client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogma.java
@@ -17,15 +17,25 @@
 package com.linecorp.centraldogma.client;
 
 import static com.linecorp.centraldogma.internal.PathPatternUtil.toPathPattern;
+import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.Objects.requireNonNull;
 
+import java.io.File;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.List;
+import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.function.Function;
 
 import javax.annotation.Nullable;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.spotify.futures.CompletableFutures;
 
 import com.linecorp.centraldogma.common.Author;
 import com.linecorp.centraldogma.common.Change;
@@ -37,6 +47,7 @@ import com.linecorp.centraldogma.common.MergedEntry;
 import com.linecorp.centraldogma.common.PushResult;
 import com.linecorp.centraldogma.common.Query;
 import com.linecorp.centraldogma.common.Revision;
+import com.linecorp.centraldogma.internal.Jackson;
 
 import io.micrometer.core.instrument.MeterRegistry;
 
@@ -60,6 +71,25 @@ public abstract class AbstractCentralDogma implements CentralDogma {
                                    @Nullable MeterRegistry meterRegistry) {
         this.blockingTaskExecutor = requireNonNull(blockingTaskExecutor, "blockingTaskExecutor");
         this.meterRegistry = meterRegistry;
+    }
+
+    private static Change<?> toChange(String repoPath, Path file) {
+        try {
+            byte[] bytes = Files.readAllBytes(file);
+            String name = file.getFileName().toString().toLowerCase();
+
+            if (name.endsWith(".json")) {
+                return Change.ofJsonUpsert(repoPath, Jackson.readTree(bytes));
+            }
+            if (name.endsWith(".yml") || name.endsWith(".yaml") ||
+                (Files.probeContentType(file) != null &&
+                 Files.probeContentType(file).startsWith("text/"))) {
+                return Change.ofTextUpsert(repoPath, new String(bytes, UTF_8));
+            }
+            return Change.ofTextUpsert(repoPath, new String(bytes, UTF_8));
+        } catch (Exception e) {
+            throw new IllegalStateException("Failed to build Change for " + file, e);
+        }
     }
 
     /**
@@ -190,6 +220,44 @@ public abstract class AbstractCentralDogma implements CentralDogma {
                                                                 .map(function)
                                                                 .mapperExecutor(executor)
                                                                 .start();
+    }
+
+    @Override
+    public CompletableFuture<ImportResult> importDir(Path dir) {
+        requireNonNull(dir, "dir");
+        if (dir.getNameCount() < 2) {
+            return CompletableFutures.exceptionallyCompletedFuture(new IllegalArgumentException(
+                    "Path must be <project>/<repo>[/…]: " + dir));
+        }
+        // todo : should refactor this method ( reference backward )
+
+        final String project = dir.getName(0).toString();
+        final String repo = dir.getName(1).toString();
+        final Path norm = dir.toAbsolutePath().normalize();
+        final CentralDogmaRepository centralDogmaRepository = forRepo(project, repo);
+
+        if (dir.getNameCount() == 2 || Files.isDirectory(norm)) {
+            return centralDogmaRepository.importDir(norm);
+        }
+
+        // single file
+        final Path rel = norm.subpath(2, norm.getNameCount());
+        final String logical = '/' + rel.toString().replace(File.separatorChar, '/');
+        final Change<?> ch = toChange(logical, norm);
+
+        return centralDogmaRepository.commit("Import " + logical, ch)
+                .push(Revision.HEAD)
+                .thenApply(ImportResult::fromPushResult);
+    }
+
+    @Override
+    public CompletableFuture<ImportResult> importResourceDir(String dir) {
+        return null;
+    }
+
+    @Override
+    public CompletableFuture<ImportResult> importResourceDir(String dir, ClassLoader classLoader) {
+        return null;
     }
 
     /**
--- a/client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogma.java
+++ b/client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogma.java
@@ -20,12 +20,9 @@ import static com.linecorp.centraldogma.internal.PathPatternUtil.toPathPattern;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.Objects.requireNonNull;
 
-import java.io.File;
-import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.List;
-import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ScheduledExecutorService;
@@ -33,8 +30,6 @@ import java.util.function.Function;
 
 import javax.annotation.Nullable;
 
-import com.fasterxml.jackson.databind.JsonNode;
-import com.fasterxml.jackson.databind.ObjectMapper;
 import com.spotify.futures.CompletableFutures;
 
 import com.linecorp.centraldogma.common.Author;
@@ -229,25 +224,15 @@ public abstract class AbstractCentralDogma implements CentralDogma {
             return CompletableFutures.exceptionallyCompletedFuture(new IllegalArgumentException(
                     "Path must be <project>/<repo>[/…]: " + dir));
         }
-        // todo : should refactor this method ( reference backward )
-
         final String project = dir.getName(0).toString();
         final String repo = dir.getName(1).toString();
         final Path norm = dir.toAbsolutePath().normalize();
         final CentralDogmaRepository centralDogmaRepository = forRepo(project, repo);
+        final CentralDogma centralDogma = centralDogmaRepository.centralDogma();
 
-        if (dir.getNameCount() == 2 || Files.isDirectory(norm)) {
-            return centralDogmaRepository.importDir(norm);
-        }
-
-        // single file
-        final Path rel = norm.subpath(2, norm.getNameCount());
-        final String logical = '/' + rel.toString().replace(File.separatorChar, '/');
-        final Change<?> ch = toChange(logical, norm);
-
-        return centralDogmaRepository.commit("Import " + logical, ch)
-                .push(Revision.HEAD)
-                .thenApply(ImportResult::fromPushResult);
+        return centralDogma.createProject(project)
+                           .thenCompose(unused -> centralDogma.createRepository(project, repo))
+                           .thenCompose(a -> centralDogmaRepository.importDir(norm));
     }
 
     @Override
--- a/client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogma.java
+++ b/client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogma.java
@@ -17,11 +17,14 @@
 package com.linecorp.centraldogma.client;
 
 import static com.linecorp.centraldogma.internal.PathPatternUtil.toPathPattern;
+import static com.spotify.futures.CompletableFutures.exceptionallyCompletedFuture;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.Objects.requireNonNull;
 
+import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executor;
@@ -219,20 +222,15 @@ public abstract class AbstractCentralDogma implements CentralDogma {
 
     @Override
     public CompletableFuture<ImportResult> importDir(Path dir) {
-        requireNonNull(dir, "dir");
         if (dir.getNameCount() < 2) {
-            return CompletableFutures.exceptionallyCompletedFuture(new IllegalArgumentException(
+            return exceptionallyCompletedFuture(new IllegalArgumentException(
                     "Path must be <project>/<repo>[/…]: " + dir));
         }
         final String project = dir.getName(0).toString();
         final String repo = dir.getName(1).toString();
         final Path norm = dir.toAbsolutePath().normalize();
-        final CentralDogmaRepository centralDogmaRepository = forRepo(project, repo);
-        final CentralDogma centralDogma = centralDogmaRepository.centralDogma();
 
-        return centralDogma.createProject(project)
-                           .thenCompose(unused -> centralDogma.createRepository(project, repo))
-                           .thenCompose(a -> centralDogmaRepository.importDir(norm));
+        return forRepo(project, repo).importDir(norm);
     }
 
     @Override
--- a/client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java
+++ b/client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java
@@ -420,7 +420,34 @@ public interface CentralDogma extends AutoCloseable {
     CompletableFuture<List<Change<?>>> getDiff(String projectName, String repositoryName,
                                                Revision from, Revision to, PathPattern pathPattern);
 
+    /**
+     * Imports files from the given absolute {@link Path} into a repository.
+     * <p>
+     * The {@code dir} must follow the format: {@code /<project>/<repository>[/path/to/files]}.
+     * If a single file is specified, only that file is imported. If a directory is specified,
+     * all eligible files under it will be imported.
+     */
     CompletableFuture<ImportResult> importDir(Path dir);
+
+    /**
+     * Imports resource files from the given directory located in the default classloader.
+     * <p>
+     * The path must follow the format: {@code <project>/<repository>[/resourcePath]}.
+     * The specified directory is resolved from the current classpath.
+     */
+    CompletableFuture<ImportResult> importResourceDir(String dir);
+
+    /**
+     * Imports resource files from the given directory using the specified {@link ClassLoader}.
+     * <p>
+     * This is useful for testing or loading from custom classpaths such as in embedded environments.
+     * The path must follow the format: {@code <project>/<repository>[/resourcePath]}.
+     *
+     * @param dir the resource directory path (project/repository/optionalSubPath)
+     * @param classLoader the class loader used to locate the resource directory
+     */
+    CompletableFuture<ImportResult> importResourceDir(String dir, ClassLoader classLoader);
+    
     /**
      * Retrieves the diffs of the files matched by the given path pattern between two {@link Revision}s.
      *
--- a/client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java
+++ b/client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java
@@ -18,6 +18,7 @@ package com.linecorp.centraldogma.client;
 import static com.linecorp.centraldogma.internal.PathPatternUtil.toPathPattern;
 import static java.util.Objects.requireNonNull;
 
+import java.nio.file.Path;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -419,6 +420,7 @@ public interface CentralDogma extends AutoCloseable {
     CompletableFuture<List<Change<?>>> getDiff(String projectName, String repositoryName,
                                                Revision from, Revision to, PathPattern pathPattern);
 
+    CompletableFuture<ImportResult> importDir(Path dir);
     /**
      * Retrieves the diffs of the files matched by the given path pattern between two {@link Revision}s.
      *
--- a/client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaRepository.java
+++ b/client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaRepository.java
@@ -19,9 +19,11 @@ import static java.util.Objects.requireNonNull;
 
 import java.io.File;
 import java.io.IOException;
+import java.net.URL;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -30,7 +32,9 @@ import java.util.List;
 import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.function.Function;
 import java.util.stream.Stream;
 
 import javax.annotation.Nullable;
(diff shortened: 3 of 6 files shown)


client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaRepository.java
```diff
@@ -298,6 +352,72 @@ public WatcherRequest<Revision> watcher(PathPattern pathPattern) {
...
+        if (Files.isRegularFile(physicalPath)) {
+            final String repoPath = '/' + logical.getFileName()
+                                                 .toString()
+                                                 .replace(File.separatorChar, '/');
+            changes.add(toChange(repoPath, physicalPath));
+        } else {
+            changes.addAll(collectImportFiles(physicalPath));
+        }
+
+        if (changes.isEmpty()) {
```
minwoox, line 413: I think we should raise an exception in this case

minwoox (commented):
Thanks and sorry for the late review. 😉

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogma.java
```diff
@@ -192,6 +201,46 @@ public <T> Watcher<T> repositoryWatcher(String projectName, String repositoryNam
...
+    public CompletableFuture<ImportResult> importDir(Path dir) {
+        if (dir.getNameCount() < 2) {
+            return exceptionallyCompletedFuture(new IllegalArgumentException(
+                    "Path must be <project>/<repo>[/…]: " + dir));
+        }
+        final String project = dir.getName(0).toString();
+        final String repo = dir.getName(1).toString();
+        final Path norm = dir.toAbsolutePath().normalize();
+
+        return forRepo(project, repo).importDir(norm);
```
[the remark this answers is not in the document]
hyunw9: I got it ! 

Currently, the `importDir()` method delegates the creation of the project and repository, but the behavior is not clearly visible or explicit.

Would it make sense to implement this logic directly in `AbstractCentraDogma` for clarity?

```
// #importDir
 return createProjectIfAbsent(project).thenCompose(unusedProjectCreated ->
             createRepositoryIfAbsent(project, repo).thenCompose(unusedRepoCreated ->
                 forRepo(project, repo).importDir(norm)));

private CompletableFuture<Void> createProjectIfAbsent(String project) {
        return listProjects().thenCompose(projectSet -> {
            if (projectSet.contains(project)) {
                return CompletableFuture.completedFuture(null);
            }
            return createProject(project);
        });
    }

    private CompletableFuture<CentralDogmaRepository> createRepositoryIfAbsent(String project, String repo) {
        return listRepositories(project).thenCompose(repoSet -> {
            if (repoSet.containsKey(repo)) {
                return CompletableFuture.completedFuture(null);
            }
            return createRepository(project, repo);
        });
    }
```

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java
```diff
@@ -419,6 +420,34 @@ <T> CompletableFuture<Change<T>> getDiff(String projectName, String repositoryNa
...
                                                Revision from, Revision to, PathPattern pathPattern);
 
+    /**
+     * Imports files from the given absolute {@link Path} into a repository.
+     * <p>
+     * The {@code dir} must follow the format: {@code /<project>/<repository>[/path/to/files]}.
+     * If a single file is specified, only that file is imported. If a directory is specified,
+     * all eligible files under it will be imported.
+     */
+    CompletableFuture<ImportResult> importDir(Path dir);
```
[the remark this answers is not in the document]
hyunw9: Now that i think about this, since our interface already enforces a `project/repo/file` style, there's no strong justification for supporting absolute paths.
Allowing only relative paths makes the API contract more explicit and avoids potential ambiguity across different operating systems. 
If we add a check like the following, we can prevent side effects. 
I'd love to hear your thoughts on this :)
```
if (dir.isAbsolute()) {
    throw new IllegalArgumentException("Only relative paths are supported: " + dir);
}
```

hyunw9:
Thanks for the review! 
I've left some replies to your comments.

Once things are clarified and we're on the same page, I'll go ahead and apply the suggested changes ! .

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogma.java
```diff
@@ -192,6 +201,46 @@ public <T> Watcher<T> repositoryWatcher(String projectName, String repositoryNam
                                                                 .start();
     }
 
+    @Override
+    public CompletableFuture<ImportResult> importDir(Path dir) {
+        if (dir.getNameCount() < 2) {
```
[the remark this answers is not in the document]
minwoox: > I want to know that this approach is reasonable ! 🤔

Absolutely. 😄 I was confused. Two names will be enough. 😉

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java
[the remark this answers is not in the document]
minwoox: Yeah let's do that. 👍

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaRepository.java
minwoox: Yeah, let's do that. 👍

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogma.java
```diff
@@ -192,6 +199,69 @@ public <T> Watcher<T> repositoryWatcher(String projectName, String repositoryNam
...
+            return exceptionallyCompletedFuture(new IllegalArgumentException(
+                    "Path must be <project>/<repo>[/…]: " + dir));
+        }
+        if (dir.isAbsolute()){
+            return exceptionallyCompletedFuture(
+                    new IllegalArgumentException("Only relative paths are supported: " + dir));
+        }
```
ikhoon: Should we support an absolute path? It would be inconvenient if we had to move the directory to the location where `java` runs.

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogma.java
```diff
@@ -192,6 +199,69 @@ public <T> Watcher<T> repositoryWatcher(String projectName, String repositoryNam
...
+    @Override
+    public CompletableFuture<ImportResult> importResourceDir(String dir) {
+        final Path path = Paths.get(dir);
+        if (path.getNameCount() < 2) {
+            return exceptionallyCompletedFuture(
+                    new IllegalArgumentException("Path must be <project>/<repo>[/…]: " + dir));
+        }
+        if (path.isAbsolute()){
+            return exceptionallyCompletedFuture(
+                    new IllegalArgumentException("Only relative paths are supported: " + dir));
```
ikhoon: Ditto.

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaRepository.java
```diff
@@ -59,6 +71,37 @@ public final class CentralDogmaRepository {
         this.meterRegistry = meterRegistry;
     }
 
+    private static Change<?> toChange(String repoPath, Path file) {
+        try {
+            final byte[] bytes = Files.readAllBytes(file);
+            final String lower = file.getFileName().toString().toLowerCase();
+
+            if (lower.endsWith(".json")) {
+                return Change.ofJsonUpsert(repoPath, Jackson.readTree(bytes));
```
ikhoon: Should we use `ObjectMapper.readTree(File)` API to convert a file into `JsonNode`?

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaRepository.java
```diff
@@ -298,6 +341,61 @@ public WatcherRequest<Revision> watcher(PathPattern pathPattern) {
...
+        if (!"file".equals(url.getProtocol())) {
+            return CompletableFutures.exceptionallyCompletedFuture(
+                    new IllegalArgumentException("Resource dir must be explodable (got " + url + ')'));
+        }
+
+        final Path physicalPath = Paths.get(url.getPath());
+        final List<Change<?>> changes = toChanges(physicalPath);
+        return commit("Import " + physicalPath.getFileName(), changes)
+                .push(Revision.HEAD)
+                .thenApply(ImportResult::fromPushResult);
```
ikhoon: Should we reuse `PushResult`? Is `ImportResult` introduced for `isEmpty()`?

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaRepository.java
hyunw9: Thanks for the review : ) How about using Jackson.readValue(byte[] data, Class<T> type)?

```
JsonNode node = Jackson.readValue(bytes, JsonNode.class);
return Change.ofJsonUpsert(repoPath, node);
```

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaRepository.java
hyunw9: I think that's right, it'd be better to fix it!

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogma.java
```diff
@@ -192,6 +196,35 @@ public <T> Watcher<T> repositoryWatcher(String projectName, String repositoryNam
...
 
+    @Override
+    public CompletableFuture<PushResult> importDir(@Nullable String projectName,
+                                                   @Nullable String repositoryName, Path dir,
+                                                   boolean createIfMissing) {
+        requireNonNull(dir, "dir");
+
+        if (!dir.isAbsolute()) {
+            return exceptionallyCompletedFuture(
+                    new IllegalArgumentException("dir must be an absolute path: " + dir)
```
minwoox: I think we can just throw the exception

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogma.java
```diff
@@ -206,4 +239,79 @@ protected final CompletableFuture<Revision> maybeNormalizeRevision(
...
+
+        if (projectName == null && repositoryName == null) {
+            return extractProjectAndRepoFromPath(dir);
+        }
+
+        if (projectName != null && repositoryName == null) {
+            return Arrays.asList(projectName, extractRepoNameFromPath(dir));
+        }
+
+        return Arrays.asList(projectName, repositoryName);
```
minwoox: How about:

```suggestion
        if (projectName == null) {
            if (repositoryName != null) {
                throw new IllegalArgumentException("projectName cannot be null when repositoryName is not null");    
            }
            return extractProjectAndRepoFromPath(dir);
        }
        
        final String repositoryName0 = repositoryName != null ? repositoryName
                                                              : extractRepoNameFromPath(dir);
        return Arrays.asList(projectName, repositoryName0);
```

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogma.java
```diff
@@ -206,4 +239,79 @@ protected final CompletableFuture<Revision> maybeNormalizeRevision(
...
+    }
+
+    /**
+     * Extracts both project and repository names from the directory path.
+     * Uses the second-to-last component as project name and the last component as repository name.
+     */
+    private List<String> extractProjectAndRepoFromPath(Path dir) {
```
minwoox:
```suggestion
    private static List<String> extractProjectAndRepoFromPath(Path dir) {
```

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogma.java
```diff
@@ -206,4 +239,79 @@ protected final CompletableFuture<Revision> maybeNormalizeRevision(
...
+        final String project = dir.getName(dir.getNameCount() - 2).toString();
+        final String repo = dir.getName(dir.getNameCount() - 1).toString();
+        return Arrays.asList(project, repo);
+    }
+
+    /**
+     * Extracts the repository name from the directory path.
+     * Uses the last component as repository name.
+     */
+    private String extractRepoNameFromPath(Path dir) {
```
minwoox: static

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogma.java
```diff
@@ -192,6 +196,35 @@ public <T> Watcher<T> repositoryWatcher(String projectName, String repositoryNam
...
+        }
+
+        final List<String> resolvedNames = resolveProjectAndRepoNames(projectName, repositoryName, dir);
+        final String finalProjectName = resolvedNames.get(0);
+        final String finalRepositoryName = resolvedNames.get(1);
+
+        final CompletableFuture<Void> ensureProjectRepo = ensureProjectAndRepo(finalProjectName,
+                                                                               finalRepositoryName,
+                                                                               createIfMissing);
+        final CentralDogmaRepository repo = forRepo(finalProjectName, finalRepositoryName);
```
minwoox: Let's create `CentralDogmaRepository` in `ensureProjectRepo.thenCompose(...)`.

client/java-armeria/src/test/java/com/linecorp/centraldogma/client/armeria/CentralDogmaImportTest.java
```diff
@@ -0,0 +1,277 @@
+/*
+ * Copyright 2020 LINE Corporation
```
minwoox: nit:
```suggestion
 * Copyright 2025 LINE Corporation
```

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaRepository.java
```diff
@@ -298,6 +352,19 @@ public WatcherRequest<Revision> watcher(PathPattern pathPattern) {
         return new WatcherRequest<>(this, pathPattern, blockingTaskExecutor, meterRegistry);
     }
 
+    /**
+     * Returns a new {@link CompletableFuture} that imports the files in the specified.
```
minwoox: How about: Imports the content of the specified {@code dir} to the repository.

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaRepository.java
```diff
@@ -59,6 +70,49 @@ public final class CentralDogmaRepository {
...
+
+            return Change.ofTextUpsert(repoPath, new String(bytes, StandardCharsets.UTF_8));
+        } catch (Exception e) {
+            throw new IllegalStateException("Failed to create Change for " + file, e);
+        }
+    }
+
+    private static List<Change<?>> toChanges(Path path) {
+        final List<Change<?>> changes = new ArrayList<>();
+        if (Files.isRegularFile(path)) {
```
minwoox: What do you think of rasing an exception if the path is not dir in `importDir` method?

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaRepository.java
```diff
@@ -59,6 +70,49 @@ public final class CentralDogmaRepository {
         this.meterRegistry = meterRegistry;
     }
 
+    private static Change<?> toChange(String repoPath, Path file) {
+        try {
+            final byte[] bytes = Files.readAllBytes(file);
+            final String lower = file.getFileName().toString().toLowerCase();
```
minwoox: I don't think the file name should only contain lowercase letters. Additionally, there will be a collision if a directory contains both files whose lowercased names are the same. e.g. a.json and A.json

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaRepository.java
```diff
@@ -59,6 +70,49 @@ public final class CentralDogmaRepository {
...
+    }
+
+    private static List<Change<?>> collectImportFiles(Path path) {
+        final List<Change<?>> changes = new ArrayList<>();
+        try (Stream<Path> s = Files.walk(path)) {
+            s.filter(Files::isRegularFile)
+             .filter(p -> !p.getFileName().toString().startsWith("."))
+             .forEach(p -> {
+                 final Path rel = path.relativize(p);
+                 final String repoPath = '/' + rel.toString().replace(File.separatorChar, '/');
```
minwoox: nit: filePath?

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaRepository.java
minwoox: But, `isEmpty()` isn't used anywhere. 😆  Can we just use `PushResult` instead of introducing a new class?

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaRepository.java
hyunw9: I agree. i will update the code  👍

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogma.java
hyunw9: It seems right ! I just updated the code. :)

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogma.java
hyunw9: It seems more simple and readable! i will accept this change :)

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaRepository.java
hyunw9: While AbstractCentralDogma already handles directory validation, adding the same validation to CentralDogmaRepository.importDir() would make the method’s contract clearer !

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogma.java
```diff
@@ -192,6 +196,34 @@ public <T> Watcher<T> repositoryWatcher(String projectName, String repositoryNam
...
+                                                   @Nullable String repositoryName, Path dir,
+                                                   boolean createIfMissing) {
+        requireNonNull(dir, "dir");
+
+        if (!dir.isAbsolute()) {
+            throw new IllegalArgumentException("dir must be an absolute path: " + dir);
+        }
+        if (!Files.exists(dir) || !Files.isDirectory(dir)) {
+            return exceptionallyCompletedFuture(
+                    new IllegalArgumentException("dir must be an existing directory: " + dir)
```
minwoox: ditto: I think we can just raise the exception

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractCentralDogma.java
```diff
@@ -206,4 +238,77 @@ protected final CompletableFuture<Revision> maybeNormalizeRevision(
...
+        return listProjects().thenCompose(projects -> {
+            if (!projects.contains(project)) {
+                if (!createIfMissing) {
+                    return exceptionallyCompletedFuture(
+                            new IllegalStateException("Project does not exist: " + project));
+                }
+                return createProject(project);
+            }
+            return CompletableFuture.completedFuture(null);
+        }).thenCompose(unused -> listRepositories(project).thenCompose(repos -> {
```
minwoox, line 250: Note to other reviewers: I will handle `ProjectExistsException` later after introducing Armeria dependency.

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaRepository.java
```diff
@@ -59,6 +70,49 @@ public final class CentralDogmaRepository {
...
+        if (Files.isRegularFile(path)) {
+            final String repoPath = '/' + path.getFileName().toString().replace(File.separatorChar, '/');
+            changes.add(toChange(repoPath, path));
+        } else {
+            changes.addAll(collectImportFiles(path));
+        }
+        return changes;
+    }
+
+    private static List<Change<?>> collectImportFiles(Path path) {
```
minwoox:
```suggestion
```
We don't need this anymore

client/java/src/main/java/com/linecorp/centraldogma/client/ImportResult.java
```diff
@@ -0,0 +1,105 @@
...
+
+import static java.util.Objects.requireNonNull;
+
+import com.linecorp.centraldogma.common.PushResult;
+import com.linecorp.centraldogma.common.Revision;
+
+/**
+ * The wrapper of the result of an import operation.
+ */
+public final class ImportResult {
```
minwoox: Let's remove this class. 😉

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java
```diff
@@ -419,6 +422,33 @@ <T> CompletableFuture<Change<T>> getDiff(String projectName, String repositoryNa
...
+     * <li><strong>(non-null, non-null)</strong>: Both project and repository names are provided and used
+     * as-is. <br>Example: {@code importDir("myproject", "myrepo", "/home/user/data")} → project:
+     * "myproject", repository: "myrepo"</li>
+     * </ul>
+     *
+     * @param projectName the project name, or {@code null} to extract from the directory path
+     * @param repositoryName the repository name, or {@code null} to extract from the directory path
+     * @param dir the directory to import
+     * @param createIfMissing whether to create the project and repository if they don't exist
+     * @return a {@link CompletableFuture} that completes with the {@link PushResult}
```
minwoox: Better to add this? If no files were imported, the future completes with null.

client/java-armeria/src/test/java/com/linecorp/centraldogma/client/armeria/CentralDogmaImportTest.java
```diff
@@ -0,0 +1,276 @@
...
+import com.linecorp.centraldogma.common.Change;
+import com.linecorp.centraldogma.common.EntryType;
+import com.linecorp.centraldogma.common.PathPattern;
+import com.linecorp.centraldogma.common.PushResult;
+import com.linecorp.centraldogma.common.Query;
+import com.linecorp.centraldogma.common.Revision;
+
+class CentralDogmaImportTest {
+
+    ScheduledExecutorService scheduler;
```
minwoox: This isn't used.

client/java-armeria/src/test/java/com/linecorp/centraldogma/client/armeria/CentralDogmaImportTest.java
```diff
@@ -0,0 +1,276 @@
...
+
+import com.linecorp.centraldogma.client.CentralDogma;
+import com.linecorp.centraldogma.common.Change;
+import com.linecorp.centraldogma.common.EntryType;
+import com.linecorp.centraldogma.common.PathPattern;
+import com.linecorp.centraldogma.common.PushResult;
+import com.linecorp.centraldogma.common.Query;
+import com.linecorp.centraldogma.common.Revision;
+
+class CentralDogmaImportTest {
```
minwoox, line 46: This test doens't work because there's no Central Dogma server.
Could you update the test using `CentralDogmaExtension`?
reference: https://github.com/line/centraldogma/blob/38dc4085303f46a6268545aaa13b5a439e7bea95/client/java-armeria/src/test/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogmaTest.java#L33

hyunw9: Thanks for suggestion! i just fixed failing tests by using CentralDogmaExtension. :)

client/java-armeria/src/test/java/com/linecorp/centraldogma/client/armeria/CentralDogmaImportTest.java
hyunw9: thanks for the review! i removed unused dependency. 👍

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogma.java
hyunw9: I updated javadoc :)

client/java-armeria/src/test/java/com/linecorp/centraldogma/client/armeria/CentralDogmaImportTest.java
```diff
@@ -0,0 +1,239 @@
...
+import com.linecorp.centraldogma.common.Revision;
+import com.linecorp.centraldogma.testing.junit.CentralDogmaExtension;
+
+class CentralDogmaImportTest {
+
+    @RegisterExtension
+    static CentralDogmaExtension dogmaExt = new CentralDogmaExtension() {
+        @Override
+        protected void scaffold(CentralDogma client) {
+        }
```
minwoox:
```suggestion
```

client/java-armeria/src/test/java/com/linecorp/centraldogma/client/armeria/CentralDogmaImportTest.java
```diff
@@ -0,0 +1,239 @@
...
+                      dogmaExt.serverAddress().getPort())
+                .build();
+    }
+
+    @AfterEach
+    void tearDown() throws Exception {
+        try {
+            dogma.close();
+        } finally {
+            removeCreatedDir(Paths.get(dir).toAbsolutePath());
```
minwoox: Have you consider using `@TempDir` which deletes the file automatically?

hyunw9: This contribution required a considerable amount of time, but exploring the Central Dogma was an exciting and rewarding experience. I look forward to more opportunities to contribute :) Thank you for your continuous reviews and guidance !

minwoox: I'm really glad to hear you had a positive experience contributing. Look forward to your next pull request! Thanks a lot. 😆

minwoox (approved):
👍 👍 👍

client/java/src/main/java/com/linecorp/centraldogma/client/CentralDogmaRepository.java
```diff
@@ -59,6 +70,38 @@ public final class CentralDogmaRepository {
...
+        } catch (Exception e) {
+            throw new IllegalStateException("Failed to create Change for " + file, e);
+        }
+    }
+
+    private static List<Change<?>> collectImportFiles(Path path) {
+        final List<Change<?>> changes = new ArrayList<>();
+        try (Stream<Path> s = Files.walk(path)) {
+            s.filter(Files::isRegularFile)
+             .filter(p -> !p.getFileName().toString().startsWith("."))
```
ikhoon, line 93: The current implementation looks good, but it would be nice to apply the rules from `.gitignore` files instead of the hard-coded rule. Maybe we can handle this as follow-up.

Example:
https://github.com/line/centraldogma/blob/3a261ba4ec9685b27d5287b239503df4bce767e5/server-mirror-git/src/main/java/com/linecorp/centraldogma/server/internal/mirror/AbstractGitMirror.java#L303-L304

ikhoon (approved):
Thanks, @hyunw9! 👍🙇‍♂️

minwoox:
@hyunw9 👍 👍 👍

PR #1142, opened 2025-05-21
Support basic access authentication

ikhoon:
Motivation:

Git over HTTP implementation in Central Dogma uses Bearer token-based authentication, which is not a standard Git HTTP authentication method.
https://git-scm.com/docs/http-protocol#_authentication
> Servers SHOULD support Basic authentication by relying upon the HTTP server placed in front of the Git server software.

Users should use `http.extraheader` to set the bearer token.
```
git clone -c http.extraheader="Authorization: Bearer <token>" https://dogma.com/foo/bar.git`
```
Supporting Basic authentication would make it eaiser to integate with systems like Git CLI and Spring Cloud Config Server that use usernames and passwords.

Modifications:

- Include `WWW-Authenticate` header field for 401 Unauthorized response to prompt the client to use Basic auth.
- Introduce `AbstractAuthorizer` to centralize the token parsing logic.
- When Basic authenitication is used:
  - `username` is fixed to `dogma`
  - `password` is set to the application token.
  -  The Authorization header will appear as: `Basic base64(dogma:appToken-...)`
- Fix GitHttpServiceTest to set username and password instead of Bearer tokens.

Result:

You can use basic authentication to access Central Dogma resources.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/admin/auth/AbstractAuthorizer.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/admin/auth/AbstractAuthorizer.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2025 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.linecorp.centraldogma.server.internal.admin.auth;
+
+import java.util.concurrent.CompletionStage;
+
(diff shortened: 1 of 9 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/admin/auth/AbstractAuthorizer.java
```diff
@@ -0,0 +1,89 @@
...
+
+        return authorize(ctx, req, token);
+    }
+
+    protected abstract CompletionStage<Boolean> authorize(ServiceRequestContext ctx, HttpRequest req,
+                                                          String accessToken);
+
+    @Nullable
+    private static String extractToken(ServiceRequestContext ctx, HttpRequest req) {
+        String token = ctx.attr(TOKEN_KEY);
```
jrhee17, line 57: nit; personally I'm not sure if this additional complexity is worth it compared to the cost of header lookups, but I understood the changes.

jrhee17 (approved):

minwoox (approved):
I should've done this before. Thanks a lot! 😄

server/src/main/java/com/linecorp/centraldogma/server/internal/admin/auth/AbstractAuthorizer.java
ikhoon: A parsed token is stored in the attributes to avoid duplicate parsing logic, not just to skip the headers lookup.

PR #1144, opened 2025-06-05
Introduce LoggerService

minwoox:
Motivation:  
To enable runtime logger level management via HTTP endpoints, a dedicated `LoggerService` is needed.  

Modifications:  
- Implemented `LoggerService` with HTTP API endpoints to get and update logger levels.
- Modified `CentralDogmaRuleDelegate` to return `CsrfToken.ANONYMOUS` instead of `null` as a default token.

Result:  
- Users can now inspect and update logger levels at runtime via HTTP.

## Summary by CodeRabbit

- **New Features**
  - Introduced a logging management API, allowing system administrators to view and update logger levels via new HTTP endpoints.
- **Tests**
  - Added tests to verify logging API behavior, including querying and updating logger levels.
- **Chores**
  - Improved internal handling of access tokens for testing, standardizing on an anonymous token instead of null.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/api/LoggerService.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/api/LoggerService.java
@@ -39,45 +39,17 @@ import com.linecorp.centraldogma.server.internal.api.auth.RequiresSystemAdminist
 import ch.qos.logback.classic.Level;
 import ch.qos.logback.classic.Logger;
 import ch.qos.logback.classic.LoggerContext;
-import ch.qos.logback.classic.spi.LoggerContextListener;
 
 @RequiresSystemAdministrator
 @ProducesJson
 public class LoggerService {
 
+    private static final org.slf4j.Logger logger = LoggerFactory.getLogger(LoggerService.class);
+
     private static final List<String> LEVELS = ImmutableList.of(
             "ALL", "TRACE", "DEBUG", "INFO", "WARN", "ERROR", "OFF");
 
     private final LoggerContext loggerContext = (LoggerContext) LoggerFactory.getILoggerFactory();
-
-    @Nullable
-    private volatile List<LoggerInfo> loggerInfos;
-
-    /**
-     * Creates a new instance.
-     */
-    public LoggerService() {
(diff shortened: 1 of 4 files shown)


jrhee17 (approved):

server/src/main/java/com/linecorp/centraldogma/server/internal/api/LoggerService.java
```diff
@@ -0,0 +1,179 @@
...
+        if (loggerInfos != null) {
+            return HttpResponse.ofJson(loggerInfos);
+        }
+
+        final Builder<LoggerInfo> builder = ImmutableList.builder();
+        for (Logger logger : loggerContext.getLoggerList()) {
+            builder.add(LoggerInfo.of(logger));
+        }
+        loggerInfos = builder.build();
+        this.loggerInfos = loggerInfos;
```
ikhoon: The API is unlikely to be called frequently, so caching may not be worthwhile.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/LoggerService.java
```diff
@@ -0,0 +1,179 @@
...
+        }
+
+        final String level = levelNode.asText();
+        final String upperCase = level.toUpperCase();
+        if (!LEVELS.contains(upperCase)) {
+            return HttpResponse.of(HttpStatus.BAD_REQUEST, MediaType.PLAIN_TEXT_UTF_8,
+                                   "Invalid log level: " + level + ". Valid levels are: " + LEVELS);
+        }
+        logger.setLevel(Level.toLevel(upperCase));
+        return HttpResponse.ofJson(LoggerInfo.of(logger));
```
ikhoon, line 88: Should we add an info or debug message to track which logging level was changed?

minwoox: That's a good suggestion. 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/api/LoggerService.java
minwoox: Agreed. Fixed. 😉

PR #1145, opened 2025-06-11
Avoid showing the sign-up prompt during the login flow.

ikhoon:
Motivation:

In the login flow, if the session set in `/api/v0/users/me` is not valid, `WWW-Authenticate` header is included in the response headers. The header could cause a brower prompt a sign-up form.

Motification:

- Fix `CentralDogmaAuthFailureHandler` not to include `WWW_AUTHENTICATE` for `/api/v0/users/me`

Result:

The login flow is not interrupted by a basic authentication prompt.

Note:

Since a new release hasn’t been made yet, it doesn’t affect users.


minwoox (approved):
👍

PR #1146, opened 2025-06-20
Introduce Data Encryption Key (DEK) per Object

minwoox:
Motivation:
Currently, if a repository's encryption key is rotated, all stored objects must be decrypted and re-encrypted. This process is expensive and time-consuming. By introducing a per-object Data Encryption Key (DEK), we can enable more efficient key rotation—only rewrapping the DEKs rather than re-encrypting the entire data.

Modifications:
- Updated encryption logic to use object-specific DEKs wrapped by the repository's Data Encryption Key.

Result:
- Will reduce overhead during key rotation by avoiding full data re-encryption.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/rocksdb/EncryptionGitStorage.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/rocksdb/EncryptionGitStorage.java
@@ -15,7 +15,9 @@
  */
 package com.linecorp.centraldogma.server.internal.storage.repository.git.rocksdb;
 
+import static com.linecorp.centraldogma.server.internal.storage.AesGcmSivCipher.KEY_SIZE_BYTES;
 import static com.linecorp.centraldogma.server.internal.storage.AesGcmSivCipher.NONCE_SIZE_BYTES;
+import static com.linecorp.centraldogma.server.internal.storage.AesGcmSivCipher.aesSecretKey;
 import static org.eclipse.jgit.lib.Constants.R_REFS;
 import static org.eclipse.jgit.lib.Constants.encode;
 import static org.eclipse.jgit.lib.ObjectReader.OBJ_ANY;
@@ -27,6 +29,7 @@ import java.nio.charset.StandardCharsets;
 
 import javax.annotation.Nullable;
 import javax.crypto.SecretKey;
+import javax.crypto.spec.SecretKeySpec;
 
 import org.eclipse.jgit.errors.IncorrectObjectTypeException;
 import org.eclipse.jgit.internal.storage.file.RefDirectory;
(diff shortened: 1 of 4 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/rocksdb/EncryptionGitStorage.java
```diff
@@ -94,18 +97,27 @@ ObjectId insertObject(ObjectId objectId, int type, byte[] data, int off, int len
         }
 
         final byte[] nonce = AesGcmSivCipher.generateNonce();
-        final byte[] nonceAndType = new byte[NONCE_SIZE_BYTES + 4];
+        // Generate a new DEK for the data so that we don't decrypt and encrypt the data again when the
+        // repository key is rotated.
```
ikhoon, line 101: Implementation looks fine. Just a couple of questions:
- Does a repository's DEK remain unchanged after it is created?
- If we use DEK per object, do you think rewrapping achieves the same goal as key rotate?

minwoox: > Does a repository's DEK remain unchanged after it is created?
If we use DEK per object, do you think rewrapping achieves the same goal as key rotate?

The repository's DEK is rotated. When the repository's DEK is rotated, the old repository's DEK decrypts this object DEK and the new repository's DEK re-encrypts this object DEK. The object DEK is never rotated and it's only used once for the corresponding object. This is used in envelope-encryption:
https://jayendrapatil.com/tag/envelope-encryption/

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/rocksdb/EncryptionGitStorage.java
```diff
@@ -94,18 +97,27 @@ ObjectId insertObject(ObjectId objectId, int type, byte[] data, int off, int len
...
+        System.arraycopy(nonce, 0, nonceAndObjectWdek, 0, NONCE_SIZE_BYTES);
+        nonceAndObjectWdek[NONCE_SIZE_BYTES] = (byte) (type >> 24);
+        nonceAndObjectWdek[NONCE_SIZE_BYTES + 1] = (byte) (type >> 16);
+        nonceAndObjectWdek[NONCE_SIZE_BYTES + 2] = (byte) (type >> 8);
+        nonceAndObjectWdek[NONCE_SIZE_BYTES + 3] = (byte) type;
+        System.arraycopy(objectWdek, 0, nonceAndObjectWdek, NONCE_SIZE_BYTES + 4, objectWdek.length);
+
+        final SecretKeySpec keySpec = aesSecretKey(objectDek);
 
         final byte[] encryptedId = encryptObjectId(nonce, objectId);
```
jrhee17: Question) If the objective is to ensure easier rotation, does it make sense that both the id and value is encrypted with the `objectWdek`?

i.e. If the `EncryptionGitStorage#dek` is modified, will all objects need to be rewritten since `encryptObjectId` is encrypted directly with `EncryptionGitStorage#dek`?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/rocksdb/EncryptionGitStorage.java
```diff
@@ -94,18 +97,27 @@ ObjectId insertObject(ObjectId objectId, int type, byte[] data, int off, int len
         }
 
         final byte[] nonce = AesGcmSivCipher.generateNonce();
-        final byte[] nonceAndType = new byte[NONCE_SIZE_BYTES + 4];
+        // Generate a new DEK for the data so that we don't decrypt and encrypt the data again when the
```
jrhee17, line 100: Note) I understood that if the metadata is stored separately in the future, this can provide an extra layer of security in case users directly try to access the objects without the metadata

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/rocksdb/EncryptionGitStorage.java
```diff
@@ -94,18 +97,27 @@ ObjectId insertObject(ObjectId objectId, int type, byte[] data, int off, int len
...
+        // repository key is rotated.
+        final byte[] objectDek = AesGcmSivCipher.generateAes256Key();
+        final byte[] objectWdek = encrypt(nonce, objectDek, 0, KEY_SIZE_BYTES);
```
jrhee17, line 103: I assume that the purpose of this change is to only rewrite metadata (and not objects) when the `dek` is rotated. Let me know if I am misunderstanding the overall intention of this PR.

jrhee17 (approved):

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/rocksdb/EncryptionGitStorage.java
minwoox: Oops, yeah the ID also must be encrypted using the `objectDek`. Thanks, updated!

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/rocksdb/EncryptionGitStorage.java
minwoox: > only rewrite metadata (and not objects) when the dek is rotated.

That's true. 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/rocksdb/EncryptionGitStorage.java
minwoox: That’s true, but I don't think the benefit outweighs the added complexity. The object DEK is already wrapped, so the current data structure will be fine.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/rocksdb/EncryptionGitStorage.java
minwoox: @ikhoon Would you mind reviewing my answer?

ikhoon: > When the repository's DEK is rotated, the old repository's DEK decrypts this object DEK and the new repository's DEK re-encrypts this object DEK.

Understood. 

Thanks for the link. I didn't know that it is a common practice to generate DEK for each data item.

minwoox:
Thanks for the review. 😉

PR #1147, opened 2025-06-27
Reduce the default maxFrameLength from 10MiB to 1MiB

ikhoon:
Motivation:

When a Central Dogma server received a large push data of over 4.2 MiB, it was observed that a ZooKeeper lock on the repository was held for more than 1 minute. If a lock is held for more than 1 minute, subsequent write operations to the repository may fail to acquire the lock and the server enters readonly mode.

The original `maxFrameLength` seems too large to be handled safely.

Modifications:

- Reduce the default `maxFrameLength` to 1MiB. users approximately write up to 10,000 lines per file which would be generally enough for a configuration file.

Result:

The default `maxFrameLength` has been reduced to 1 MiB.

## Summary by CodeRabbit

* **Documentation**
  * Updated docs to state the default maximum request size is now 1 MiB.

* **New Features**
  * Default maximum request size of 1 MiB applied when unspecified.
  * Clear handling and messaging for requests that exceed the size limit.
  * Improved error response rendering for better client feedback.

* **Bug Fixes**
  * Consistent enforcement of the request size limit and appropriate exceptions.

* **Tests**
  * Added test verifying oversized requests are rejected with the specific exception.

* **Chores**
  * Introduced the new default to ensure consistent server behavior.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/api/HttpApiUtil.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/api/HttpApiUtil.java
@@ -214,8 +214,11 @@ public final class HttpApiUtil {
             ResponseHeadersBuilder headersBuilder,
             @Nullable Throwable cause, @Nullable String message) {
         final HttpStatus status = headersBuilder.status();
-        checkArgument(!status.isContentAlwaysEmpty(),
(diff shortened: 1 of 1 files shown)


minwoox:
Oops, please check out the CI failures.

ikhoon:
> Oops, please check out the CI failures.

Fixed.

PR #1149, opened 2025-07-02
Do not enter readonly mode on lock timeout

ikhoon:
Motivation:

A lock could be held by a request for a long time when the size is large. While this is not desiable, it isn't necessarily a system error. So, failing the request makes more sense rather than entering readonly mode.

Modifications:

- Do not stop ZooKeeper replication when a lock can be acquired within one minute.
  - Change the log level to `WARN` from `ERROR`.
  - If a request timeout is configured to be longer than the lock timeout, `LockAcquireTimeoutException` will be returned.
- Add `ApiRequestTimeoutException` for converting Armeria `RequestTimeoutException` to a `CentralDogmaException`.
  - Also update the client error factories.

Result:

A lock timeout no longer triggers readonly mode.

## Summary by CodeRabbit

* **New Features**
  * Introduced new error messages for request timeouts and lock acquisition timeouts to provide clearer feedback when operations exceed their allowed time.
* **Bug Fixes**
  * Improved server responses for timeout scenarios, returning more appropriate status codes and error details.
* **Tests**
  * Added comprehensive tests to verify correct handling and reporting of timeout-related errors.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/api/HttpApiExceptionHandler.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/api/HttpApiExceptionHandler.java
@@ -29,14 +29,17 @@ import com.linecorp.armeria.common.HttpStatus;
 import com.linecorp.armeria.common.util.Exceptions;
 import com.linecorp.armeria.server.HttpResponseException;
 import com.linecorp.armeria.server.HttpStatusException;
+import com.linecorp.armeria.server.RequestTimeoutException;
 import com.linecorp.armeria.server.ServerErrorHandler;
 import com.linecorp.armeria.server.ServiceRequestContext;
 import com.linecorp.armeria.server.annotation.ExceptionHandlerFunction;
+import com.linecorp.centraldogma.common.ApiRequestTimeoutException;
 import com.linecorp.centraldogma.common.AuthorizationException;
 import com.linecorp.centraldogma.common.ChangeConflictException;
 import com.linecorp.centraldogma.common.EntryNoContentException;
 import com.linecorp.centraldogma.common.EntryNotFoundException;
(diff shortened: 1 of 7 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/api/HttpApiExceptionHandler.java
```diff
@@ -162,6 +167,12 @@ public HttpResponse onServiceException(ServiceRequestContext ctx, Throwable caus
             return newResponse(ctx, HttpStatus.SERVICE_UNAVAILABLE, peeledCause);
         }
 
+        if (peeledCause instanceof RequestTimeoutException) {
+            ctx.setShouldReportUnloggedExceptions(false);
+            return newResponse(ctx, HttpStatus.SERVICE_UNAVAILABLE,
+                               new ApiRequestTimeoutException("Request timed out", peeledCause));
```
jrhee17, line 173: Question) Is there a reason a separate `ApiRequestTimeoutException` is defined instead of using `RequestTimeoutException` directly?

```suggestion
                               peeledCause);
```

jrhee17 (approved):

minwoox (approved):
👍 👍 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/api/HttpApiExceptionHandler.java
ikhoon: `ApiRequestTimeoutException` was added so that the client can deserialize the exception and throw it:
- `RequestTimeoutException` does not have a public constructor to create with a message.
- `CentralDogma` client is designed to deserialize only `CentralDogmaException` types, and I preferred to keep the behavior unchanged.

PR #1156, opened 2025-07-28
Update credential pattern from webapp

sh-cho:
https://github.com/line/centraldogma/blob/29250c7233fc2899388bb250f4055d4c6d59bd13/common/src/main/java/com/linecorp/centraldogma/internal/CredentialUtil.java#L28-L29

On server-side, credential ID pattern is like above, but on webapp pattern is different and can cause error like below

```json
{
    "exception": "java.lang.IllegalArgumentException",
    "message": "invalid project credentialName: projects/local-proj/credentials/1111 (expected: ^projects/([^/]+)/credentials/([a-z](?:[a-z0-9-_.]{0,61}[a-z0-9])?)$)",
    "detail": "java.lang.IllegalArgumentException: invalid project credentialName: projects/local-proj/credentials/1111 (expected: ^projects/([^/]+)/credentials/([a-z](?:[a-z0-9-_.]{0,61}[a-z0-9])?)$)\n\tat com.google.common.base.Preconditions.checkArgument(Preconditions.java:446)\n\tat com.linecorp.centraldogma.internal.CredentialUtil.validateProjectCredentialName(CredentialUtil.java:63)\n\..."
}
```

### DIFF
--- a/webapp/src/dogma/features/project/settings/credentials/CredentialForm.tsx
+++ b/webapp/src/dogma/features/project/settings/credentials/CredentialForm.tsx
@@ -135,7 +135,7 @@ const CredentialForm = ({
               readOnly={!isNew}
               placeholder="The credential ID"
               defaultValue={defaultValue.id}
-              {...register('id', { required: true, pattern: /^[a-zA-Z0-9-_.]+$/ })}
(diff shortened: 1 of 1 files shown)


minwoox (approved):
Oops Thanks!

PR #1157, opened 2025-07-29
Abort watching on the initial failure

ikhoon:
Motivation:

If the initial fetch of `Watcher` fails with non-`CentralDogmaException`, it is likely that some configuration is wrong or the file is malformed. So earily completing the initialValueFuture with the exception rather than retrying indefinitely would help users to debug the reason.

Modifications:

- Complete `initialValueFuture` exceptionally when:
  - `initialValueFuture` is not completed yet
  - The initial attempts fail three times in a row

Result:

`Watcher` client stops without retrying if the first attempt fails with a non-`CentralDogmaException`.

## Summary by CodeRabbit

* **New Features**
  * Improved error handling for watcher setup failures by limiting initial fetch retries and providing clearer failure feedback.

* **Tests**
  * Added tests to verify that watcher errors during initialization are correctly propagated to users.

### DIFF
--- a/client/java/src/main/java/com/linecorp/centraldogma/client/AbstractWatcher.java
+++ b/client/java/src/main/java/com/linecorp/centraldogma/client/AbstractWatcher.java
@@ -65,6 +65,8 @@ abstract class AbstractWatcher<T> implements Watcher<T> {
             "centraldogma.client.watcher.latest.received.time";
     private static final AtomicLong WATCHER_ID = new AtomicLong();
 
+    private static final int MAX_INITIAL_FETCH_ATTEMPTS = 2;
+
     private enum State {
         INIT,
         STARTED,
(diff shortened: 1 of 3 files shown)


client/java/src/main/java/com/linecorp/centraldogma/client/AbstractWatcher.java
```diff
@@ -329,11 +329,20 @@ private void doWatch(int numAttemptsSoFar) {
...
-                     // Cancelled by close()
-                     return null;
+                     if (!initialValueFuture.isDone()) {
+                         // If the initial fetch failed with non-CentralDogmaException, it is likely that some
+                         // configuration is wrong or the file is malformed. So we complete the
+                         // initialValueFuture exceptionally rather than retrying indefinitely.
+                         initialValueFuture.completeExceptionally(thrown);
+                         close();
+                         return null;
+                     }
```
YongjoonSeo: Previously, it seems that if the network was temporarily unstable when the initial value had not been fetched, it would retry. However, with the new logic, it simply fails. Can this case be ignored, or should it be considered as well?

jrhee17 (approved):

client/java/src/main/java/com/linecorp/centraldogma/client/AbstractWatcher.java
ikhoon: Good point. I configured a `RetryingClient` to automatically retry on network failures or server errors.

ikhoon: On second thought, `RetryingClient` is not a good approach since Central Dogma watch uses long polling. `RetryingClient` will retain the references of requests for a long time. I switched to the implementation to retry up to 3 times when `initialValueFuture` is not completed.

PR #1159, opened 2025-08-11
Drop Java 8 support

ikhoon:
Motivation:

We had been maintaining Java 8 for LY legacy systems but now their Java versions have been upgraded to 11. Therefore, we no longer need to stick with Java 8.

Modifications:

- Use Java 11 for the base Java requirement
- Replace `java11` flags with `java` in `settings.gradle` since `java` now implies `java11`.
- Remove Java 8 from the CI build matrix

Result:

Java 8 is no longer supported.

## Summary by CodeRabbit

* **Chores**
  * Raised the project Java baseline to Java 11 for builds and compiler settings.
  * Standardized module profiles to use a generic Java profile instead of a java11-specific tag.
  * CI updated to stop Java 8 testing on Ubuntu while retaining Java 11, 17 and a Java 21 snapshot where present.


jrhee17 (approved):

minwoox (commented):
Do you also have a plan to put the server-mirror-git module to server module?

minwoox:
I think we need to also remove jgit5 usage.

Issue #1148, opened 2025-07-02
Admin page to update server status


ikhoon:
- A server's executor status can be updated only via the REST API in `ServerStatusService`.
https://github.com/line/centraldogma/blob/c2d559d418e1b32130395aebb1db7d561e8fa284/server/src/main/java/com/linecorp/centraldogma/server/internal/api/sysadmin/ServerStatusService.java#L58-L67
- A server's health can't be updated externally. 
https://github.com/line/centraldogma/blob/3d08000d62317537f34776bfe2dbdd0621f020ca/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java#L235

If an admin page is provided to manage the statuses, administrators can easily monitor and quickly update the statuses, saving time on operations.
https://github.com/line/centraldogma/blob/efb85f5870d8204151142580087c7a35d8704cf6/webapp/src/pages/app/settings/index.tsx#L20

minwoox:
That's going to be very useful. 👍

ikhoon:
Closed by #1163

PR #1163, opened 2025-08-13
Add admin menu to update server status

sh-cho:
## Motivation
- https://github.com/line/centraldogma/issues/1148
  - Currently server executor's status can only be updated via REST api

## Modification
- Add admin menu to update server status

## Result

- Server status can be updated in admin page

https://github.com/user-attachments/assets/960e9b4a-900d-4c46-883f-d9804591e16a

## Summary by CodeRabbit

* **New Features**
  * Admin-only “Server Status” tab in Settings.
  * View current server status with color-coded badges (Writable, Replication Only, Read Only).
  * Change server status via confirmation modal with loading state.
  * Choose propagation scope for updates (ALL or LOCAL) with contextual guidance.
  * Success, warning, and error notifications for update flows.
  * Display automatically refreshes after successful changes.

### DIFF
--- a/webapp/src/dogma/features/settings/server-status/UpdateServerStatus.tsx
+++ b/webapp/src/dogma/features/settings/server-status/UpdateServerStatus.tsx
@@ -79,15 +79,12 @@ export const UpdateServerStatus = ({
   const isUpdatingToSameStatusWithAllScope = currentStatus === selectedStatus && selectedScope === 'ALL';
   const isUpdatingReadOnlyToAllScope =
     currentStatus === 'READ_ONLY' && currentStatus !== selectedStatus && selectedScope === 'ALL';
-  const isUpdateDisabled = !selectedStatus || (currentStatus && (isUpdatingToSameStatusWithAllScope || isUpdatingReadOnlyToAllScope));
+  const isUpdateDisabled =
+    !selectedStatus || (currentStatus && (isUpdatingToSameStatusWithAllScope || isUpdatingReadOnlyToAllScope));
 
   return (
     <>
-      <Button
-        colorScheme="blue"
-        onClick={onOpen}
-        isDisabled={isUpdateDisabled}
-      >
+      <Button colorScheme="blue" onClick={onOpen} isDisabled={isUpdateDisabled}>
         Update Server Status
       </Button>
       <Modal isOpen={isOpen} onClose={onClose}>


webapp/src/dogma/common/components/Navbar.tsx
```diff
@@ -124,6 +124,11 @@ export const Navbar = () => {
                 <MenuItem as={RouteLink} href="/app/settings/tokens">
                   Application tokens
                 </MenuItem>
+                {user.systemAdmin && (
```
minwoox: ditto. 🤔

minwoox (approved):
This is exactly what I wanted to add. Thanks a lot, @sh-cho! 🙇

webapp/src/dogma/common/components/Navbar.tsx
sh-cho: ok removed :)

webapp/src/pages/app/settings/server-status/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+                onChange={(e) => setSelectedScope(e.target.value as ServerStatusScope)}
+                width={{ base: '100%', md: '300px' }}
+              >
+                <option value="ALL">ALL (Propagate to all cluster servers)</option>
+                <option value="LOCAL">LOCAL (Apply only to this server)</option>
+              </Select>
+              <Text fontSize="sm" color="gray.500" mt="1">
+                Note: Setting &#39;Replication Only&#39; or &#39;Writable&#39; with{' '}
+                <Code variant="outline">ALL</Code> scope is not allowed if the cluster is currently not
+                replicating. Use <Code variant="outline">LOCAL</Code> scope for individual instance changes.
```
jrhee17, line 95: Optional) Since we know the current state and the next state, we could optionally disable the button if a user tries to change the state from READ-ONLY a different state with ALL.
Having said this, as this is a page only for admins, I guess the current as-is is fine as well.

jrhee17 (approved):

webapp/src/pages/app/settings/server-status/index.tsx
sh-cho: Making button disable based on state will be more convenient
I'll apply it

sh-cho: https://github.com/user-attachments/assets/df75b30f-c5ca-43bc-a5ea-4e2ce7452a57

[`a06e65c` (#1163)](https://github.com/line/centraldogma/pull/1163/commits/a06e65ced6f5b44dba177b419643f3a5f90bd93a)

done :)

ikhoon (approved):
It looks super useful for us. Thanks, @sh-cho! 🙇‍♂️🙇‍♂️

minwoox:
🙇

PR #1166, opened 2025-08-19
Fix redirection after SAML authentication

minwoox:
Motivation:
Redirection was broken after SAML authentication. Previously, the `ref` parameter was used to convey the redirection URL,
but it is no longer handled since the migration to Next.js.

Modifications:
- Restored handling of the `ref` parameter to properly redirect users after SAML authentication.

Result:
- Users are redirected to the intended page after successful SAML login.

## Summary by CodeRabbit

* New Features
  * Login and logout now preserve the user’s previous page, returning there after authentication in both hosted and local environments.
* Bug Fixes
  * Hardened SSO redirect handling: validates relay paths, defaults invalid inputs to home, and escapes values to prevent incorrect redirects and injection.
* Documentation
  * Added third‑party license notice for the OWASP Encoder.
* Chores
  * Introduced OWASP Encoder as a new dependency and added its license.
* Tests
  * Expanded tests covering redirect path validation and escaping.


webapp/src/dogma/common/components/Navbar.tsx
```diff
@@ -129,10 +129,13 @@ export const Navbar = () => {
                   onClick={async () => {
                     await dispatch(logout());
                     if (typeof window !== 'undefined') {
+                      const returnPath = encodeURIComponent(
+                        `${window.location.pathname}${window.location.search}`,
```
ikhoon, line 133: Should we also add a URL fragment to `returnPath` if it exists?

minwoox: URL fragment is not set to the server so we need to treat it specially.
But, I thought we didn't use the fragment so we don't have to take care of it. Please let me know if I am wrong.

ikhoon: The Chrome browser specially handles `#:~text=...` fragments to scroll down to a text in a page. https://chromestatus.com/feature/4733392803332096

If we don't store the encoded fragment in `returnPath`, users can't use the feature when they first log in to Central Dogma.

minwoox: It's Chrome browser's feature. Now I get it.
Supporting that feature is not critical and the relayState cannot be exceeded 80 characters, let's implement it later when it becomes really needed.

minwoox: FYI: https://github.com/line/armeria/blob/b2aa9f49b46a7b0e03d8b8d753809cd1e8e2016c/saml/src/main/java/com/linecorp/armeria/server/saml/HttpRedirectBindingUtil.java#L113-L115

ikhoon: > let's implement it later when it becomes really needed.

I see.

> FYI: [line/armeria@b2aa9f4/saml/src/main/java/com/linecorp/armeria/server/saml/HttpRedirectBindingUtil.java#L113-L115](https://github.com/line/armeria/blob/b2aa9f49b46a7b0e03d8b8d753809cd1e8e2016c/saml/src/main/java/com/linecorp/armeria/server/saml/HttpRedirectBindingUtil.java#L113-L115)

Regarding the length of `relayState`, does a login fail with an exception if `returnPath` is longer than 80 bytes? `returnPath` may easily exceed 80 bytes if it points to a file in a deeply nested directory.

minwoox: >  does a login fail with an exception if returnPath is longer than 80 bytes? returnPath may easily exceed 80 bytes if it points to a file in a deeply nested directory.

That's correct but I thought the chances are low.
If that becomes a problem, we can just ignore the relayState instead of raising an exception.

minwoox: Discussed with the team, and we've decided to make the relayState limit configurable.
Currently, we ignore the relayState whose length is greater than 80:
https://github.com/line/centraldogma/blob/3ac766a99242010efee2c3a984851f99f8c4a89f/server-auth/saml/src/main/java/com/linecorp/centraldogma/server/auth/saml/SamlAuthSsoHandler.java#L100-L103
So let me merge this for now. 

Will make configurable after updating Armeria version that includes this patch:
https://github.com/line/armeria/pull/6381

PR #1169, opened 2025-08-25
Make logback a required implementation dependency

minwoox:
Motivation:
After introducing `LoggerService`, the logback dependency must not be optional; it should be a required implementation.

Additionally, since Java 8 support has been dropped, we can upgrade to logback 1.5 with SLF4J 2.x.

Modifications:
- Changed logback dependency scope from optional to implementation.
- Upgraded logback to 1.5 (compatible with SLF4J 2.x).

Result:
- Logging is now consistently available without optional dependency issues.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -78,28 +78,20 @@ configure(projectsWithFlags('java')) {
         // Jetty ALPN support
         compileOnly libs.jetty.alpn.api
 
-        // Logging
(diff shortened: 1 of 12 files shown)


minwoox:
> I would rather make LoggerService optional, i.e., only activate it when Logback is the current logging provider.

Thanks! Addressed. 😉

ikhoon (approved):
👍 👍

PR #1172, opened 2025-09-01
Optimize docker image

sh-cho:
Motivation:

Currently docker image has all client binaries inside, also copied twice under `bin/*` and `bin/native/*`. Only single client binary file is enough. Image below is layer info inside `ghcr.io/line/centraldogma:0.77.2` (used [dive](https://github.com/wagoodman/dive))
Also linux/arm64 image doesn't have linux/arm64 client binary.

Modification:

- Use centraldogma-go 0.1.1 version. Since 0.1.1, `linux/arm64` binary is available.
- In dockerfile, copy only one client binary for same target architecture
- Other changes
  - Remove gradle docker plugin, as it is not used since #1158
  - Client binary name is changed. It uses underscore now. (ex. `dogma.linux-amd64` -> `dogma.linux_amd64`)

Result:

- Only one client binary is included in docker image. image size is shrinked from 379M to 331M (-12.6%, checked with `docker inspect`)

- linux/arm64 image contains linux/arm64 client binary

## Summary by CodeRabbit

* Chores
  * Removed Docker-related entries from the build configuration to streamline the toolchain and reduce maintenance.
  * May slightly improve build times and avoid fetching unused artifacts in local and CI environments.
  * No user-facing functionality or interfaces were changed; application behavior remains the same.

### DIFF
--- a/dist/Dockerfile
+++ b/dist/Dockerfile
@@ -3,6 +3,9 @@
 #
 FROM openjdk:11-jre
 
+# Multi-platform build arguments (pre-defined by docker)
+ARG TARGETARCH
+
 # Environment variables.
 ENV CENTRALDOGMA_HOME "/opt/centraldogma"
 ENV CENTRALDOGMA_OPTS ""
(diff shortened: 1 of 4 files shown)


minwoox:
> If client binary name shouldn't be changed, it's also ok but string replace codes would be placed, because client artifact name is currently using underscores (ex. dogma_0.1.1_darwin_amd64.tar.gz)

I just read this. 👍

dist/Dockerfile
```diff
@@ -21,12 +24,16 @@ RUN mkdir -p                      \
...
-COPY build/dist/bin/native/* "$CENTRALDOGMA_HOME"/bin/native/
+COPY build/dist/bin/common* build/dist/bin/dogma build/dist/bin/shutdown build/dist/bin/startup \
+  "$CENTRALDOGMA_HOME"/bin/
 COPY build/dist/conf/*       "$CENTRALDOGMA_HOME"/conf/
 COPY build/dist/lib/*        "$CENTRALDOGMA_HOME"/lib/
 COPY build/dist/licenses/*   "$CENTRALDOGMA_HOME"/licenses/
 
+COPY --chmod=755 \
+  build/dist/bin/native/dogma.linux_${TARGETARCH} \
+  "$CENTRALDOGMA_HOME"/bin/native/
```
ikhoon, line 35: 👍

PR #1176, opened 2025-09-10
Assign ownership roles to access tokens when creating projects and repositories

minwoox:
Motivation:
Currently, when projects or repositories are created using an access token, the token is not automatically granted the corresponding owner/admin roles. This results in inconsistent permissions, as the token that initiated the creation does not have proper control afterward.

Modifications:
- Grant the token project owner role when creating a project.
- Grant the token repository admin role when creating a repository.
  - Since any project member can create a repository, explicit ADMIN permission is required.
- Standardize token user email format by appending `@dogma-token.local` to the token app ID, instead of adding the remote host string.
  - The remote host string was meaningless in this context.
  - This change allows clear distinction between user accounts and token accounts without introducing a separate `AuthorType`, since authors must still be retrieved from Git logs.

Result:
- Access tokens used to create projects and repositories now have appropriate owner/admin roles.
- Improved clarity in author identification between users and tokens.

### DIFF
--- a/common/src/main/java/com/linecorp/centraldogma/common/Author.java
+++ b/common/src/main/java/com/linecorp/centraldogma/common/Author.java
@@ -32,12 +32,12 @@ public class Author {
     /**
      * The system author.
      */
-    public static final Author SYSTEM = new Author("System", "system@localhost.localdomain");
+    public static final Author SYSTEM = new Author("system", "system@localhost.localdomain");
 
     /**
      * The default author which is used when security is disabled.
      */
-    public static final Author DEFAULT = new Author("User", "user@localhost.localdomain");
+    public static final Author DEFAULT = new Author("user", "user@localhost.localdomain");
 
     /**
      * An unknown author.
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/api/RepositoryServiceUtil.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/api/RepositoryServiceUtil.java
@@ -15,15 +15,25 @@
  */
 package com.linecorp.centraldogma.server.internal.api;
 
+import static com.linecorp.centraldogma.internal.Util.TOKEN_EMAIL_SUFFIX;
+import static com.linecorp.centraldogma.server.metadata.RepositoryMetadata.DEFAULT_PROJECT_ROLES;
+
+import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 
 import javax.annotation.Nullable;
 
+import com.google.common.collect.ImmutableMap;
+
 import com.linecorp.centraldogma.common.Author;
+import com.linecorp.centraldogma.common.RepositoryRole;
 import com.linecorp.centraldogma.common.Revision;
 import com.linecorp.centraldogma.server.command.Command;
 import com.linecorp.centraldogma.server.command.CommandExecutor;
 import com.linecorp.centraldogma.server.metadata.MetadataService;
+import com.linecorp.centraldogma.server.metadata.RepositoryMetadata;
+import com.linecorp.centraldogma.server.metadata.Roles;
+import com.linecorp.centraldogma.server.metadata.UserAndTimestamp;
 import com.linecorp.centraldogma.server.storage.encryption.EncryptionStorageException;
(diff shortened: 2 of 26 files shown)


common/src/main/java/com/linecorp/centraldogma/common/Author.java
```diff
@@ -32,12 +32,12 @@ public class Author {
     /**
      * The system author.
      */
-    public static final Author SYSTEM = new Author("System", "system@localhost.localdomain");
+    public static final Author SYSTEM = new Author("system", "system@localhost.localdomain");
```
minwoox, line 37: There's no reason to have different username with the email especially we are using the username when creating an Author from an email:
https://github.com/line/centraldogma/blob/6bcc0a1580c79cf02c8280c2de240284bf4b1a41/common/src/main/java/com/linecorp/centraldogma/internal/Util.java#L240-L243

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/project/DefaultProject.java
```diff
@@ -229,11 +232,25 @@ private void initializeMetadata(long creationTimeMillis, Author author) {
...
+                final Member member = new Member(author, ProjectRole.OWNER, userAndTimestamp);
+                members = ImmutableMap.of(member.id(), member);
+                tokens = ImmutableMap.of();
+            }
+
             final ProjectMetadata metadata = new ProjectMetadata(name,
                                                                  ImmutableMap.of(),
-                                                                 ImmutableMap.of(member.id(), member),
-                                                                 ImmutableMap.of(),
+                                                                 members,
```
jrhee17, line 251: Question) Before this change, if a project/repository was created using a token, was the repository not accessible if `Requires[Repository|Project]Role` was used? (since the tokens map was queried instead of the members map)

minwoox: Exactly. 😅

server/src/main/java/com/linecorp/centraldogma/server/internal/api/RepositoryServiceUtil.java
```diff
@@ -33,17 +43,34 @@ public static CompletableFuture<Revision> createRepository(
             CommandExecutor commandExecutor, MetadataService mds,
             Author author, String projectName, String repoName, boolean encrypt,
             @Nullable EncryptionStorageManager encryptionStorageManager) {
+        final Map<String, RepositoryRole> users;
+        final Map<String, RepositoryRole> tokens;
+        if (author.email().endsWith(TOKEN_EMAIL_SUFFIX)) {
```
ikhoon: What do you think of using `User` information to check whether the current request is authorized by a token?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/project/DefaultProject.java
```diff
@@ -229,11 +232,25 @@ private void initializeMetadata(long creationTimeMillis, Author author) {
             logger.info("Initializing metadata of project: {}", name);
 
             final UserAndTimestamp userAndTimestamp = UserAndTimestamp.of(author);
-            final Member member = new Member(author, ProjectRole.OWNER, userAndTimestamp);
+            final Map<String, Member> members;
+            final Map<String, TokenRegistration> tokens;
+            if (author.email().endsWith(TOKEN_EMAIL_SUFFIX)) {
```
ikhoon: Should we add `author.isUser()` to the member methods of `Author` for better encapsulation? 
```
if (!author.isUser()) {

}
```

server/src/main/java/com/linecorp/centraldogma/server/internal/api/RepositoryServiceUtil.java
minwoox: Let me just use the `author.isToken()` to minimize the changes.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/project/DefaultProject.java
minwoox: Good suggestion. 👍 `author.isToken()` is added because `isUser()` might or might not contain the system.

PR #1177, opened 2025-09-12
Reduce noise by suppressing logging of common exceptions

minwoox:
Motivation:
While monitoring logs, I found that certain expected exceptions were being logged unnecessarily.

Modifications:
- Suppress logging for the following exceptions:
  - `ProjectNotFoundException`, `ProjectExistsException`
  - `RepositoryNotFoundException`, `RepositoryExistsException`
- Do not log `TokenNotFoundException` in `ApplicationTokenAuthorizer`.

Result:
- Cleaner logs with fewer noise.

## Summary by CodeRabbit

- New Features
  - Standardized HTTP error responses for project/repository exists/not-found: responses include an "exception" type and concise messages (e.g., "project not found: name", "repository already exists: project/repo"). Status codes unchanged.

- Refactor
  - Unified exception construction via static factory methods for project/repo exceptions to produce consistent messages.
  - Simplified error-response construction and centralized payload creation.
  - Adjusted application-token logging to use clearer debug/warn paths.

- Tests
  - Updated tests to expect new message text and payload shape; minor test mock adjustments.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/api/HttpApiUtil.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/api/HttpApiUtil.java
@@ -188,8 +188,8 @@ public final class HttpApiUtil {
         requireNonNull(status, "status");
         requireNonNull(exceptionType, "exceptionType");
         requireNonNull(message, "message");
-        return newResponse2(HttpResponse::of, null, ResponseHeaders.builder(status),
-                            null, exceptionType, message);
+        return newResponseWithoutLogging(HttpResponse::of, null, ResponseHeaders.builder(status),
+                                         null, exceptionType, message);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/api/auth/ApplicationTokenAuthorizer.java
```diff
@@ -79,15 +78,14 @@ protected CompletionStage<Boolean> authorize(ServiceRequestContext ctx, HttpRequ
...
-            final LogLevel level;
             if (cause instanceof IllegalArgumentException ||
                 cause instanceof TokenNotFoundException) {
-                level = LogLevel.DEBUG;
+                logger.debug("Failed to authorize an application token: token={}, addr={}",
+                             maskToken(accessToken), ctx.clientAddress());
             } else {
-                level = LogLevel.WARN;
+                logger.warn("Failed to authorize an application token: token={}, addr={}",
+                             maskToken(accessToken), ctx.clientAddress(), cause);
```
ikhoon, line 88: I understood that the log level and message are the same, but they are refactored.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/HttpApiUtil.java
```diff
@@ -278,6 +281,23 @@ private static <O> O newResponse1(
             }
         }
 
+        return newResponse2(responseFactory, ctx, headersBuilder, cause, exceptionType, m);
+    }
+
+    private static <O> O newResponse2(BiFunction<ResponseHeaders, HttpData, O> responseFactory,
```
ikhoon: Would it make sense to give a more descriptive name for the original methods, such as `newResponseWithLogging0()`?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/auth/ApplicationTokenAuthorizer.java
minwoox: It doesn't log the cause when the exception is `TokenNotFoundException`. 😉

server/src/main/java/com/linecorp/centraldogma/server/internal/api/HttpApiUtil.java
minwoox: Updated. 😉

server/src/main/java/com/linecorp/centraldogma/server/internal/api/auth/ApplicationTokenAuthorizer.java
ikhoon: I missed it. Noted that `cause` is the difference.

PR #1178, opened 2025-09-15
Fix open redirection vulnerability and session initialization bug

minwoox:
Motivation:
This commit addresses two distinct issues.
The first is a security vulnerability where the application was susceptible to open redirection. A malicious actor could craft a URL with a `returnTo` parameter pointing to an external, harmful website, potentially leading to phishing attacks. The second issue was a bug in Central Dogma's read-only mode. The `InternalProjectInitializer` was not being correctly invoked, which prevented the session token from being initialized. This would lead to authentication failures or other session-related errors for users operating in a read-only environment.

Modifications:
- Implemented a strict validation mechanism for the `returnTo` redirect parameter.
  - The new logic parses the `returnTo` URL and explicitly checks if the host is `127.0.0.1` or `localhost`. Redirects to any other hosts are ignored.
- Modified the server's startup sequence to ensure `InternalProjectInitializer.initializeInReadOnlyMode()` is always called when the server is in read-only mode.

Result:
- User redirects after login are now restricted to safe, internal paths.
- Session tokens are now correctly initialized in read-only mode.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/auth/AuthProvider.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/auth/AuthProvider.java
@@ -18,6 +18,7 @@ package com.linecorp.centraldogma.server.auth;
 import static com.linecorp.centraldogma.internal.api.v1.HttpApiV1Constants.API_V0_PATH_PREFIX;
 import static com.linecorp.centraldogma.internal.api.v1.HttpApiV1Constants.API_V1_PATH_PREFIX;
 
+import java.net.URI;
 import java.util.Set;
 
 import javax.annotation.Nullable;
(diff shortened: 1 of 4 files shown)


server/src/main/java/com/linecorp/centraldogma/server/auth/AuthProvider.java
```diff
@@ -86,13 +87,25 @@ public interface AuthProvider {
...
+            String finalRedirectPath = BUILTIN_WEB_LOGIN_PATH;
+
+            if (returnTo != null && !returnTo.trim().isEmpty()) {
+                try {
+                    final URI frontendOrigin = new URI(returnTo);
+                    final String host = frontendOrigin.getHost();
+
+                    // Allow only localhost for security reason.
+                    if ("127.0.0.1".equals(host) || "localhost".equals(host)) {
+                        // e.g. http://localhost:3000 + /web/auth/login -> http://localhost:3000/web/auth/login
```
ikhoon, line 100: Question) Is `returnTo` only used for local development or testing? `http://localhost:3000/web/auth/login` seems valid when a server is running locally.

PR #1179, opened 2025-09-16
Refactor session management to use secure cookies and CSRF tokens

minwoox:
Motivation:
Previously, the application utilized `localStorage` for storing session IDs. Storing session tokens in `localStorage` is vulnerable to Cross-Site Scripting (XSS) attacks.
 This change enhances security by migrating to a more robust session management mechanism using `HttpOnly` session cookies and Cross-Site Request Forgery (CSRF) tokens.

Modifications:
- Stored session ID in an `HttpOnly`, `Secure` cookie.
  - The session cookie is now automatically sent with every subsequent request to the server, authenticating the user.
- Introduced a CSRF token mechanism.
  - For state-changing requests, the client-side application is now required to include the CSRF token in the request headers. 

Result:
- The application's defense against XSS attacks is significantly strengthened, as session tokens are no longer accessible to JavaScript.

## Summary by CodeRabbit

- New Features
  - Switched to cookie-based sessions with CSRF protection; session cookie is TLS-aware and X-CSRF-Token is returned on valid sessions.
  - Web app auto-redirects to login on 401/403 and after logout; added createLoginUrl and CSRF meta tag management.
  - API now supports token-based access to internal projects; users/me includes CSRF header when applicable.

- Bug Fixes
  - Git clone requests now return proper WWW-Authenticate headers when unauthorized.
  - Tightened CORS to explicit localhost origins.

- Chores
  - Dev script runs Next.js with --hostname 127.0.0.1.


server/src/main/java/com/linecorp/centraldogma/server/internal/admin/auth/SessionCookieAuthorizer.java
```diff
@@ -0,0 +1,137 @@
...
+        return sessionManager.get(sessionId).thenApply(session -> {
+            if (session == null) {
+                logger.trace("Session not found (or expired): {}, ctx={}", sessionId, ctx);
+                return AuthorizationStatus.of(false);
+            }
+            // Check the token when the method is not safe:
+            // https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#javascript-automatically-including-csrf-tokens-as-an-ajax-request-header
+            if (!isSafeMethod(req)) {
+                final String csrfToken = req.headers().get(X_CSRF_TOKEN);
+                if (isNullOrEmpty(csrfToken) || !csrfToken.equals(session.csrfToken())) {
```
jrhee17, line 105: Understood that during rollout, users may not be able to push via UI depending on which host is being accessed - once completely rolled out, refreshing will likely fix accessibility issues

server-auth/saml/src/main/java/com/linecorp/centraldogma/server/auth/saml/HtmlUtil.java
```diff
@@ -19,19 +19,33 @@
...
 final class HtmlUtil {
 
-    private static final String BEGIN =
+    private static final String HTML_HEAD_START =
             "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n" +
             "<!DOCTYPE html PUBLIC \"-//W3C//DTD XHTML 1.0 Strict//EN\" \"DTD/xhtml1-strict.dtd\">" +
             "<html xmlns=\"http://www.w3.org/1999/xhtml\" xml:lang=\"en\" lang=\"en\">" +
-            "<head><meta http-equiv=\"content-type\" content=\"text/html;charset=utf-8\" /></head>" +
-            "<body onload=\"";
+            "<head><meta http-equiv=\"content-type\" content=\"text/html;charset=utf-8\" />";
```
jrhee17, line 24: Understood that the current approach is preferred over the "double-submit cookie pattern" to avoid relying on cookie behavior in browsers. (Although I do think the double-submit pattern is probably simpler to implement in our case as the project is served via an SPA and cookies are shared across tabs, and csrf isn't terribly important in this project)

server-auth/shiro/src/main/java/com/linecorp/centraldogma/server/auth/shiro/ShiroLoginService.java
```diff
@@ -107,18 +137,26 @@ protected HttpResponse doPost(ServiceRequestContext ctx, HttpRequest req) throws
...
-                                       new AccessToken(sessionId, sessionValidDuration.getSeconds());
+                               logger.debug("{} Logged in: {}", ctx, usernamePassword.getUsername());
+
+                               final Cookie cookie = createSessionIdCookie(sessionId, tlsEnabled,
+                                                                           cookieMaxAgeSecond);
+                               final ResponseHeaders responseHeaders =
+                                       ResponseHeaders.builder(HttpStatus.OK)
+                                                      .contentType(MediaType.JSON_UTF_8)
+                                                      .set(HttpHeaderNames.CACHE_CONTROL,
+                                                           ServerCacheControl.DISABLED.asHeaderValue())
```
jrhee17, line 148: Understood this is to notify proxies not to store the csrf token

server/src/main/java/com/linecorp/centraldogma/server/internal/admin/auth/SessionCookieAuthorizer.java
```diff
@@ -0,0 +1,137 @@
...
+        }
+        return sessionManager.get(sessionId).thenApply(session -> {
+            if (session == null) {
+                logger.trace("Session not found (or expired): {}, ctx={}", sessionId, ctx);
+                return AuthorizationStatus.of(false);
+            }
+            // Check the token when the method is not safe:
+            // https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#javascript-automatically-including-csrf-tokens-as-an-ajax-request-header
+            if (!isSafeMethod(req)) {
+                final String csrfToken = req.headers().get(X_CSRF_TOKEN);
```
ikhoon: nit:
```suggestion
                final String csrfToken = req.headers().get(X_CSRF_TOKEN, "");
```

server/src/main/java/com/linecorp/centraldogma/server/internal/admin/auth/SessionUtil.java
```diff
@@ -0,0 +1,87 @@
...
+
+    public static String sessionCookieName(boolean tlsEnabled) {
+        // Prefix the cookie name with "__Host-Http-".
+        // https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Cookies#cookie_prefixes
+        return tlsEnabled ? "__Host-Http-session-id" : "session-id";
+    }
+
+    @Nullable
+    public static String getSessionIdFromCookie(ServiceRequestContext ctx, String sessionCookieName) {
+        final String cookieValue = ctx.request().headers().get(HttpHeaderNames.COOKIE);
```
ikhoon: Should we use `.headers().cookies()`?

webapp/src/dogma/features/auth/Authorized.tsx
```diff
@@ -14,43 +14,56 @@
...
   useEffect(() => {
-    const validateSession = async () => {
-      const action = await dispatch(checkSecurityEnabled());
-      if (isFulfilled(action)) {
-        dispatch(getUser());
-      }
-    };
-    validateSession();
-  }, [dispatch]);
+    if (!isInitialized) {
```
ikhoon, line 35: Was this block invoked more than once?

webapp/src/dogma/features/auth/authSlice.ts
```diff
@@ -21,24 +21,36 @@ import ErrorMessageParser from 'dogma/features/services/ErrorMessageParser';
...
 axios.defaults.baseURL = process.env.NEXT_PUBLIC_HOST || '';
+axios.defaults.withCredentials = true;
 
-export const getUser = createAsyncThunk('/auth/user', async (_, { getState, dispatch, rejectWithValue }) => {
+const getCsrfTokenFromMeta = (): string | null => {
+  if (typeof window === 'undefined') {
+    return null;
+  }
+  const element = document.querySelector('meta[name="csrf-token"]');
+  return element ? element.getAttribute('content') : null;
```
ikhoon: Optional) Could use `element?.getAttribute('content') ?? null`

webapp/src/dogma/features/api/apiSlice.ts
```diff
@@ -98,19 +100,42 @@ export type MirrorConfig = {
...
+});
+
+const baseQueryWithReauth: BaseQueryFn<string | FetchArgs, unknown, FetchBaseQueryError> = async (
+  args,
+  api,
+  extraOptions,
+) => {
+  const result = await baseQuery(args, api, extraOptions);
+
+  if (result.error && result.error.status === 401) {
```
ikhoon: Redirecting to the login page on 403 Forbidden may help mitigate potential session issues during rollout.

server-auth/saml/src/main/java/com/linecorp/centraldogma/server/auth/saml/HtmlUtil.java
```diff
@@ -19,19 +19,33 @@
...
+        sb.append(HTML_HEAD_END_BODY_START);
+        sb.append(redirectScript);
+        if (!redirectScript.endsWith(";")) {
+            sb.append(';');
+        }
+        sb.append(END);
+        return sb.toString();
+    }
+
     static String getHtmlWithOnload(String... statements) {
```
ikhoon: This method is unused.

server/src/main/java/com/linecorp/centraldogma/server/internal/admin/service/UserService.java
```diff
@@ -35,19 +48,47 @@
...
+        sessionCookieName = sessionCookieName(tlsEnabled);
+        this.sessionManager = sessionManager;
     }
 
     /**
      * GET /users/me
      * Returns a login {@link User} if the user is authorized. Otherwise, {@code 401 Unauthorized} HTTP
-     * response is sent.
+     * response is sent. If the user has a valid session, the CSRF token is included in the response
+     * header {@code X-CSRF-Token}.
```
ikhoon, line 67: Noted that `/users/me` can be accessible without `X-CSRF-Token` as its method is safe.

server-auth/saml/src/main/java/com/linecorp/centraldogma/server/auth/saml/HtmlUtil.java
minwoox: Do you remember that I'm going to implement this session cookie authentication for read-only mode as well?
I'm going to use [Signed Double-Submit Cookie](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#signed-double-submit-cookie-recommended) for that. 😆 
The reason why I used `Synchronizer Token Pattern` instead of `Signed Double-Submit Cookie` is:
- There's [no trouble maintaining the CSRF token on the server-side](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#alternative-using-a-double-submit-cookie-pattern)
- Other companies that have not yet applied the encryption system cannot use the `Signed Double-Submit Cookie` because of the lack of maintaining session keys.

server-auth/shiro/src/main/java/com/linecorp/centraldogma/server/auth/shiro/ShiroLoginService.java
minwoox: That is correct. 👍

webapp/src/dogma/features/api/apiSlice.ts
minwoox: That's a good suggestion. 👍

webapp/src/dogma/features/auth/Authorized.tsx
minwoox: It's invoked only once. 😆 
I originally added this guarding logic while fixing another issue. It turned out to be unnecessary for that fix, but I left it in because I thought it was harmless and might signal to reviewers that this block is only invoked once.
I'm happy to remove it if you think it's confusing or adds unnecessary complexity. What is your preference?

minwoox:
Let me merge this to implement the session management in read-only mode.

PR #1181, opened 2025-09-18
Introduce SessionMasterKey and improve session deserialization

minwoox:
Motivation:
This commit introduces two improvements:
- Introduce a `SessionMasterKey` that we can derive distinct keys for signing and encryption keys.
- The session deserialization logic has been improved to support forward and backward compatibility for #1179.

Modifications:
- Introduced a `SessionMasterKey` to serve as the primary secret for session cookie security.
  - A session signing key and a session encryption key are derived from the master key via the HMAC-based Key Derivation Function (HKDF).
- Enhanced the session class deserializer to handle different versions of the session object.

Result:
- Ready to sign and encrypt Session cookies.


server/src/main/java/com/linecorp/centraldogma/server/storage/encryption/DefaultEncryptionStorageManager.java
```diff
@@ -171,12 +175,109 @@ public boolean enabled() {
...
+    @Override
+    public void storeSessionMasterKey(SessionMasterKey sessionMasterKey) {
+        requireNonNull(sessionMasterKey, "sessionMasterKey");
+        final ColumnFamilyHandle wdekCf = columnFamilyHandlesMap.get(WDEK_COLUMN_FAMILY);
+        final int version = sessionMasterKey.version();
+        final byte[] masterKeyKey = sessionMasterKeyKey(version);
+        try {
+            final byte[] existing = rocksDb.get(wdekCf, masterKeyKey);
+            if (existing != null) {
+                throw new EncryptionEntryExistsException(
```
jrhee17, line 210: Question) If this exception is thrown, is there a way to recover from this?
I'm imagining a scenario where two servers started up at the same time and both call `InternalProjectInitializer#initializeSessionKey`.

jrhee17 (approved):
👍

server/src/main/java/com/linecorp/centraldogma/server/storage/encryption/DefaultEncryptionStorageManager.java
minwoox: > If this exception is thrown, is there a way to recover from this?

There's no way to recover if an exception is thrown, so we should recover it manually. 😓 
Because we rely on the ZooKeeper lock for creating and storing the session master key,  
we have no choice but to trust that the ZooKeeper lock is working correctly.

server/src/main/java/com/linecorp/centraldogma/server/auth/SessionKey.java
```diff
@@ -0,0 +1,85 @@
...
+ * A session key used to sign and encrypt session cookies.
+ */
+public final class SessionKey {
+
+    /**
+     * Derives a new {@link SessionKey} from the specified {@code masterKey} and {@code salt}.
+     */
+    public static SessionKey of(byte[] masterKey, byte[] salt, int version) {
+        final HKDFBytesGenerator hkdfBytesGenerator = new HKDFBytesGenerator(new SHA256Digest());
+        final byte[] signingKeyInfo = "session-signing-key".getBytes(StandardCharsets.UTF_8);
```
ikhoon: Make this a static variable?

server/src/main/java/com/linecorp/centraldogma/server/storage/project/InternalProjectInitializer.java
```diff
@@ -88,6 +129,7 @@ public void initialize() {
         try {
             initialize0(INTERNAL_PROJECT_DOGMA);
             initializeTokens();
+            initializeSessionKey();
```
ikhoon, line 132: Couldn't we safely call this method until `Command.createSessionMasterKey()` is deployed?

minwoox: We have to do this in this order:
```
// dogma.json
{
  "encryptionAtRest: {
    "enabled": true,
    "encryptSessionCookie": false
  }
}
```
```
// dogma.json
{
  "encryptionAtRest: {
    "enabled": true,
    "encryptSessionCookie": true
  }
}
```

server/src/main/java/com/linecorp/centraldogma/server/auth/SessionKey.java
minwoox: Addressed. 🫡

ikhoon (approved):

minwoox:
Thanks for reviewing. 😉

PR #1182, opened 2025-09-18
Log ZooKeeper lock acquisition duration for performance monitoring

minwoox:
Motivation:
Without specific metrics, it is difficult to determine if slow operations are caused by lock contention, network issues, or problems within the ZooKeeper ensemble itself.

Modifications:
-   Identified the critical section of code responsible for acquiring a distributed ZooKeeper lock.
- Implemented to record duration for acquiring a ZooKeeper lock.
  - Record only for push commands using projectName to reduce the cardinality

Result:
- The ZooKeeper lock acquision time is now recorded.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
@@ -82,6 +82,7 @@ import com.google.common.escape.Escapers;
 import com.google.common.util.concurrent.MoreExecutors;
 import com.google.common.util.concurrent.Uninterruptibles;
 
+import com.linecorp.armeria.common.metric.MoreMeters;
 import com.linecorp.armeria.common.util.SafeCloseable;
 import com.linecorp.centraldogma.common.LockAcquireTimeoutException;
 import com.linecorp.centraldogma.common.Revision;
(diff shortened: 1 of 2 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -923,11 +929,19 @@ private SafeCloseable safeLock(Command<?> command) {
...
-            final long deadlineNanos = System.nanoTime() + remainingTimeNanos;
+            final long startTime = System.nanoTime();
+            final long deadlineNanos = startTime + remainingTimeNanos;
             for (;;) {
                 try {
                     if (mtx.acquire(remainingTimeNanos, TimeUnit.NANOSECONDS)) {
                         lockAcquired = true;
+                        if (command instanceof AbstractPushCommand) {
+                            final String projectName = ((AbstractPushCommand<?>) command).projectName();
+                            record(projectName, startTime);
```
ikhoon, line 940: Would we also record the time when a lock acquisition fails?

minwoox: If it fails to acquire a lock within 60 seconds, an exception is thrown (entering read-only in some situation), so we could probably notice that.
So I thought it wasn't necessary. But I will add it if you think we need it. 😉

ikhoon: If most lock acquisition attempts exceeds 60 seconds, we may not realize from that metrics that it has slowed. Although we could know the problem from the logs. It would be worth recording all lock acquisition time to the metrics.

minwoox: That makes sense. 👍 
Please note that I didn't add the extra result tag because I don't want to create another timer for the failure.

minwoox: On my second thought, I can add `acquired` because failing to acquire the lock rarely happens

jrhee17 (approved):

ikhoon (approved):
🙇‍♂️🙇‍♂️

PR #1186, opened 2025-09-23
Remove Thrift support from testing modules

minwoox:
Motivation:
This change prepares Central Dogma for the eventual removal of Thrift service support entirely. Currently, including Thrift-related code and dependencies in the testing modules creates a significant problem:
- It can lead to dependency conflicts when users integrate Central Dogma into projects that already use a different version of Thrift.

Modifications:
- Added a new configuration option, `CentralDogmaBuilder.enableThriftService`, allowing users to explicitly disable the Thrift service during Central Dogma setup, particularly useful for testing.
- Removed all direct Thrift dependencies from the Central Dogma testing modules.
- Deprecated and removed Thrift-related APIs from `CentralDogmaExtension` and `CentralDogmaRule`.

Result:
- (Breaking) Central Dogma testing modules do not support the Thrift service anymore.

## Summary by CodeRabbit

* New Features
  * Added a config flag to enable/disable the Thrift service (enabled by default) and runtime detection to expose Thrift endpoints only when available.

* Refactor
  * Removed legacy Thrift client support from testing APIs and simplified related integration tests and flows.
  * Centralized Thrift service wiring and guarded it with the new toggle.

* Chores
  * Updated test dependencies to use the modern client and exclude legacy Thrift artifacts.

* Tests
  * Removed numerous Thrift-specific converter, backward-compatibility and Thrift-based compression tests.

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -36,8 +36,8 @@ jobs:
         env:
           # Should not use '-P' option with 'secrets' that can cause unexpected results
           # if secret values contains white spaces or new lines.
-          ORG_GRADLE_PROJECT_ossrhUsername: ${{ secrets.OSSRH_USER_NAME }}
-          ORG_GRADLE_PROJECT_ossrhPassword: ${{ secrets.OSSRH_PASSWORD }}
+          ORG_GRADLE_PROJECT_ossrhUsername: ${{ secrets.MAVEN_CENTRAL_USERNAME }}
+          ORG_GRADLE_PROJECT_ossrhPassword: ${{ secrets.MAVEN_CENTRAL_PASSWORD }}
           ORG_GRADLE_PROJECT_signingKeyId: ${{ secrets.GPG_KEY_ID }}
           ORG_GRADLE_PROJECT_signingKey: ${{ secrets.GPG_PRIVATE_KEY }}
           ORG_GRADLE_PROJECT_signingPassword: ${{ secrets.GPG_PASSWORD }}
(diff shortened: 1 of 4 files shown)


it/server/src/test/java/com/linecorp/centraldogma/it/ClientType.java
```diff
@@ -23,8 +23,7 @@
 
 enum ClientType {
```
minwoox, line 24: I will probably remove this class in the future.

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -1103,14 +1086,12 @@ private static Function<? super HttpService, EncodingService> contentEncodingDec
                         switch (subtype) {
                             case "json":
                             case "xml":
-                            case "x-thrift":
```
minwoox, line 1106: I believe we don't need to support encoding for Thrift service anymore even though it might be used.

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -736,7 +741,9 @@ private Server startServer(ProjectManager pm, CommandExecutor executor,
         final ProjectApiManager projectApiManager =
                 new ProjectApiManager(pm, executor, mds, encryptionStorageManager);
 
-        configureThriftService(sb, projectApiManager, executor, watchService, mds);
+        if (THRIFT_FOUND && cfg.enableThriftService()) {
```
ikhoon: Should we warn if `enableThriftService == true` and `THRIFT_FOUND == false`?

minwoox: Good suggestion. Addressed. 😉

PR #1190, opened 2025-09-30
Implement JWT-based encrypted session cookies

minwoox:
Motivation:
By signing and encrypting the session cookie into a JSON Web Token (JWT), we ensure both its integrity and confidentiality.

Modifications:
- Introduced a feature flag, `encryptSessionCookie`, which, when enabled, transforms the session cookie into a signed and encrypted JWT.
- Implemented a new CSRF token strategy for read-only mode:
  - The CSRF token is now generated using a hash of the JWT's signature, linking the token's validity directly to the session cookie.
  - Validation in this mode involves re-deriving the token from the cookie on the server and comparing it to the token provided by the client.
- For normal (writable) mode, the existing stateful CSRF mechanism is maintained:
  - A cryptographically secure CSRF token is generated independently.
  - This token is stored in the server-side session manager.
  - Validation is performed by comparing the incoming CSRF token from the client against the one stored in the session.

Result:
- Session cookie data is now confidential and tamper-proof when encryption is enabled.
- The application now provides robust, stateless CSRF protection in read-only mode, where it is not possible to rely on server-side session storage.
- (Breaking) The default session timeout is now 8 hours. Previously, it was 7 days.

## Summary by CodeRabbit

- **New Features**
  - Encrypted session cookies (JWT/JWE), centralized session service, and unified CSRF handling with read-only session support.
  - New auth/session utilities and handlers to manage session lifecycle and responses.

- **Changes**
  - Default session timeout changed from 7 days to 8 hours.
  - Built-in metadata for the "dogma" project; New Repository UI hidden for that project.
  - Auth flows surface encryption manager availability and session-writability checks.

- **Bug Fixes**
  - SAML web login now redirects to “/” (302).

- **Dependencies**
  - Added Nimbus JOSE JWT library.

- **Tests**
  - New end-to-end tests and test helpers validating session/JWT/JWE and CSRF flows.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
@@ -55,9 +55,9 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.BiFunction;
+import java.util.function.BooleanSupplier;
 import java.util.function.Consumer;
 import java.util.function.Function;
-import java.util.function.Supplier;
 
 import javax.annotation.Nullable;
 
@@ -851,7 +851,7 @@ public class CentralDogma implements AutoCloseable {
         }
 
         checkState(sessionManager != null, "SessionManager is null");
-        final Supplier<Boolean> sessionPropagatorWritableChecker = commandExecutor::isWritable;
+        final BooleanSupplier sessionPropagatorWritableChecker = commandExecutor::isWritable;
         final AuthProviderParameters parameters = new AuthProviderParameters(
                 // Find application first, then find the session token.
                 new ApplicationTokenAuthorizer(mds::findTokenBySecret)
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/admin/auth/SessionCookieHandler.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/admin/auth/SessionCookieHandler.java
@@ -23,7 +23,7 @@ import static com.linecorp.centraldogma.server.internal.admin.auth.SessionUtil.g
 import static com.linecorp.centraldogma.server.internal.admin.auth.SessionUtil.sessionCookieName;
 import static java.util.Objects.requireNonNull;
 
-import java.util.function.Supplier;
+import java.util.function.BooleanSupplier;
 
 import javax.annotation.Nullable;
 
@@ -40,7 +40,7 @@ import com.linecorp.centraldogma.server.storage.encryption.EncryptionStorageMana
 
 public final class SessionCookieHandler {
 
-    private final Supplier<Boolean> sessionPropagatorWritableChecker;
+    private final BooleanSupplier sessionPropagatorWritableChecker;
     private final String sessionCookieName;
 
     @Nullable
@@ -50,7 +50,7 @@ public final class SessionCookieHandler {
     @Nullable
     private final JWEDecrypter decrypter;
 
-    public SessionCookieHandler(Supplier<Boolean> sessionPropagatorWritableChecker, boolean tlsEnabled,
+    public SessionCookieHandler(BooleanSupplier sessionPropagatorWritableChecker, boolean tlsEnabled,
                                 EncryptionStorageManager encryptionStorageManager) {
         this.sessionPropagatorWritableChecker =
                 requireNonNull(sessionPropagatorWritableChecker, "sessionPropagatorWritableChecker");
@@ -91,7 +91,7 @@ public final class SessionCookieHandler {
             if (objectSessionId instanceof String) {
                 return new SessionInfo((String) objectSessionId, null, null);
             } else {
-                if (sessionPropagatorWritableChecker.get()) {
+                if (sessionPropagatorWritableChecker.getAsBoolean()) {
                     return null;
                 }
                 // In read-only mode, we support authentication using only the username claim.
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/admin/auth/SessionUtil.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/admin/auth/SessionUtil.java
@@ -18,6 +18,7 @@ package com.linecorp.centraldogma.server.internal.admin.auth;
 
 import static com.google.common.base.Strings.isNullOrEmpty;
 
+import java.nio.charset.StandardCharsets;
 import java.time.Duration;
 import java.util.Date;
 
@@ -149,7 +150,8 @@ public final class SessionUtil {
     }
 
     static String csrfTokenFromSignedJwt(SignedJWT signedJwt) {
-        return Hashing.sha256().hashBytes(signedJwt.getSignature().decode()).toString();
+        return Hashing.sha256().hashString(signedJwt.getSignature().toString(), StandardCharsets.UTF_8)
+                      .toString();
     }
 
     public static boolean validateCsrfToken(ServiceRequestContext ctx, HttpRequest req,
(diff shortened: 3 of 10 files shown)


webapp/src/test/java/com/linecorp/centraldogma/webapp/SamlIdpServer.java
```diff
@@ -0,0 +1,224 @@
...
+import com.linecorp.armeria.common.HttpRequest;
+import com.linecorp.armeria.common.HttpResponse;
+import com.linecorp.armeria.common.HttpStatus;
+import com.linecorp.armeria.common.MediaType;
+import com.linecorp.armeria.common.QueryParams;
+import com.linecorp.armeria.server.HttpService;
+import com.linecorp.armeria.server.Server;
+import com.linecorp.armeria.server.ServiceRequestContext;
+
+final class SamlIdpServer {
```
minwoox, line 43: This test was written with the help of AI and only implements the basic elements for UI SAML testing. Therefore, there are parts that do not comply with the spec, and I don't think you need to review this file. Unless you want to.

server/src/main/java/com/linecorp/centraldogma/server/internal/admin/auth/SessionCookieHandler.java
```diff
@@ -0,0 +1,141 @@
...
+            }
+            final JWTClaimsSet jwtClaimsSet = getJwtClaimsSetFromSignedJwt(ctx, signedJwt, verifier);
+            if (jwtClaimsSet == null) {
+                return null;
+            }
+            final Object objectSessionId = jwtClaimsSet.getClaim("sessionId");
+            if (objectSessionId instanceof String) {
+                return new SessionInfo((String) objectSessionId, null, null);
+            } else {
+                if (sessionPropagatorWritableChecker.get()) {
```
jrhee17: Question) Given that servers in a cluster may be in a mixed state of READONLY and WRITABLE, is it intentional that unlucky users who have cookies issued from READONLY servers cannot call APIs for WRITABLE servers?

jrhee17 (approved):

server/src/main/java/com/linecorp/centraldogma/server/internal/admin/auth/SessionUtil.java
```diff
@@ -35,26 +61,135 @@ public final class SessionUtil {
...
+            jweObject.decrypt(decrypter);
+        } catch (Exception e) {
+            logger.trace("Failed to parse the session cookie. ctx={}", ctx, e);
+            return null;
+        }
+
+        final Payload payload = jweObject.getPayload();
+        final String jwsString = payload.toString();
+        try {
+            return SignedJWT.parse(jwsString);
```
ikhoon, line 124: nit: payload.toSignedJWT() could be used as a shortcut.

server/src/main/java/com/linecorp/centraldogma/server/internal/admin/auth/SessionUtil.java
```diff
@@ -80,5 +214,67 @@ private static Cookie findSessionCookie(Cookies cookies, String sessionCookieNam
...
+    public static String createJwe(String signedJwt, String sessionKeyVersion, JWEEncrypter encrypter) {
+        final JWEHeader jweHeader = new JWEHeader.Builder(JWEAlgorithm.DIR, EncryptionMethod.A256GCM)
+                .contentType("JWT")
+                .keyID(sessionKeyVersion)
+                .build();
+
+        final Payload payload = new Payload(signedJwt);
+        final JWEObject jweObject = new JWEObject(jweHeader, payload);
+        try {
+            jweObject.encrypt(encrypter);
```
ikhoon, line 274: Q) Our JWT does not include sensitive data. Will using JWE provide any additional protection?

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -839,17 +851,20 @@ private AuthProvider createAuthProvider(
         }
 
         checkState(sessionManager != null, "SessionManager is null");
+        final Supplier<Boolean> sessionPropagatorWritableChecker = commandExecutor::isWritable;
```
ikhoon: nit: Could use `BooleanSupplier`?

server/src/main/java/com/linecorp/centraldogma/server/internal/admin/auth/SessionUtil.java
```diff
@@ -35,26 +61,135 @@ public final class SessionUtil {
...
+            verifier.verify(jwtClaimsSet, null);
+            return jwtClaimsSet;
+        } catch (BadJWTException e) {
+            logger.trace("Invalid claim set in the inner JWS. ctx={}", ctx, e);
+            return null;
+        }
+    }
+
+    static String csrfTokenFromSignedJwt(SignedJWT signedJwt) {
+        return Hashing.sha256().hashBytes(signedJwt.getSignature().decode()).toString();
```
ikhoon: For hashing, we may use the base64 encoded value as is.
```suggestion
        return Hashing.sha256().hashBytes(signedJwt.getSignature()).toString();
```

ikhoon (approved):
👍👍

server/src/main/java/com/linecorp/centraldogma/server/internal/admin/auth/SessionCookieHandler.java
minwoox: If so, the user will get a 404 response and re-login immediately. So I think it's going to be okay.

server/src/main/java/com/linecorp/centraldogma/server/internal/admin/auth/SessionUtil.java
minwoox: `payload.toSignedJWT()` swallowed the exception and just returns null. Let me use it as is. 😉

server/src/main/java/com/linecorp/centraldogma/server/internal/admin/auth/SessionUtil.java
minwoox: > Will using JWE provide any additional protection?

No, it won't. 😉 
I chose to use JWE for the following reasons:
- It aligns with the mechanism used for the read-only mode cookie, which simplifies our logic.
- Since this is only used by browsers, the performance overhead of encryption is negligible.
- JWE provides the flexibility to add more properties to the cookie in the future.

server/src/main/java/com/linecorp/centraldogma/server/internal/admin/auth/SessionUtil.java
minwoox: Good suggestion. Fixed. 😉

server/src/main/java/com/linecorp/centraldogma/server/internal/admin/auth/SessionUtil.java
ikhoon: I see. I just worried the drawback of JWE that browsers can't read the payload which may contain some useful information for UI control such as permission and scope.

ikhoon: As we have our own mechanism for UI control, there is no blocking factor even if we choose JWE.

minwoox: That is true. 
Also, the company's guideline for authentication cookies says it should be encrypted, although I don't think we don't have to encrypt the cookie for our case:
```
The requirements for cookies used for authentication are as follows:
    Excluding existing cookies, the issuance of cookies for authentication must be restricted to the authentication server, and a digital signature must be applied for integrity and non-repudiation.
    The applied digital signature must be encrypted along with the Cookie-Value.
    The cryptographic key for the digital signature must be rotatable.
```

PR #1191, opened 2025-10-01
Record ZooKeeperCommandExecutor timings

ikhoon:
Motivation:

Currently, there are no metrics to identify wihch part of a `Command` execution is slow. To improve the write performance of Central Dogma, I propose adding three new metrics in addition to the existing lock acquisition metric.

- Command execution time
- Log replay time
- Log store time

Modifications:

- Add `ReplicationTimings` to recode the timings.
- Add `ReplicationMetrics` to export the recorded metrics per project through Micrometer.

Result:

You can now use the following four metrics to measure the write performance of Central Dogma.
- `replication.lock.waiting`
- `replication.command.execution`
- `replication.log.replay`
- `replication.log.store`

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
@@ -82,7 +82,6 @@ import com.google.common.escape.Escapers;
 import com.google.common.util.concurrent.MoreExecutors;
 import com.google.common.util.concurrent.Uninterruptibles;
 
-import com.linecorp.armeria.common.metric.MoreMeters;
 import com.linecorp.armeria.common.util.SafeCloseable;
 import com.linecorp.centraldogma.common.LockAcquireTimeoutException;
 import com.linecorp.centraldogma.common.Revision;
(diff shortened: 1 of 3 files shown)


minwoox (approved):
👍 👍 👍 

I'm not an expert of ZooKeeper but how about also collecting the lock release time since we collect all kinds of Replication metrics?

https://github.com/line/centraldogma/blob/6a58a9f9ad71b121c5a7acfb146612d4be7b8075/server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java#L973

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -1163,48 +1157,71 @@ protected <T> CompletableFuture<T> doExecute(Command<T> command) throws Exceptio
...
         }
         executor.execute(() -> {
+            final ReplicationTimings timings = newReplicationTimings(command);
             try {
-                future.complete(blockingExecute(command));
+                future.complete(blockingExecute(command, timings));
             } catch (Throwable t) {
                 future.completeExceptionally(t);
+            } finally {
+                timings.record();
```
jrhee17, line 1118: Understood that if failure occurs while replaying, certain metrics (e.g. `commandExecutionTimer`) will record 0 instead of not recording at all

jrhee17 (approved):

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
ikhoon: Good point. Added some flags to skip unmeasured metrics.

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -1097,6 +1082,18 @@ private static String pathFromRevision(long revision) {
         return String.format("%010d", revision);
     }
 
+    private <T> ReplicationTimings newReplicationTimings(Command<T> command) {
+        ReplicationMetrics metrics = null;
+        if (command instanceof RepositoryCommand) {
```
jrhee17, line 1087: nit; If only `RepositoryCommand`s are recorded, I don't see a reason not to also record per repository. As is seems fine though.

ikhoon: We used to record metrics per repository in `MetricCollectingService`, but it caused excessive memory usage and frequent GC, so I took a more conservative approach. If we later find the project-level metrics are insufficient, we may consider tagging repositories.

minwoox (approved):
👍

PR #1192, opened 2025-10-01
Compress ZooKeeper replication logs

ikhoon:
Motivation:

When a file is newly created or has a large diff, the replication log size can grow significantly. With compression, up to about 7–8 MB of data can be stored in a single znode, improving replication performance.

Modifcations:

- Add `zstd-jni` dependency to the server module.
- Compress replication logs using zstd before storing them in ZooKeeper.
- Decompress replication logs when compressed prior to replay.

Result:

Central Dogma server now compresses replication logs with zstd.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
@@ -16,6 +16,7 @@
 
 package com.linecorp.centraldogma.server.internal.replication;
 
+import static com.google.common.base.MoreObjects.firstNonNull;
 import static com.google.common.collect.ImmutableList.toImmutableList;
 import static java.util.Objects.requireNonNull;
 
@@ -72,6 +73,7 @@ import org.slf4j.LoggerFactory;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.github.luben.zstd.Zstd;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Joiner;
 import com.google.common.base.MoreObjects;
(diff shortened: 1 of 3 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -1029,18 +1031,21 @@ private static class LogMeta {
...
         private final long timestamp;
         private final int size;
+        private final boolean compressed;
         private final List<Long> blocks = new ArrayList<>();
 
         @JsonCreator
         LogMeta(@JsonProperty(value = "replicaId", required = true) int replicaId,
                 @JsonProperty(value = "timestamp", defaultValue = "0") Long timestamp,
-                @JsonProperty("size") int size) {
+                @JsonProperty("size") int size, @JsonProperty("compressed") Boolean compressed) {
```
minwoox: `@Nullable`?
Shouldn't we create a PR for forward compatibility first?

ikhoon: Fixed.

minwoox (approved):
👍 👍 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -1029,18 +1031,21 @@ private static class LogMeta {
...
         private final long timestamp;
         private final int size;
+        private final boolean compressed;
         private final List<Long> blocks = new ArrayList<>();
 
         @JsonCreator
         LogMeta(@JsonProperty(value = "replicaId", required = true) int replicaId,
                 @JsonProperty(value = "timestamp", defaultValue = "0") Long timestamp,
-                @JsonProperty("size") int size) {
+                @JsonProperty("size") int size, @Nullable @JsonProperty("compressed") Boolean compressed) {
```
minwoox: I realized that I need an option to distinguish the encrypted log.
Could you add an enum so that we can also specify the encryption option?
```
public enum ReplicationLogOption {
  NONE,
  COMPRESSED,
  ENCRYPTED,
  COMPRESSED_AND_ENCRYPTED;
}
```

jrhee17 (approved):

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
ikhoon: Compression and encryption are different characteristics. In addition, `ReplicationLogOption` does not seem scalable. If we add an new option, we may add `COMPRESSED_AND_ENCRYPTED_AND_OTHER`.


Instead, let me add `encrypted` flag to `LogMeta`.

minwoox: Yeah that also works. 😉

PR #1194, opened 2025-10-19
Migrate to JSpecify annotations

### LINKED ISSUE #1193: Migrate to JSpecify annotations for nullability constraints
A lot of projects are migrating from various annotations to JSpecify

e.g.
- https://github.com/spring-projects/spring-framework/issues/28797
- https://github.com/jspecify/jspecify/issues/239

sh-cho:
Motivation:
jsr305 is dormant and now JSpecify is recommended annotation sets for variety of static analyses (for now, it supports only nullness) and many open sources are moving to JSpecify (e.g. Spring Framework, Guava, ...)
- https://jspecify.dev/
- https://spring.io/blog/2025/03/10/null-safety-in-spring-apps-with-jspecify-and-null-away

Modification:
- Replace `javax.annotation.Nullable`, `com.linecorp.armeria.common.annotation.Nullable` with `org.jspecify.annotations.Nullable`
- Use `@NullMarked` (`org.jspecify.annotations.NullMarked`) instead of `NonNullByDefault` and apply it to `package-info.java`s
- Use `@Nullable` within inner or fully-qualified type names (Java enforces this to `@Target(ElementType.TYPE_USE)`)
  - as-is: `@Nullable public static com.linecorp.centraldogma.common.EntryType convertEntryType(...)`
  - to-be: `public static com.linecorp.centraldogma.common.@Nullable EntryType convertEntryType(...)`
  - (ref: https://docs.oracle.com/javase/specs/jls/se19/html/jls-9.html#jls-9.7.4)
- Use `@Nullable` between type and `[]` where it means 'NULL or array'
  - as-is: `@Nullable byte[] foo`
  - to-be: `byte @Nullable [] foo`
  - (ref: https://jspecify.dev/docs/user-guide/#type-use-annotation-syntax)
- Add checkstyle rule to only allow JSpecify's `@Nullable` annotation import

Result:
resolves #1193

### DIFF
--- a/client/java-armeria/src/main/java/com/linecorp/centraldogma/internal/client/armeria/ArmeriaCentralDogma.java
+++ b/client/java-armeria/src/main/java/com/linecorp/centraldogma/internal/client/armeria/ArmeriaCentralDogma.java
@@ -1086,7 +1086,7 @@ public final class ArmeriaCentralDogma extends AbstractCentralDogma {
         final JsonNode node;
         try {
             node = Jackson.readTree(content);
-        } catch (JsonParseException e) {
+        } catch (JsonProcessingException e) {
             throw new CentralDogmaException("failed to parse the response JSON", e);
         }
 
@@ -1198,7 +1198,7 @@ public final class ArmeriaCentralDogma extends AbstractCentralDogma {
                     return unsafeCast(Entry.ofJson(revision, entryPath, rawContent.asText()));
                 } catch (JsonParseException e) {
                     // Should never reach here as the raw JSON text was already validated by the server.
-                    throw new IllegalStateException(e);
+                    throw new IllegalStateException("Failed to parse JSON content in " + entryPath, e);
                 }
             }
             logger.warn("The server does not support raw content. Using Entry#content() instead. path: {}",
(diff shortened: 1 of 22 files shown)


minwoox:
@sh-cho Thanks for the PR! 👍 
I want to apply this to Armeria first.
Would you check if the annotation works correctly, in the situation that we mentioned in this issue, please?
https://github.com/line/armeria/issues/2793

sh-cho:
@minwoox Ok let me check armeria first

sh-cho:
I'll update this PR, in this weekend

dependencies.toml
```diff
@@ -293,6 +294,10 @@ version.ref = "json-unit"
 module = "org.jsoup:jsoup"
 version.ref = "jsoup"
 
+[libraries.jspecify]
```
jrhee17, line 298: nit; this is defined but the actual dependency isn't added - it seems like the current code compiles due to guava's transitive dependencies.

I'm wondering if we may as well just include the dep as an `api` scope in the common deps
https://github.com/line/centraldogma/blob/7105683469acb4a4d7035137dc7000e12f2209b3/build.gradle#L51

ref: https://jspecify.dev/docs/using#gradle

sh-cho: Thanks, I added with api scope

PR #1200, opened 2025-10-29
Increase ZooKeeper leadership release timeout to 60 seconds

minwoox:
Motivation:
The current leader needs to wait for the leader plugins to be gracefully complete before it stops. We observed that certain internal plugins can take longer than the existing 10-second timeout to complete.

Modifications:
- Increased the timeout value from 10 seconds to 60 seconds.
- Added logging for recording the elapsed time to stop the plugin.

Result:
- The leader waits for the leader plugins to complete up to 60 seconds.

## Summary by CodeRabbit

* **Bug Fixes**
  * Extended shutdown timeout and added graceful shutdown checks to prevent unnecessary processing during service shutdown.

* **Chores**
  * Enhanced logging and performance timing measurements for plugin startup and shutdown operations.
  * Code formatting updates for consistency.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/PluginGroup.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/PluginGroup.java
@@ -191,14 +191,14 @@ final class PluginGroup {
                     plugin -> {
                         logger.info("Starting plugin: {}", plugin);
                         final long start = System.nanoTime();
-                        plugin.start(arg)
-                              .thenAccept(unused -> logger.info(
-                                      "Plugin started: {} in {} seconds", plugin,
-                                      Duration.ofNanos(System.nanoTime() - start).getSeconds()))
-                              .exceptionally(cause -> {
-                                  logger.info("Failed to start plugin: {}", plugin, cause);
-                                  return null;
-                              })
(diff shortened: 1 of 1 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/DefaultMirroringServicePlugin.java
```diff
@@ -108,10 +114,14 @@ public synchronized CompletionStage<Void> start(PluginContext context) {
 
     @Override
     public synchronized CompletionStage<Void> stop(PluginContext context) {
+        logger.info("Stopping the mirroring service...");
+        final long start = System.nanoTime();
```
ikhoon: What do you think of logging the execution time of starting and stopping in a centralized location such as:
https://github.com/line/centraldogma/blob/efb85f5870d8204151142580087c7a35d8704cf6/server/src/main/java/com/linecorp/centraldogma/server/PluginGroup.java#L190-L208

minwoox: Good suggestion. Addressed. 😉

PR #1204, opened 2025-10-31
Disallow trailing tokens in JSON files

ikhoon:
Motivation:

We observed an unexpected deserialization outcome with malformed JSON. Jackson does not fail on that data but silently drops part of the data.

For example, we expected Jackson to fail to parse the data.
```json5
{
  "a" : {},
  "b" : {},
  "c" : {
    },
    "c1" : true
  }  // a malformed tailing JSON token
  "d" : {},
  "e": {}
}
```
But Jackson just drops `d` and `e` fields and parses the JSON into:
```json
{
  "a":{},
  "b":{},
  "c":{},
  "c1":true
}
```

The maintainer of Jackson said it is an expected behavior but not a bug. https://github.com/FasterXML/jackson-core/issues/808#issuecomment-1211057640
> But this is intended behavior, not a bug: Jackson only reads as much
> content as it needs for a valid JSON Value. It does not try to consume
> everything there might be available, given that the white-space-separated
> content streams are commonly used.

He suggested enabling `DeserializationFeature.FAIL_ON_TRAILING_TOKENS` to fail on the malformed data added after the end of the top-level JSON.

Modifications:

- Enable `DeserializationFeature.FAIL_ON_TRAILING_TOKENS` to the default Jackson `ObjectMapper` in Central Dogma

Result:

Central Dogma now correctly rejects malformed JSON when parsing JSON data.


jrhee17 (approved):

minwoox (approved):
👍

build.gradle
```diff
@@ -108,15 +108,6 @@ configure(projectsWithFlags('java')) {
...
 
-    // Target Java 8 except for spring-boot3
-    if (!project.name.startsWith("java-spring-boot3")) {
-        tasks.withType(JavaCompile) {
-            if (JavaVersion.current() >= JavaVersion.VERSION_1_9) {
-                options.release = 8
-            }
-        }
-    }
-
```
ikhoon: Removed this hard-coded logic. I believe `options.release` is properly set through `java.gradle` according to `javaTargetCompatibility` and `javaN` flags.

ikhoon:
I found that `DeserializationFeature.FAIL_ON_TRAILING_TOKENS` throws an exception on valid JSON if `@JsonDeserialize` is applied to an element of a list.
https://github.com/line/centraldogma/blob/3f7cac9e4543fd504b91da2d1734d3e4a49c98e3/server/src/main/java/com/linecorp/centraldogma/server/CentralDogmaConfig.java#L278-L280

```
Trailing token (of type END_ARRAY) found after value (bound as `com.fasterxml.jackson.databind.JsonNode`): not allowed as per `DeserializationFeature.FAIL_ON_TRAILING_TOKENS`
 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 15, column: 3] (through reference chain: com.linecorp.centraldogma.server.CentralDogmaConfig["ports"]->java.util.ArrayList[0])
com.fasterxml.jackson.databind.exc.MismatchedInputException: Trailing token (of type END_ARRAY) found after value (bound as `com.fasterxml.jackson.databind.JsonNode`): not allowed as per `DeserializationFeature.FAIL_ON_TRAILING_TOKENS`
 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 15, column: 3] (through reference chain: com.linecorp.centraldogma.server.CentralDogmaConfig["ports"]->java.util.ArrayList[0])
```

Although it could be worked around for that case, I wasn't sure if enabling `DeserializationFeature.FAIL_ON_TRAILING_TOKENS` was a good idea given its false-positive behavior.

If there is consensus that applying a workaround and changing the internal deserializers is preferable, I'm happy to follow the direction. Otherwise, I will close this PR.

PR #1209, opened 2025-11-14
Provide a way to rotate wdek and session master key

minwoox:
Motivation:
Central Dogma needs to support key rotation for both wrapped data encryption keys (WDEKs) and session master keys to maintain security best practices.

Modifications:
- Renamed `KeyManagementService` to `KeyWrapper`
- Introduced `KeyManagementService` for key APIs
  - Implemented WDEK rotation logic that generates new wrapped keys with incremented versions
  - Implemented session master key rotation with version increment and concurrency protection
- Splitted `DefaultEncryptionStorageManager` into `rocksDbStorage`, `sessionKeyStorage` and `repositoryEncryptionStorage`.

Result:
- System administrators can now rotate DEKs and session master key

## Summary by CodeRabbit

* **New Features**
  * Added support for key rotation of encryption keys and session master keys
  * Added REST API endpoints for key management and encryption administration
  * Enhanced session cookie encryption configuration

* **Refactor**
  * Refactored internal encryption storage architecture to support key versioning and lifecycle management


server/src/main/java/com/linecorp/centraldogma/server/storage/encryption/KeyWrapper.java
```diff
@@ -20,15 +20,23 @@
...
     /**
      * Unwraps the given wrapped data encryption key (WDEK) using the key management service.
      */
-    CompletableFuture<byte[]> unwrap(byte[] wdek);
+    CompletableFuture<byte[]> unwrap(String wdek, String kekId);
+
+    /**
+     * Re-wraps the given wrapped data encryption key (WDEK).
+     */
+    CompletableFuture<String> rewrap(String wdek, String oldKekId, String newKekId);
```
minwoox, line 41: Some KMS providers, such as AWS KMS, embed version information directly within the wrapped key. In contrast, other providers do not support versioning, which requires users to manually specify a different key to perform a key rotation.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/sysadmin/KeyManagementService.java
```diff
@@ -0,0 +1,117 @@
...
+     */
+    @Post("/masterkeys/session/rotate")
+    @Blocking
+    public CompletableFuture<Void> rotateSessionMasterKey() {
+        if (!encryptionStorageManager.encryptSessionCookie()) {
+            throw new IllegalStateException("Session cookie encryption is disabled.");
+        }
+        final SessionMasterKey currentSessionMasterKey = encryptionStorageManager.getCurrentSessionMasterKey();
+        // If concurrent rotation happened, only one of them will succeed to create a new session master key
+        // because the storage manager accepts only the key with the version which is exactly one greater.
```
jrhee17: Question) While this may be true in terms of keeping consistency, given that an exception will be thrown when checking this condition would replicas go into read-only mode?

I believe I left a similar comment previously as well: https://github.com/line/centraldogma/pull/1181/files/a10997166edba92551567a1a833936c0439ea57a#r2361488383

server/src/main/java/com/linecorp/centraldogma/server/storage/encryption/KeyWrapper.java
jrhee17: Note) No issue with this API, but it seems like it isn't called from anywhere

jrhee17 (approved):
Understood that with the current changes, the rotated key is used to encrypt new sessions/objects.

Conversely, each session/object is tagged with the master dek version and queries the appropriate dek when unencrypting.

server/src/main/java/com/linecorp/centraldogma/server/storage/encryption/KeyWrapper.java
minwoox: Yeah, let me use it in the follow-up PR. 😉

server/src/main/java/com/linecorp/centraldogma/server/auth/SessionMasterKey.java
```diff
@@ -25,48 +27,81 @@
...
+    private final String kekId;
+    private final String creation;
 
     /**
      * Creates a new instance.
+     *
+     * @param wrappedMasterKey the wrapped session master key
+     * @param version the version of the session master key
+     * @param salt the salt used to derive session keys from the master key. It's encoded in base64.
+     * @param kekId the key encryption key (KEK) ID used to wrap the
```
ikhoon: nit: an incomplete sentence?

server/src/main/java/com/linecorp/centraldogma/server/auth/SessionMasterKey.java
```diff
@@ -25,48 +27,81 @@
  */
 public final class SessionMasterKey {
 
-    private final byte[] wrappedMasterKey;
-    private final byte[] salt;
+    private final String wrappedMasterKey;
     private final int version;
+    private final String salt;
+    private final String kekId;
+    private final String creation;
```
ikhoon, line 37: How about using `Instant` for type-safety?

server/src/main/java/com/linecorp/centraldogma/server/command/CreateProjectCommand.java
```diff
@@ -80,20 +86,20 @@ public boolean equals(Object obj) {
...
+        return (projectName.hashCode() * 31 + Objects.hashCode(wdekDetails)) * 31 + super.hashCode();
     }
 
     @Override
     ToStringHelper toStringHelper() {
         final ToStringHelper toStringHelper = super.toStringHelper()
                                                    .add("projectName", projectName);
-        if (wdek != null) {
-            toStringHelper.add("wdek", "[***]");
+        if (wdekDetails != null) {
```
ikhoon: nit: A null value can be set to `.add()`

server/src/main/java/com/linecorp/centraldogma/server/internal/api/sysadmin/KeyManagementService.java
```diff
@@ -0,0 +1,117 @@
...
+        requireNonNull(repoName, "repoName");
+        if (reencrypt) {
+            // TODO(minwoox): Implement re-encryption logic.
+            throw new UnsupportedOperationException();
+        } else {
+            final SecretKeyWithVersion currentDek =
+                    encryptionStorageManager.getCurrentDek(projectName, repoName);
+            return encryptionStorageManager.generateWdek().thenCompose(wdek -> {
+                // If concurrent rotation happened, only one of them will succeed to create a new WDEK because
+                // the storage manager accepts only the WDEK with the version which is exactly one greater.
```
ikhoon: Question) If the same command is accidentally submitted to two different replicas and both succeed locally, could the ZooKeeper replication be stopped?

server/src/main/java/com/linecorp/centraldogma/server/internal/admin/auth/SessionCookieHandler.java
```diff
@@ -17,94 +17,113 @@
...
-                }
-                // In read-only mode, we support authentication using only the username claim.
-                final String subject = jwtClaimsSet.getSubject();
-                if (isNullOrEmpty(subject)) {
-                    return null;
-                }
-                final String csrfTokenFromSignedJwt = csrfTokenFromSignedJwt(signedJwt);
-                return new SessionInfo(null, subject, csrfTokenFromSignedJwt);
+
+            if (sessionKeyVersion == sessionKey.version()) {
```
ikhoon, line 87: I understand that a session issued by the new key may not be able to access servers whose key has not been updated yet while the new key is being replicated.

Technically, we could specify the exact time when the sessionMasterKey should take effect, but since key rotation rarely happens and replication completes within a short period, I don’t think it’s a significant issue.

ikhoon (approved):
👍👍

server/src/main/java/com/linecorp/centraldogma/server/auth/SessionMasterKey.java
minwoox: Fixed. 😉

server/src/main/java/com/linecorp/centraldogma/server/command/CreateProjectCommand.java
minwoox: It's not added if `wdekDetails` is null. Am I missing something? 🤔

server/src/main/java/com/linecorp/centraldogma/server/internal/admin/auth/SessionCookieHandler.java
minwoox: Thanks for the advice. Yeah, let's revisit this if it becomes a problem. 😉

server/src/main/java/com/linecorp/centraldogma/server/command/CreateProjectCommand.java
ikhoon: Sorry, my explanation wasn't enough. I meant we conventionally specify null values to `ToStringHelper` and use `omitNullValues()` to exclude them.

minwoox: Got it. Fixed. 😉

server/src/main/java/com/linecorp/centraldogma/server/auth/SessionMasterKey.java
```diff
@@ -53,6 +52,7 @@ public SessionMasterKey(@JsonProperty("wrappedMasterKey") String wrappedMasterKe
         this.salt = requireNonNull(salt, "salt");
         this.kekId = requireNonNull(kekId, "kekId");
         this.creation = requireNonNull(creation, "creation");
+        creationInstant = Instant.from(DateTimeFormatter.ISO_INSTANT.parse(creation));
```
ikhoon: Do we need to manually parse the value although `InstantDeserializer` is registered to the default `ObjectMapper`?
https://github.com/line/centraldogma/blob/1cafcd1c5f6e407d80ad4a2b28c80b03b3b55104/common/src/main/java/com/linecorp/centraldogma/internal/Jackson.java#L81-L82

minwoox:
Thanks for reviewing. 😉

PR #1212, opened 2025-11-20
Add API endpoint to rewrap all encryption keys with a new KEK

minwoox:
Motivation:
This change introduces a critical security feature for managing the lifecycle of the Key Encryption Key (KEK). If a KEK is ever compromised or needs to be rotated as part of a security policy, all data encrypted by it becomes vulnerable. To mitigate this, all Wrapped Data Encryption Keys (WDEKs) and session master keys that were encrypted with the old KEK must be re-encrypted (or "rewrapped") with a new, secure KEK.

Modifications:
- Added a new `rewrapAllKeys()` API endpoint, `POST /api/v1/keys/rewrap`, to the `KeyManagementService`.

Result:
- System administrators can now trigger a full re-encryption of all application data keys via a secure API call.

## Summary by CodeRabbit

* **New Features**
  * Admin endpoint to rewrap all encryption keys to a new KEK.
  * API to create/rotate session master keys with author attribution.

* **Improvements**
  * Key rotation and rewrap operations now record author info for auditability.
  * DEK listener/caching for faster key access and better resource cleanup.
  * Encryption tasks (rotate/rewrap/create) executed asynchronously to reduce blocking.

* **Tests**
  * End-to-end tests validating rewrap across restarts and KEK propagation.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/command/StandaloneCommandExecutor.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/command/StandaloneCommandExecutor.java
@@ -20,6 +20,7 @@ import static java.util.Objects.requireNonNull;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executor;
 import java.util.function.Consumer;
+import java.util.function.Function;
 
 import javax.annotation.Nullable;
 
@@ -389,18 +390,20 @@ public class StandaloneCommandExecutor extends AbstractCommandExecutor {
         }
 
         logger.info("Rewrapping all keys with kek: {}", encryptionStorageManager.kekId());
-        return CompletableFuture.supplyAsync(encryptionStorageManager::rewrapAllKeys, repositoryWorker)
-                .thenCompose(future -> future)
-                .handle((unused, cause) -> {
-                    if (cause != null) {
-                        logger.warn("Failed to rewrap all keys with kek: {}",
-                                    encryptionStorageManager.kekId(), cause);
-                        Exceptions.throwUnsafely(cause);
-                    } else {
-                        logger.info("All keys rewrapped.");
-                    }
-                    return null;
-                });
+        return CompletableFuture.supplyAsync(() -> encryptionStorageManager.rewrapAllKeys(repositoryWorker),
+                                             repositoryWorker)
+                                .thenCompose(Function.identity())
+                                .handle((unused, cause) -> {
+                                    if (cause != null) {
+                                        logger.warn("Failed to rewrap all keys with kek: {}",
+                                                    encryptionStorageManager.kekId(),
+                                                    cause);
+                                        Exceptions.throwUnsafely(cause);
+                                    } else {
+                                        logger.info("All keys rewrapped.");
+                                    }
+                                    return null;
+                                });
     }
 
     private CompletableFuture<Void> rotateWdek(RotateWdekCommand c) {
--- a/server/src/main/java/com/linecorp/centraldogma/server/storage/encryption/EncryptionStorageManager.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/storage/encryption/EncryptionStorageManager.java
@@ -22,6 +22,7 @@ import java.nio.file.Path;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Executor;
 import java.util.function.Consumer;
 
 import javax.annotation.Nullable;
@@ -225,6 +226,8 @@ public interface EncryptionStorageManager extends SafeCloseable {
     /**
      * Rewraps all wrapped data encryption keys (WDEKs) and session master keys
      * with the {@link EncryptionAtRestConfig#kekId()} specified in the configuration.
+     *
+     * @param executor the {@link Executor} to use for storing re-wrapped keys.
      */
-    CompletableFuture<Void> rewrapAllKeys();
+    CompletableFuture<Void> rewrapAllKeys(Executor executor);
 }
--- a/server/src/main/java/com/linecorp/centraldogma/server/storage/encryption/RepositoryEncryptionStorage.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/storage/encryption/RepositoryEncryptionStorage.java
@@ -36,6 +36,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
 
@@ -623,7 +624,7 @@ final class RepositoryEncryptionStorage {
         return ("wdeks/" + projectName + '/' + repoName + "/current").getBytes(StandardCharsets.UTF_8);
     }
 
-    CompletableFuture<Void> rewrapAllWdeks() {
+    CompletableFuture<Void> rewrapAllWdeks(Executor executor) {
         final List<WrappedDekDetails> allWdeks = wdeks();
         if (allWdeks.isEmpty()) {
             logger.info("No WDEKs to rewrap");
@@ -670,7 +671,7 @@ final class RepositoryEncryptionStorage {
         }
 
         return CompletableFuture.allOf(rewrapFutures.toArray(new CompletableFuture[0]))
-                                .thenAccept(unused -> {
+                                .thenAcceptAsync(unused -> {
                                     final List<WrappedDekDetails> collected =
                                             rewrapFutures.stream().map(CompletableFuture::join)
                                                          .filter(Objects::nonNull)
@@ -712,7 +713,7 @@ final class RepositoryEncryptionStorage {
                                         throw new EncryptionStorageException(
                                                 "Failed to store re-wrapped WDEKs", e);
                                     }
-                                });
+                                }, executor);
     }
 
     private static boolean startsWith(byte[] array, byte[] prefix) {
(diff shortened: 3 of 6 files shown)


server/src/main/java/com/linecorp/centraldogma/server/storage/encryption/EncryptionStorageManager.java
```diff
@@ -151,7 +151,7 @@ static EncryptionStorageManager of(Path path, boolean encryptSessionCookie, Stri
     /**
      * Removes the wrapped data encryption key (WDEK) for the specified project and repository.
      */
-    void removeWdek(String projectName, String repoName, int version);
+    void removeWdek(String projectName, String repoName, int version, boolean removeCurrent);
```
jrhee17, line 155: nit; it seems like `removeCurrent` is always true

server/src/main/java/com/linecorp/centraldogma/server/storage/encryption/RepositoryEncryptionStorage.java
```diff
@@ -578,6 +623,98 @@ private static byte[] repoCurrentWdekDbKey(String projectName, String repoName)
...
+                                                  "Skipping rewrap.",
+                                                  projectRepoVersion(wdekDetails.projectName(),
+                                                                     wdekDetails.repoName(),
+                                                                     wdekDetails.dekVersion()),
+                                                  kekId);
+                                      return null;
+                                  }
+
+                                  return new WrappedDekDetails(wrappedDek,
+                                                               wdekDetails.dekVersion(),
```
jrhee17, line 664: Question) I may be misunderstanding, but given that the active dek is queried by version, is there no need to increment the version for the newly wrapped dek?

server/src/main/java/com/linecorp/centraldogma/server/storage/encryption/SessionKeyStorage.java
```diff
@@ -247,4 +254,102 @@ void addSessionKeyListener(Consumer<SessionKey> listener) {
...
+        final byte[] prefix = "session/master/".getBytes(StandardCharsets.UTF_8);
+        try (RocksIterator iterator = rocksDbStorage.newIterator(
+                rocksDbStorage.getColumnFamilyHandle(WDEK_COLUMN_FAMILY))) {
+            iterator.seek(prefix);
+            while (iterator.isValid()) {
+                final byte[] key = iterator.key();
+                final String keyStr = new String(key, StandardCharsets.UTF_8);
+
+                if (!keyStr.startsWith("session/master/")) {
+                    break;
```
jrhee17, line 270: Question) Just making sure, is a break done instead of continue because it is assumed that iterators are done in alphabetic order?

server/src/main/java/com/linecorp/centraldogma/server/storage/encryption/SessionKeyStorage.java
```diff
@@ -247,4 +254,102 @@ void addSessionKeyListener(Consumer<SessionKey> listener) {
...
+                                                continue;
+                                            }
+                                            final byte[] keyBytes =
+                                                    sessionMasterKeyKey(newSessionMasterKey.version());
+                                            writeBatch.put(
+                                                    rocksDbStorage.getColumnFamilyHandle(WDEK_COLUMN_FAMILY),
+                                                    keyBytes, sessionMasterKeyBytes);
+                                        }
+
+                                        rocksDbStorage.write(writeOptions, writeBatch);
```
jrhee17, line 349: Question) Is there no need to update the `currentSessionKey`?

server/src/main/java/com/linecorp/centraldogma/server/storage/encryption/SessionKeyStorage.java
```diff
@@ -247,4 +254,102 @@ void addSessionKeyListener(Consumer<SessionKey> listener) {
...
+                                                  sessionMasterKey.version(), cause);
+                                      return null;
+                                  }
+                                  if (oldKekId.equals(kekId) &&
+                                      sessionMasterKey.wrappedMasterKey().equals(newWrappedKey)) {
+                                      return null;
+                                  }
+
+                                  return new SessionMasterKey(newWrappedKey,
+                                                              sessionMasterKey.version(),
```
jrhee17, line 308: Question) Similar to the case of `RepositoryEncryptionStorage`, is there no need to increment the version?

jrhee17 (approved):

server/src/main/java/com/linecorp/centraldogma/server/storage/encryption/EncryptionStorageManager.java
minwoox: That is correct. 👍 It will be used later. 😉

server/src/main/java/com/linecorp/centraldogma/server/storage/encryption/RepositoryEncryptionStorage.java
minwoox: >  I may be misunderstanding, but given that the active dek is queried by version, is there no need to increment the version for the newly wrapped dek?

No, it's not. 😉 
Because the dek isn't changed, the version is also not changed. It's just rewraping the dek with another kek. 😉

server/src/main/java/com/linecorp/centraldogma/server/storage/encryption/SessionKeyStorage.java
minwoox: I believe so: 
https://github.com/facebook/rocksdb/wiki/Iterator
```
All data in the database is logically arranged in sorted order. An application can specify a key comparison method that specifies a total ordering of keys. An Iterator API allows an application to do a range scan on the database. The Iterator can seek to a specified key and then the application can start scanning one key at a time from that point. 
```

server/src/main/java/com/linecorp/centraldogma/server/storage/encryption/SessionKeyStorage.java
minwoox: It's just rewraping so the `currentSessionKey` isn't changed. 😉

server/src/main/java/com/linecorp/centraldogma/server/storage/encryption/RepositoryEncryptionStorage.java
minwoox: Because KeK is compromised, we don't have to recreate DEKs. We can simpley rewrap the DEKs with the new KEK.
When DEKs are compromised, we need to recreate the DEKs. It will be handled in this line:
https://github.com/line/centraldogma/blob/9bc952d5200ed76a950fd2847c350b3e850e9dce/server/src/main/java/com/linecorp/centraldogma/server/internal/api/sysadmin/KeyManagementService.java#L70-L73

server/src/main/java/com/linecorp/centraldogma/server/command/StandaloneCommandExecutor.java
```diff
@@ -380,22 +383,64 @@ private Repository repo(RepositoryCommand<?> c) {
...
     }
 
+    private CompletableFuture<Void> rewrapAllKeys() {
+        if (!encryptionStorageManager.enabled()) {
+            throw new IllegalStateException("Encryption is not enabled.");
+        }
+
+        logger.info("Rewrapping all keys with kek: {}", encryptionStorageManager.kekId());
+        return CompletableFuture.supplyAsync(encryptionStorageManager::rewrapAllKeys, repositoryWorker)
+                .thenCompose(future -> future)
```
ikhoon: nit
```suggestion
                .thenCompose(Fuction.identity())
```

server/src/main/java/com/linecorp/centraldogma/server/storage/encryption/RepositoryEncryptionStorage.java
```diff
@@ -578,6 +623,98 @@ private static byte[] repoCurrentWdekDbKey(String projectName, String repoName)
...
+                                                               kekId,
+                                                               wdekDetails.creationInstant(),
```
ikhoon, line 666: Question) Is there a special reason to use the old `creation` time for the new `WrappedDekDetails`? If it is newly created, would it make sense to set it the actual creation?

server/src/main/java/com/linecorp/centraldogma/server/storage/encryption/RepositoryEncryptionStorage.java
```diff
@@ -578,6 +623,98 @@ private static byte[] repoCurrentWdekDbKey(String projectName, String repoName)
...
+                                            rewrapFutures.stream().map(CompletableFuture::join)
+                                                         .filter(Objects::nonNull)
+                                                         .collect(toImmutableList());
+
+                                    logger.info("Storing {} re-wrapped WDEKs", collected.size());
+                                    final ColumnFamilyHandle wdekCf =
+                                            rocksDbStorage.getColumnFamilyHandle(WDEK_COLUMN_FAMILY);
+
+                                    int failedCount = 0;
+                                    try (WriteBatch writeBatch = new WriteBatch();
```
ikhoon, line 685: Question) Is it safe to perform these operations in the thread where rewrap executes?

ikhoon (approved):
👍👍

server/src/main/java/com/linecorp/centraldogma/server/storage/encryption/RepositoryEncryptionStorage.java
minwoox: Because the DEK is not changed. I thought I needed to use the original timestamp. Otherwise, all keys will eventually have the same timestamp, which means that we don't have to keep the creation timestamp.

server/src/main/java/com/linecorp/centraldogma/server/storage/encryption/RepositoryEncryptionStorage.java
minwoox: Thanks for catching this. 😆 I need to fix it.

minwoox:
Thanks for reviewing. 😉

PR #1213, opened 2025-11-24
Add validation to prevent disabling encryption once enabled

minwoox:
Motivation:

Once encryption at rest is enabled, the data in the repository is encrypted. Disabling encryption after it has been enabled could lead to data corruption or make the encrypted data inaccessible.

Modifications:

- Add `validateEncryptionState()` method to `EncryptionAtRestConfig` that:
  - Creates a marker file (`.encryption-enabled`) in the data directory when encryption is first enabled
  - Checks for the presence of this marker file on subsequent startups
- Renamed `EncryptionAtRestConfig` class to `EncryptionConfig`
  - Because it now includes not only encryption at rest settings but also session cookie encryption configuration.

Result:

- System administrators cannot disable encryption once it has been enabled, preventing potential data loss or corruption

## Summary by CodeRabbit

* **Refactor**
  * Unified and renamed the server encryption configuration API and builder method across server and tests.

* **New Features**
  * Introduced a generalized encryption config with validation that enforces and persists encryption state (prevents disabling once enabled).

* **Bug Fixes / Documentation**
  * Updated related documentation and references to reflect the new encryption property name.

* **Tests**
  * Added unit tests covering encryption state, marker-file behavior, re-enable/disable rules, and validation.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/EncryptionConfig.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/EncryptionConfig.java
@@ -79,7 +79,15 @@ public final class EncryptionConfig {
             // Create the marker file to indicate encryption has been enabled
             try {
                 Files.createDirectories(dataDir.toPath());
-                Files.createFile(markerFilePath);
+                final String markerContent =
+                        "# DO NOT DELETE THIS FILE\n" +
+                        "#\n" +
+                        "# This file indicates that encryption has been enabled for this " +
+                        "Central Dogma server.\n" +
+                        "# Once encryption is enabled, it cannot be disabled.\n" +
+                        "#\n" +
+                        "# Deleting this file may cause the server to malfunction or fail to start.\n";
+                Files.write(markerFilePath, markerContent.getBytes());
             } catch (IOException e) {
                 throw new IllegalStateException(
                         "Failed to create encryption marker file: " + markerFilePath, e);


server/src/main/java/com/linecorp/centraldogma/server/EncryptionConfig.java
```diff
@@ -51,7 +58,37 @@ public EncryptionAtRestConfig(@JsonProperty("enabled") @Nullable Boolean enabled
...
+            throw new IllegalStateException(
+                    "Cannot disable encryption after it has been enabled. " +
+                    "Encryption was previously enabled. To proceed, you must keep encryption enabled.");
+        }
+
+        if (enabled && !wasEncryptionEnabled) {
+            // Create the marker file to indicate encryption has been enabled
+            try {
+                Files.createDirectories(dataDir.toPath());
+                Files.createFile(markerFilePath);
```
ikhoon: Optional) What do you think of writing a brief description that this file does in the file itself so that an administrator does not accidentally delete it.

ikhoon (approved):
👍👍

server/src/main/java/com/linecorp/centraldogma/server/EncryptionConfig.java
minwoox: Good suggestion. 👍

PR #1214, opened 2025-11-26
Add functionality to re-encrypt repository data with a new DEK

minwoox:
Motivation:
In the event that a repository's Data Encryption Key (DEK) is compromised, all data encrypted with that key must be considered exposed. To remediate this security vulnerability, we need to re-encrypt all associated data with a new, secure key.

Modifications:
- Added a new method, `EncryptionStorageManager.reencryptRepositoryData()`, to manage the re-encryption process.

Result:
- Administrators can now rotate a repository's encryption key and ensure all its data is re-encrypted.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -1,0 +1,7 @@
+reviews:
+  profile: "chill"
+  high_level_summary: false
+  auto_review:
+    enabled: true
+    drafts: false
+    ignore_title_keywords: ["WIP"]


server/src/main/java/com/linecorp/centraldogma/server/command/StandaloneCommandExecutor.java
```diff
@@ -422,6 +422,19 @@ private CompletableFuture<Void> rotateWdek(RotateWdekCommand c) {
                             c.projectName(), c.repositoryName(), c.wdekDetails().dekVersion(), t);
                 Exceptions.throwUnsafely(t);
             }
+            if (c.reencrypt()) {
+                try {
+                    logger.info("Re-encrypting all data in {}/{} with the new WDEK.",
+                                c.projectName(), c.repositoryName());
+                    encryptionStorageManager.reencryptRepositoryData(c.projectName(), c.repositoryName());
+                    logger.info("All data re-encrypted in {}/{}.", c.projectName(), c.repositoryName());
```
ikhoon, line 430: nit) It may be worth recording the elapsed time for re-encryption.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/sysadmin/KeyManagementService.java
```diff
@@ -129,6 +131,7 @@ public CompletableFuture<Void> rotateSessionMasterKey(Author author) {
      */
     @Post("/keys/rewrap")
     @Blocking
+    @RequestTimeout(60000)
     public CompletableFuture<Void> rewrapAllKeys(Author author) {
```
ikhoon, line 135: Question) 204 No Content status is returned for the successful response?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/sysadmin/KeyManagementService.java
```diff
@@ -64,25 +67,24 @@ public List<WrappedDekDetails> listWdeks() {
...
-            return encryptionStorageManager.generateWdek().thenCompose(wdek -> {
-                // If concurrent rotation happened, only one of them will succeed to create a new WDEK because
-                // the storage manager accepts only the WDEK with the version which is exactly one greater.
-                final int newVersion = currentDek.version() + 1;
-                final WrappedDekDetails wdekDetails = new WrappedDekDetails(
-                        wdek, newVersion, encryptionStorageManager.kekId(),
-                        projectName, repoName);
-                return execute(Command.rotateWdek(author, projectName, repoName, wdekDetails));
-            });
+            ctx.setRequestTimeout(Duration.ofSeconds(60));
```
ikhoon, line 74: What do you think of disabling the request timeout. The processing time is unpredictable since it depends on the size of data and it would be embarrassing for the system administrator if this critical operation is halted due to the timeout.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/sysadmin/KeyManagementService.java
minwoox: Yeah, 204 is returned. Do you have any concerns? 🤔 
FYI:
```
The 204 (No Content) status code indicates that the server has
successfully fulfilled the request and that there is no additional
content to send in the response payload body.
```   
https://datatracker.ietf.org/doc/html/rfc7231#section-6.3.5

Or, do you want to return something?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/sysadmin/KeyManagementService.java
minwoox: Disabled it. 😉 By the way, the operation is not halted but continued until it's done. 😉

server/src/main/java/com/linecorp/centraldogma/server/internal/api/sysadmin/KeyManagementService.java
ikhoon: I prefer 200 OK if it performs successfully. 204 No Content is a cacheable status so it may not be proper for POST.

ikhoon: I don't think we necessarily have to return 200 OK, so using 204 should be fine as well.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/sysadmin/KeyManagementService.java
ikhoon: > the operation is not halted but continued until it's done. 

I know, but the result is a black box from the caller's side if the request finishes due to the timeout.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/sysadmin/KeyManagementService.java
minwoox: POST is not cached by default so it shouldn't be a problem. 😉 
https://www.rfc-editor.org/rfc/rfc9110.html#section-9.3.3
```
Responses to POST requests are only cacheable when they include explicit freshness information (see [Section 4.2.1](https://www.rfc-editor.org/rfc/rfc9111#section-4.2.1) of [[CACHING](https://www.rfc-editor.org/rfc/rfc9110.html#CACHING)]) and a [Content-Location](https://www.rfc-editor.org/rfc/rfc9110.html#field.content-location) header field that has the same value as the POST's target URI ([Section 8.7](https://www.rfc-editor.org/rfc/rfc9110.html#field.content-location)). A cached POST response can be reused to satisfy a later GET or HEAD request. In contrast, a POST request cannot be satisfied by a cached POST response because POST is potentially unsafe; see [Section 4](https://www.rfc-editor.org/rfc/rfc9111#section-4) of [[CACHING](https://www.rfc-editor.org/rfc/rfc9110.html#CACHING)].
```

PR #1218, opened 2025-12-02
Provide a way to view raw JSON content

ikhoon:
Motivation:

Exposing raw content has two benefits:

- It makes it easier to compare the user's original data with the data stored in Central Dogma.
- It simplifies supporting JSON5. When a JSON content is normalized via `JsonNode` before, we lose information present only in the original, such as JSON5 comments, which makes JSON5 support challenging.

In this PR, the content pushed by the user or mirrored by a scheduler is stored in the Central Dogma repository in its original form and a new feature is added to return the raw content when `viewRaw` option is specified on files read or watch operations.

Modifications:

- Push side)
  - Added the `rawContent()` method to `Change`.
    - It stores the original value and is now part of the serialized output.
    - Refactored `Change` implementations to handle `rawContent()` easily.
  - Implemented a new serializer and deserializer for `Change` to correctly handle the new `rawContent` field.
    - If `rawContent` exits, `content` is excluded from the serialized output.
    - Migrated duplicate serializers and deserializers to use the new implementations.
  - `Command.push()` now returns `PushAsIsCommand` since normalization is no longer necessary.
  - `ExecutionContext` is added to provide additional information when executing `Command`.
    - Currently, it only indicates whether the `Command` is executed by a replay.
  - Refactor internal Git repository implementations to store the original content.
- Read side)
  - Added the `rawContent()` method to `Entry`.
    - It stores the original content retrieved from Central Dogma Git repository.
  - Added `viewRaw` parameter to the `getFiles` REST API.
    - When `viewRaw` is true, `EntryDto` includes the original content.
    - `viewRaw` option is disabled by default in the client API for backward compatibility.
    - `viewRaw` option is always enabled in the web application so that users can easily compare the content with the upstream data.
  - Fixed front-end Typescript code to render the raw content in the viewer and store the edited data as-is.

Result:

Central Dogma now support storing and retrieving the original JSON content.

### DIFF
--- a/client/java-armeria/src/main/java/com/linecorp/centraldogma/internal/client/armeria/ArmeriaCentralDogma.java
+++ b/client/java-armeria/src/main/java/com/linecorp/centraldogma/internal/client/armeria/ArmeriaCentralDogma.java
@@ -873,6 +873,8 @@ public final class ArmeriaCentralDogma extends AbstractCentralDogma {
         requireNonNull(query, "query");
         checkArgument(timeoutMillis > 0, "timeoutMillis: %s (expected: > 0)", timeoutMillis);
         if (viewRaw && query.type() == QueryType.JSON_PATH) {
+            // JSON_PATH query cannot return raw content because the raw content is normalized
+            // when applying JSON_PATH.
             throw new IllegalArgumentException("JSON_PATH query cannot be used with raw view");
         }
         try {
@@ -888,7 +890,8 @@ public final class ArmeriaCentralDogma extends AbstractCentralDogma {
                 path.setLength(path.length() - 1);
             }
             if (viewRaw) {
-                path.append(path.indexOf("?") >= 0 ? "&" : "?").append("viewRaw=true");
+                // The query type can't be JSON_PATH here as checked above.
+                path.append("?viewRaw=true");
             }
 
             return watch(lastKnownRevision, timeoutMillis, path.toString(), query.type(),
--- a/common/src/main/java/com/linecorp/centraldogma/common/ChangeDeserializer.java
+++ b/common/src/main/java/com/linecorp/centraldogma/common/ChangeDeserializer.java
@@ -109,6 +109,7 @@ final class ChangeDeserializer extends StdDeserializer<Change<?>> {
                 }
                 break;
             case RENAME:
+                assert content != null;
                 result = Change.ofRename(path, content.asText());
                 break;
             case APPLY_JSON_PATCH:
--- a/common/src/main/java/com/linecorp/centraldogma/common/ChangeSerializer.java
+++ b/common/src/main/java/com/linecorp/centraldogma/common/ChangeSerializer.java
@@ -46,7 +46,7 @@ final class ChangeSerializer extends StdSerializer<Change<?>> {
         if (change.type().contentType() == String.class) {
             if (content != null) {
                 // For text-based changes, always use 'content' field.
-                gen.writeStringField("content", (String) change.content());
+                gen.writeStringField("content", (String) content);
             }
             gen.writeEndObject();
             return;
--- a/xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsResourceManager.java
+++ b/xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsResourceManager.java
@@ -42,6 +42,7 @@ import com.linecorp.centraldogma.common.ChangeConflictException;
 import com.linecorp.centraldogma.common.Markup;
 import com.linecorp.centraldogma.common.RedundantChangeException;
 import com.linecorp.centraldogma.common.Revision;
+import com.linecorp.centraldogma.internal.Jackson;
 import com.linecorp.centraldogma.server.command.Command;
 import com.linecorp.centraldogma.server.command.CommandExecutor;
 import com.linecorp.centraldogma.server.storage.project.Project;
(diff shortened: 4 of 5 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/DefaultChangesApplier.java
```diff
@@ -73,13 +73,23 @@ int doApply(Revision unused, DirCache dirCache,
 
             switch (change.type()) {
                 case UPSERT_JSON: {
-                    final JsonNode oldJsonNode = oldContent != null ? Jackson.readTree(oldContent) : null;
+                    final String rawContent = change.rawContent();
                     final JsonNode newJsonNode = firstNonNull((JsonNode) change.content(),
                                                               JsonNodeFactory.instance.nullNode());
+                    final boolean hasChanges;
```
jrhee17, line 79: Note) Understood that the changes here and at `*Change.java` is the actual change when writing - the rest is refactoring

client/java-armeria/src/main/java/com/linecorp/centraldogma/internal/client/armeria/ArmeriaCentralDogma.java
```diff
@@ -852,11 +866,15 @@ private static Revision watchRepository(AggregatedHttpResponse res, QueryType un
...
                                                      Revision lastKnownRevision, Query<T> query,
-                                                     long timeoutMillis, boolean errorOnEntryNotFound) {
+                                                     long timeoutMillis, boolean errorOnEntryNotFound,
+                                                     boolean viewRaw) {
         validateProjectAndRepositoryName(projectName, repositoryName);
         requireNonNull(lastKnownRevision, "lastKnownRevision");
         requireNonNull(query, "query");
         checkArgument(timeoutMillis > 0, "timeoutMillis: %s (expected: > 0)", timeoutMillis);
+        if (viewRaw && query.type() == QueryType.JSON_PATH) {
+            throw new IllegalArgumentException("JSON_PATH query cannot be used with raw view");
```
jrhee17, line 878: Question) I assumed that json path works since raw content is also a valid json. Do you mind leaving why json path is not supported? I couldn't find any explanation in the PR description.

jrhee17: On second thought, I assume that this is disabled since although a request with `rawView` is sent, a normalized result will be returned. Let me know if I'm misunderstanding

client/java-armeria/src/main/java/com/linecorp/centraldogma/internal/client/armeria/ArmeriaCentralDogma.java
```diff
@@ -869,21 +887,24 @@ public <T> CompletableFuture<Entry<T>> watchFile(String projectName, String repo
                 // Remove the trailing '?' or '&'.
                 path.setLength(path.length() - 1);
             }
+            if (viewRaw) {
```
minwoox, line 892: nit: we could probably just do
```
} else if (viewRaw) {
    path.append('?').append("viewRaw=true");
}
```

common/src/main/java/com/linecorp/centraldogma/common/ChangeDeserializer.java
```diff
@@ -0,0 +1,145 @@
...
+                break;
+            case UPSERT_TEXT:
+                if (rawContent != null) {
+                    result = Change.ofTextUpsert(path, rawContent);
+                } else {
+                    result = Change.ofTextUpsert(path, content.asText());
+                }
+                break;
+            case RENAME:
+                result = Change.ofRename(path, content.asText());
```
minwoox, line 113: let's add an `assert`

common/src/main/java/com/linecorp/centraldogma/common/ChangeSerializer.java
```diff
@@ -0,0 +1,65 @@
...
+        if (change.type().contentType() == Void.class) {
+            gen.writeEndObject();
+            return;
+        }
+
+        final Object content = change.content();
+        if (change.type().contentType() == String.class) {
+            if (content != null) {
+                // For text-based changes, always use 'content' field.
+                gen.writeStringField("content", (String) change.content());
```
minwoox: nit:
```suggestion
                gen.writeStringField("content", (String) content);
```

minwoox (approved):
Thanks! 👍 👍 👍 
Could you also change these lines to use JsonNode instead of rawContent?

https://github.com/line/centraldogma/blob/c2d559d418e1b32130395aebb1db7d561e8fa284/xds/src/main/java/com/linecorp/centraldogma/xds/k8s/v1/XdsKubernetesEndpointFetchingService.java#L318
https://github.com/line/centraldogma/blob/7fd2e775c590f29247d60e006661c9e1e77d0f39/xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsResourceManager.java#L160

client/java-armeria/src/main/java/com/linecorp/centraldogma/internal/client/armeria/ArmeriaCentralDogma.java
ikhoon: JSON Path only operates on the normalized JSON content. It throws an exception because it is not intended to see raw content. 
Although the content is a valid JSON, the JSON that JSON-Path operates on is not raw content due to reordering and formatting.

I will leave a comment about that.

PR #1221, opened 2025-12-04
Fix docker failure

jrhee17:
Fixes https://github.com/line/centraldogma/actions/runs/19922217590

- Run with the same JRE set up as the target toolchain used in docker
- In case some other location is missing, apply `foojay-resolver-convention` so that the necessary toolchain can be downloaded.

### DIFF
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -23,7 +23,7 @@ jobs:
         uses: actions/setup-java@v4
         with:
           distribution: 'zulu'
-          java-version: '25'
+          java-version: 25
 
       - name: Setup Gradle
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-docker.yml
```diff
@@ -19,11 +19,11 @@ jobs:
...
         uses: actions/checkout@v4
 
-      - name: Set up JDK 17
+      - name: Set up JDK 25
         uses: actions/setup-java@v4
         with:
-          distribution: 'temurin'
-          java-version: 17
+          distribution: 'zulu'
+          java-version: '25'
```
ikhoon: nit:
```suggestion
          java-version: 25
```

PR #1222, opened 2025-12-04
Support for JSON5

### LINKED ISSUE #538: Support JSON5

ikhoon:
Motivation: 

This PR enables JSON5 features available in Jackson, preserves JSON5 content as-is, and still provide the same support such as JSON Path that was available for JSON.

Limitation: 

Comments in JSON5 content will be lost when a JSON Patch operation is applied. If you want to preserve the original content, you must use `UPSERT_JSON`.

Related: #538

Modifications:

- Java implementations)
  - `Json5` is introduced to deserialize JSON5 content.
    - Hexadecimal integers (0xdecaf) and trailing decimal points (8675309.) are unsupported due to the limitation of Jackson.
    - Thanks to @ks-yim, most of the code comes from #655
    - `Json5` is used in `Entry` and `JsonChange` to convert JSON5 file into `JsonNode`.
  - Updated `Util.JSON_FILE_PATH_PATTERN` to allow the `.json5` extension.
  - `CentralDogmaRepository.importDir()` now uploads raw JSON files without normalization.
  - Fixed `AbstractGitMirror` to mirror raw JSON content instead of the normalized one.
- Webapp implementations)
  - Implemented JSON5 tokenizer for Monaco editor to provide syntax highlighting.
    - It aligns Jackson's JSON5 features.
  - Add `json5` dependency to validate JSON5 content.
  - Fix `NewFile.tsx` to extract a file extension and set it to `language` for Monaco editor.
    - Previously, it was hardcoded to `json` so other languages such as YAML were not supported.

Result:

- Central Dogma now provides first-class support for JSON5.
- Closes #538

Co-authored-by: KS. Yim <point1304@gmail.com>

### DIFF
--- a/server/src/test/java/com/linecorp/centraldogma/server/Json5CrudTest.java
+++ b/server/src/test/java/com/linecorp/centraldogma/server/Json5CrudTest.java
@@ -225,14 +225,12 @@ class Json5CrudTest {
         assertThatJson(usersEntry.content()).node("[1].name").isEqualTo("Bob");
         assertThatJson(usersEntry.content()).node("[2].name").isEqualTo("Charlie");
 
-        // 다른 JSON Path 쿼리
         final Entry<JsonNode> themeEntry = repo.file(Query.ofJsonPath("/users.json5", "$.settings.theme"))
                                                .get()
                                                .join();
 
         assertThatJson(themeEntry.content()).isEqualTo("\"dark\"");
 
-        // 배열 필터링
         final Entry<JsonNode> adminEntry = repo.file(Query.ofJsonPath("/users.json5",
                                                                       "$.users[?(@.role == 'admin')]"))
                                                .get()
@@ -249,13 +247,11 @@ class Json5CrudTest {
         final Change<JsonNode> change = Change.ofJsonUpsert("/test.json5", json5Text);
         repo.commit("Add test file", change).push().join();
 
-        // JSON_PATH 쿼리는 viewRaw=true와 함께 사용할 수 없음
         final Query<JsonNode> jsonPathQuery = Query.ofJsonPath("/test.json5", "$.key");
         assertThatThrownBy(() -> repo.file(jsonPathQuery).viewRaw(true).get().join())
                 .isInstanceOf(IllegalStateException.class)
                 .hasMessageContaining("JSON_PATH query cannot be used with raw view");
 
-        // viewRaw=false (기본값)는 정상 작동
         final Entry<JsonNode> entry = repo.file(jsonPathQuery).get().join();
         assertThatJson(entry.content()).isEqualTo("\"value\"");
     }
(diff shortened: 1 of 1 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/DefaultChangesApplier.java
```diff
@@ -177,6 +174,7 @@ int doApply(Revision unused, DirCache dirCache,
 
                     // Apply only when the contents are really different.
                     if (!newJsonNode.equals(oldJsonNode)) {
+                        // NB: Some JSON5 features, such as comments, will be lost when using JSON Patch.
```
jrhee17, line 177: Note) Understood the limitation. It seems reasonable for users to assume that json patch will result in lost formatting

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
```diff
@@ -419,7 +419,7 @@ private Map<String, Entry<?>> blockingFind(
                 } else {
                     switch (entryType) {
                         case JSON:
-                            entry = Entry.ofJson(normRevision, path, Jackson.nullNode);
+                            entry = Entry.ofJson(normRevision, path, "");
```
jrhee17, line 422: Understood that with this change, `Entry`s that are of JSON type generated by the server are guaranteed to always have non-null `rawContent`.

ikhoon:
> It seems trivial to also add support for yaml as well.

I'm on the same page.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/DefaultChangesApplier.java
```diff
@@ -83,7 +86,7 @@ int doApply(Revision unused, DirCache dirCache,
                         hasChanges = !rawContent.equals(oldRawContent);
                     } else {
                         // Otherwise, compare the parsed JSON nodes.
-                        final JsonNode oldJsonNode = oldContent != null ? Jackson.readTree(oldContent) : null;
+                        final JsonNode oldJsonNode = toJsonNode(changePath, oldContent);
```
minwoox, line 89: Shouldn't we compare the raw content for JSON5 because a comment might be added or modified?

minwoox (approved):
Left just one comment. Thanks!

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/DefaultChangesApplier.java
ikhoon: `rawContent` is null in this branch. If raw content is provided for changes, it is used to compare.

PR #1223, opened 2025-12-05
Do not fail on no discovered test

minwoox:


minwoox:
> Ah, it fails. 

Glad that I'm not the only one who fails the build. 😉 
Let's wait for #1224 if it's the cache issue or not.

PR #1225, opened 2025-12-05
Add support for mTLS-based application authentication

minwoox:
Motivation:
This change introduces mutual TLS (mTLS) as a first-class authentication method for client applications. mTLS provides a stronger security posture by authenticating the client machine itself through a cryptographic certificate. By supporting mTLS, we can offer a more secure, automatable, and manageable authentication mechanism for server-to-server communication.

Modifications:
- Introduced a new `Application` interface to abstract the concept of a client machine that connects to Central Dogma.
  - Created two implementations of this interface: `Token` for legacy token-based auth, and `Certificate` for the new mTLS-based auth.
- An `appId` is now used as the primary identifier, which maps to a certificate's identity (extracted from the SPIFFE URI by default) or a token ID. Permissions are granted to this `appId`.
- Renamed `Tokens` to `ApplicationRegistry` to reflect that it now manages both tokens and certificates.
- Renamed the corresponding service from `TokenService` to `ApplicationRegistryService`.
- Refactored the codebase by extracting token-related logic from `MetadataService` into the new `ApplicationService` for better encapsulation and clarity.
- Renamed `ProjectMetadata.tokens()` and `Roles.tokens()` to `ProjectMetadata.applications()` and `Roles.applications()` respectively.
- Deprecated the existing token-based REST APIs in favor of new, more generic `/applications` APIs that handle both types.

Result:
- Client applications can now authenticate with the Central Dogma server using mTLS client certificates in addition to API tokens.

To-do:
- Prepare the backward and forward compatiblity logic because tokens are renamed to applications in various places.


server/src/main/java/com/linecorp/centraldogma/server/metadata/ApplicationRegistry.java
```diff
@@ -37,46 +37,53 @@
...
     /**
-     * A mapping of secret and {@link Token#id()}.
+     * A mapping of secret and {@link Token#appId()}.
      */
     private final Map<String, String> secrets;
 
+    /**
+     * A mapping of certificate ID and {@link Certificate#appId()}.
+     */
+    private final Map<String, String> certificateIds;
```
jrhee17, line 61: Note) Understood that this is in sync with `appIds`, and a design choice was made to store persistently instead of computing based on `appIds` in the constructor. (If this were computed transiently, I think `ApplicationService` could possibly be simplified)

server/src/main/java/com/linecorp/centraldogma/server/internal/api/auth/ApplicationCertificateAuthorizer.java
```diff
@@ -0,0 +1,126 @@
...
+        if (peerCertificates.length == 0) {
+            return UnmodifiableFuture.completedFuture(false);
+        }
+
+        String certificateId = null;
+        for (java.security.cert.Certificate peerCert : peerCertificates) {
+            logger.trace("Peer certificate: addr={}, cert={}", ctx.clientAddress(), peerCert);
+            if (!(peerCert instanceof X509Certificate)) {
+                continue;
+            }
```
jrhee17, line 90: Optional) if we would like to disallow intermediate/ca certs from being the source of trust

```
            if (((X509Certificate) peerCert).getBasicConstraints() != -1) {
                continue;
            }
```

server/src/main/java/com/linecorp/centraldogma/server/internal/api/auth/CertificateAuthorizer.java
```diff
@@ -0,0 +1,126 @@
...
+        if (certificateId == null) {
+            logger.trace("No certificateId found in certificate: addr={}", ctx.clientAddress());
+            return UnmodifiableFuture.completedFuture(false);
+        }
+
+        try {
+            final Certificate certificate = certificateLookupFunc.apply(certificateId);
+            if (certificate != null && certificate.isActive()) {
+                final String appId = certificate.appId();
+                ctx.logBuilder().authenticatedUser("appCert/" + appId);
```
ikhoon: Appending the application identifiers to the `appId` (e.g., `app/<appId>/cert` and  `app/<appId>/token`) would make it easier to search for appId accesses using LIKE pattern in OpenSearch or other storage systems.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/sysadmin/ApplicationRegistryService.java
```diff
@@ -0,0 +1,455 @@
...
+            throw new IllegalArgumentException(
+                    "Cannot create a CERTIFICATE type application when mTLS is disabled.");
+        }
+
+        if (!loginUser.isSystemAdmin()) {
+            checkArgument(!isSystemAdmin,
+                          "Only system administrators are allowed to create a system admin-level application.");
+            checkArgument(secret == null,
+                          "Only system administrators are allowed to create a new application token from " +
+                          " the given secret string");
```
ikhoon: nit
```suggestion
                          "Only system administrators are allowed to create a new application token from " +
                          "the given secret string");
```

server/src/main/java/com/linecorp/centraldogma/server/metadata/ApplicationRegistry.java
minwoox: > Understood that this is in sync with appIds, and a design choice was made to store persistently instead of computing based on appIds in the constructor.

Yeah, for fast lookup. 😉

server/src/main/java/com/linecorp/centraldogma/server/internal/api/auth/ApplicationCertificateAuthorizer.java
minwoox: Nice suggestion. 👍

server/src/main/java/com/linecorp/centraldogma/server/metadata/ApplicationRegistry.java
minwoox: Let me think about it a little more.

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -709,23 +717,50 @@ private Server startServer(ProjectManager pm, CommandExecutor executor,
...
+                    if (!mtlsEnabled) {
+                        sb.tls(tlsKeyPair);
+                    } else {
+                        final ServerTlsConfigBuilder serverTlsConfigBuilder =
+                                ServerTlsConfig.builder().clientAuth(ClientAuth.OPTIONAL);
+                        if (!authConfig.mtlsConfig().caCertificateFiles().isEmpty()) {
+                            final List<X509Certificate> caCertificates =
+                                    authConfig.mtlsConfig().caCertificates();
+                            if (!caCertificates.isEmpty()) {
+                                serverTlsConfigBuilder.tlsCustomizer(b -> b.trustManager(caCertificates));
```
ikhoon: Optional) Could use `TlsProviderBuilder` to set the trusted certificates.
```java
TlsProvider
        .builder()
        .trustedCertificates(...)
```

server/src/main/java/com/linecorp/centraldogma/server/metadata/ApplicationRegistry.java
minwoox: I realized that we've used this approach not to compute the maps again when the token or certificate is added.
When a new application is added, we use `JsonPatchOperation.testAbsence(certificateId)` to fail if there's a duplicate key. If we don't store it separately, we need to compute the `appIds` every time an application is updated.

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -846,21 +881,27 @@ static HttpFile webAppTitleFile(@Nullable String webAppTitle, String hostname) {
...
         final BooleanSupplier sessionPropagatorWritableChecker = commandExecutor::isWritable;
+        Authorizer<HttpRequest> authorizer = new ApplicationTokenAuthorizer(mds::findTokenBySecret)
+                .orElse(new SessionCookieAuthorizer(
+                        sessionManager, sessionPropagatorWritableChecker,
+                        tlsEnabled, encryptionStorageManager,
+                        authCfg.systemAdministrators()));
+        if (mtlsEnabled) {
+            // Find the certificate lastly because it raises an exception if no certificate
+            // is found in the connection.
+            authorizer = authorizer.orElse(new CertificateAuthorizer(mds::findCertificateById));
```
ikhoon: I understand that although it is not a common case, when both a token and a mTLS certificate are provided by a client, the token takes precedence.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/auth/ApplicationCertificateAuthorizer.java
```diff
@@ -0,0 +1,126 @@
...
+import io.netty.handler.ssl.OpenSslSession;
+
+/**
+ * An authorizer which extracts certificate ID from mTLS peer certificate and validates it.
+ */
+public final class CertificateAuthorizer implements Authorizer<HttpRequest> {
+
+    private static final Logger logger = LoggerFactory.getLogger(CertificateAuthorizer.class);
+
+    // TODO(minwoox): Make it configurable via SPI.
```
ikhoon, line 53: When it becomes configurable, exposing this API to the builder method would also make testing easier.

server/src/main/java/com/linecorp/centraldogma/server/metadata/Certificate.java
```diff
@@ -0,0 +1,104 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.base.MoreObjects.ToStringHelper;
+import com.google.common.base.Objects;
+
+/**
+ * Specifies details of a certificate-based application identity.
+ */
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonInclude(Include.NON_NULL)
+public final class Certificate extends AbstractApplication {
```
ikhoon: Because `Certificate` is too broad a name, it caused confusion with JDK `Certificate`. How about renaming it with `ApplicationCertificate`?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/sysadmin/ApplicationRegistryService.java
```diff
@@ -0,0 +1,455 @@
...
+    /**
+     * PATCH /applications/{appId}
+     *
+     * <p>Activates or deactivates the application of the specified {@code appId}.
+     */
+    @Patch("/applications/{appId}")
+    @Consumes("application/json-patch+json")
+    public CompletableFuture<Application> updateApplication(ServiceRequestContext ctx,
+                                                            @Param String appId,
+                                                            JsonNode node, Author author, User loginUser) {
```
ikhoon, line 201: Optional) Although JSON-Path expression has been used for PATCH operations, I don't find it convenient for REST API. How about accepting the desired status with a simple JSON notion?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/sysadmin/ApplicationRegistryService.java
```diff
@@ -0,0 +1,455 @@
...
+     * DELETE /applications/{appId}/removed
+     *
+     * <p>Purges an application of the specified ID that was deleted before.
+     */
+    @Delete("/applications/{appId}/removed")
+    @RequiresSystemAdministrator
+    public CompletableFuture<Application> purgeApplication(ServiceRequestContext ctx,
+                                                           @Param String appId,
+                                                           Author author, User loginUser) {
+        return getApplicationOrRespondForbidden(ctx, appId, loginUser).thenApplyAsync(
```
ikhoon, line 182: Optional) It may also be possible to move `getApplicationOrRespondForbidden(ctx, appId, loginUser)` into a request converter to inject `Application` from the method parameters.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/sysadmin/ApplicationRegistryService.java
minwoox: I just followed the old API, but yeah, it's not easy to use. 😓 
We can probably just use `{"status": "active"}` and `{"status": "inactive"}`.
Do you have any other preference?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/sysadmin/ApplicationRegistryService.java
minwoox: With the current design, it's not easy because `getApplicationOrRespondForbidden` is an async operation.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/auth/CertificateAuthorizer.java
minwoox: That's a good suggestion. 👍

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
minwoox: Thanks!
So we can set the `caCertificates` to either `tlsProviderBuilder` or `serverTlsConfigBuilder`, right?

server/src/main/java/com/linecorp/centraldogma/server/metadata/Certificate.java
minwoox: That's a good suggestion. Fixed. 😉

server/src/main/java/com/linecorp/centraldogma/server/internal/api/auth/ApplicationCertificateAuthorizer.java
minwoox: Let me add this condition after we fix the `SignedCertificate` since we always create the certificate with CA option.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/sysadmin/ApplicationRegistryService.java
minwoox: @ikhoon Please let me know if this is what you have expected. 😉 
https://github.com/line/centraldogma/pull/1225/commits/0f0b4824ffd162af31843ea845d3bb74d969f633#diff-ad109b475c83ca8082c1de1f61822706e945f6713516e97c994ef7282ab09060R202

ikhoon: It looks good.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/sysadmin/ApplicationRegistryService.java
ikhoon: I see. We may introduce `AsyncRequestConverterFunction` later if we encounter similar patterns again.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/auth/ApplicationCertificateAuthorizer.java
```diff
@@ -0,0 +1,136 @@
...
+
+    private final Function<String, ApplicationCertificate> certificateLookupFunc;
+
+    public ApplicationCertificateAuthorizer(Function<String, ApplicationCertificate> certificateLookupFunc) {
+        this.certificateLookupFunc = requireNonNull(certificateLookupFunc, "certificateLookupFunc");
+    }
+
+    @Override
+    public CompletionStage<Boolean> authorize(ServiceRequestContext ctx, HttpRequest data) {
+        final SSLSession sslSession = ctx.sslSession();
```
ikhoon, line 64: It occurred to me that we may cache the authorization information in `Channel.attr()` because all requests in the same connection have the same permission.

minwoox: That's brilliant. 👍

PR #1227, opened 2025-12-10
Compress ZooKeeper replication logs

ikhoon:
Motivation:

When a file is newly created or has a large diff, the replication log size can grow significantly. With compression, up to about 7–8 MB of data can be stored in a single znode, improving replication performance.

Modifcations:

- Add `zstd-jni` dependency to the server module.
- Compress replication logs using zstd before storing them in ZooKeeper.

Result:

Central Dogma server now compresses replication logs with zstd.


minwoox (approved):
👍 👍 👍

PR #1228, opened 2025-12-12
Add support for mTLS-based application authentication

minwoox:
Motivation:
This change introduces mutual TLS (mTLS) as a first-class authentication method for client applications. mTLS provides a stronger security posture by authenticating the client machine itself through a cryptographic certificate. By supporting mTLS, we can offer a more secure, automatable, and manageable authentication mechanism for server-to-server communication.

Modifications:
- Introduced a new `AppIdentity` interface to abstract the concept of a client machine that connects to Central Dogma.
  - Created two implementations of this interface: `Token` for legacy token-based auth, and `CertificateAppIdentity` for the new mTLS-based auth.
- An `appId` is now used as the primary identifier, which maps to a certificate's identity (extracted from the SPIFFE URI by default) or a token ID. Permissions are granted to this `appId`.
- Renamed `Tokens` to `AppIdentityRegistry` to reflect that it now manages both tokens and certificates.
- Renamed the corresponding service from `TokenService` to `AppIdentityRegistryService`.
- Refactored the codebase by extracting token-related logic from `MetadataService` into the new `AppIdentityService` for better encapsulation and clarity.
- Renamed `ProjectMetadata.tokens()` and `Roles.tokens()` to `ProjectMetadata.appIds()` and `Roles.appIds()` respectively.
- Deprecated the existing token-based REST APIs in favor of new, more generic `/appIdentities` APIs that handle both types.

Result:
- Client applications can now authenticate with the Central Dogma server using mTLS client certificates in addition to API tokens.

To-do:
- Prepare the backward and forward compatiblity logic because tokens are renamed to applications in various places.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/auth/AuthConfig.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/auth/AuthConfig.java
@@ -202,7 +202,7 @@ public final class AuthConfig {
     /**
      * Returns the mTLS configuration.
      */
-    @JsonProperty
+    @JsonProperty("mtls")
     public MtlsConfig mtlsConfig() {
         return mtlsConfig;
     }
--- a/common/src/main/java/com/linecorp/centraldogma/common/Author.java
+++ b/common/src/main/java/com/linecorp/centraldogma/common/Author.java
@@ -16,8 +16,8 @@
 
 package com.linecorp.centraldogma.common;
 
-import static com.linecorp.centraldogma.internal.Util.APPLICATION_EMAIL_SUFFIX;
-import static com.linecorp.centraldogma.internal.Util.LEGACY_APPLICATION_EMAIL_SUFFIX;
+import static com.linecorp.centraldogma.internal.Util.APP_IDENTITY_EMAIL_SUFFIX;
+import static com.linecorp.centraldogma.internal.Util.LEGACY_APP_IDENTITY_EMAIL_SUFFIX;
 import static com.linecorp.centraldogma.internal.Util.emailToUsername;
 import static java.util.Objects.requireNonNull;
 
@@ -96,11 +96,12 @@ public class Author {
     }
 
     /**
-     * Returns {@code true} if this author is an application.
+     * Returns {@code true} if this author is an app identity.
      */
     @JsonIgnore
-    public boolean isApplication() {
-        return email().endsWith(APPLICATION_EMAIL_SUFFIX) || email().endsWith(LEGACY_APPLICATION_EMAIL_SUFFIX);
+    public boolean isAppIdentity() {
+        return email().endsWith(APP_IDENTITY_EMAIL_SUFFIX) ||
+               email().endsWith(LEGACY_APP_IDENTITY_EMAIL_SUFFIX);
     }
 
     @Override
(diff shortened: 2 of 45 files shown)


server/src/main/java/com/linecorp/centraldogma/server/auth/AuthConfig.java
```diff
@@ -191,6 +199,14 @@ public String sessionValidationSchedule() {
         return sessionValidationSchedule;
     }
 
+    /**
+     * Returns the mTLS configuration.
+     */
+    @JsonProperty
```
ikhoon: I don't know if the serialized format is used anywhere except for logging, but the serialized field is different from the name `@JsonCreator`

server/src/main/java/com/linecorp/centraldogma/server/internal/api/sysadmin/ApplicationRegistryService.java
```diff
@@ -0,0 +1,462 @@
...
+                }, ctx.blockingTaskExecutor());
+    }
+
+    /**
+     * PATCH /applications/{appId}
+     *
+     * <p>Activates or deactivates the application of the specified {@code appId}.
+     */
+    @Patch("/applications/{appId}")
+    @Consumes("application/json-patch+json")
```
ikhoon: Does it now consume json?
```suggestion
    @Consumes("application/json")
```

server/src/main/java/com/linecorp/centraldogma/server/internal/api/sysadmin/ApplicationRegistryService.java
```diff
@@ -0,0 +1,462 @@
...
+                }
+        );
+    }
+
+    /**
+     * PATCH /applications/{appId}/level
+     *
+     * <p>Updates a level of the application of the specified ID.
+     */
+    @Patch("/applications/{appId}/level")
```
ikhoon: Add `@ConsumesJson`?

server/src/main/java/com/linecorp/centraldogma/server/metadata/ApplicationType.java
```diff
@@ -13,10 +13,21 @@
...
+
+package com.linecorp.centraldogma.server.metadata;
+
 /**
- * Central Dogma internal classes.
+ * The type of application.
  */
-@NonNullByDefault
-package com.linecorp.centraldogma.server.internal.command;
+public enum ApplicationType {
```
ikhoon: I think these are the possible types that can authenticate applications.
```suggestion
public enum ApplicationAuthType {
```

server/src/main/java/com/linecorp/centraldogma/server/auth/AuthConfig.java
minwoox: 😱

server/src/main/java/com/linecorp/centraldogma/server/metadata/ApplicationType.java
minwoox: I found the name `Application` to be misleading, as this class represents the application's identity rather than the application itself.
To improve clarity, I've renamed all occurrences to `AppIdentity`. Please take a look.

ikhoon: The new naming looks good.

PR #1229, opened 2025-12-15
Support for YAML

### LINKED ISSUE #151: Better YAML support

ikhoon:
Motivation:

This PR adds YAML support in a similar way to JSON5 and enables the use of advances features such as JSON Path and JSON Patch.

Modifications:

- Adds `ChangeType.UPSERT_YAML` and `EntryType.YAML` for representing YAML files.
  - If the content is YAML, `YAMLMapper` of Jackson is used to (de)serialize the content. The default configuration is copied from #616
- `JsonChange` is renamed to `StructuredChange` to include structured formats such as JSON, JSON5, YAML.
- YAML content can be converted to `JsonNode`, after which JSON Path and Patch operations can be applied.
- Deprecated `type` field in `EntryDto` and fixed the client logic to guess the `EntryType` from the file extension.
  - Previously, YAML content was represented as EntryType.TEXT, but if it is returned as EntryType.YAML, older clients may fail to work.
  - To maintain backward compatibility, when `viewRaw` is not set, the server returns `EntryType.TEXT` instead of `EntryType.YAML`.
  - By removing `EntryDto.type`, we aim to reduce the migration effort required when additional entry types are added later.
  - A trailing slash (`/`) is added to directory paths to make it easier to determine the `EntryType` from the path.
- Used pretty format when writing `JsonNode` to Git to improve readability in web UI.
- Front-end) 
  - Fix to handle the new directory path properly.
  - Enable JSON-Path search for YAML content.
  - Add validation for YAML content.

Result:

- Central Dogma now provides first-class support for YAML.
- Closes #151

---

Co-Authored-By: chacham <ccm0702@naver.com>


server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/AbstractChangesApplier.java
```diff
@@ -80,26 +80,17 @@ static void applyPathEdit(DirCache dirCache, PathEdit edit) {
...
         public void apply(DirCacheEntry ent) {
             try {
-                final byte[] jsonBytes;
-                if (jsonText != null) {
-                    // Use the raw JSON text if available to preserve formatting and comments.
-                    jsonBytes = jsonText.getBytes(UTF_8);
-                } else {
-                    jsonBytes = Jackson.writeValueAsBytes(jsonNode);
-                }
+                final byte[] jsonBytes = Jackson.writeValueAsPrettyString(jsonNode).getBytes(UTF_8);
```
jrhee17, line 93: Note) Understood that the intention of this change is that usage of `ContentTransformer` is intended for internal writes only - hence formatting is discarded.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/PathPatternFilter.java
```diff
@@ -87,12 +87,17 @@ private PathPatternFilter(String pathPattern) {
...
         if (p.charAt(0) != '/') {
             normalized = "/**/" + p;
         } else {
             normalized = p;
         }
+
+        // The path given in match() does not have a trailing '/' for directories.
+        if (normalized.endsWith("/")) {
+            normalized = normalized.substring(0, p.length() - 1);
+        }
```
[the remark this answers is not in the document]
jrhee17: Question) should this comment be addressed?

client/java-armeria/src/main/java/com/linecorp/centraldogma/internal/client/armeria/ArmeriaCentralDogma.java
```diff
@@ -1062,25 +1063,45 @@ private static String toString(AggregatedHttpResponse res) {
...
                 return entryAsText(revision, node, entryPath, viewRaw);
             case IDENTITY_JSON:
             case JSON_PATH:
-                if (receivedEntryType != EntryType.JSON) {
-                    throw new CentralDogmaException("invalid entry type. entry type: " + receivedEntryType +
-                                                    " (expected: " + queryType + ')');
+                switch (receivedEntryType) {
+                    case JSON:
+                        return entryAsJson(revision, node, entryPath, viewRaw);
+                    case YAML:
```
jrhee17, line 1085: Question) Just making sure, while unlikely it seems like the `IDENTITY_JSON` branch result in a yaml file. e.g. querying with JSON_PATH returns a YAML type entry.
Just making sure this is intentional

common/src/main/java/com/linecorp/centraldogma/common/Entry.java
```diff
@@ -80,6 +84,29 @@ public static Entry<JsonNode> ofJson(Revision revision, String path, String cont
         return new Entry<>(revision, path, EntryType.JSON, jsonNode, content);
     }
 
+    /**
+     * Returns a newly-created {@link Entry} of a YAML file with the given content.
+     * @param revision the revision of the YAML file
+     * @param path the path of the YAML file
+     * @param content the content of the YAML file
+     */
+    public static Entry<JsonNode> ofYaml(Revision revision, String path, JsonNode content) {
```
jrhee17, line 93: Question) It isn't easy to track whether `rawContent` is null or non-null. What do you think of removing this and just always populating `rawContent` for yaml types?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/DtoConverter.java
```diff
@@ -77,14 +77,27 @@ public static RepositoryDto newRepositoryDto(Repository repository, RepositorySt
...
         requireNonNull(entry, "entry");
         if (withContent && entry.hasContent()) {
             if (viewRaw) {
                 // Raw content is not null if content is present.
                 return newEntryDto(repository, revision, entry.path(), entry.type(), null, entry.rawContent());
             }
+            if (entry.type() == EntryType.YAML) {
+                if (entry.rawContent() == null) {
+                    // JSON_PATH is used to read the part of YAML files, which is a new feature.
+                    return newEntryDto(repository, revision, entry.path(), entry.type(), entry.content(), null);
```
jrhee17, line 91: Question) Shouldn't `EntryType.YAML` enum never be passed to the client even for the JSON_PATH case for backwards compat?
```suggestion
                    return newEntryDto(repository, revision, entry.path(), EntryType.TEXT, entry.content(), null);
```

jrhee17 (approved):

server/src/main/java/com/linecorp/centraldogma/server/internal/api/DtoConverter.java
jrhee17: On second thought, I guess the assumption is that previously JSON_PATH wouldn't have been called for yaml files. The current change seems reasonable.

webapp/src/dogma/features/file/StructuredFileSupport.ts
```diff
@@ -0,0 +1,51 @@
...
+    case 'json':
+      return JSON.parse(content);
+    case 'json5':
+      return JSON5.parse(content);
+    case 'yaml':
+      return YAML.parse(content);
+    default:
+      throw new Error(`Unsupported structured file format: ${language}`);
+  }
+}
```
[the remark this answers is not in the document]
ikhoon: `yml` is not a language name. `.yml` extension is normalized into `yaml`.
https://github.com/line/centraldogma/blob/d92095ac71ad236a5e887c98b4987acaf69a1837/webapp/src/dogma/common/components/editor/FileEditor.tsx#L53-L54

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/PathPatternFilter.java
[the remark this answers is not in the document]
ikhoon: I think so.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/AbstractChangesApplier.java
ikhoon: Right. In addition, the original content is not preserved when using `ContentTransformer`

common-legacy/src/main/java/com/linecorp/centraldogma/internal/thrift/EntryConverter.java
```diff
@@ -46,6 +46,9 @@ public static Entry convert(com.linecorp.centraldogma.common.Entry<?> entry) {
             case TEXT:
                 file.setContent((String) entry.content());
                 break;
+            case YAML:
+                file.setContent(entry.rawContent());
+                break;
```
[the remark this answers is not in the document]
ikhoon: Thrift does not support YAML.

client/java-armeria/src/main/java/com/linecorp/centraldogma/internal/client/armeria/ArmeriaCentralDogma.java
ikhoon: It is a mistake. I will fix it.

common/src/main/java/com/linecorp/centraldogma/common/Entry.java
ikhoon: `rawContent` is not provided if `viewRaw` is false. This factory method is used for that case. The `rawContent` always exists at the server side and it is nullable at the client side depending on the `viewRaw` option. 
https://github.com/line/centraldogma/blob/cdd0286fd0f1032aca5041e73a5c6663c99d21d4/client/java-armeria/src/main/java/com/linecorp/centraldogma/internal/client/armeria/ArmeriaCentralDogma.java#L1171-L1176

Do you prefer filling the serialized YAML content even for `viewRaw=false` option for simplicity?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/DtoConverter.java
ikhoon: As you said, the client may know `EntryType.YAML` when JSON path is used for YAML.

client/java-armeria/src/main/java/com/linecorp/centraldogma/internal/client/armeria/ArmeriaCentralDogma.java
```diff
@@ -1062,25 +1063,46 @@ private static String toString(AggregatedHttpResponse res) {
...
+                        return entryAsYaml(revision, node, entryPath, viewRaw);
+                    default:
+                        throw new CentralDogmaException("invalid entry type. entry type: " + receivedEntryType +
+                                                        " (expected: " + queryType + ')');
```
minwoox: nit: expected json or yaml?

client/java-armeria/src/main/java/com/linecorp/centraldogma/internal/client/armeria/ArmeriaCentralDogma.java
```diff
@@ -1062,25 +1063,46 @@ private static String toString(AggregatedHttpResponse res) {
...
                     case JSON:
                         return entryAsJson(revision, node, entryPath, viewRaw);
+                    case YAML:
+                        return entryAsYaml(revision, node, entryPath, viewRaw);
                     case TEXT:
                         return entryAsText(revision, node, entryPath, viewRaw);
                     case DIRECTORY:
                         return unsafeCast(Entry.ofDirectory(revision, entryPath));
+                    default:
+                        throw new CentralDogmaException("invalid entry type. entry type: " + receivedEntryType +
```
minwoox: should never reach here?

common/src/main/java/com/linecorp/centraldogma/common/Change.java
```diff
@@ -216,24 +228,33 @@ static Change<JsonNode> ofJsonPatch(String path, @Nullable JsonNode oldJsonNode,
...
+                                    JsonPatch.generate(oldJsonNode, newJsonNode, ReplaceMode.SAFE)
+                                             .toJson());
     }
 
     /**
      * Returns a newly-created {@link Change} whose type is {@link ChangeType#APPLY_JSON_PATCH}.
+     * The JSON patch operation can be applied to JSON, JSON5, YAML files.
+     *
+     * <p>Note that the JSON patch operation normalizes the original data before applying the patch,
+     * so contextual information such as comments cannot be retained when the patch is applied.
```
minwoox, line 247: Let's also add this Javadoc to the above method.

client/java-armeria/src/main/java/com/linecorp/centraldogma/internal/client/armeria/ArmeriaCentralDogma.java
```diff
@@ -1126,6 +1148,35 @@ private static <T> Entry<T> entryAsJson(Revision revision, JsonNode node, String
...
+            final JsonNode rawContent = node.get("rawContent");
+            if (rawContent != null) {
+                try {
+                    return unsafeCast(Entry.ofYaml(revision, entryPath, rawContent.asText()));
+                } catch (JsonParseException e) {
+                    // Should never reach here as the raw JSON text was already validated by the server.
+                    throw new IllegalStateException(e);
+                }
+            }
+            // Should never reach here as the server already supports YAML entries.
```
minwoox: We should deploy this version to the server before a user uses this client. 😉

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/AbstractChangesApplier.java
minwoox: Then, can we just use `writeValueAsBytes`?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/DefaultChangesApplier.java
```diff
@@ -91,12 +91,33 @@ int doApply(Revision unused, DirCache dirCache,
                     }
 
                     if (hasChanges) {
-                        applyPathEdit(dirCache,
-                                      new InsertJson(changePath, inserter, newJsonNode, change.rawContent()));
+                        String newJson = rawContent;
+                        if (newJson == null) {
+                            // Use pretty format for readability in the web UI.
```
minwoox, line 96: So now, we always store a JSON file with the pretty format. Is that correct? If so, do you think it's okay to store a big file (e.g. 50000 lines of JSON)?

client/java-armeria/src/main/java/com/linecorp/centraldogma/internal/client/armeria/ArmeriaCentralDogma.java
ikhoon: Correct. Let me leave a warning instead.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/DefaultChangesApplier.java
ikhoon: > So now, we always store a JSON file with the pretty format. Is that correct?

I guess a formatted JSON content is usually passed as an input for `ofJsonUpsert` by git mirroring or push API. If `ofJsonPatch()` collapses it, the diff becomes noisy and harder to read when using the `viewRaw` option. 

> do you think it's okay to store a big file (e.g. 50000 lines of JSON)?

I don't think storing itself is a problem. Instead, the web UI could be unresponsive and frozen when rendering a large single line.

minwoox: Thanks for the explanation. I was just curious about it. 😉

PR #1231, opened 2025-12-22
Add support for `spring-boot4` modules

jrhee17:
**Motivation**

Following https://github.com/line/armeria/pull/6549, this PR adds support for Spring Boot 4 so users using Armeria and CentralDogma together can use Spring Boot 4.

**Modifications**

- Modified to use `Autowired` in tests to avoid incompatibility between `javax` and `jakarta` annotations
- `spring-boot-validation` is now used instead of `javax.validation` for spring-boot4. A different version of `CentralDogmaSettings` is maintained across versions to handle this.

**Result**
- Support for Spring Boot 4 integration modules


minwoox (approved):
👍 👍 👍

PR #1233, opened 2025-12-26
Extract ssh key files for mirroring test to GitHub secrets

minwoox:
Because GitHub complains about exposing the keys even if they are only used for testing.

### DIFF
--- a/.github/workflows/actions_build.yml
+++ b/.github/workflows/actions_build.yml
@@ -74,47 +74,90 @@ jobs:
       uses: gradle/actions/setup-gradle@v5
 
     - name: Setup SSH test keys
+      env:
+        # ECDSA 256
+        SSH_TEST_ECDSA_256_OPENSSH: ${{ secrets.SSH_TEST_ECDSA_256_OPENSSH }}
+        SSH_TEST_ECDSA_256_OPENSSH_PASSWORD: ${{ secrets.SSH_TEST_ECDSA_256_OPENSSH_PASSWORD }}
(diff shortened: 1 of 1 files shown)


.github/workflows/actions_build.yml
```diff
@@ -73,15 +73,91 @@ jobs:
...
+        SSH_TEST_ED25519_OPENSSH: ${{ secrets.SSH_TEST_ED25519_OPENSSH }}
+        SSH_TEST_ED25519_OPENSSH_PASSWORD: ${{ secrets.SSH_TEST_ED25519_OPENSSH_PASSWORD }}
+        # RSA
+        SSH_TEST_RSA_OPENSSH: ${{ secrets.SSH_TEST_RSA_OPENSSH }}
+        SSH_TEST_RSA_OPENSSH_PASSWORD: ${{ secrets.SSH_TEST_RSA_OPENSSH_PASSWORD }}
+        SSH_TEST_RSA_PEM: ${{ secrets.SSH_TEST_RSA_PEM }}
+        SSH_TEST_RSA_PEM_PASSWORD: ${{ secrets.SSH_TEST_RSA_PEM_PASSWORD }}
+      run: |
+        # Check if secrets are available (e.g., skip on forked PRs)
+        if [ -z "$SSH_TEST_ECDSA_256_OPENSSH" ]; then
```
minwoox, line 103: Mirroring tests that use SSH keys are only run in
- The PR that is created from line/centraldogma branch. e.g. #1234
- The merge commit to the main branch

If this becomes a problem, I will probably modify the test to use a local git server by creating SSH keys every time the test is executed, so that we don't have to store the keys in secrets.

ikhoon (approved):
👍 👍

PR #1237, opened 2025-12-31
Bundle Monaco Editor workers

### LINKED ISSUE #1236: Monaco Editor seems to require internet connection to jsdelivr CDN
Hi,
we're using central dogma very heavily for feature flags and remote configuration inside our production environment.
the problem is, we updated the version all the way from 0.64.2 to 0.78.0, and the new web UI seems to have issues when opened in a environment where the browser can't access the jsdelivr CDN.

Like the image below my work desktop which is inside a environment where it can't access internet, has issues with loading the Monaco editor script which is hosted in jsdelivr CDN.

Image

Image

It would be very helpful if you can make the JS script for monaco editor to be bundled inside the UI package, so that we can use the fresh new UI without internet access

ikhoon:
Motivation:

Central Dogma should be used in an isolated network environment without accessing the internet. However, Monaco Editor dynamically loads workers from internet. See #1236 for details.

Modifications:

- Use `monaco-editor-webpack-plugin` to load Monaco Editor resources with WebPack.
- Use `loader` from '@monaco-editor/react' to load resources locally instead of CDN
- Lazily loading `monaco-editor` in the client side only to avoid accessing the `window` object on the server.

Result:

- You can now use Central Dogma UI without internet access.
- Closes #1236

### DIFF
--- a/webapp/src/dogma/features/file/MonacoLoader.ts
+++ b/webapp/src/dogma/features/file/MonacoLoader.ts
@@ -1,0 +1,46 @@
+import { useEffect, useState } from 'react';
+import { loader } from '@monaco-editor/react';
+
+// Helper to define the worker paths (matches next.config.js)
+const setupMonacoEnv = () => {
+  if (typeof window !== 'undefined' && !window.MonacoEnvironment) {
+    window.MonacoEnvironment = {
+      //eslint-disable-next-line @typescript-eslint/no-explicit-any
+      getWorkerUrl: function (_moduleId: any, label: string) {
+        switch (label) {
+          case 'json':
+            return '/_next/static/json.worker.js';
+          case 'javascript':
+          case 'typescript':
+            return '/_next/static/ts.worker.js';
+          default:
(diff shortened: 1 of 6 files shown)


ikhoon:
image

webapp/src/dogma/features/file/MonacoLoader.ts
```diff
@@ -0,0 +1,46 @@
...
+
+      import('monaco-editor')
+        .then((monacoInstance) => {
+          // Tell the wrapper to use our local instance instead of CDN
+          loader.config({ monaco: monacoInstance });
+
+          setMonaco(monacoInstance);
+        })
+        .catch((err) => {
+          console.error('Failed to load local monaco-editor:', err);
```
jrhee17, line 40: Noted that if the editor is stuck at `<Loading />`, checking the console might help debug issues

jrhee17 (approved):

minwoox (approved):
👍

PR #1243, opened 2026-01-12
Provide a way to store variables via REST API

ikhoon:
Motivation:

If template files are supported, specific values in a configuration file can be overridden, and sensitive information such as database passwords or tokens can be mirrored without being exposed to the upstream Git repository. For example:
```json
{
  "env": "${vars.phase}"
  "db": {
    "host": "${vars.db.host}",
    "pw": "${secrets.db.password}",
  }
}
```

As a first step, this PR implements REST API to store variables in Central Dogma. Follow-up work includes template processing, a UI for variable UI, and secret support.

Modifications:

- Implemented `VariableServiceV1` to store variables at the project level and repository level.
  - The prefix of each API is as follows:
    - `/api/projects/{projectName}/variables`
    - `/api/projects/{projectName}/repos/{repoName}/variables`
  - `JSON` and `STRING` types are supported for variables.
- Introduced `CrudOperation` to perform CRUD operations on arbitrary repository and paths.
  - `GitCrudRepository` now delegates the logic to `CrudOperation`
- Moved `HasRevision` to `com.linecorp.centraldogma.server.storage.repository` to expose it as a public API

Result:

Variables can be stored and retrieved via REST API at the project and repository level.

PR #1244, opened 2026-01-12
Render template files using variables

ikhoon:
Motivation:

As a follow-up to #1243, this PR introduces a templating feature to generate a configuration with variables.

Modifications:

- Server side)
  - [FreeMarker](https://freemarker.apache.org/index.html) is used to generate content from template files.
    - FreeMarker usage is encapsulated in `Templater`.
  - Along with the variables REST API, support for variable files has been added.
    - The following files are supported to store variables in a repository. 
      - `.variables.json`
      - `.variables.json5`
      - `.variables.yaml`
      - `.variables.yml`
    - If there are multiple `.variables.xxx` files at the same level, they take precedence in the order listed above.
    - Variable files located in the same directory as the template file take the highest priority.
    - Variable files in the root directory have the next priority.
    - Variables stored through the REST API have lower priority than those defined in variable files.
    - All variable files are merged, such that values from higher-priority files override those from lower-priority ones.
    - The variables have a `vars.` prefix when accessing them.
      ```json5
      // in .variables.json
      {
        "phase": "alpha"
      }
      // in configuration.json
      {
         "host": "domga-${vars.phase}.com"
      }
      ```
  - Introduced `EntryTransformer` to render template files with variables.
    - It can be specified when fetching entities from `Repository`.
    - `ContentServiceV1` uses `EntryTransformer` to wrap `Templater` and render templates.
  - Added `applyTemplate`, `variableFile` and `variableRevision` parameters to the `getFiles` API to indicate whether and how to render a template file.
  - `Entry.variableRevision` was added to propagate the variable version used when rendering a template.
    - This variable version is required to regenerate the old data on the server-side.
    - The watch API detects changes by comparing the old data with the new data.
- Client side)
  - Added `applyTemplate(boolean)` and `applyTemplate(variableFile)` to the fluent APIs.
  - Cleaned up `ArmeriaCentralDogma` to use `QueryParams` instead of raw query encoding.
  - The file watcher API now relays `variableRevision` along with `lastKnownRevision` to detect new changes triggered by variable updates.

Result:

Template rendering with variables is now supported.

### DIFF
--- a/it/server/src/test/java/com/linecorp/centraldogma/it/FileHistoryAndDiffTest.java
+++ b/it/server/src/test/java/com/linecorp/centraldogma/it/FileHistoryAndDiffTest.java
@@ -78,7 +78,7 @@ class FileHistoryAndDiffTest {
                         .get(Revision.HEAD, Revision.INIT)
                         .join();
         summaries = extractSummaryAsInt(commits);
-        assertThat(summaries).containsExactly(9, 8, 7, 6, 5, 4, 3, 2, 1);
+        assertThat(summaries).containsExactly(9, 8, 7, 6, 5, 4, 3, 2, 1, 0);
     }
 
     private static ImmutableList<Integer> extractSummaryAsInt(List<Commit> commits) {


common/src/main/java/com/linecorp/centraldogma/common/Entry.java
```diff
@@ -181,6 +203,27 @@ public Revision revision() {
...
     }
 
+    /**
+     * Returns the revision of the variables that were used to render the template and generate this
+     * {@link Entry}.
+     *
+     * <p>{@code null} if this {@link Entry} was not created by applying variables to a template.
+     */
+    @Nullable
+    public Revision variableRevision() {
```
jrhee17: Naming) If secrets are also implemented, I understood that this will also use `"variable"Revision` as the representative revision.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/variable/Templater.java
```diff
@@ -0,0 +1,340 @@
...
+        return repo.get(entry.revision(), variableFile).thenApply(entry0 -> {
+            if (entry0.type().type() != JsonNode.class) {
+                throw new TemplateProcessingException(
+                        "The variable file must be a JSON or YAML type: " + variableFile);
+            }
+            return parseVariableFile(entry0);
+        });
+    }
+
+    private static CompletableFuture<Map<String, Object>> findVariableFile(
```
jrhee17: nit; this seems unused

server/src/main/java/com/linecorp/centraldogma/server/internal/api/variable/Templater.java
```diff
@@ -0,0 +1,340 @@
...
+            return UnmodifiableFuture.completedFuture(entry);
+        }
+        if (variableRevision == null) {
+            variableRevision = Revision.HEAD;
+        }
+
+        final String projectName = repo.parent().name();
+        // TODO(ikhoon): Optimize by caching the rendering result for the same set of variables and template.
+        return mergeVariables(crudRepo.findAll(crudContext(projectName, variableRevision)),
+                              crudRepo.findAll(crudContext(projectName, repo.name(), variableRevision)),
```
jrhee17: Understood the variables fetched may originate from different revisions depending on timing.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/variable/Templater.java
```diff
@@ -0,0 +1,340 @@
...
+    private static Object parseValue(Variable variable) {
+        switch (variable.type()) {
+            case JSON:
+                try {
+                    return Jackson.readValue(variable.value(), Object.class);
+                } catch (JsonProcessingException e) {
+                    // Should not reach here as the value has been validated before storing.
+                    throw new IllegalStateException(e);
+                }
+            case STRING:
```
jrhee17, line 288: Note) Understood that the `STRING` type is kept going forward

server/src/main/java/com/linecorp/centraldogma/server/storage/repository/RepositoryWatcher.java
```diff
@@ -0,0 +1,189 @@
...
+            return repo.getOrNull(pair.repoRev, query, transformer).thenAccept(newResult -> {
+                if (errorOnEntryNotFound && newResult == null) {
+                    // The entry is removed.
+                    watchResult.completeExceptionally(
+                            new EntryNotFoundException(pair.repoRev, query.path()));
+                    return;
+                }
+
+                if (newResult == null ||
+                    oldResult != null && Objects.equals(oldResult.content(), newResult.content())) {
```
jrhee17, line 129: Note) Understood that watch terminates if the template-applied contents differ

common/src/main/java/com/linecorp/centraldogma/common/Entry.java
ikhoon: Considering secrets, a more neutral name like templateRevision might be a better choice.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/variable/Templater.java
ikhoon: Nice catch.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/variable/Templater.java
ikhoon: I kept `STRING` type because it is more convenient when users enter values via an input box in the UI.

PR #1247, opened 2026-01-19
Run a mirror when its configuration or a local file is updated

### LINKED ISSUE #1183: Mirror is not updated when only the path has changed
I set up a repository mirror with the following properties:
- Remote: `git+ssh://github.com/KarboniteKream/foo.git`
- Branch: `master`
- Path: `/config/master`

I wanted to update path to `/config/staging`, expecting the Central Dogma repository to be populated with new files, but nothing happened. After manually running the mirror, I get the following error:
```
No changes
Repository 'foo' already at deadbeef, git+ssh://github.com/KarboniteKream/foo.git#master
```

This makes sense because there were no new commits to the GitHub repository. However, I would have expected the mirror operation to continue because even though the commit hash is the same, the requested path is different.

This can be fixed by adding the full remote path or just the repository path to `mirror_state.json`:
```json
{
  "sourceRevision": "deadbeefdeadbeefdeadbeefdeadbeefdeadbeefde",
  "repositoryPath": "/config/master"
}
```

ikhoon:
Motivation:

See #1183 for details.

Modifications:

- Add `remoteRevision`, `remotePath`, `localRevision` and `localPath` fields to `.mirror_state.json` to capture the configuration used when a mirror is run.
- Move `RepositoryUri` to public API and expose it in `Mirror.java` to allow access to the full remote URL in mirror tasks.
- Set the depth of Git fetch to 2 to read parent commit.
  - This is required to detect changes made on the remote in local-to-remote mode.
- If mirrored data becomes dirty due to a manual push, it is overwritten in the next mirror job.

Result:

- Git mirroring now runs correctly when the configuration changes or the mirrored content is manually overridden.
- Closes #1183

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirrorState.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirrorState.java
@@ -36,14 +36,14 @@ import com.linecorp.centraldogma.server.mirror.MirrorDirection;
  *
  * <p>The semantics of {@code remoteRevision} and {@code localRevision} depend on the {@link MirrorDirection}.
  *
- * <h3>Remote-to-local mirroring ({@link MirrorDirection#REMOTE_TO_LOCAL})
+ * <h3>Remote-to-local mirroring ({@link MirrorDirection#REMOTE_TO_LOCAL})</h3>
  * When a new commit is pushed to the local repository:
  * <ul>
  *   <li>{@code remoteRevision} - the current remote HEAD (e.g, Git commit SHA)</li>
  *   <li>{@code localRevision} - the previous local HEAD (e.g, Central Dogma revision)</li>
  * </ul>
  *
- * <h3>Local-to-remote mirroring ({@link MirrorDirection#LOCAL_TO_REMOTE})
+ * <h3>Local-to-remote mirroring ({@link MirrorDirection#LOCAL_TO_REMOTE})</h3>
  * When a new commit is pushed to the remote repository:
  * <ul>
  *   <li>{@code remoteRevision} - the previous remote HEAD (e.g, Git commit SHA)</li>


server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirrorState.java
```diff
@@ -18,22 +18,103 @@
...
+        this.localPath = localPath;
     }
 
+    /**
+     * Returns the source revision. It could be a remote commit hash or a revision version of a local repo
+     * depending on the mirror type.
+     *
+     * @deprecated Use {@link #remoteRevision()} or {@link #localRevision()} instead.
+     */
+    @Deprecated
```
jrhee17, line 85: nit; the proposed logic always gets the previous revision of the target (local for remote2local, remote for local2remote).

It might be easier to understand if a separate `previousTargetRevision()` method is added.

server-mirror-git/src/main/java/com/linecorp/centraldogma/server/internal/mirror/AbstractGitMirror.java
```diff
@@ -173,6 +171,31 @@ GitWithAuth openGit(File workDir,
         }
     }
 
+    boolean shouldRunMirror(@Nullable MirrorState oldMirrorState, Revision localHead,
```
jrhee17: Question) If there are multiple mirrors configured for a single repository, even if a file in directory A is changed will all mirrors be triggered due to also looking at the target revision?

I wonder if adding a hash of the mirror configuration for each mirror to `MirrorState` is a simpler/safer approach

ikhoon: This is an important point. I hadn't considered that multiple mirrors can be configured for a single repository.

>  if adding a hash of the mirror configuration for each mirror to MirrorState is a simpler/safer approach

A mirror configuration hash alone can't tell us whether we should run the mirror when there has been a manual push to the local repo. However, it would be worth adding the hash to detect any changes in mirror configuration.

Although, the logic is complex, the following approach seems to be the most appropriate under the Central Dogma revision model which does not provide the HEAD revision for a directory or a file.

Proposal: For remote-to-local mirroring, it should be safe to rely on the local HEAD when the local path is the root (`/`). If the local path is non-root (e.g, `/subdir`) and `previousLocalRevision` + 1 doesn't match the current HEAD, we could run `repository.diff(previousLocalRevision, headRevision, "/subdir")` to check whether there are actual changes happened before running a mirror task.

jrhee17: > if adding a hash of the mirror configuration for each mirror to MirrorState is a simpler/safer approach

I interpreted the original issue as mirror doesn't run properly when the path (MirrorState) changes. I didn't think it was necessary that the revision of the target is considered for mirroring.

Whatever approach is taken, I'm fine as long as it's consistent for both directions.

ikhoon: I made new changes to properly overwrite a manual push in the next mirror job. PTAL.

server-mirror-git/src/main/java/com/linecorp/centraldogma/server/internal/mirror/AbstractGitMirror.java
```diff
@@ -173,6 +174,79 @@ GitWithAuth openGit(File workDir,
...
+            return MirrorDecision.RUN;
+        }
+        final String previousRemoteCommitId = oldMirrorState.remoteRevision();
+        final String previousLocalRevision = oldMirrorState.localRevision();
+        if (previousRemoteCommitId == null || previousLocalRevision == null) {
+            // Run the mirror to update the legacy mirror state file.
+            return MirrorDecision.RUN;
+        }
+
+        if (!localHead.text().equals(previousLocalRevision)) {
```
jrhee17: Question) I may have not understood as the logic is complex, but is `localHead` ever equal to `previousLocalRevision` given that 
- `localHead` is the current `HEAD` revision
- `previousLocalRevision` is from the `MirrorState`, which means it is the previous local revision?

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirrorState.java
```diff
@@ -18,22 +18,120 @@
...
+    @Nullable
+    private final MirrorDirection direction;
+    @Nullable
+    private final String configHash;
 
+    // TODO(ikhoon): Remove sourceRevision and use either remoteRevision or localRevision.
     @JsonCreator
-    MirrorState(@JsonProperty(value = "sourceRevision", required = true) String sourceRevision) {
+    public MirrorState(@JsonProperty(value = "sourceRevision", required = true) String sourceRevision,
+                       @JsonProperty("remoteRevision") @Nullable String remoteRevision,
```
jrhee17, line 68: Understood that:
- For `localToRemote`
  - `remoteRevision` is the revision before the mirror push (so the previous version)
  - `localRevision` is the previous revision before mirror push
- For `remoteToLocal`
  - `remoteRevision` is the current revision (since no push is done to remote)
  - `localRevision` is the previous revision before mirror push

It might be worth documenting this somewhere since I don't think this is well expressed in the currently proposed code.

server-mirror-git/src/main/java/com/linecorp/centraldogma/server/internal/mirror/AbstractGitMirror.java
```diff
@@ -173,6 +174,79 @@ GitWithAuth openGit(File workDir,
...
+
+            // If the local path is not the root, check whether there are changes under the local path only.
+            return MirrorDecision.COMPARE_AND_RUN;
+        }
+
+        return MirrorDecision.SKIP;
+    }
+
+    private MirrorDecision shouldRunLocalToRemote(@Nullable MirrorState oldMirrorState, Revision localHead,
+                                                  @Nullable ObjectId remoteCommitId)
```
jrhee17: Optional) Naming if a commit id is previous or not may help understand the condition more clearly. Ditto for `shouldRunRemoteToLocal`


```suggestion
                                                  @Nullable ObjectId prevRemoteCommitId)
```

where the previous `prevRemoteCommitId` could be `mirror*`

jrhee17 (approved):
👍 👍

server-mirror-git/src/main/java/com/linecorp/centraldogma/server/internal/mirror/AbstractGitMirror.java
ikhoon: Local-to-remote mirroring pushes a new commit that includes `mirror_state.json` to the remote repository which does not increase the local head revision. If no commits are pushed to either the local or remote repository, `previousLocalRevision` equals `localHead` in next mirroring jobs.

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirrorState.java
ikhoon: `remoteRevision` and `localRevision` represent the revisions at the time a mirroring job updates a change. 
- For local-to-remove, a new commit is pushed to the remote repository.
  - `remoteRevision` indicates the previous head revision (e.g, commit sha236)
  - `localRevision` still equals to the local HEAD.
- For remote-to-local, a new commit is pushed to the local repository.
  - `localRevision` indicates the previous head revision (1234)
  - `remoteRevision` still equals to the remote HEAD.

I will document this behavior in the source code.

ikhoon: Done.

server-mirror-git/src/main/java/com/linecorp/centraldogma/server/internal/mirror/AbstractGitMirror.java
ikhoon: The naming seems confusing. I updated the naming to be consistent.

PR #1248, opened 2026-01-23
Add Micrometer metrics to track xDS resource snapshot versions

ikhoon:
Motivation:

To monitor and track xDS resource changes, it's useful to expose the
version of each resource type as Micrometer metrics. This allows
operators to observe version changes in monitoring systems like
Prometheus.

Modifications:

- Added info gauges in `ControlPlaneService` for each resource type
  with version as a tag
  - Metric name: `xds.control.plane.snapshot.version`
  - Tags: `resource` (cluster/endpoint/listener/route),
    `version` (SHA-256 hash)
- Added version getter methods to `CentralDogmaSnapshot`:
  `clustersVersion()`, `endpointsVersion()`, `listenersVersion()`,
  `routesVersion()`
- Made `allResourceVersion()` package-private in
  `CentralDogmaSnapshotResources`

Result:

The version of xDS resources can now be tracked via Micrometer metrics.

### DIFF
--- a/xds/src/main/java/com/linecorp/centraldogma/xds/internal/ControlPlaneService.java
+++ b/xds/src/main/java/com/linecorp/centraldogma/xds/internal/ControlPlaneService.java
@@ -57,9 +57,7 @@ import io.envoyproxy.envoy.service.discovery.v3.DiscoveryResponse;
 import io.grpc.MethodDescriptor;
 import io.grpc.MethodDescriptor.Marshaller;
 import io.grpc.MethodDescriptor.PrototypeMarshaller;
-import io.micrometer.core.instrument.Gauge;
 import io.micrometer.core.instrument.MeterRegistry;
-import io.micrometer.core.instrument.Tags;
 import io.micrometer.core.instrument.binder.jvm.ExecutorServiceMetrics;
 import io.netty.util.concurrent.DefaultThreadFactory;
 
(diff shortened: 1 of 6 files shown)


minwoox (approved):
Thanks! 👍 👍 👍

xds/src/main/java/com/linecorp/centraldogma/xds/internal/ControlPlaneService.java
```diff
@@ -213,7 +232,73 @@ protected void onFileRemoved(String groupName, String path) {
...
+        final String clustersVersion = snapshot.clustersVersion();
+        if (clustersVersionGauge == null ||
+            !clustersVersion.equals(clustersVersionGauge.getId().getTag("version"))) {
+            if (clustersVersionGauge != null) {
+                clustersVersionGauge.close();
+                meterRegistry.remove(clustersVersionGauge);
+            }
+            clustersVersionGauge = Gauge.builder(VERSION_METRIC_NAME, () -> 1)
+                                        .tags(Tags.of("resource", "cluster", "version", clustersVersion))
+                                        .register(meterRegistry);
```
jrhee17: - It may be worth recording 0 to the gauge to differentiate between 1) a version not loaded at all 2) a server down (e.g. due to maintenance)
- FWIU version strings are a hash, which means the number of `MeterId` may be unbounded with the current implementation.

jrhee17 (approved):
👍 👍

xds/src/main/java/com/linecorp/centraldogma/xds/internal/ControlPlaneService.java
ikhoon: > It may be worth recording 0 to the gauge to differentiate between 1) a version not loaded at all 2) a server down (e.g. due to maintenance)

Good idea.

> The number of MeterId may be unbounded with the current implementation.

When a resource version is changed, the previous meters are closed. Therefore, the number of meters is fixed at 5 regardless of version strings.

jrhee17: > The number of MeterId may be unbounded with the current implementation.

For the record, I was concerned more about downstream aggregators rather than the application

minwoox: I agree with you. Since the hash value is not important, how about using counters instead? We can probably just increase the counter when it's changed, and I think it will be enough.

ikhoon: I'm not sure how to correctly set up an alarm using a counter when the xDS resources is not initialized properly. If the counter does not increase for a certain period time, should that be regarded as an abnormal situation?

If you are concerned the cardinality of the metrics, we may add two metrics.
- A guage to return 0 if it a resource is not initialized.
- A counter to increase when a resource is updated.

minwoox: > If the counter does not increase for a certain period time, should that be regarded as an abnormal situation?

No, I don't think we should rely on the counter to spot that issue. The same applies to using a gauge. If we want to detect that situation, we would need a different approach, such as checking the counter after a change is committed.

ikhoon: There are only five resource types, so I don't think frequent changes to resource versions would overload the downstream aggregators. It would be better to explore other approaches, but integrating change timestamps with the monitoring systems doesn't seem easy. For now, I will set up two metrics suggested above.

minwoox: >  For now, I will set up two metrics suggested above.

Yeah, that will work. 👍

PR #1251, opened 2026-01-29
Add variable UI

ikhoon:
Motivation:

As a final step in template file support, add a UI for manipulating variables so users can manage variables and preview rendered files in the Central Dogma webapp.

Modifications:

- Add CRUD pages for variables in the project and repository settings page.
- Add `Preview with Variables` tab in the file viewer to preview the rendered file with applied variables.
- Rename `applyTemplate` to `renderTemplate`, which seems to convey the intention more clearly for both the method and parameter name.
- Lower the permission level of variable management to `MEMBER` and `WRITE` from `ProjectRole.OWNER` and `RepositoryRole.ADMIN`.
  - If users can modify files, they can also modify `.variables.xxx` file, so granting the same permission seems to make sense.
- Change the file pattern in the file editor to allow dot files.
- Redirect `localhost:3000` to `127.0.0.1:3000` to avoid cross-origin issues in development environments.

Result:

You can now manage variables using the variable UI in Central Dogma web app.

### DIFF
--- a/webapp/src/dogma/common/components/editor/FileEditor.tsx
+++ b/webapp/src/dogma/common/components/editor/FileEditor.tsx
@@ -146,8 +146,6 @@ const FileEditor = ({
     }
   }, [cachedRevision, revision, displayContent]);
 
-  console.log(`file content: ${fileContent}, ${originalContent}, revision: ${revision}`);
-
   const monaco = useLocalMonaco();
   if (!monaco) {
     return <Loading />;


webapp/src/dogma/features/project/settings/variables/VariableForm.tsx
```diff
@@ -0,0 +1,230 @@
...
+            <FormHelperText>
+              A string value renders directly inside your templates. A JSON value must be valid JSON so your
+              templates can access their keys safely.
+            </FormHelperText>
+          </FormControl>
+          <Spacer />
+
+          <FormControl isRequired isInvalid={errors.value != null}>
+            <FormLabel>
+              <LabelledIcon icon={variableType === 'STRING' ? CiText : TbJson} text="Value" />
```
jrhee17, line 165: Note) For the record, I'm still unsure why a separate `String` type is necessary when the UI could map variable name to each json key.

jrhee17 (approved):
👍 👍

webapp/src/dogma/features/project/settings/variables/VariableForm.tsx
ikhoon: There is a slight difference in perspective on this. We may need to look at real use cases later, but personally, I don't think JSON encoding in Web UI without IDE assistance is convenient. 

To provide a better user experience, I thought offering a STRING, so users can simply enter values in an input box without worrrying about encoding, would be helpful. If we only provide JSON type, users would have to add quotes even for simple value, which could make the UI inconvenient.

PR #1257, opened 2026-02-09
Serve invalid YAML files as TEXT type

ikhoon:
Motivation:

We ran a canary deployment for the YAML feature and found that there were more invalid YAML files than expected. When executing queries such as `findAll` (`/**`), a YAML parsing error in a single file could prevent all files from being served.

To avoid an issue in one file affecting the entire repository, we changed the behavior so that invalid YAML files can still be read as plain text. This also makes it easier to identify and fix problematic files.

Modifications:

- Used `Entry.ofText()` if a YAML file is invalid when reading the file from Git repository on the server.
- Improved error messages on the client side to easily identify invalid files.
- Changed Jackson deserializers to throw `JsonProcessingException` instead of `JsonParsingException` to avoid unexpected `IOError`.

Result:

Invalid YAML files are now served as plain text instead of failing.

### DIFF
--- a/common/src/main/java/com/linecorp/centraldogma/common/ContentHolder.java
+++ b/common/src/main/java/com/linecorp/centraldogma/common/ContentHolder.java
@@ -114,7 +114,7 @@ public interface ContentHolder<T> {
         } catch (JsonParseException e) {
             throw e;
         } catch (JsonProcessingException e) {
-            throw new JsonParseException(null, "Failed to parse the content as JSON", e);
+            throw new JsonParseException(null, "Failed to parse the content", e);
         }
     }
 


server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
```diff
@@ -385,62 +386,75 @@ private Map<String, Entry<?>> blockingFind(
...
-                if (treeWalk.isSubtree()) {
-                    if (matches) {
-                        // Add the directory itself to the result set if its path matches the pattern.
-                        result.put(path, Entry.ofDirectory(normRevision, path));
-                    }
-
-                    treeWalk.enterSubtree();
-                    continue;
-                }
+                try {
```
ikhoon, line 389: Recommend reviewing with whitespace changes hidden.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
```diff
@@ -385,62 +386,75 @@ private Map<String, Entry<?>> blockingFind(
...
+                        switch (entryType) {
+                            case JSON:
+                                entry = Entry.ofJson(normRevision, path, string);
+                                break;
+                            case YAML:
+                                Entry<?> maybeYaml;
+                                try {
+                                    maybeYaml = Entry.ofYaml(normRevision, path, string);
+                                } catch (JsonProcessingException e) {
+                                    // Fall back to text entry if the content is not valid YAML.
```
minwoox, line 422: How about adding a debug log?

common/src/main/java/com/linecorp/centraldogma/internal/Yaml.java
```diff
@@ -35,20 +34,14 @@ public final class Yaml {
...
+    public static JsonNode readTree(String data) throws JsonProcessingException {
+        return mapper.readTree(data);
     }
 
-    public static JsonNode readTree(byte[] data) throws JsonParseException {
+    public static JsonNode readTree(byte[] data) throws JsonProcessingException {
         try {
             return mapper.readTree(data);
-        } catch (JsonParseException e) {
+        } catch (JsonProcessingException e) {
```
ikhoon, line 44: I found that Jackson could throw `MismatchedInputException` (rather than a JsonParseException) when duplicate keys are detected, and this was converted to `IOError`. Since `IOError` is not a proper type for a validation error, I update the code to catch all `JsonProcessingException`. 
```
Duplicate field 'required' for `ObjectNode`: not allowed when `DeserializationFeature.FAIL_ON_READING_DUP_TREE_KEY` enabled
 at [Source: (StringReader); line: 98, column: 11]java.io.IOError: com.fasterxml.jackson.databind.exc.MismatchedInputException: Duplicate field 'required' for `ObjectNode`: not allowed when `DeserializationFeature.FAIL_ON_READING_DUP_TREE_KEY` enabled
 at [Source: (StringReader); line: 98, column: 11]
	at com.linecorp.centraldogma.endpoint.Yaml.readTree(Yaml.java:33)
```

PR #1259, opened 2026-02-10
Handle YAML upsert as text for backward compatibility

ikhoon:
Motivation:

Old Central Dogma server earlier than 0.78.0 need to handle `YAML_UPSERT`. During a rolling update:
- `YAML_UPSERT` requests may be requested to old servers.
- A YAML file may be created in the new web app, but the request can be routed to an old server.

To address this issue, added logic to treat `YAML_UPSERT` as `TEXT_UPSERT`.

Modifications:

- Fall back to `UPSERT_TEXT` if an unknown change type is set.
- Fall back to `rawContent` if `content` is null for `UPSERT_TEXT`

Result:

Central Dogma 0.78.1 is forward-compatible with 0.79.0.

### DIFF
--- a/common/src/test/java/com/linecorp/centraldogma/common/ChangeDeserializeTest.java
+++ b/common/src/test/java/com/linecorp/centraldogma/common/ChangeDeserializeTest.java
@@ -57,7 +57,7 @@ class ChangeDeserializeTest {
     @Test
     void handleYamlUpsertAsUpsert() throws JsonProcessingException {
         final String yaml =
-                "{ \"type\": \"YAML_JSON\", \"path\": \"/foo.yaml\", " +
+                "{ \"type\": \"UPSERT_YAML\", \"path\": \"/foo.yaml\", " +
                 "\"rawContent\": \"key: value\" }";
         final Change<String> change = mapper.readValue(yaml, Change.class);
         assertThat(change.type()).isEqualTo(ChangeType.UPSERT_TEXT);


common/src/main/java/com/linecorp/centraldogma/common/DefaultChange.java
```diff
@@ -28,29 +28,44 @@
...
                                         @JsonProperty("path") String path,
                                         @JsonProperty("content") @Nullable JsonNode content,
                                         @JsonProperty("rawContent") @Nullable String rawContent) {
         requireNonNull(type, "type");
-        final Class<?> contentType = type.contentType();
+        ChangeType changeType;
+        try {
+            changeType = ChangeType.parse(type);
+        } catch (IllegalArgumentException e) {
+            // Fallback for backward compatibility. `UPSERT_YAML` or other types are treated as `UPSERT_TEXT`.
```
minwoox: How about fallback only when the string is `UPSERT_YAML` and rethrowing the exception if it's not?

ikhoon: We could do more precise validation if we only handled `UPSERT_YAML`, but thinking about how we should handle unknown `ChangeType`s, which may be added in the future versions, on the main branch going forward, falling back to `UPSERT_TEXT` seems better for compatibilty.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/converter/ChangesRequestConverter.java
```diff
@@ -91,7 +98,11 @@ private static Change<?> readChange(JsonNode node) {
 
         final String path = node.get("path").textValue();
         if (changeType == ChangeType.UPSERT_TEXT) {
-            return Change.ofTextUpsert(path, node.get("content").textValue());
+            final JsonNode content = node.get("content");
+            if (content != null) {
+                return Change.ofTextUpsert(path, content.textValue());
+            }
+            return Change.ofTextUpsert(path, node.get("rawContent").textValue());
```
jrhee17, line 98: Question) I was under the impression that `rawContent` has higher priority than `content`. Is there a reason `content` is preferred over `rawContent` here?

common/src/main/java/com/linecorp/centraldogma/common/DefaultChange.java
minwoox: Let's say we add `REMOVE_YAML` in the future. That will be very weird if it's changed to `UPSERT_TEXT`.
How about checking if the string starts with `UPSERT` at least?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/converter/ChangesRequestConverter.java
ikhoon: There is no particular priority. If both rawContent and content are present, it can be considered invalid input. Although this validation exists in the main branch, I only applied a minimal fix in the current logic.

https://github.com/line/centraldogma/blob/d9964365722d7030e6a256961a7c4389b0f5b1fb/common/src/main/java/com/linecorp/centraldogma/common/ChangeDeserializer.java#L85-L92

common/src/main/java/com/linecorp/centraldogma/common/DefaultChange.java
ikhoon: I considered that `UPSERT_TOML` is more likely to be added than `REMOVE_YAML`. However, both are hypothetical scenarios that are unlikely to happen in practice, so I will only apply a fix for the current state.

common/src/test/java/com/linecorp/centraldogma/common/ChangeDeserializeTest.java
```diff
@@ -46,11 +47,22 @@ void deserializeRawContentAsJson() throws JsonProcessingException {
...
-        final ObjectMapper mapper = JacksonUtil.newDefaultObjectMapper();
         final Change<JsonNode> change = mapper.readValue(json, Change.class);
         assertThat(change.type()).isEqualTo(ChangeType.UPSERT_JSON);
         assertThat(change.path()).isEqualTo("/foo.json");
         assertThat(change.content()).isInstanceOf(ObjectNode.class);
         assertThat(change.content().get("key").asText()).isEqualTo("value");
     }
+
+    @Test
+    void handleYamlUpsertAsUpsert() throws JsonProcessingException {
```
minwoox, line 58: Oops, we need to fix this test.

PR #1261, opened 2026-02-11
Make `DiffEditor` readonly

ikhoon:
Motivation:

`DiffEditor` was added to preview changes but it was accidently editable. As a result, edits maded in diff view weren't applied to the original content. Also, the diff shown in `Preview with Variables` doesn't represent an actual change.

Modifications:

- Make `DiffEditor` readonly

Result:

The diff tab is now readonly to prevent edits in the file viewer.

### DIFF
--- a/webapp/src/dogma/common/components/editor/FileEditor.tsx
+++ b/webapp/src/dogma/common/components/editor/FileEditor.tsx
@@ -263,6 +263,7 @@ const FileEditor = ({
               original={fileContent}
               modified={modifiedContent}
               options={{
+                readOnly: true,
                 autoIndent: 'full',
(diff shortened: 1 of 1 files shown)


jrhee17 (approved):
👍👍

minwoox:
Can't we display `Preview with Variables` only when the file is a template? 🤔

ikhoon:
It would be difficult to determine whether a template exists in a large file. In such case, `Preview with Variables` seems useful. If is were shown only conditionally, users may get confused by when it appears. That's why I prefer the current design.

minwoox:
Had a chat, and we decided to show `No template variables detected.` in the Preview with Variables tab.

PR #1262, opened 2026-02-12
Show No template variables detected in the preveiw tab

minwoox:


minwoox (commented):
Please use hide whitespace when reviewing.

PR #1293, opened 2026-04-15
Fix stale repository references after encryption migration

minwoox:
Motivation:
- When `migrateToEncryptedRepository()` or `fallbackToFileRepository()` replaces a `Repository` instance, code holding references to the old instance becomes stale. Three cases were identified:
  1. `RepositoryListener`s registered on the old `GitRepository` are not transferred to the new one, causing metadata and token change detection to stop working.
  2. `DefaultProject.metaRepo` keeps wrapping the old repository, so all `MetaRepository` operations fail after migration.
  3. `XdsResourceWatchingService.watchRepository()` captures a `Repository` in a recursive async loop, causing the watch to permanently fail after the repository is replaced.

Modifications:
- Transfer all `RepositoryListener`s from the old repository to the new one in both `migrateToEncryptedRepository()` and `fallbackToFileRepository()` before closing the old repository.
- Add `setPostMigrationCallback()` to the `RepositoryManager` interface and implement it in `GitRepositoryManager` and `RepositoryManagerWrapper`. The callback is invoked after a successful migration or fallback with the repository name and the new `Repository` instance.
- Register a post-migration callback in `DefaultProject` to replace `metaRepo` with a new `DefaultMetaRepository` wrapping the new repository when `REPO_DOGMA` is migrated.
- Change `XdsResourceWatchingService.watchRepository()` to accept a group name instead of a `Repository` instance, re-fetching the latest repository from `xdsProject.repos().get()` on each watch cycle.

Result:
- `RepositoryListener`s (including metadata and app-identity-registry listeners) continue to work after migration without re-registration.
- `DefaultProject.metaRepo` is refreshed to wrap the new repository.
- `XdsResourceWatchingService` watch loops automatically pick up the new repository instance after migration.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/project/DefaultProject.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/project/DefaultProject.java
@@ -111,6 +111,7 @@ public class DefaultProject implements Project {
                 author = Author.ofEmail(creation.user());
                 attachMetadataListener();
                 metaRepo = new DefaultMetaRepository(repos.get(REPO_DOGMA));
+                registerMigrationCallback();
             } else {
                 creationTimeMillis = repos.get(REPO_DOGMA).creationTimeMillis();
                 author = repos.get(REPO_DOGMA).author();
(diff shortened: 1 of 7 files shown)


ikhoon (approved):
👍👍

jrhee17 (approved):
👍 👍

PR #1294, opened 2026-04-15
Remove the method filter in CentralDogmaBeanFactory and deprecate bidirectional

zzhen0v0:
## Motivation
- The `MethodFilter` set on the Javassist `ProxyFactory` in `CentralDogmaBeanFactory` only intercepted methods with zero parameters(when `bidirectional = false`, the default).
- Any method with parameters was not intercepted and would read the default field value directly from the proxy instance, ignoring updates from the Central Dogma watcher.

## Modifications
 - Remove `ProxyFactory.setFilter()` call in `CentralDogmaBeanFactory`so that all methods are intercepted by the proxy
 - Deprecate `@CentralDogmaBean.bidirectional`, which is no longer referenced
 - Add `methodWithParameterShouldBeIntercepted` integration test to verify that methods with parameters return values from the latest watched value

## Result
All methods on a `@CentralDogmaBean` proxy are now intercepted and delegated to the latest watched value, regardless of their parameter count.

### DIFF
--- a/client/java/src/main/java/com/linecorp/centraldogma/client/updater/CentralDogmaBean.java
+++ b/client/java/src/main/java/com/linecorp/centraldogma/client/updater/CentralDogmaBean.java
@@ -70,6 +70,10 @@ public @interface CentralDogmaBean {
     /**
      * If {@code true}, the change of each bean property will be pushed to Central Dogma.
      * Use this property with caution because it can result in unnecessarily large number of commits.
+     *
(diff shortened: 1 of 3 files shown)


jrhee17 (approved):
👍 👍

ikhoon (approved):
Thanks, @zzhen0v0!

minwoox (approved):
Thanks, @zzhen0v0!

PR #1295, opened 2026-04-17
Allow the dogma project's dogma repository to be migrated to an encrypted repository

minwoox:
Motivation:
- Previously, the dogma project and its repositories were excluded from encryption migration. Modifications:
- Remove the restriction that blocks the dogma project from encryption migration in `RepositoryServiceV1.validateMigrationPrerequisites()`.
- Only the dogma repository is allowed to be migrated in the dogma project; other repositories in the dogma project are rejected with a bad request.
- Skip the read-only status transition for the dogma project because its project metadata is stored in the dogma repository itself, making it impossible to change the repository status.
- Stop creating the deprecated meta repository in `InternalProjectInitializer.initializeInternalRepos()`. The meta repository is no longer used, and `isInternalRepo("meta")` still returns `true` to prevent name collisions.

Result:
- The dogma project's dogma repository can now be migrated to an encrypted repository.
- The meta repository is no longer created on new server initialization.


ikhoon (approved):
👍👍

jrhee17 (approved):
👍 👍

PR #1296, opened 2026-04-23
Validate file content compatibility on rename across entry types

minwoox:
Motivation:
- When renaming a text file to a JSON/YAML path (e.g., `/a.txt` to `/a.json`) via `Change.ofRename()`, the rename succeeded without validating that the file content was valid for the target entry type.

Modifications:
- Added `validateRenameContent()` in `DefaultChangesApplier` to check that the existing file content is parseable as the target entry type when a rename changes the file extension from text to JSON/YAML.

Result:
- Renaming a file to a structured file path (JSON/YAML) now fails fast if the content is not valid for the target format.


jrhee17 (approved):
👍

ikhoon (approved):
👍 👍

PR #1297, opened 2026-04-23
Fix pretty print not being applied to single-line JSON files

minwoox:
Motivation:
- When the server stores JSON content, it appends a trailing newline (`\n`). The existing single-line check `originalContent.indexOf('\n') === -1` fails because of this trailing newline, so compact JSON files are displayed as a single unformatted line.

Modifications:
- Trim the content before checking for newlines so that a trailing newline does not prevent the pretty print logic from being applied.

Result:
- Single-line JSON files are now correctly pretty-printed in the file viewer.


jrhee17 (approved):
👍 👍

ikhoon (approved):

PR #1299, opened 2026-04-23
Persist pagination state in URL query parameters

minwoox:
Motivation:
When navigating to a repository from a paginated list and pressing the browser back button, the pagination state (page number and page size) was lost and reset to page 1 with the default page size of 10.

Modifications:
- Sync pagination state (page index and page size) with URL query parameters.
  - Default values (page=1, pageSize=10) are omitted from the URL to keep it clean.

Result:
- Browser back/forward navigation now restores the correct pagination state across all pages that use DataTableClientPagination.


ikhoon (approved):
👍👍

jrhee17 (approved):
👍 👍

PR #1301, opened 2026-04-30
Allow custom Node IP extraction for xDS Kubernetes endpoints via SPI

ikhoon:
Motivation:

The xDS Kubernetes integration always uses a Node's `InternalIP` when building endpoints for a `ServiceEndpointWatcher`. Operators sometimes need a different address (for example, an external IP carried in a Node label/annotation, or one derived from per-watcher configuration), and no extension point exists for that today.

Modifications:

- Add a `KubernetesNodeIpExtractor` SPI to extract an IP address from `Node`.
- Add an `additional_properties` map to the `ServiceEndpointWatcher` proto so SPI implementations can be configured per watcher.
- The loaded `KubernetesNodeIpExtractor` is wired into `KubernetesEndpointGroupBuilder.nodeIpExtractor` if specified in `XdsKubernetesService`
- Add a new `:it:xds-k8s-node-ip-extractor` integration module and test with a label-based extractor.

Result:

- Users can plug in a `KubernetesNodeIpExtractor` via `META-INF/services` to override how a Node's IP is resolved, and tune behaviour per watcher through `ServiceEndpointWatcher.additional_properties`.

### DIFF
--- a/it/xds-k8s-node-ip-extractor/src/test/java/com/linecorp/centraldogma/it/xds/k8s/LabelBasedNodeIpExtractor.java
+++ b/it/xds-k8s-node-ip-extractor/src/test/java/com/linecorp/centraldogma/it/xds/k8s/LabelBasedNodeIpExtractor.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2026 LINE Corporation
+ *
+ * LINE Corporation licenses this file to you under the Apache License,
+ * version 2.0 (the "License"); you may not use this file except in compliance
+ * with the License. You may obtain a copy of the License at:
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.linecorp.centraldogma.it.xds.k8s;
+
+import java.util.Map;
+import java.util.Queue;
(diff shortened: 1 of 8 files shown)


it/xds-k8s-node-ip-extractor/src/test/java/com/linecorp/centraldogma/it/xds/k8s/LabelBasedNodeIpExtractor.java
```diff
@@ -0,0 +1,67 @@
...
+    public String extract(ServiceEndpointWatcher watcherConfig, Node node) {
+        invocations.add(watcherConfig);
+        final String labelKey = watcherConfig.getAdditionalPropertiesOrDefault(NODE_IP_LABEL_PROPERTY, null);
+        if (labelKey == null) {
+            return internalIp(node);
+        }
+        final Map<String, String> labels = node.getMetadata().getLabels();
+        if (labels == null) {
+            return null;
+        }
```
jrhee17, line 51: Understood:
- if labelKey is null, it is assumed the default nodeIpExtractor behavior is used
- if a node doesn't have the target label, the node will not be considered for endpoint selection

jrhee17 (approved):
👍 👍

it/xds-k8s-node-ip-extractor/src/test/java/com/linecorp/centraldogma/it/xds/k8s/LabelBasedNodeIpExtractor.java
ikhoon: Correct. Although this is test code, I thought the pattern may be used in real-world implementations.

minwoox (approved):
👍 👍 👍

PR #1303, opened 2026-05-12
Fix replication replay failure after copying data directory from another replica

ikhoon:
Motivation:

When a replica's data directory was copied from another replica (e.g., for disaster recovery) and the replica was restarted, replication replay could silently skip self-originated commands and leave the local data inconsistent with the ZooKeeper log. A subsequent push from another replica then failed with `mismatching replay result` and the restored replica entered read-only mode.

The root cause was that `ZooKeeperCommandExecutor.replayLogs` skipped log entries whose originating replica id matched the running replica's id. https://github.com/line/centraldogma/blob/ab4a05a51c48df8e8905c270b61eeba69f534057/server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java#L1044-L1054 https://github.com/line/centraldogma/blob/ab4a05a51c48df8e8905c270b61eeba69f534057/server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java#L832-L847

The skip assumed the local state already had those changes — true under normal operation because `blockingExecute` applies the command locally before storing the log to ZooKeeper, but no longer true after a cross-replica restore.

Modifications:

- Add a new per-replica progress file `<dataDir>/local_last_revision` that tracks the highest ZooKeeper revision whose effect is actually in the local data. It is updated only by `storeLog`, reflecting this replica's own self-origination history.
- Change the replay skip decision to be data-driven instead of based on the originating replica id, so a restored data directory replays every revision the local data is missing.
- On startup, fall back to `last_revision` when `local_last_revision` is absent so binaries upgraded from older versions keep working.
- Expose a `replica.last.local.applied.revision` gauge alongside the existing `replica.last.replayed.revision`.
- Add unit tests (data-restore regression, missing/corrupt `local_last_revision` handling) and integration tests covering commit, restart, follower catch-up, per-replica state, and gauge behavior.

Result:

Replicas that have been restored from another replica's data directory now replay any revisions missing from local data instead of silently skipping them, eliminating the read-only-mode incident triggered by `mismatching replay result`.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
@@ -33,7 +33,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Objects;
-import java.util.Optional;
 import java.util.Properties;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
@@ -135,6 +134,7 @@ public final class ZooKeeperCommandExecutor
 
     private final ZooKeeperReplicationConfig cfg;
     private final File revisionFile;
+    private final File localRevisionFile;
     private final File zkConfFile;
     private final File zkDataDir;
     private final File zkLogDir;
@@ -347,7 +347,10 @@ public final class ZooKeeperCommandExecutor
     }
 
     private static final class ListenerInfo {
+        // Highest ZK revision processed by the replay loop.
         long lastReplayedRevision;
+        // Highest ZK revision applied to local data by storeLog.
+        long localLastAppliedRevision;
         @Nullable
         final Runnable onTakeLeadership;
         @Nullable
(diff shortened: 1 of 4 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -383,6 +387,8 @@ public ZooKeeperCommandExecutor(ZooKeeperReplicationConfig cfg,
         this.cfg = requireNonNull(cfg, "cfg");
         requireNonNull(dataDir, "dataDir");
         revisionFile = new File(dataDir.getAbsolutePath() + File.separatorChar + "last_revision");
+        localRevisionFile = new File(dataDir.getAbsolutePath() + File.separatorChar +
```
jrhee17, line 394: I understood `local_last_revision` should also be copied over when applying changes

jrhee17 (approved):
👍 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -829,11 +872,11 @@ private synchronized void replayLogs(long targetRevision) {
             }
             ReplicationLog<?> l = null;
             try {
-                final Optional<ReplicationLog<?>> log = loadLog(nextRevision, true);
-                Command<?> command = null;
-                if (log.isPresent()) {
-                    l = log.get();
-                    command = l.command();
+                l = loadLog(nextRevision);
```
minwoox: let's call loadLog in the `if (shouldExecute) {...}`

minwoox (approved):
Just one nit. 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
ikhoon: fixed.

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -666,7 +666,7 @@ private EmbeddedZooKeeper startZooKeeper() throws Exception {
             peer.start();
 
             // Wait until the ZooKeeper joins the cluster.
-            for (;;) {
+            for (; ; ) {
```
minwoox: revert? 😉

ikhoon: Oops... 😅

PR #1305, opened 2026-05-14
Scope replication-failure read-only to repo/project, not the whole server

ikhoon:
Motivation:

Central Dogma is a multi-tenant store: a single replica hosts many projects and repositories belonging to unrelated tenants. When a follower fails to replay a single log (state divergence, lock acquisition timeout, or transient I/O error), the existing failure path puts the whole server into read-only mode. A problem in one repository then propagates to every other tenant on that replica: writes stop everywhere until an operator investigates and lifts the server-wide read-only state.

The goal of this change is to shrink that blast radius. A replay failure flips only the smallest enclosing scope (repository, project, or server) to read-only, and the executor keeps replicating unrelated logs for everyone else. The read-only state is persisted so it survives restarts — in `dogma/dogma` for repo/project scope, and in a properties file for server scope. The existing administrative API can lift it once the operator has investigated.

Modifications:

- Add `UpdateProjectStatusCommand` and `UpdateRepositoryStatusCommand`, routed through the same replication pipeline as ordinary commands so scope changes converge across the cluster. Their effect lands in `dogma/dogma/status/{project}/{repo}.json` — the same storage as ordinary commands, intentionally: a replay failure of the status update itself is then treated as a `dogma/dogma` failure and escalates to server scope.
- Carry `commandType`, `projectName`, and `repoName` on every `LogMeta`.
- Add `ReplicationLogContext` attached to `ReplicationException` so `ZooKeeperCommandExecutor.handleReplicationFailure()` can pick the read-only scope:
  - null project name (root commands or `dogma/dogma`) → server
  - non-null project name with null repo (or the dogma metadata repo) → project
  - otherwise → repository
- Add `RepoStatusManager` to manage the read-only status of projects and repositories; remove `MetadataService.updateRepositoryStatus()` and migrate callers.
- On a replay error, skip past the failing log (so subsequent unrelated logs keep replaying) and issue the scope-appropriate `Update(Project|Repository)StatusCommand`. Fall back to a server-wide read-only state only if recovery itself fails.
- Wire `StandaloneCommandExecutor` to consult `RepoStatusManager` on every push, so a per-repo / per-project read-only state rejects writes with `ReadOnlyException` before reaching the storage layer.
- Rename `DefaultCrud*` → `ReplicatingCrud*` and introduce `StandaloneCrud*` to bypass replication when the standalone executor applies state updates directly.
- Move `ServerStatusManager` to the internal package.
- Add `ReplicationStatus` to describe the status of projects and repositories, and migrate `RepositoryStatus` in `RepositoryDto` to `ReplicationStatus`.

Result:

- A replication failure no longer puts the whole server into read-only mode; only the affected project or repository becomes read-only.


it/xds-k8s-node-ip-extractor/build.gradle
```diff
@@ -3,6 +3,7 @@
 dependencies {
     testImplementation project(':server')
     testImplementation project(':xds')
+    testImplementation libs.armeria.grpc
```
jrhee17: Question) So is gRPC actually used in the tests?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/RepositoryServiceV1.java
```diff
@@ -477,33 +449,35 @@ private void validateMigrationPrerequisites(ServiceRequestContext ctx, Project p
...
-                      } else {
-                          logger.info("Changed repository status: project={}, repository={}, status={}",
-                                      projectName, repoName, status);
-                      }
-                      return null;
-                  });
+                    projectName, repoName, newStatus);
+        final Command<Void> command = Command.updateRepositoryStatus(project.name(), repoName, author,
+                                                                     newStatus);
+        return executor().execute(command).handle((unused, cause) -> {
```
jrhee17, line 470: 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/crud/StandaloneCrudRepository.java
```diff
@@ -0,0 +1,70 @@
...
+
+import com.linecorp.centraldogma.common.Author;
+import com.linecorp.centraldogma.common.Revision;
+import com.linecorp.centraldogma.server.storage.project.ProjectManager;
+import com.linecorp.centraldogma.server.storage.repository.HasRevision;
+
+/**
+ * A {@link CrudRepository} implementation which stores JSON objects in a Git repository.
+ */
+public final class StandaloneCrudRepository<T> implements CrudRepository<T> {
```
jrhee17, line 32: Question) I couldn't find usages of this - was this introduced to be used later on?

server/src/main/java/com/linecorp/centraldogma/server/internal/management/RepoStatusManager.java
```diff
@@ -0,0 +1,196 @@
...
+        key = getKey(projectName, repoName);
+        return statusMap.get(key);
+    }
+
+    private static String getKey(String projectName, String repoName) {
+        return projectName + '/' + repoName;
+    }
+
+    public boolean isWritable(String projectName, String repoName) {
+        if (!statusManager.serverStatus().writable()) {
```
jrhee17, line 163: Note) I understood that `statusManager.serverStatus()` triggers a disk read every time it is invoked, and `isWritable` is called per each command executed

server/src/main/java/com/linecorp/centraldogma/server/internal/management/RepoStatusManager.java
```diff
@@ -0,0 +1,196 @@
...
+            if (repoState != null && repoState.status() == ReplicationStatus.READ_ONLY) {
+                return false;
+            }
+        }
+
+        // Both the project and the repository are writable.
+        return true;
+    }
+
+    public CompletableFuture<Void> updateProjectStatus(String projectName, Author author,
```
jrhee17, line 184: Note) Understood that
- Marking project/repository as readonly is done independently (if project/repo are both read_only, we must individually mark each project/repo as writable to recover)
- When reading, both project/repo must be writable to be writable overall

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/LogMeta.java
```diff
@@ -81,6 +95,24 @@ int size() {
         return size;
     }
 
+    @Nullable
+    @JsonProperty("commandType")
+    String commandType() {
```
jrhee17, line 100: Question) Where is `commandType` used? If possible, it might be worth minimizing the size of `LogMeta` as much as possible

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -1185,6 +1279,11 @@ protected <T> CompletableFuture<T> doExecute(ExecutionContext ctx, Command<T> co
             try {
                 timings.startExecutorExecution();
                 future.complete(blockingExecute(ctx, command, timings));
+            } catch (ScopedReadOnlyAppliedException e) {
+                future.completeExceptionally(e.getCause());
+            } catch (ReplicationException e) {
+                handleReplicationFailure(e, false);
```
jrhee17: Question) Can this branch actually be reached since `blockingExecute` always seems to wrap in a `ScopedReadOnlyAppliedException`?

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -1109,20 +1139,82 @@ private long storeLog(ReplicationLog<?> log) {
...
+                delegate.execute(ExecutionContext.empty(), command).get(30, TimeUnit.SECONDS);
+                final ReplicationLog<?> log = new ReplicationLog<>(replicaId(), command, null);
+                final long revision = storeLog(log);
+                logger.info("Successfully applied read-only mode to {}. revision: {}", scope, revision);
+            } catch (Exception e) {
+                logger.error("Failed to apply read-only mode for {}; stopping command executor...", scope, e);
+                stopLater();
+            }
+        } else {
+            execute(command).handle((unused, cause) -> {
```
jrhee17, line 1117: Note) Understood that for followers, there exists a timing where more commands can be processed before the actual `StatusCommand` goes through.

I'm also fine with applying the same approach for `directExecution` for simplicity since the important point is that further writes stop at some point so we can operate manually

jrhee17 (approved):
Understood the changes.
Also understood that UI/metric related changes will ensue in future PRs

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
jrhee17: Follow up - it seems like `ReplicationException` is thrown from `safeLock` when lock acquisition fails.
Could this potentially mean that lock acquisition failure triggers readonly mode?

it/xds-k8s-node-ip-extractor/build.gradle
ikhoon: It was added to resolve a complication error about `Message` dependency. Let me retest after removing it.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/crud/StandaloneCrudRepository.java
ikhoon: `StandaloneCrudOperation` is currently being used but `StandaloneCrudRepository` was introduced to match the style of `ReplicatingCrud{Operation,Repository}`. If it remained unsed, we can remove it.

server/src/main/java/com/linecorp/centraldogma/server/internal/management/RepoStatusManager.java
ikhoon: Correct. We need to make `dogma`(for project) and the target repository writable.

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/LogMeta.java
ikhoon: It was added for debugging purposes so that it can be included in `toString()`. When a replication log fails to replay, a system administer can use the value as a clue to diagnose the root cause.

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
ikhoon: > Question) Can this branch actually be reached since blockingExecute always seems to wrap in a ScopedReadOnlyAppliedException?

The reason it is split is that, after acquiring a lock, we need to enter readonly mode while holding it to prevent any subsequent commands in the same scope from running.

> Could this potentially mean that lock acquisition failure triggers readonly mode?

If a command fails to acquire a lock simply due to another long-running command, it does not enter readonly mode because `LockAcquireTimeoutException` is thrown instead. However, if Curator raises an exception while acquiring a lock, the lock acquisition failure causes the scope to enter readonly mode.

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
ikhoon: It will work as you described. There is a chance already queued commands may be executed before readonly mode effects. However, as you said, manual recovery is required for readonly mode, I chose this approach rather than introducing a complex rescheduling mechanism.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/RepositoryServiceV1.java
```diff
@@ -307,16 +289,9 @@ public CompletableFuture<RepositoryDto> updateStatus(Project project,
...
-        final RepositoryStatus newStatus = statusRequest.status();
-        if (oldStatus == newStatus) {
-            // No need to update the status, just return the current status.
-            return CompletableFuture.completedFuture(newRepositoryDto(repository, oldStatus));
-        }
-
-        return mds.updateRepositoryStatus(author, project.name(),
-                                          normalizeRepositoryName(repository), newStatus)
-                  .thenApply(unused -> newRepositoryDto(repository, newStatus));
+        final ReplicationStatus newStatus = statusRequest.status();
```
minwoox, line 292: Skipping comparing the old status?

server/src/main/java/com/linecorp/centraldogma/server/internal/management/RepositoryState.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(projectName, repoName, status, updatedAt);
+    }
+
+    @Override
+    public String toString() {
+        return MoreObjects.toStringHelper(this)
```
minwoox, line 96: omit nullable?

server/src/main/java/com/linecorp/centraldogma/server/internal/management/RepoStatusManager.java
minwoox: Is this okay for accessing the disk every time?

server/src/main/java/com/linecorp/centraldogma/server/internal/management/ServerStatusManager.java
```diff
@@ -39,21 +43,21 @@ public final class ServerStatusManager {
             Executors.newSingleThreadExecutor(
                     ThreadFactories.newThreadFactory("server-status-manager", true));
 
-    private final Path serverStatusFile;
+    private final File dataDir;
```
minwoox, line 54: What's the rational behind this change? Can't we just use serverStatusFile?

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -817,9 +826,12 @@ private synchronized void replayLogs(long targetRevision, boolean force) {
             if (!force && !canReplicate) {
                 break;
             }
+            ReplicationLogContext logContext = null;
             ReplicationLog<?> l = null;
             try {
-                l = loadLog(nextRevision);
+                logContext = loadLog(nextRevision);
+                l = logContext.log();
```
minwoox: Now we can do:
```suggestion
                final ReplicationLog<?> l = logContext.log();
```

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -1020,27 +1048,90 @@ private long storeLog(ReplicationLog<?> log) {
...
+            }
+        }
+
+        logger.error("Failed to replicate a log; attemping to switch to read-only mode. " +
+                     "scope: {}, logContext: {}", scope, logContext, exception);
+        if (directExecution) {
+            // When a replication failure occurs during command execution, the replication should enter
+            // read-only mode while holding the lock for the original command.
+            try {
+                delegate.execute(ExecutionContext.empty(), command).get(30, TimeUnit.SECONDS);
```
minwoox:
```suggestion
                delegate.execute(command).get(30, TimeUnit.SECONDS);
```

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -1020,27 +1048,90 @@ private long storeLog(ReplicationLog<?> log) {
...
+            if (repoName == null || repoName.equals(Project.REPO_DOGMA)) {
+                command = Command.updateProjectStatus(projectName, ReplicationStatus.READ_ONLY);
+                scope = projectName;
+            } else {
+                command = Command.updateRepositoryStatus(projectName, repoName, ReplicationStatus.READ_ONLY);
+                scope = projectName + '/' + repoName;
+            }
+        }
+
+        logger.error("Failed to replicate a log; attemping to switch to read-only mode. " +
```
minwoox:
```suggestion
        logger.error("Failed to replicate a log; attempting to switch to read-only mode. " +
```

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -1103,6 +1196,11 @@ protected <T> CompletableFuture<T> doExecute(ExecutionContext ctx, Command<T> co
             try {
                 timings.startExecutorExecution();
                 future.complete(blockingExecute(ctx, command, timings));
```
minwoox, line 1207: So this command is missing if it fails to replay logs even though this command might not related to the read-only repository. Is that correct?

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -842,26 +854,22 @@ private synchronized void replayLogs(long targetRevision, boolean force) {
...
                 if (t instanceof ReplicationException) {
                     throw (ReplicationException) t;
                 }
-                final StringBuilder sb = new StringBuilder();
-                sb.append("failed to replay a log at revision " + nextRevision);
-                if (l != null) {
-                    sb.append(". replay log: ").append(l);
-                }
-                throw new ReplicationException(sb.toString(), t);
+                throw new ReplicationException("failed to replay a log at revision " + nextRevision, t,
```
minwoox: Even though there are unapplied stored logs left, they will be replayed by the next blocking execute call. Is that correct?

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -842,26 +854,22 @@ private synchronized void replayLogs(long targetRevision, boolean force) {
...
                 }
             } catch (Throwable t) {
-                if (l != null) {
-                    logger.error(
-                            "Failed to replay a log at revision {}; entering read-only mode. replay log: {}",
-                            nextRevision, l, t);
-                } else {
-                    logger.error("Failed to replay a log at revision {}; entering read-only mode.",
-                                 nextRevision, t);
+                try {
```
minwoox, line 851: If there's a bug in ZooKeeper or somewhere, all the following logs will probably fail to replay. Shouldn't we set the whole clusters to read-only mode in that case?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/crud/StandaloneCrudOperation.java
```diff
@@ -0,0 +1,95 @@
...
+import com.linecorp.centraldogma.common.Revision;
+import com.linecorp.centraldogma.internal.Jackson;
+import com.linecorp.centraldogma.server.command.CommitResult;
+import com.linecorp.centraldogma.server.storage.project.ProjectManager;
+import com.linecorp.centraldogma.server.storage.repository.HasRevision;
+import com.linecorp.centraldogma.server.storage.repository.Repository;
+
+public final class StandaloneCrudOperation<T> implements CrudOperation<T> {
+
+    // TODO(ikhoon): Implement RocksDB based CrudOperation
```
minwoox, line 38: What do we need this for?

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -1103,6 +1196,11 @@ protected <T> CompletableFuture<T> doExecute(ExecutionContext ctx, Command<T> co
             try {
                 timings.startExecutorExecution();
                 future.complete(blockingExecute(ctx, command, timings));
+            } catch (ScopedReadOnlyAppliedException e) {
+                future.completeExceptionally(e.getCause());
+            } catch (ReplicationException e) {
+                handleReplicationFailure(e, false);
+                future.completeExceptionally(e);
```
minwoox: Can't we complete this future after the handle in `handleReplicationFailure` is invoked?

minwoox (commented):
Basically looks great. 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/management/ServerStatusManager.java
ikhoon: It is not directly related to this PR but I found out the current implementation does not work correctly in the following condition:

If `server-status.properties` is removed or newly created by a system administrator while Central Dogma is running, the cached `serverStatusFile` doesn't detect the changes and throw an unexpected exception.

server/src/main/java/com/linecorp/centraldogma/server/internal/management/RepoStatusManager.java
ikhoon: I didn't expect it to take that long or be a heavy task compared to Git write/read operations or Zookeeper replication jobs.

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
ikhoon: In that case, `loadLog()` will return an incomplete `RelicationLogContext` that is classified as cluster-wide read-only mode.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/crud/StandaloneCrudOperation.java
ikhoon: I don't think we need Git storage just for CRUD operations. RocksDB is faster and we don't need such a long Git history.

We may store metadata such as settings and tokens in the RocksDB. This approach would make Central Dogma more scalable.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/RepositoryServiceV1.java
ikhoon: It is safe since `RedundantChangeException` is already handled in RepoStatusManager.
https://github.com/line/centraldogma/blob/02d9fc8964258c151881b7c58dd9b73e7aba3fc4/server/src/main/java/com/linecorp/centraldogma/server/internal/management/RepoStatusManager.java#L117-L121

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
ikhoon: Good point. I didn't consider that case. I will update it so that the command is executed even if `replayLogs()` fails.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/crud/StandaloneCrudOperation.java
minwoox: I thought we could add something like `RocksDBNoGitRepository` that implements `Repository` for that case but you don't want to use the `Repository` due to API mismatch. Is that right?

ikhoon: If we can implement a lightweight `Repository` backed by RocksDB, we won't need a RocksDB implementation of `CrudOperation`.

server/src/main/java/com/linecorp/centraldogma/server/internal/management/RepoStatusManager.java
ikhoon: I found out a code path that invokes this method in an event loop. Let me fix `statusManager.serverStatus()` to cache `ServerStatus` and refresh it every second.

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
ikhoon: It wasn't. I fixed to continue replaying logs even if a replication fails.

ikhoon:
All review comments have been addressed. PTAL.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/sysadmin/ServerStatusService.java
```diff
@@ -78,6 +79,10 @@ public CompletableFuture<ServerStatus> updateStatus(UpdateServerStatusRequest st
 
             return CompletableFuture.supplyAsync(() -> {
                 executor().statusManager().updateStatus(newStatus);
+                if (executor() instanceof ZooKeeperCommandExecutor) {
```
minwoox, line 82: This looks a bit weird to me. Can't we only create the statusManager in the StandAloneCommandExecutor and delegate it?

ikhoon: For `Scope.ALL`, if `UpdateServerStatusCommand` is executed, it is first replicated to other replicas, followed by stopping `StandaloneCommandExecutor` and `ZookeeperCommandExecutor` sequentially.

For `Scope.LOCAL`, the remaining behavior should be the same except that the command replication is skipped. While adding tests, I found and fixed a bug where `ZookeeperCommandExecutor` was stopped but `StandaloneCommandExecutor` was not. In replication mode, since `ZookeeperCommandExecutor` first checks whether it is writable, this bug probably would not have caused issues in practice.

minwoox: So I guess keeping both status managers is inevitable. Got it.

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -828,41 +839,43 @@ private synchronized void replayLogs(long targetRevision, boolean force) {
...
-                if (l != null) {
-                    sb.append(". replay log: ").append(l);
+                    re = (ReplicationException) t;
+                } else {
+                    re = new ReplicationException("failed to replay a log at revision " + nextRevision,
+                                                  t, logContext);
                 }
-                throw new ReplicationException(sb.toString(), t);
+                // Apply read-only asynchronously and keep replaying the next log.
+                handleReplicationFailure(re, false, null);
```
minwoox, line 872: Should't we just return if no log context?

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -1020,27 +1048,94 @@ private long storeLog(ReplicationLog<?> log) {
...
+        if (projectName == null || projectName.equals(InternalProjectInitializer.INTERNAL_PROJECT_DOGMA)) {
+            command = Command.updateServerStatus(ServerStatus.REPLICATION_ONLY);
+            scope = "SERVER(all)";
+        } else {
+            if (repoName == null || repoName.equals(Project.REPO_DOGMA)) {
+                command = Command.updateProjectStatus(projectName, ReplicationStatus.READ_ONLY);
+                scope = projectName;
+            } else {
+                command = Command.updateRepositoryStatus(projectName, repoName, ReplicationStatus.READ_ONLY);
+                scope = projectName + '/' + repoName;
```
minwoox, line 1098: Can we mark this project or repository somewhere in the memory so that replaylogs skip if the repository is read-only mode before the command is applied?

server/src/main/java/com/linecorp/centraldogma/server/internal/api/sysadmin/ServerStatusService.java
ikhoon: Yeah, because the timing of the stop operation is different.

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
ikhoon: There shouldn't be a big difference since `repoStatusManager` already caches the status. The only difference is whether to raise an exception. But the rest of the code path should be the same since we need to recode `logContext` anyway.

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
ikhoon: Make sense.

minwoox (approved):
👍 👍 👍

PR #1306, opened 2026-05-14
Revert "Cache an entire repository to increase cache hit rate (#1098)"

ikhoon:
This reverts the whole-repository caching logic implemented in #1098 but intentionally retains a few non-cache changes that were bundled in the original PR:

- Removed the overridden `getOrNull()` that caches the entire repository in `CachingRepository`.
- `FindOption.fullName` removal: kept (the field was unused).
- `CentralDogmaBuilder` cache size 256MB: kept.
- `PathPatternFilter` leading-`/` strip: kept as defensive code (no current callers exercise it, but harmless).


jrhee17 (approved):
👍👍

minwoox:
So this logic isn't needed anymore?
https://github.com/line/centraldogma/pull/1098/changes#diff-1f8c93e4e14434464d9124d6f5cfe19a891a4c7812e7b25199a743daa9588406L106

minwoox (approved):
👍 👍

ikhoon:
> So this logic isn't needed anymore? [#1098 (changes)](https://github.com/line/centraldogma/pull/1098/changes#diff-1f8c93e4e14434464d9124d6f5cfe19a891a4c7812e7b25199a743daa9588406L106)

The reasons I removed the method are:
- The original code needs to be updated to support templating and YAML.
- The previous code cached JSON path processing results separately but I questioned whether it was worth storing those results while maintaining two similar code paths.

PR #1309, opened 2026-05-21
Allow adding repository roles without project-level registration

### LINKED ISSUE #930: Provide a way to allow a user or access token to access a certain repository and nothing else in a project.
Currently, to allow a user or access token to access a certain repository, a project owner must add the user or access token to the project metadata. The user or access token being added can have MEMBER or OWNER role.

Let's say we have the following project structure

- my_proj
  - repo_foo (allows MEMBER access)
  - repo_bar (allows MEMBER access)
  - repo_qux (allows MEMBER access)

.. and the following users:

- alice (role: OWNER)
- bob (role: MEMBER)
- charlie (role: MEMBER)

It is currently not possible to disallow `charlie` to access other repositories than `repo_qux` in Central Dogma, because `charlie` has the MEMBER role. Note the `charlie` can't have the GUEST role here because a project member is always a MEMBER or a OWNER.

It is also not possible to remove `charlie` from the project member list, because a user (or access token) must be a member of the project to have the per-repository permissions.

To address this issue, we could:

- Allow granting repository-level permissions to a user or access token even if the user or access token is not a member of the project; or
- Allow assigning the GUEST role when adding a user or access token to the project.

To me it seems like the first option is more intuitive and user-friendly because it sounds odd to have a project 'member' whose role is 'GUEST'.

minwoox:
Motivation:
- Previously, users and app identities had to be registered at the project level before they could be assigned a repository role, which was an unnecessary extra step.

Modifications:
- Remove ensureProjectMember() and ensureProjectAppIdentity() checks from addUserRepositoryRole() and addAppIdentityRepositoryRole() in MetadataService.
- Update NewUserRepositoryRole UI to use free text input instead of a project-member dropdown.
- Update NewAppIdentityRepositoryRole UI to fetch all app identities from the system instead of only project-registered ones.

Result:
- Close #930
- Users and app identities can be assigned repository roles directly without prior project-level registration.


minwoox:
image
This is displayed when the project-level role has higher access than repository-level role

jrhee17 (approved):
👍 👍

ikhoon:
It seems this approach comes with trade-offs. Allowing users to be registered at the repository level would improve flexibility, but it may not be ideal from a governance perspective since the project owner may find it difficult to see the members of each repository.

The downside may be mitigated by allowing repository members to be registered from the Member tab of the project settings page and by listing the members of each repository there.

minwoox:
> The downside may be mitigated by allowing repository members to be registered from the Member tab of the project settings page and by listing the members of each repository there.

Thanks for the concern. Here's my take:

The proposal to allow registering repository members from the project Member tab has some trade-offs:
- Concept mixing — The Member tab manages project-level membership. Adding repository member registration there would blur the distinction between "project member" and "repository member."
- Circles back to the original constraint — If we make project owners register repository members from the project settings, it essentially recreates the same centralized registration flow we're trying to remove. Only the entry point changes.
- The real need is visibility, not a second registration path

A cleaner approach would be to add a read-only aggregated view (e.g., a "Repository Members" section) to the project settings page that lists all members across all repositories. What do you think?

ikhoon:
> A cleaner approach would be to add a read-only aggregated view (e.g., a "Repository Members" section) to the project settings page that lists all members across all repositories. What do you think?

This approach sounds good to me.

minwoox:
> This approach sounds good to me.

If so, I will implement it in the follow-up PR. 😉

ikhoon (approved):
🙇‍♂️🙇‍♂️

PR #1312, opened 2026-06-10
Prevent open redirect in SamlAuthSsoHandler via RelayState normalization bypass

minwoox:
Motivation:
- `SamlAuthSsoHandler.loginSucceeded()` reflects the attacker-controlled `RelayState` into a `window.location.href='...'` redirect after a successful SAML login. The existing guard only checked that the value starts with a single `/` (`startsWith("/")` && `!startsWith("//")`), which validates the raw string but not the URL a browser actually navigates to.

Modifications:
- Extract the RelayState validation into `SamlAuthSsoHandler.isSafeRelayState()` and additionally reject backslashes and ISO control characters, so the validated string matches what the browser will navigate to. Unsafe values fall back to `/`.

Result:
- `RelayState` values that a browser would resolve off-origin are no longer used as a redirect target; the handler falls back to `/`.


jrhee17 (approved):
👍 👍

PR #1313, opened 2026-06-11
Persist local_last_revision in storeLog even while shutting down

minwoox:
Motivation:
- After a cluster reboot, a replica entered read-only mode while replaying the latest ZooKeeper log: the mirror PushAsIsCommand replayed at a revision whose effect was already in the local data, so it failed with a RedundantChangeException ("changes did not change anything ...").
- Root cause: ZooKeeperCommandExecutor.doStop() sets listenerInfo to null at the very beginning, before draining the in-flight blockingExecute via shutdown(executor). storeLog() persisted local_last_revision only when listenerInfo != null. When a command was being originated at shutdown time, its change had already been applied to local data and the ZooKeeper log node was still created, but the local_last_revision update was silently skipped. This left the local data ahead of both progress files (git HEAD at the new revision while last_revision and local_last_revision lagged by one).
- On the next start-up, replay re-executed that already-applied revision. Because the mirror command is an unnormalized PushAsIsCommand whose base revision is Revision.HEAD, re-applying it against the already-advanced head produced an empty diff and threw RedundantChangeException, tripping read-only mode. This was the gap introduced together with local_last_revision in #1303.

Modifications:
- Decouple the local_last_revision disk write from the listenerInfo guard in storeLog(): always call updateLocalLastAppliedRevision(revision), and gate only the in-memory ListenerInfo.localLastAppliedRevision update on listenerInfo.

Result:
- A command being originated while the replica is shutting down now durably records its progress in local_last_revision, so the data and the progress files stay consistent and the replica no longer re-applies an already-applied revision (and no longer enters read-only mode) on the next start-up.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
@@ -1096,13 +1096,14 @@ public final class ZooKeeperCommandExecutor
                                .forPath(absolutePath(LOG_PATH) + '/', logMetaBytes);
                 revision = revisionFromPath(logPath);
 
-                // The command has already been applied to the local data by blockingExecute before this
-                // method is called, so the new revision must be persisted to local_last_revision regardless
-                // of listenerInfo.
-                updateLocalLastAppliedRevision(revision);
                 final ListenerInfo info = listenerInfo;
-                if (info != null && revision > info.localLastAppliedRevision) {
-                    info.localLastAppliedRevision = revision;
+                if (info != null) {
+                    if (revision > info.localLastAppliedRevision) {
+                        updateLocalLastAppliedRevision(revision);
+                        info.localLastAppliedRevision = revision;
+                    }
+                } else if (revision > getLocalLastAppliedRevision()) {
+                    updateLocalLastAppliedRevision(revision);
                 }
             }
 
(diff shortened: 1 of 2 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -1095,9 +1096,12 @@ private long storeLog(ReplicationLog<?> log) {
                                .forPath(absolutePath(LOG_PATH) + '/', logMetaBytes);
                 revision = revisionFromPath(logPath);
 
+                // The command has already been applied to the local data by blockingExecute before this
+                // method is called, so the new revision must be persisted to local_last_revision regardless
+                // of listenerInfo.
+                updateLocalLastAppliedRevision(revision);
```
jrhee17: Note) While the condition itself looks like the file's `revision != localLastAppliedRevision` could be true, I understood in practice they can't diverge unless `info` is `null`.

e.g. this fix would also work
```
final ListenerInfo info = listenerInfo;
if (info != null) {
  if (revision > info.localLastAppliedRevision) {
    updateLocalLastAppliedRevision(revision);
    info.localLastAppliedRevision = revision;
  } else {
    // log the ksip
  }
} else {
    // Shutting down — the command was already applied locally,
    // so we must still persist the revision to avoid replay on restart.
    updateLocalLastAppliedRevision(revision);
}
```

jrhee17 (approved):
👍 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
minwoox: That's a good point. Fixed.

ikhoon (approved):
Oops... 🥲

PR #1314, opened 2026-06-18
Replay pending logs in storeLog so no revision is skipped

ikhoon:
Motivation:
A replica entered read-only mode after a `mismatching replay result`: it had
silently skipped a log appended by another replica, so its local data diverged
from the ZooKeeper log and a later replay no longer matched the recorded result.

The skip was caused by the `local_last_revision` high-water mark added in https://github.com/line/centraldogma/pull/1303.
`storeLog` bumped the mark to the revision it had just created, and `replayLogs`
skipped every revision at or below it as "already applied". So when another
replica appended a lower-numbered log between this replica's replay and its own
`storeLog`, the mark jumped over that log and it was never applied locally.

Modifications:
The two preceding commits revert https://github.com/line/centraldogma/pull/1313 and https://github.com/line/centraldogma/pull/1303. That restores the original
replay, which instead skips a log whose originating replica id matches this
replica (`loadLog(.., skipIfSameReplica=true)`) — correct in normal operation
but wrong after a data directory is restored from another replica, where such
logs are not actually in local data. Rather than rely on either skip, this
commit keeps the replay progress contiguous by replaying in `storeLog`:

- `storeLog`, under one monitor, creates the log node, calls `replayLogs` up to
  the new revision to apply any logs other replicas appended in the meantime,
  then records the new revision as the last replayed one. The command was
  already applied locally before `storeLog`, so it is recorded without being
  re-executed.
- `replayLogs` no longer skips anything and executes every revision above
  `lastReplayedRevision`; `loadLog` drops its `skipIfSameReplica` parameter.
  Self-originated revisions are recorded at store time, so the loop never
  reaches them.
- `lastReplayedRevision` moves from `ListenerInfo` to the executor so `storeLog`
  can record progress even after the listener is cleared on stop.
- A `force` flag keeps the `blockingExecute` and `storeLog` replay running while
  the executor is shutting down, so an in-flight command finishes against
  up-to-date state, while childEvent-driven background replay still stops.
- Restoring from another replica now needs only `last_revision`; rewinding it
  replays every revision the local data is missing.
- Add regression tests: self-originated commands are not re-executed,
  cross-replica commands converge on every replica, a `last_revision` rewind
  replays everything, and an in-flight command at stop records its progress and
  is not replayed on restart.

Result:
- A replica no longer skips a log appended by another replica, so replicas stay
  consistent and do not enter read-only mode.


server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -1088,25 +1018,17 @@ private long storeLog(ReplicationLog<?> log) {
...
-                    if (revision > info.localLastAppliedRevision) {
-                        updateLocalLastAppliedRevision(revision);
-                        info.localLastAppliedRevision = revision;
-                    }
-                } else if (revision > getLocalLastAppliedRevision()) {
-                    updateLocalLastAppliedRevision(revision);
-                }
+                replayLogs(revision - 1, true);
+                updateLastReplayedRevision(revision);
+                lastReplayedRevision = revision;
```
ikhoon, line 1026: This part is the core change in this PR.

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -1088,25 +1018,17 @@ private long storeLog(ReplicationLog<?> log) {
...
-                final ListenerInfo info = listenerInfo;
-                if (info != null) {
-                    if (revision > info.localLastAppliedRevision) {
-                        updateLocalLastAppliedRevision(revision);
-                        info.localLastAppliedRevision = revision;
-                    }
-                } else if (revision > getLocalLastAppliedRevision()) {
-                    updateLocalLastAppliedRevision(revision);
-                }
+                replayLogs(revision - 1, true);
```
jrhee17, line 1024: Understood that this is the main change

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -838,77 +788,57 @@ private static long readRevisionFile(File file) throws Exception {
...
-    private synchronized void replayLogs(long targetRevision) {
-        final ListenerInfo info = listenerInfo;
-        if (info == null) {
+    private void replayLogs(long targetRevision) {
+        replayLogs(targetRevision, false);
+    }
+
+    // force keeps origination-path replay running during shutdown; childEvent replay stops.
+    private synchronized void replayLogs(long targetRevision, boolean force) {
+        if (!force && listenerInfo == null) {
```
ikhoon, line 807: Added a workaround to replay all logs up to the point where the executed command’s revision is persisted when ZooKeeperCommandExecutor is being stopped.

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -838,77 +788,57 @@ private static long readRevisionFile(File file) throws Exception {
...
 
-    private synchronized void replayLogs(long targetRevision) {
-        final ListenerInfo info = listenerInfo;
-        if (info == null) {
+    private void replayLogs(long targetRevision) {
+        replayLogs(targetRevision, false);
+    }
+
+    // force keeps origination-path replay running during shutdown; childEvent replay stops.
+    private synchronized void replayLogs(long targetRevision, boolean force) {
```
jrhee17, line 806: Understood that `force` is to replay logs so that the write directly done to this replica goes through even during shutdown

jrhee17 (approved):
👍 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -838,77 +788,57 @@ private static long readRevisionFile(File file) throws Exception {
...
+                logger.error("Failed to update lastReplayedRevision to: {}", lastReplayedRevision);
             }
-            throw e;
         }
     }
 
-    private synchronized void replayLogs(long targetRevision) {
-        final ListenerInfo info = listenerInfo;
-        if (info == null) {
+    private void replayLogs(long targetRevision) {
```
minwoox: Let's remove this method and just call `replayLogs(targetRevision, false); in `childEvent`
    }`

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
jrhee17, line 1024: I skimmed through this, but shouldn't replayLogs be executed before the execution for this replica takes place?

Just to make sure I am understanding correctly because this is how I interpret it at the moment
```
client -> replica X -> `delegate.execute()` and store as revision 5 -> replay up to revision 4
```

minwoox (approved):
👍 👍 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
ikhoon: `blockingExecute()` already runs `replayLogs()` when it starts. In this part, we only replay the logs added while `delegate.execute()` is running. Please let me know if I misunderstood your question.

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
ikhoon: Done

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
ikhoon:
```
step1  replayLogs(98)            
step2  delegate.execute(command)
(Meanwhile, replica 2 stores revision 99 in ZooKeeper)
step3  storeLog:
         synchronized (this) {
           create → revision 100
           replayLogs(99)  → 99(replica 2's log, not exists in local)
                           → lastReplayedRevision = 99
           lastReplayedRevision = 100   ← Mark lastReplayedRevision as 100 to skip
                                          the self-originated log
         }
```

PR #1316, opened 2026-06-19
Fix mirror loading failure when a mirror has an invalid credential

minwoox:
Motivation:
- The SSH host key verification change made SshGitMirror's constructor throw a MirrorException whenever the credential is not an SshKeyCredential (e.g. a missing credential that resolves to Credential.NONE).

Modifications:
- Wrap the per-mirror MirrorConverter.convertToMirror call in handleAllMirrors with a try/catch that logs a warning and skips the bad mirror (returns null) instead of failing the whole list.

Result:
- A mirror with an invalid or missing credential no longer prevents the rest of the project's mirrors from being loaded and scheduled; only the bad mirror is skipped with a warning.


jrhee17 (approved):
👍 👍

ikhoon (approved):
👍👍

PR #1317, opened 2026-06-23
Add a web UI for the xDS control plane

minwoox:
Motivation:
- The xds module exposes Envoy control plane features (LDS/RDS/CDS/EDS plus group, credential and Kubernetes-aggregator management) over gRPC and HTTP/JSON, but had no UI. Operators had to call the HTTP API by hand to create a group and manage its resources, which is error-prone and hard to inspect.

Modifications:
- Add a standalone Next.js web app under xds/webapp (Chakra UI, Redux Toolkit / RTK Query, Monaco editor), served under '/xds' with its own layout: a group selector, a per-group resource sidebar, and a JSON editor. Wire it into the Gradle build and serve it from ControlPlaneService when bundled.
- Group management (a group is a repository under the internal '@xds' project): list with paging and search, create, and ADMIN-only deletion in a dedicated "Danger Zone" section. Per-group Permissions, Credentials and Kubernetes endpoint aggregator management.
- Resource management for LDS/RDS/CDS/EDS: list, create from templates, read-only view with an explicit Edit toggle, update and delete. A "References" panel links a resource to the children it references (LDS -> RDS/CDS, RDS -> CDS, CDS -> EDS), and missing resources/groups are reported with explicit messages instead of raw exceptions.
- Surface Kubernetes-aggregator-generated endpoints as read-only.
- Authentication/authorization: reuse the main Central Dogma session auth, redirect unauthenticated users to the main login, and add an "xDS" navbar link in the main web app when the xDS web is enabled. Scope the discovery API per app identity (mTLS certificate or access token) and expose an ungated EDS read API so endpoints are readable without the READ role.
- Add a runnable XdsTestServer (embedded fabric8 Kubernetes mock and sample data) for manual testing, plus integration tests for mTLS/token discovery scoping, EDS read permission, write permission, and web app serving. You can run the server via `./gradlew :xds:runXdsTestServer`

Result:
- Operators can manage xDS groups and their LDS/RDS/CDS/EDS resources, credentials, Kubernetes endpoint aggregators and permissions through a dedicated web UI served under '/xds', instead of hand-crafting HTTP calls.


xds/src/main/java/com/linecorp/centraldogma/xds/internal/ControlPlaneService.java
```diff
@@ -217,9 +294,111 @@ protected void onFileRemoved(String groupName, String path) {
...
+            public HttpResponse serve(ServiceRequestContext ctx, HttpRequest req) {
+                // The authorizer sets the authenticated user on the context when the certificate or token is
+                // valid. The boolean result is ignored on purpose.
+                return HttpResponse.of(authorizer.authorize(ctx, req).handle((authorized, cause) -> {
+                    try {
+                        return unwrap().serve(ctx, req);
+                    } catch (Exception e) {
+                        return Exceptions.throwUnsafely(e);
+                    }
+                }));
```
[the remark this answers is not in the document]
minwoox: This is for bypassing the authorization to provide backward compatibility. It will be fixed after migration is done.

xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsResourceManager.java
```diff
@@ -141,12 +146,30 @@ public CommandExecutor commandExecutor() {
...
+        if (metadata != null) {
+            final RepositoryRole role = MetadataService.findRepositoryRole(metadata, group, user);
+            if (role != null && role.has(RepositoryRole.WRITE)) {
+                return;
+            }
+        }
+        throw Status.PERMISSION_DENIED.withDescription(
+                "You must have the WRITE repository role to modify resources in the group: " + group)
+                                      .asRuntimeException();
     }
```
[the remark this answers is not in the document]
minwoox: Any authenticated user can create groups. That's the intention.

xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsResourceManager.java
```diff
@@ -141,12 +146,30 @@ public CommandExecutor commandExecutor() {
...
-        // TODO(minwoox): check the write permission.
         if (!xdsProject.repos().exists(group)) {
             throw Status.NOT_FOUND.withDescription("Group not found: " + group).asRuntimeException();
         }
+        checkWritePermission0(group);
+    }
+
+    private void checkWritePermission0(String group) {
+        final User user = AuthUtil.currentUserOrNull();
+        if (user == null || user.isSystemAdmin()) {
```
jrhee17: Note) Understood that this is for no-auth mode.  It might be clearer if `XdsResourceManager` were instantiated as a field.

xds/src/main/java/com/linecorp/centraldogma/xds/internal/ControlPlaneService.java
```diff
@@ -217,9 +294,111 @@ protected void onFileRemoved(String groupName, String path) {
 
     @Override
     protected void onDiffHandled() {
+        updateAllSnapshots();
```
jrhee17: Note) Understood that all caches will be updated even if a single group within the project is updated

xds/src/main/java/com/linecorp/centraldogma/xds/internal/ControlPlaneService.java
```diff
@@ -217,9 +294,111 @@ protected void onFileRemoved(String groupName, String path) {
...
         metrics.onSnapshotUpdate(snapshot);
+        scopedClients.forEach((key, scopedClient) -> {
+            final Set<String> groups = scopedClient.readableGroups;
+            if (groups != null) {
+                cache.setSnapshot(key, centralDogmaXdsResources.snapshot(groups));
+            }
+        });
+    }
+
+    private String cacheKey() {
```
jrhee17, line 358: Note) Understood that `cacheKey` is used as the group name for each discovery req.
Hence, each group is scoped in terms of what groups it can subscribe to.

jrhee17 (approved):
Understood the java-side main changes as:
- Metadata/User level permission is also managed for the `@xds` repo
- AppIdentity requests are scoped in terms of what resources are in it's group

xds/src/main/java/com/linecorp/centraldogma/xds/internal/ControlPlaneService.java
minwoox: That is true and it's on my todo list:
```
// TODO(minwoox): Implement better cache implementation that updates only changed resources
//                instead of this snapshot based implementation.
```

xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsResourceManager.java
minwoox: > Understood that this is for no-auth mode. 

That's correct. After the migration is done, it will require the auth.

> It might be clearer if XdsResourceManager were instantiated as a field.

Could you elaborate more on this?

xds/src/main/java/com/linecorp/centraldogma/xds/internal/ControlPlaneService.java
minwoox: That's correct. `cacheKey` is used for `NodeGroup` and it can subscribe to the Central Dogma `groups`.

xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsResourceManager.java
jrhee17: I assumed that no-auth mode is a server state, and hence it doesn't need to be calculated per auth validation.
Hence, I was imagining there was a field such as `XdsResourceManager#authEnabled`.

This is purely from a readability perspective, so feel free to ignore if you feel differently.

minwoox: Now I get it.
Let me just throw an exception if the user is null. We provide the example servers with authentication enabled, so it should be no problem.

xds/src/main/java/com/linecorp/centraldogma/xds/internal/ControlPlaneService.java
```diff
@@ -217,9 +294,111 @@ protected void onFileRemoved(String groupName, String path) {
...
+            MetadataService mds) {
+        final Authorizer<HttpRequest> authorizer =
+                new ApplicationCertificateAuthorizer(mds::findCertificateById)
+                        .orElse(new ApplicationTokenAuthorizer(mds::findTokenBySecret));
+        return delegate -> new SimpleDecoratingHttpService(delegate) {
+            @Override
+            public HttpResponse serve(ServiceRequestContext ctx, HttpRequest req) {
+                // The authorizer sets the authenticated user on the context when the certificate or token is
+                // valid. The boolean result is ignored on purpose.
+                return HttpResponse.of(authorizer.authorize(ctx, req).handle((authorized, cause) -> {
```
ikhoon, line 325: It was noted that `authorizer` does not complete exceptionally.

xds/src/main/java/com/linecorp/centraldogma/xds/internal/CentralDogmaSnapshotResources.java
```diff
@@ -51,6 +51,18 @@ public static <T extends Message> SnapshotResources<T> create(
                 versionedResourcesMap.build(), resourcesMap.build(), resourceType);
     }
 
+    /**
+     * Creates {@link SnapshotResources} from an already flattened (resource name to {@link VersionedResource})
```
ikhoon: For me, it took some time to understand what `flattened` means in this context. From my understanding, the snapshot are created from the filtered resources.
```suggestion
     * Creates {@link SnapshotResources} from an already filtered (resource name to {@link VersionedResource})
```

xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsEndpointReadService.java
```diff
@@ -0,0 +1,100 @@
...
+import com.linecorp.armeria.server.annotation.Param;
+import com.linecorp.armeria.server.annotation.ProducesJson;
+import com.linecorp.centraldogma.common.Entry;
+import com.linecorp.centraldogma.common.EntryType;
+import com.linecorp.centraldogma.common.Revision;
+import com.linecorp.centraldogma.server.storage.project.Project;
+import com.linecorp.centraldogma.server.storage.repository.Repository;
+
+/**
+ * Serves read access to EDS (endpoint) resources of the xDS project without requiring a repository role.
```
ikhoon, line 38: I’m a bit concerned that access control for CD may become fragmented, but I think this is the most appropriate implementation for now.
I’ll also think more about a more general implementation for file-level and directory-level access control.

ikhoon (approved):
👍👍

xds/src/main/java/com/linecorp/centraldogma/xds/internal/ControlPlaneService.java
minwoox: It could be, I intentionally didn't check it because the authorization will be enabled anyway after the migration is done.

xds/src/main/java/com/linecorp/centraldogma/xds/internal/CentralDogmaSnapshotResources.java
minwoox: create method uses `Map<String, Map<String, VersionedResource<T>>> resources` but this method uses 
`Map<String, VersionedResource<T>> versionedResources` which is flattened from `Map<String, Map<String, VersionedResource<T>>> resources`. However, it's filtered as well so let me update it as `filtered and flattened`

PR #1319, opened 2026-06-23
Make it possible to permit empty endpoints in CD endpoint groups

m50d:
We have a use case where this is desirable for a testing-only endpoint group (which will be empty when there are no tests currently active, and will then be populated before testing starts - we'd like the server to start up successfully even if no tests are currently active).

### DIFF
--- a/client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/CentralDogmaEndpointGroup.java
+++ b/client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/CentralDogmaEndpointGroup.java
@@ -104,21 +104,6 @@ public final class CentralDogmaEndpointGroup<T> extends DynamicEndpointGroup {
      * Query<T> query = ... // The query to the entry that contains the list of endpoints.
      * Watcher watcher = centralDogma.fileWatcher(projectName, repositoryName, query);
      * }</pre>
-     */
-    public static <T> CentralDogmaEndpointGroupBuilder<T> builder(Watcher<T> watcher,
-                                                                  EndpointListDecoder<T> endpointListDecoder) {
-        return new CentralDogmaEndpointGroupBuilder<>(watcher, endpointListDecoder, false);
-    }
-
-    /**
-     * Returns a new {@link CentralDogmaEndpointGroupBuilder} with the {@link Watcher}
-     * and {@link EndpointListDecoder}. You can create a {@link Watcher} using {@link CentralDogma}:
-     *
-     * <pre>{@code
-     * CentralDogma centralDogma = ...
-     * Query<T> query = ... // The query to the entry that contains the list of endpoints.
-     * Watcher watcher = centralDogma.fileWatcher(projectName, repositoryName, query);
-     * }</pre>
      *
      * @param allowEmptyEndpoints if true, allow the group to start up with no endpoints instead of erroring,
      *                            and allow the list of endpoints to become empty instead of always retaining
--- a/client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/CentralDogmaEndpointGroup.java
+++ b/client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/CentralDogmaEndpointGroup.java
@@ -140,7 +140,7 @@ public final class CentralDogmaEndpointGroup<T> extends DynamicEndpointGroup {
         instanceListWatcher.watch((revision, instances) -> {
             try {
                 final List<Endpoint> newEndpoints = endpointListDecoder.decode(instances);
-                if (newEndpoints.isEmpty()) {
+                if (newEndpoints.isEmpty() && !allowsEmptyEndpoints()) {
                     logger.info("Not refreshing the endpoint list of {} because it's empty. {}",
                                 instanceListWatcher, revision);
                     return;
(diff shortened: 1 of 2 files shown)


client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/CentralDogmaEndpointGroup.java
```diff
@@ -107,13 +107,30 @@ public static <T> CentralDogmaEndpointGroup<T> of(CentralDogma centralDogma,
...
+        return new CentralDogmaEndpointGroupBuilder<>(watcher, endpointListDecoder, allowEmptyEndpoints);
     }
 
     CentralDogmaEndpointGroup(EndpointSelectionStrategy strategy,
                               Watcher<T> instanceListWatcher,
-                              EndpointListDecoder<T> endpointListDecoder) {
-        super(strategy);
+                              EndpointListDecoder<T> endpointListDecoder,
+                              boolean allowEmptyEndpoints) {
+        super(strategy, allowEmptyEndpoints);
```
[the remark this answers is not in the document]
m50d: Hmm fair point

m50d:
Build error looks like an unrelated flake?

client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/CentralDogmaEndpointGroup.java
```diff
@@ -107,13 +107,30 @@ public static <T> CentralDogmaEndpointGroup<T> of(CentralDogma centralDogma,
...
+     *
+     * <pre>{@code
+     * CentralDogma centralDogma = ...
+     * Query<T> query = ... // The query to the entry that contains the list of endpoints.
+     * Watcher watcher = centralDogma.fileWatcher(projectName, repositoryName, query);
+     * }</pre>
+     */
+    public static <T> CentralDogmaEndpointGroupBuilder<T> builder(Watcher<T> watcher,
+                                                                  EndpointListDecoder<T> endpointListDecoder,
+                                                                  boolean allowEmptyEndpoints) {
```
minwoox: Would you mind adding the description of `allowEmptyEndpoints` to Javadoc?

minwoox (approved):
👍 👍

client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/CentralDogmaEndpointGroup.java
```diff
@@ -107,13 +107,34 @@ public static <T> CentralDogmaEndpointGroup<T> of(CentralDogma centralDogma,
...
+     * Watcher watcher = centralDogma.fileWatcher(projectName, repositoryName, query);
+     * }</pre>
+     *
+     * @param allowEmptyEndpoints if true, allow the group to start up with no endpoints instead of erroring,
+     *                            and allow the list of endpoints to become empty instead of always retaining
+     *                            the last non-empty version
+     */
+    public static <T> CentralDogmaEndpointGroupBuilder<T> builder(Watcher<T> watcher,
+                                                                  EndpointListDecoder<T> endpointListDecoder,
+                                                                  boolean allowEmptyEndpoints) {
```
jrhee17: Optional) Given that we're already using a builder pattern and `allowEmptyEndpoints`'s default value is reasonable, I wonder if it's a better idea to add a builder method instead of adding to the builder factory method.

jrhee17 (approved):
👍 👍

client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/CentralDogmaEndpointGroup.java
```diff
@@ -123,7 +144,7 @@ private void registerWatcher() {
         instanceListWatcher.watch((revision, instances) -> {
             try {
                 final List<Endpoint> newEndpoints = endpointListDecoder.decode(instances);
-                if (newEndpoints.isEmpty()) {
+                if (newEndpoints.isEmpty() && !allowsEmptyEndpoints()) {
```
ikhoon, line 127: I understand this is the main change because empty endpoints are allowed in `DynamicEndpointGroup` by default.

ikhoon (approved):
LGTM if @jrhee17 comments are addressed.

client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/CentralDogmaEndpointGroup.java
m50d: Done

client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/CentralDogmaEndpointGroup.java
m50d: Hmm I guess so. Didn't dig in far enough, just saw it would exception when trying to start if there were no endpoints.

client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/CentralDogmaEndpointGroup.java
m50d: The use sites I've seen don't really use it as a builder, they all call the factory method and `build()` immediately, so this seemed more in line with existing use? But happy to follow the builder style if you prefer.

jrhee17: My line of thought is just from an API management point - each param could possibly add 2^n overload variants.

client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/CentralDogmaEndpointGroup.java
m50d: Checked the consequences of setting to false rather than true, seems to just be early exit from update methods that would otherwise set it to empty.

client/java-armeria/src/main/java/com/linecorp/centraldogma/client/armeria/CentralDogmaEndpointGroup.java
m50d: Makes sense. Doing it that way.

PR #1320, opened 2026-06-25
Improve xDS web UI: navigation, pagination, history diff, and client …

minwoox:
…subscriptions

Motivation:
- Selecting a group landed on the Listeners page instead of the Overview page, making it hard to get a quick summary of the group at a glance.
- The Resources and History pages had no pagination, making large groups hard to browse.
- The group-wide History page showed commit summaries as plain text with no way to inspect what actually changed in a commit.
- The References page gave no indication that inbound references from other groups are not shown, which could be confusing.
- The control plane Clients table had no visibility into what resource names each client subscribed to.

Modifications:
- Change the default navigation target when selecting a group (group list, sidebar dropdown, breadcrumb, and the `useXdsRoute` fallback) from `listeners` to `overview`.
- Add client-side pagination (default 10 rows) to the Resources and History pages.
- Add a Subscriptions column to the Clients table.
- Track subscribed resource names in `XdsClientStatusTracker`.

Result:
- Selecting a group now lands on the Overview page.
- Resources and History pages are paginated.
- Every commit in the group-wide History is inspectable with a before/after diff.
- The References page clearly communicates its cross-group limitation.


jrhee17 (approved):
Understood java-related changes are
- exposing APIs for administrative purposes
- applying auth for native gRPC paths

webapp/src/pages/app/xds/group.tsx
```diff
@@ -98,15 +101,21 @@ const GroupDetailPage = () => {
...
+          ? 'K8s Aggregators'
+          : section === 'credentials'
+            ? 'Credentials'
+            : section === 'dangerZone'
+              ? 'Danger Zone'
+              : section === 'history'
+                ? 'History'
+                : section === 'references'
+                  ? 'References'
+                  : `${XDS_RESOURCE_META[section].label}s`;
```
ikhoon: Should we use a static map instead of these nested ternary operators?

xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsClientStatusTracker.java
```diff
@@ -0,0 +1,224 @@
...
+    @Override
+    public void onStreamOpen(long streamId, String typeUrl) throws RequestException {
+        streams.put(streamId, new StreamState(System.currentTimeMillis()));
+        // The controlplane invokes onStreamClose/onStreamCloseWithError only for a graceful close or an error
+        // surfaced through the request stream; an abrupt client disconnect is delivered as a cancellation
+        // (onCancelHandler), which does NOT trigger those callbacks and would leak the entry. Tie removal to
+        // the end of the underlying gRPC call instead, which covers every termination path. onStreamOpen is
+        // invoked synchronously from the service method with the Armeria context mounted.
+        final ServiceRequestContext ctx = ServiceRequestContext.currentOrNull();
+        if (ctx != null) {
```
ikhoon: If ctx is absent in the thread local, should we treat it as a bug?
```
assert ctx != null;
```

xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsClientStatusTracker.java
```diff
@@ -0,0 +1,224 @@
...
+     * Returns the current state of every connected client as a JSON array. Each element describes one stream
+     * and the ACK/NACK state of every resource type subscribed on it. The view is eventually consistent: it
+     * may observe concurrent updates from the gRPC worker threads.
+     */
+    ArrayNode toClientsJson() {
+        final ArrayNode array = JsonNodeFactory.instance.arrayNode();
+        streams.forEach((streamId, stream) -> {
+            final ObjectNode streamNode = array.addObject();
+            streamNode.put("streamId", streamId);
+            streamNode.put("nodeId", stream.nodeId);
```
ikhoon: I noticed that `toClientsJson` could be called while `stream` is being updated, so it may not serialize a constant snapshot of `stream`.

xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsControlPlaneStatusService.java
```diff
@@ -0,0 +1,80 @@
...
+    @Get("/xds/clients")
+    public ArrayNode clients() {
+        return clientStatusTracker.toClientsJson();
+    }
+
+    /**
+     * Lists the application identities that have connected to the discovery API, with the groups each can read.
+     */
+    @Get("/xds/apps")
+    public ArrayNode apps() {
```
ikhoon: Optional) I prefer defining DTO object to track changes in the serialzed output more easily.

ikhoon (approved):
👍 👍

webapp/src/pages/app/xds/group.tsx
minwoox: Yeah, it looks ugly. 😓

xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsClientStatusTracker.java
minwoox: Nice catch. Updated.

xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsControlPlaneStatusService.java
minwoox: Fixed. 😉

PR #1321, opened 2026-06-30
Store xDS resources as YAML instead of JSON

minwoox:
Motivation:
- xDS resources were stored as JSON files. YAML is more human-readable and is the preferred format for configuration in the Kubernetes/Envoy ecosystem.

Modifications:
- `XdsResourceManager`: `fileName()` now returns `.yaml`; `push()` uses `Change.ofYamlUpsert()`; `updateOrDelete()` falls back to the legacy `.json` path when the `.yaml` file is absent; `update()` atomically removes the old `.json` and creates the new `.yaml` in a single commit.
- `XdsResourceWatchingService`: changed `handleXdsResource()` signature from `String contentAsText` to `JsonNode content`; both call sites now pass `entry.content()` / `change.content()` directly instead of `contentAsText()`. Also accepts `EntryType.YAML` and handles `UPSERT_YAML`.

Result:
- Newly created xDS resources are stored as `.yaml` files.
- Existing `.json` files remain readable; they are atomically migrated to `.yaml` on the first update.


xds/src/main/java/com/linecorp/centraldogma/xds/k8s/v1/XdsKubernetesEndpointFetchingService.java
```diff
@@ -179,31 +183,49 @@ protected void onGroupRemoved(String groupName) {
...
+                logger.warn("Failed to find endpoint file for {} in {}", endpointBase, groupName, findCause);
+                return null;
+            }
+            final String actualPath = entries.keySet().stream()
+                    .filter(p -> p.endsWith(".yaml") || p.endsWith(".json"))
+                    .findFirst().orElse(null);
+            if (actualPath == null) {
+                return null; // Already removed or never created.
+            }
+            removeEndpointFile(groupName, actualPath);
```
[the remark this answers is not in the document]
minwoox: Only one of them can exist

xds/src/main/java/com/linecorp/centraldogma/xds/endpoint/v1/XdsEndpointUpdateScheduler.java
```diff
@@ -195,47 +198,92 @@ private void flush() {
...
-                                      if (cause2 != null) {
-                                          final Throwable peeled = Exceptions.peel(cause2);
-                                          if (!(peeled instanceof RedundantChangeException)) {
-                                              copied.forEach(pendingUpdate -> pendingUpdate
-                                                      .streamObserver.onError(peeled));
-                                              return null;
-                                          }
-                                          // If the change is redundant, we just ignore it and complete
-                                          // the stream observer without error.
+                if (resolvedFileName.endsWith(".json")) {
```
jrhee17, line 206: Note) Understood the assumption is that only one of yaml or json had been stored

xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsResourceWatchingService.java
```diff
@@ -232,6 +233,15 @@ private void handleDiff(String groupName, Revision newRevision,
                         }
                         break;
                     case REMOVE:
+                        // Skip if this is an atomic .json → .yaml migration commit: the .json is being
+                        // removed while the same resource is upserted as .yaml in the same diff.
+                        if (path.endsWith(".json")) {
```
jrhee17, line 238: 👍 👍

jrhee17 (approved):
Question: It seems like migration is done lazily (on each push for a file)
Will there be any attempt to migrate at once? Or would it be helpful if we can at least visualize the number of json files vs yaml files that are used in the xds repo?

minwoox:
> Will there be any attempt to migrate at once?

Yeah, I'm preparing. 😉 

> Or would it be helpful if we can at least visualize the number of json files vs yaml files that are used in the xds repo?

I have a command for checking it. 😉

ikhoon (approved):
👍 👍

PR #1324, opened 2026-07-01
Add YAML/JSON backward compatibility for xDS resource files

minwoox:
Motivation:
- xDS resources are currently stored as .json files, but are planned to be migrated to .yaml format.
- To support safe rollout and rollback, the system must handle both .json and .yaml files coexisting in the same repository without errors.

Modifications:
- XdsResourceWatchingService: accept EntryType.YAML alongside JSON during the initial repository scan; handle UPSERT_YAML change type in the diff watcher; normalize YAML content to JSON before invoking handleXdsResource() so all implementations continue to use JSON_MESSAGE_MARSHALLER without format-specific logic.
- XdsResourceManager: add alternativeFileName() helper that swaps between .json and .yaml extensions; change updateOrDelete() to search for both the requested filename and its alternative so that update and delete operations locate the file regardless of which format it was stored in; pass the resolved filename back to the task via Function<String, Runnable> so the correct file is mutated; make push() emit Change.ofYamlUpsert when the resolved filename ends with .yaml.

Result:
- .yaml and .json xDS resource files can coexist in the same group repository.

### DIFF
--- a/xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsEndpointReadService.java
+++ b/xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsEndpointReadService.java
@@ -56,12 +56,12 @@ public final class XdsEndpointReadService {
                          .thenApply(entries -> {
             final ArrayNode array = JsonNodeFactory.instance.arrayNode();
             for (Entry<?> entry : entries.values()) {
-                if (entry.type() != EntryType.JSON) {
+                if (entry.type() != EntryType.JSON && entry.type() != EntryType.YAML) {
                     continue;
                 }
                 final ObjectNode node = array.addObject();
                 node.put("path", entry.path());
-                node.put("type", "JSON");
+                node.put("type", entry.type().name());
                 node.put("revision", entry.revision().major());
             }
             return array;
@@ -73,7 +73,7 @@ public final class XdsEndpointReadService {
      */
     @Get("/xds/groups/{group}/endpoints/{*id}")
     public CompletableFuture<JsonNode> getEndpoint(@Param String group, @Param String id) {
-        return read(group, ENDPOINTS_DIRECTORY + id + ".json");
+        return readByBase(group, ENDPOINTS_DIRECTORY + id);
     }
 
     /**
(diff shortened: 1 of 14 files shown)


xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsEndpointReadService.java
```diff
@@ -82,17 +82,27 @@ public CompletableFuture<JsonNode> getEndpoint(@Param String group, @Param Strin
...
+                         .thenCompose(entries -> {
+                             if (entries.isEmpty()) {
+                                 // Delegate to get() so the caller receives a proper EntryNotFoundException.
+                                 return repository.get(Revision.HEAD, pathBase + ".json")
+                                                  .thenApply(XdsEndpointReadService::toNode);
+                             }
+                             return CompletableFuture.completedFuture(
+                                     toNode(entries.values().iterator().next()));
+                         });
     }
```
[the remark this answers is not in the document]
minwoox: They can't coexist.

xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsResourceManager.java
```diff
@@ -279,8 +315,19 @@ public void updateOrDelete(StreamObserver<?> responseObserver, String group, Str
                                                          .asRuntimeException());
                 return null;
             }
-            task.run();
+            final String resolvedFileName = entries.keySet().iterator().next();
+            taskProvider.apply(resolvedFileName).run();
```
ikhoon: Question) If `Runnable` is invoked immediately, should we simply take `Function<String, Void>` or `Consumer<String>`?

xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsResourceManager.java
```diff
@@ -179,12 +180,44 @@ private void checkWritePermission0(String group) {
...
+    private <T extends Message> void doPush(
+            StreamObserver<T> responseObserver, String group, String resourceName, String fileName,
+            String summary, T resource, Author author, boolean create) {
         final Change<JsonNode> change;
         try {
             final String jsonText = JSON_MESSAGE_MARSHALLER.writeValueAsString(resource);
             final JsonNode jsonNode = Jackson.readTree(jsonText);
             if (create) {
                 change = Change.ofJsonPatch(fileName, null, jsonNode);
+            } else if (fileName.endsWith(".yaml")) {
```
ikhoon, line 219: I understand that only `json` files will be pushed in this version.

xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsResourceManager.java
```diff
@@ -179,12 +180,44 @@ private void checkWritePermission0(String group) {
...
+                                    .withDescription("Resource already exists: " + resourceName)
+                                    .asRuntimeException());
+                    return null;
+                }
+                doPush(responseObserver, group, resourceName, fileName, summary, resource, author, true);
+                return null;
+            });
+            return;
+        }
+        doPush(responseObserver, group, resourceName, fileName, summary, resource, author, false);
```
ikhoon, line 207: Question) I'm wondering if there is a case where `a.json` is updated while `a.json` is being migrated to `a.yaml`? It seems I may not have understood the current mechanism correctly.

jrhee17 (approved):
👍 👍 

Understood the changes as no migration logic, but just supporting read for both json/yaml formats

xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsResourceManager.java
minwoox: `a.json` will be removed and `a.yaml` will be created at once:
https://github.com/line/centraldogma/pull/1321/changes#diff-e5c2302f8f9692ff7ce5ea7df6b42d7a4241ac562a37195614ca7deba434154bR225-R228

xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsResourceManager.java
minwoox: That's correct because this is for compatibility.

xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsResourceManager.java
minwoox: Thanks! Fixed. :wink:

xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsResourceManager.java
ikhoon: I understand `a.json` can't be modified due to the removal after `a.yaml` is created.

ikhoon (approved):
👍 👍

PR #1327, opened 2026-07-08
Add a Repository Status page and surface read-only status across the web UI

ikhoon:
Motivation:

Central Dogma can put an individual repository or a whole project into read-only mode (#1305), but the web UI neither showed that status nor acted on it. Every write button stayed enabled, so a user discovered the read-only state only by having the server reject the write, and there was no way to see or change the status from the UI.

Modifications:

- Add `status` to `ProjectDto` so that `GET /api/v1/projects` reports whether a project is read-only.
- Show a `Read-only` / `Writable` tag on the project list, the project detail page, the repository list and the repository tree page.
- Disable the write actions of a read-only project or repository, each with a tooltip that explains why: New Repository, New File, Edit, Delete (both a file and a repository), Revert, Project Settings and Repository Settings. A read-only project stays browsable.
- Add a "Repository Status" settings tab (system administrator only) that lists the read-only projects and repositories and lets an administrator make one read-only or writable, guarded by re-typing the full `project/repository` name. Selecting the internal `dogma` repository changes the whole project, and the confirmation now says so.
- Add `GET /api/v1/status/repos/read-only`, serialize `RepositoryState.updatedAt` as an ISO-8601 string, and expose `repository.read.only.count` and `repository.read.only` metrics.
- Invalidate the project and repository caches on a status change, so the lists do not keep showing the previous status.

Result:

- A user can see whether a project or a repository is read-only, and the write actions that would fail are disabled with an explanation of why.
- A system administrator can view every read-only project and repository and toggle the status from the web UI, with a type-to-confirm safeguard and an explicit warning when the change is project-wide.

### DIFF
--- a/common/src/main/java/com/linecorp/centraldogma/internal/api/v1/ProjectDto.java
+++ b/common/src/main/java/com/linecorp/centraldogma/internal/api/v1/ProjectDto.java
@@ -34,6 +34,7 @@ import com.google.common.base.MoreObjects.ToStringHelper;
 
 import com.linecorp.centraldogma.common.Author;
 import com.linecorp.centraldogma.common.ProjectRole;
+import com.linecorp.centraldogma.common.ReplicationStatus;
 
 @JsonInclude(Include.NON_NULL)
 public class ProjectDto {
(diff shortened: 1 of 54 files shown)


ikhoon:
05-confirm-project-writable
04-project-detail
03-projects-list
02-repo-status-list
01-confirm-project-read-only

jrhee17 (approved):
👍 👍

minwoox (approved):
👍 👍 👍

PR #1328, opened 2026-07-08
Fix UDP socket leak in CentralDogma-to-CentralDogma mirroring

minwoox:
Motivation:

CentralDogma-to-CentralDogma mirrors were creating a new
ArmeriaCentralDogmaClient on every scheduler tick, allocating a fresh
Armeria WebClient, connection pool, and UDP socket each time. These
sockets were not reclaimed promptly, leading to too many open
files.

Modifications:

- Add CentralDogma.withAccessToken(String) as a default interface method
  that returns a derived client sharing the same underlying WebClient.
  Calling close() on the derived client is a no-op; the base client owns
  the connection resources.
- Add Mirror.setBaseClientPool(ConcurrentHashMap<String, Object>) as a
  default no-op method so MirrorSchedulingService can inject the pool
  without coupling the server module to client types.

Result:

- Mirrors pointing at the same remote host share one Armeria WebClient,
  eliminating per-run DNS/TLS/connection overhead and stopping UDP socket
  accumulation.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirrorSchedulingService.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirrorSchedulingService.java
@@ -24,9 +24,12 @@ import java.io.File;
 import java.time.Duration;
 import java.time.Instant;
 import java.time.ZonedDateTime;
+import java.util.HashSet;
 import java.util.List;
 import java.util.ServiceLoader;
+import java.util.Set;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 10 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirrorSchedulingService.java
```diff
@@ -265,11 +273,35 @@ private void scheduleMirrors() {
...
                           }
-                          for (Mirror m : mirrors) {
-                              if (m.schedule() == null) {
+                          for (Mirror freshMirror : mirrors) {
+                              if (freshMirror.schedule() == null) {
                                   continue;
                               }
 
+                              final String key = mirrorCacheKey(freshMirror);
+                              currentMirrorKeys.add(key);
```
ikhoon: What do you think of adding a new API that derives a Central Dogma with a new access token, and caching CentralDogma clients per endpoint URL?

```java
Map<String, CentralDogma> clients = ...;

String key = dogmaUrl;
CentralDogma defaultClient = clients.get(key);
CentralDogma derivedClient = defaultClient.withAccessToken(credential.accessToken());
// Perform CD-to-CD mirroring with the derivedClient
```

Even if we cache clients for active mirrors, I am concerned that the number of `DnsEndpointGroup`s  will increase as CD-to-CD mirroring increases. It could lead to leaks.

server/src/main/java/com/linecorp/centraldogma/server/mirror/Mirror.java
```diff
@@ -131,4 +131,10 @@ default String remoteBranch() {
      */
     MirrorResult mirror(File workDir, CommandExecutor executor, int maxNumFiles, long maxNumBytes,
                         Instant triggeredTime);
+
+    /**
+     * Closes the resources held by this {@link Mirror}, such as cached connections to the remote.
+     * The default implementation is a no-op.
+     */
+    default void close() {}
```
jrhee17: Question) Will mirrors created by `MirrorServiceV1` also be closed/cleaned up?

jrhee17 (approved):
👍 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/MirrorSchedulingService.java
minwoox: Thanks, addressed!
Please note that there are some limitations though:
- `MirrorSchedulingService`, which manages the resources, do not have the client dependency so we need to use `Object`.
- This won't work for mLTS auth so I might refactor later when I implement it:
  https://github.com/line/centraldogma/blob/main/server-mirror-dogma/src/main/java/com/linecorp/centraldogma/server/internal/mirror/CentralDogmaMirror.java#L112
- Might use the cache for mirror again When I need to reuse the SSH client.

server/src/main/java/com/linecorp/centraldogma/server/mirror/Mirror.java
minwoox: It should be closed. 😓 Thanks!
By the way, close method is gone now.

ikhoon (approved):
👍 👍

PR #1329, opened 2026-07-09
Fix xDS reference links broken when resource id contains slashes

minwoox:
Motivation:

`resolveReference()` built the `id` for navigation links by taking only the last path segment of a resource name such as
`groups/dx/clusters/centraldogma-oss-mdc/alpha`, producing `alpha` instead of the correct `centraldogma-oss-mdc/alpha`.

Modifications:

- In `resolveReference()`, replace `segments[segments.length - 1]` with `segments.slice(idStart).join('/')`, where `idStart` skips the resource-type prefix (1 for regular paths, 2 for k8s paths).

Result:

- Reference links for resource names containing `/` (e.g. `groups/dx/clusters/centraldogma-oss-mdc/alpha`) now navigate to the correct resource.


ikhoon (approved):
👍 👍

ikhoon:
`xdsReferences.ts` is shown as a binary file. This looks like a GitHub bug but could you double check whether the file header or content is not broken?
image

minwoox:
> xdsReferences.ts is shown as a binary file. This looks like a GitHub bug but could you double check whether the file header or content is not broken?

Yeah, I fixed it in this PR. The previous file has a null byte that Git considers as a binary file.

PR #1330, opened 2026-07-10
Add MirrorFileValidator SPI and xDS mirroring UI

minwoox:
Motivation:

When a Git repository is mirrored into an @xds repository, there was no mechanism to reject invalid xDS resource files before they were committed. A typo or structurally-wrong protobuf file would silently land in the repository and break the control plane at push time rather than at the source. Additionally, the xDS UI had no way to manage the mirrors and credentials that back an xDS group.

Modifications:

- Add `MirrorFileValidator`, a `@FunctionalInterface` SPI loaded via `ServiceLoader`. Implementations are invoked in `AbstractMirror` inside a new `validateChanges()` method called by both `AbstractGitMirror` and `CentralDogmaMirror` before each push. All validation errors are collected and surfaced as a single `MirrorException`; mirror state files are excluded from validation.
- Add `XdsMirrorFileValidator`, the xDS-specific implementation. It rejects changes to the `@xds` project that target paths reserved for the Kubernetes controller (`/k8s/endpoints/`), paths outside the recognised xDS resource directories, or content that cannot be parsed as the expected protobuf message type (Cluster, Listener, RouteConfiguration, ClusterLoadAssignment, KubernetesEndpointAggregator).
- Enforce in `DefaultMetaRepository` that xDS mirrors must use `localPath: /`.
- Add `XdsMirroringTab` and xDS-scoped pages for creating, viewing, and editing mirrors (`/app/xds/mirrors/*`) and credentials (`/app/xds/credentials/*`), reusing the existing `MirrorForm`, `MirrorList`, `MirrorView`, `CredentialForm`, and `CredentialList` components with xDS-specific URL routing.

Result:

- Mirroring an invalid or structurally-wrong file into an `@xds` repository now fails fast.
- Users can create, view, and edit xDS mirrors and credentials directly from the xDS group UI.


server/src/main/java/com/linecorp/centraldogma/server/internal/mirror/AbstractMirror.java
```diff
@@ -252,6 +269,39 @@ protected final IgnoreNode ignoreNode() {
...
+        for (Change<?> change : changes.values()) {
+            // Skip mirror state files — they are internal bookkeeping, not user content.
+            if (change.path().endsWith(MIRROR_STATE_FILE_NAME)) {
+                continue;
+            }
+            for (MirrorFileValidator validator : FILE_VALIDATORS) {
+                try {
+                    validator.validate(projectName, repoName, change);
+                } catch (MirrorException e) {
+                    errors.add(e.getMessage());
```
ikhoon, line 294: Should we add the `MirrorException` as a suppressed exception so that we could see the original exception in the `centraldogma.log`?

xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsMirrorFileValidator.java
```diff
@@ -0,0 +1,114 @@
...
+            } else {
+                JSON_MESSAGE_MARSHALLER.mergeValue(change.contentAsText(), builder);
+            }
+        } catch (IOException e) {
+            throw new MirrorException(
+                    filePath + ": not a valid " + builder.getDescriptorForType().getName(), e);
+        }
+    }
+
+    private static Message.@Nullable Builder builderForPath(String filePath) {
```
ikhoon, line 100: 👍

ikhoon (approved):
👍 👍

jrhee17 (approved):
👍 👍

PR #1331, opened 2026-07-10
Fix file descriptor leaks in ServerStatusManager

minwoox:
Motivation:

`Files.newInputStream()` and `Files.newOutputStream()` allocate real OS file descriptors. `Properties.load()` and `Properties.store()` do not close the stream passed to them, so the caller is responsible for closing it. The descriptors are eventually reclaimed by GC, but GC is triggered by memory pressure rather than FD exhaustion.

Modifications:

- `ServerStatusManager.serverStatus()`: wrap `Files.newInputStream()` in try-with-resources so the stream is closed after `Properties.load()`.
- `ServerStatusManager.updateStatus()`: same fix for the load path; also wrap `Files.newOutputStream()` in try-with-resources so the stream is closed after `Properties.store()`.

Result:

- File descriptors opened for property file I/O are now released deterministically instead of relying on GC finalisation.


ikhoon (approved):
👍 👍

jrhee17 (approved):
👍 👍

PR #1332, opened 2026-07-13
Allow users to specify a commit summary when mutating xDS resources

minwoox:
Motivation:

There was no way to specify a commit for xds resource update.

Modifications:

- Add an optional `summary` field to all xDS Create/Update/Delete request proto messages (cluster, listener, route, endpoint, group, k8s). The field is a query parameter in the HTTP-JSON transcoding because the request body is already mapped to the resource message.

Result:

- You can now enter a free-form commit summary when creating, updating, or deleting an xDS resource.


jrhee17 (approved):
👍 👍

ikhoon (approved):
👍 👍

PR #1333, opened 2026-07-13
Recover a diverged repository from a source replica

ikhoon:
Motivation:

A replication failure can put one repository or project into read-only mode (#1305, #1327), but the replication log cannot repair
a repository that has already diverged across replicas. It cannot move a Git ref backward, and the required old log entries may
already be gone, so recovery currently requires a manual replica re-sync.

Modifications:

- Add `POST /api/v1/projects/{projectName}/repos/{repoName}/recover` for system administrators on a ZooKeeper-replicated server.
  - The request selects a source replica, an explicit revision range, and the highest revision observed across the replicas.
  - The selected commits and compatibility padding are limited to 100 revisions.
- Originate recovery from the selected source through the normal replication flow.
  - `RequestRepositoryRecoveryCommand` is applied as a replicated no-op. The source builds and publishes the apply command only
    after recording the request revision.
  - The response is asynchronous. If the source is unavailable when it replays the request, the request must be submitted again.
- Apply one self-contained `ApplyRepositoryRecoveryCommand` identically on every replica, including the source.
  - Derive the reset and final revisions from the ordered `ReplayCommit` list, reset to the revision before the first commit, replay
    the selected source history, and add empty commits through `maxRevision + 1`.
  - Leave a replica untouched when every replayed revision already has the expected tree. Refuse recovery when the replica lacks
    the shared base history or its non-converged head has reached the final recovery revision.
  - A commit added after payload creation can be discarded only while the head remains below the final padding revision.
- Preserve each commit's author, timestamp, summary, detail, markup, self-contained changes, and expected tree ID.
  - Verify every replayed revision by tree ID because commit IDs can differ for identical content when metadata repositories create
    their early commits locally on each replica.
  - Stop on a mismatch and leave the partial history readable. The administrator can run recovery again.
- Keep readers and watchers consistent across a history rewrite.
  - Hold the repository write lock across reset and replay, truncate the commit-ID database, and bump the cache generation.
  - Persist `lastRecoveryRevision` and notify pending watches only after the complete recovery succeeds.
  - Let existing watch clients advance past recovered same-content revisions without a recovery-specific wire protocol.
- Enforce the recovery prerequisites once, before origination.
  - Require a replicated server, a source ID in its replica roster, an unencrypted repository, and an existing read-only scope.
  - Keep apply-time behavior dependent only on the command payload so every replica makes the same replay decision.
- Add `GET /api/v1/projects/{projectName}/repos/{repoName}/head` with the serving replica's revision, commit ID, and tree ID.
- Add `GET /api/v1/replicas` with the configured replica roster.
- Remove purged or soft-removed projects and repositories from read-only status results and metrics, and delete status files on purge.
- Let CodeRabbit review pull requests that target an integration branch.

Result:

- A system administrator can initiate repair of a diverged repository from a healthy replica and compare each replica's tree ID
  before making the repository writable again.
- The recovery response reports only the expected final `recoveryRevision`. Completion and convergence are verified separately with
  `GET .../head`.

The recovery web UI follows in a separate pull request.


server/src/main/java/com/linecorp/centraldogma/server/internal/api/RepositoryServiceV1.java
```diff
@@ -294,6 +303,129 @@ public CompletableFuture<RepositoryDto> updateStatus(Project project,
...
+            throw new StorageException("failed to resolve the head commit of " +
+                                       repository.parent().name() + '/' + repository.name(), e);
+        }
+        if (commitId == null) {
+            throw new StorageException("no head commit in " +
+                                       repository.parent().name() + '/' + repository.name());
+        }
+        return new RepositoryHead(headRevision, commitId.name());
+    }
+
```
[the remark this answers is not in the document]
ikhoon: This is a system-administrator-only endpoint, and it is meant to be called during a recovery, while the repository is already read-only, so no commit can move the head between the two reads.

`GitRepository.head()` also reads the revision and the commit ID together under the read lock, so the pair always describes the same commit.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepositoryManager.java
```diff
@@ -269,6 +272,227 @@ public void fallbackToFileRepository(String repositoryName) {
...
+                    if (!expectedCommitId.equals(actualCommitId)) {
+                        throw new StorageException(
+                                "commit id mismatch while recovering '" +
+                                projectRepositoryName(repositoryName) + "' at " + revision + " (expected: " +
+                                expectedCommitId + ", actual: " + actualCommitId + "). Revisions up to " +
+                                resetToRevision + " may have diverged, or the content is not reproducible " +
+                                "byte-identically (e.g. written by a content transformer); the repository " +
+                                "was rolled back.");
+                    }
+                }
```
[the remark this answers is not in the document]
ikhoon: Wontfix (no longer applicable): recovery no longer validates commit IDs in `GitRepositoryManager`. The implementation now lives in `RepositoryRecovery` and verifies the tree ID after each successful replay, so the flagged `.name()` call no longer exists.

webapp/javaTest/java/com/linecorp/centraldogma/webapp/ReplicatedShiroCentralDogmaTestServer.java
```diff
@@ -0,0 +1,116 @@
...
+        // A two-node quorum needs both peers; start them concurrently.
+        final var start1 = server1.start();
+        final var start2 = server2.start();
+        start1.join();
+        start2.join();
+        scaffold();
+        Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+            server1.close();
+            server2.close();
+        }));
```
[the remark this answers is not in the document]
ikhoon: Wontfix (removed from this PR): the recovery web UI, including `ReplicatedShiroCentralDogmaTestServer`, was moved to a separate pull request. This startup path is no longer part of this diff.

webapp/src/dogma/features/api/apiSlice.ts
```diff
@@ -483,6 +484,23 @@ export const apiSlice = createApi({
...
+      RecoverRepositoryResponse,
+      { projectName: string; repoName: string; fromRevision: number; sourceServerId: number }
+    >({
+      query: ({ projectName, repoName, fromRevision, sourceServerId }) => ({
+        url: `/api/v1/projects/${projectName}/repos/${repoName}/recover`,
+        method: 'POST',
+        body: { fromRevision, sourceServerId },
+      }),
+      // Recovery rewrites the repository history on the other replicas.
+      invalidatesTags: ['Repo'],
```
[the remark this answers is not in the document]
ikhoon: Wontfix (removed from this PR): the recovery web UI and its API mutation were moved to a separate pull request. `apiSlice.ts` no longer contains `recoverRepository` in this diff.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/CommitWatchers.java
```diff
@@ -90,7 +98,7 @@ void notify(Revision revision, String path) {
                  mapIt.hasNext();) {
 
                 final Entry<PathPatternFilter, Set<Watch>> entry = mapIt.next();
-                if (!entry.getKey().matches(path)) {
+                if (!force && !entry.getKey().matches(path)) {
```
jrhee17, line 101: nit) seems like `path` is `Nullable` whereas matches expects non-null

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
```diff
@@ -1167,10 +1296,54 @@ void notifyWatchers(Revision newRevision, List<DiffEntry> diffEntries) {
...
+
+    private static Revision lastRecoveryRevisionOf(org.eclipse.jgit.lib.Repository repository) {
+        return new Revision(repository.getConfig().getInt(
+                RECOVERY_CONFIG_SECTION, null, LAST_RECOVERY_REVISION_CONFIG_KEY,
+                Revision.INIT.major()));
+    }
+
+    void setLastRecoveryRevision(Revision revision) {
+        requireNonNull(revision, "revision");
+        if (revision.major() <= lastRecoveryRevision.major()) {
```
jrhee17, line 1325: Note) Understood that if certain replicas fail for some reason, we must retry with a higher `maxRevision`

server/src/main/java/com/linecorp/centraldogma/server/internal/api/RecoverRepositoryRequest.java
```diff
@@ -0,0 +1,111 @@
...
+        return toRevision;
+    }
+
+    @JsonProperty("sourceServerId")
+    public int sourceServerId() {
+        return sourceServerId;
+    }
+
+    @JsonProperty("maxRevision")
+    public int maxRevision() {
```
jrhee17, line 80: Optional) I'm not sure how this will be represented in the UI, but `maxRevision` itself seems a little vague in terms of naming. What do you think about directly accepting the new `recoveryRevision`?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
```diff
@@ -1094,6 +1205,24 @@ public CompletableFuture<Revision> watch(Revision lastKnownRevision, String path
         return future;
     }
 
+    private boolean tryCompleteWatchAfterRecovery(Revision lastKnownRevision, String pathPattern,
+                                                  boolean errorOnEntryNotFound,
+                                                  CompletableFuture<Revision> future) {
+        final Revision currentHeadRevision = headRevision;
+        final Revision recoveryRevision = lastRecoveryRevision;
+        if (recoveryRevision.compareTo(currentHeadRevision) <= 0 &&
```
jrhee17, line 1213: Question) Is there ever a case where `head revision > recovery revision`?

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
```diff
@@ -1074,6 +1181,10 @@ public CompletableFuture<Revision> watch(Revision lastKnownRevision, String path
             failFastIfTimedOut(this, logger, ctx, "watch", lastKnownRevision, pathPattern);
             readLock();
             try {
+                if (tryCompleteWatchAfterRecovery(normLastKnownRevision, pathPattern,
```
jrhee17, line 1184: Note) I understood this is done to move client watch revisions in case clients need to re-fetch rewritten files

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/RepositoryRecovery.java
```diff
@@ -0,0 +1,331 @@
...
+            if (currentHead.compareTo(recoveryRevision) >= 0) {
+                throw new StorageException(
+                        "recovery revision " + recoveryRevision + " must be greater than the current head " +
+                        currentHead + " of " + repoPath);
+            }
+            checkResetBase(repoPath, resetToRevision, currentHead);
+            final ObjectId resetToCommitId = commitIdDatabase.get(resetToRevision);
+
+            try {
+                rewindTo(repo, resetToCommitId, resetToRevision);
```
jrhee17, line 94: Noted that if a `fromRevision` which already diverges is received as an input, the recovery process may immediately fail in a reset state. I'm wondering if it's worth also passing the `recoveryRevision` expected treeId for safety

jrhee17 (approved):
Understood the changes 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
ikhoon: `setLastRecoveryRevision` is set only after the replay job for recovery commits has been completed. If a replica fails to recover and we retry the recovery with the same revision, the successfully recovered replicas will skip the command, while the failed replica will retry it.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/RecoverRepositoryRequest.java
ikhoon: `maxRevision` means `max(replica 1 head, replica 2 head, ...)`. However, from the perspective of `RecoverRepositoryRequest`, `maxRevision` sounds vague because there is no context for what `max` refers to. I agree that `recoveryRevision` is clearer in `RecoverRepositoryRequest`.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/CommitWatchers.java
ikhoon: `path` should not be null if `force == false`. I will add an assertion.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
ikhoon: Recover may rewrite the content of the same revision so the head revision is advanced to notify clients to re-read the content.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/GitRepository.java
ikhoon: It does not happen during recovery, but after recovery is complete, when the readonly repository transitions to writable and new commits are added.

server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/git/RepositoryRecovery.java
ikhoon: The final recovery revision's tree ID is already carried by the last padding `ReplayCommit` and verified during replay. However, that check happens after the rewind, so it does not protect against a diverged reset base. It is a good idea to  carry the expected tree ID for fromRevision - 1 and validate it before rewinding.

PR #1334, opened 2026-07-14
Allow dots and forbid slashes in new xDS resource IDs

minwoox:
Motivation:

We have decided to allow dots in a xDS group name and resource ID. At the same time, slashes are disallowed in a resource ID.

Existing resources whose IDs contain slashes (created before this change) must continue to be updatable and deletable, so backward compatibility must be preserved for update/delete operations.

Modifications:

- Allow dots and forbid slashes in an xDS group name and resource ID.

Result:

- Dots are now allowed in xDS group names and resource IDs.
- Slashes are rejected when creating new resources.


ikhoon (approved):
👍 👍

jrhee17 (approved):

PR #1336, opened 2026-07-16
Fix an exception thrown when an SSH mirror has an invalid credential type

minwoox:
…type

Motivation:
When an SSH mirror was configured with a non-SSH_KEY credential (e.g. ACCESS_TOKEN), `SshGitMirror` threw a `MirrorException` during mirror conversion. Although the exception was caught by `handleAllMirrors()` and the bad mirror was skipped, the mirror was invisible in the UI as a result, making it impossible to update or delete the mirror.

Modifications:

- Move the credential type validation from `SshGitMirror` constructor to `GitMirrorProvider.newMirror()`. When the credential type is wrong, log a warning and return `null` instead of throwing `MirrorException`.

Result:

- A mirror with an invalid credential type is loaded correctly and shown in the UI, allowing users to update or delete it.

### DIFF
--- a/server-mirror-git/src/main/java/com/linecorp/centraldogma/server/internal/mirror/GitMirrorProvider.java
+++ b/server-mirror-git/src/main/java/com/linecorp/centraldogma/server/internal/mirror/GitMirrorProvider.java
@@ -25,12 +25,18 @@ import static java.util.Objects.requireNonNull;
 
 import java.net.URI;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.linecorp.centraldogma.server.internal.credential.SshKeyCredential;
 import com.linecorp.centraldogma.server.mirror.Mirror;
 import com.linecorp.centraldogma.server.mirror.MirrorContext;
 import com.linecorp.centraldogma.server.mirror.MirrorProvider;
 import com.linecorp.centraldogma.server.mirror.RepositoryUri;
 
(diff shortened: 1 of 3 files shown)


jrhee17 (approved):
👍 👍

server-mirror-git/src/main/java/com/linecorp/centraldogma/server/internal/mirror/GitMirrorProvider.java
```diff
@@ -44,6 +50,10 @@ public Mirror newMirror(MirrorContext context) {
 
         switch (scheme) {
             case SCHEME_GIT_SSH: {
+                if (!(context.credential() instanceof SshKeyCredential)) {
```
ikhoon, line 53: Question) Could we validate the configuration correctness also when a mirror is created.?

ikhoon (approved):
👍 👍

server-mirror-git/src/main/java/com/linecorp/centraldogma/server/internal/mirror/GitMirrorProvider.java
minwoox: Yeah, it's already validated. 😉 
https://github.com/line/centraldogma/blob/main/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/repository/DefaultMetaRepository.java#L398

PR #1337, opened 2026-07-20
Migrate xDS resource services from gRPC to YAML-over-HTTP annotated services

minwoox:
Motivation:

The xDS resource management APIs (cluster, endpoint, listener, route, group, kubernetes) were implemented as gRPC services backed by service-specific `.proto` files. Migrating them to Armeria annotated HTTP services that consume `application/yaml` removes the protobuf service layer, simplifies the API surface, and makes the YAML-native storage format a first-class concern — clients send YAML directly and it is stored as-is (with `name`/`clusterName` injected).

Modifications:

- Convert `XdsClusterService`, `XdsEndpointService`, `XdsListenerService`, `XdsRouteService` from `*ImplBase` gRPC stubs to plain annotated HTTP services (`@Post`/`@Put`/`@Delete`, `@Consumes("application/yaml")`).
- Add `RequiresXdsGroupRole` annotation and `RequiresXdsGroupRoleDecorator` to replace the inline `checkWritePermission` calls with a declarative, per-method authorization mechanism.
- Refactor `XdsResourceManager` so that `push`/`update`/`delete` operate on raw YAML strings instead of protobuf messages, and add helpers `parseYaml`, `normalizeYamlKeys`, `injectYamlField`, and `errorResponse`.

Result:

- xDS resource CRUD is now served via plain HTTP with YAML bodies, removing the gRPC service layer and the associated proto wrapper definitions.


xds/src/main/java/com/linecorp/centraldogma/xds/internal/ControlPlaneService.java
```diff
@@ -152,43 +149,23 @@ void start(PluginInitContext pluginInitContext) {
         final ServerBuilder serverBuilder = pluginInitContext.serverBuilder();
         serverBuilder.service(grpcService, optionalAppIdentityAuth(mds));
         final XdsResourceManager xdsResourceManager = new XdsResourceManager(xdsProject(), commandExecutor);
+        serverBuilder.dependencyInjector(
+                DependencyInjector.ofSingletons(
+                        new RequiresXdsGroupRoleDecoratorFactory(xdsProject())), false);
+        serverBuilder.annotatedService().pathPrefix("/api/v1").build(
```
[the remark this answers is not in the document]
minwoox: It doesn't replace. It chains the injector instead.

minwoox:
Will make changes for the UI in the follow-up PR. :wink:

xds/src/test/java/com/linecorp/centraldogma/xds/internal/XdsTestUtil.java
```diff
@@ -273,8 +274,8 @@ public static AggregatedHttpResponse createEndpoint(
                                        "/api/v1/xds/" + groupName + "/endpoints?endpoint_id=" +
                                        endpointId)
                               .set(HttpHeaderNames.AUTHORIZATION, "Bearer anonymous")
-                              .contentType(MediaType.JSON_UTF_8).build();
-        return webClient.execute(headers, JSON_MESSAGE_MARSHALLER.writeValueAsString(endpoint))
+                              .contentType(MediaType.parse("application/yaml")).build();
```
ikhoon, line 277: Should we create a constant value of `MediaType.parse("application/yaml")`?

xds/src/main/java/com/linecorp/centraldogma/xds/cluster/v1/XdsClusterService.java
```diff
@@ -51,69 +57,91 @@ public XdsClusterService(XdsResourceManager xdsResourceManager) {
...
+     *
+     * <p>Creates a new cluster.
+     */
+    @Post("/xds/groups/{group}/clusters")
+    @Consumes("application/yaml")
+    @RequiresXdsGroupRole(RepositoryRole.WRITE)
+    public CompletableFuture<HttpResponse> createCluster(
+            @Param("group") String group,
+            @Param("cluster_id") String clusterId,
+            @Param("summary") @Nullable String summary,
```
ikhoon, line 71: A summary seems like an arbitrary value set by users. Is it safe to set in the query parameters? I'm concerned about potential issues related to encoding or string length.

xds/src/main/java/com/linecorp/centraldogma/xds/cluster/v1/XdsClusterService.java
```diff
@@ -51,69 +57,91 @@ public XdsClusterService(XdsResourceManager xdsResourceManager) {
...
+            String body) {
+        final String clusterName = clusterName(group, clusterId);
+        if (!CLUSTER_NAME_PATTERN.matcher(clusterName).matches()) {
+            return CompletableFuture.completedFuture(
+                    XdsResourceManager.errorResponse(HttpStatus.BAD_REQUEST,
+                                                     "Invalid cluster name: " + clusterName));
+        }
+        try {
+            XdsResourceManager.parseYaml(body, Cluster.newBuilder());
```
ikhoon, line 112: Question) Would it make sense to fork `ProtobufRequestConverterFunction` so that YAML request-body parsing can be handled as a cross-cutting concern?

xds/src/main/java/com/linecorp/centraldogma/xds/k8s/v1/XdsKubernetesService.java
```diff
@@ -111,55 +117,203 @@ public XdsKubernetesService(XdsResourceManager xdsResourceManager) {
...
+        return xdsResourceManager.delete(group, aggregatorName, deleteSummary, currentAuthor());
+    }
+
+    /**
+     * POST /xds/groups/{group}/k8s/endpointAggregators:preview
+     *
+     * <p>Previews the endpoints that would be resolved for a Kubernetes endpoint aggregator
+     * without persisting it.
+     */
+    @Blocking
```
ikhoon: Question) Is `@Blocking` still necessary? The code seems to run asynchronously.

ikhoon (approved):

xds/src/test/java/com/linecorp/centraldogma/xds/internal/XdsTestUtil.java
minwoox: It's added for production code:
https://github.com/line/centraldogma/pull/1337/changes#diff-e5c2302f8f9692ff7ce5ea7df6b42d7a4241ac562a37195614ca7deba434154bR75
Also, I'm planning to add this to Armeria.

xds/src/main/java/com/linecorp/centraldogma/xds/cluster/v1/XdsClusterService.java
minwoox: > Is it safe to set in the query parameters? I'm concerned about potential issues related to encoding or string length.

Yeah, it's going to be encoded. Also the length limit is more than 2000 chars so I think it should be fine.
https://medium.com/suyeonme/effective-strategies-for-handling-long-query-strings-b790e1fddd65

Let's revisit this if this becomes an issue. 😉

xds/src/main/java/com/linecorp/centraldogma/xds/cluster/v1/XdsClusterService.java
minwoox: I think there's not so much gain for forking the class. If you think, this will be useful, let's consider adding yaml support to the `ProtobufRequestConverterFunction` on upstream.

xds/src/main/java/com/linecorp/centraldogma/xds/k8s/v1/XdsKubernetesService.java
minwoox: We don't need it anymore. 😉

jrhee17 (commented):
👍 👍

jrhee17 (approved):
👍 👍

PR #1339, opened 2026-07-22
Use port 443 by default when TLS is enabled and no port is specified

ikhoon:
Motivation:

`AbstractCentralDogmaBuilder.host(String)` always assumed the default port 36462 even when `useTls()` was enabled, so a client configured with TLS and no explicit port silently sent HTTPS requests to the cleartext port instead of 443.

Modifications:

- Deferred default port resolution from `host(String)` to `hosts()` so that the result no longer depends on the order of `host()` and `useTls()` calls. A host added without a port is stored with a sentinel port and resolved to 443 if TLS is enabled, or 36462 otherwise.
- Documented the new default port in the Javadoc and the Java client tutorial.

Result:

- A `CentralDogma` client built with `useTls()` and `host(host)` without a port number now connects to port 443 instead of 36462.
- Behavioral change: users who serve TLS on port 36462 (or any non-443 port) should specify the port explicitly, e.g. `.host("example.com", 36462)`.

### DIFF
--- a/client/java-armeria/src/test/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogmaBuilderTest.java
+++ b/client/java-armeria/src/test/java/com/linecorp/centraldogma/client/armeria/ArmeriaCentralDogmaBuilderTest.java
@@ -101,6 +101,15 @@ class ArmeriaCentralDogmaBuilderTest {
     }
 
     @Test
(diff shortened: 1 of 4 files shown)


jrhee17 (approved):
👍 👍

minwoox (approved):
👍

PR #1340, opened 2026-07-24
Allow regenerating the secret of an application token

ikhoon:
Motivation:

When the secret of an application token is leaked, the only remedy has been to
deactivate or delete the token and create a new one. This loses the roles and
permissions granted to the application ID and forces every project to register
the new token again. There should be a way to revoke the leaked secret and
issue a new one in place.

Modifications:

- Add `POST /api/v1/appIdentities/{appId}/secret` which issues a newly-generated
  secret to a deactivated token. Only the token creator or a system
  administrator is allowed to call it, the same as deletion.
- A token must be deactivated first and the new secret does not authenticate
  until the token is activated, so the rotation procedure is: deactivate,
  regenerate, distribute the new secret and activate.
- Fail with a conflict when the token was recreated or regenerated concurrently
  after the caller was authorized, so that nobody distributes a secret that
  will never work.
- Add a 'Regenerate secret' action with a confirmation dialog to the
  application identities settings page. The new secret is displayed once; the
  button is enabled only for deactivated tokens.
- Stop including the whole file content, which may contain secrets, in the
  exception message raised when a content transformer fails.

Result:

- Users can rotate a leaked token secret in place through the staged
  deactivate → regenerate → distribute → activate procedure; the application ID
  keeps its roles and permissions.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/api/sysadmin/AppIdentityRegistryService.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/api/sysadmin/AppIdentityRegistryService.java
@@ -251,9 +251,10 @@ public final class AppIdentityRegistryService extends AbstractService {
     /**
      * POST /appIdentities/{appId}/secret
      *
-     * <p>Regenerates the secret of the token of the specified {@code appId} and returns the token with
-     * a newly-generated secret. The old secret is revoked in the same commit, but it may take a short
-     * time for the revocation to be propagated to the authorization cache.
+     * <p>Regenerates the secret of the deactivated token of the specified {@code appId} and returns the
+     * token with a newly-generated secret. The token must be deactivated first and the new secret does
+     * not authenticate until the token is activated, so that the new secret can be distributed to the
+     * clients before it takes effect.
      */
     @Post("/appIdentities/{appId}/secret")
     public CompletableFuture<Token> regenerateTokenSecret(ServiceRequestContext ctx,
@@ -265,9 +266,13 @@ public final class AppIdentityRegistryService extends AbstractService {
                         throw new IllegalArgumentException(
                                 "You can't regenerate the secret of the token scheduled for deletion.");
                     }
-                    // Pass the creation metadata of the authorized token so that a token recreated
-                    // with the same application ID in the meantime is not rotated.
-                    return mds.regenerateTokenSecret(author, appId, token.creation());
+                    if (token.isActive()) {
+                        throw new IllegalArgumentException(
+                                "You can't regenerate the secret of an active token. Deactivate it first.");
+                    }
+                    // Pass the authorized token so that the regeneration fails if the token is
+                    // recreated or regenerated concurrently in the meantime.
+                    return mds.regenerateTokenSecret(author, appId, token);
                 });
     }
 
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/api/sysadmin/AppIdentityRegistryService.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/api/sysadmin/AppIdentityRegistryService.java
@@ -249,6 +249,28 @@ public final class AppIdentityRegistryService extends AbstractService {
     }
 
     /**
+     * POST /appIdentities/{appId}/secret
+     *
+     * <p>Regenerates the secret of the token of the specified {@code appId}. The old secret is revoked
+     * immediately and the token with a newly-generated secret is returned.
+     */
+    @Post("/appIdentities/{appId}/secret")
+    public CompletableFuture<Token> regenerateTokenSecret(ServiceRequestContext ctx,
+                                                          @Param String appId,
+                                                          Author author, User loginUser) {
+        return getTokenOrRespondForbidden(ctx, appId, loginUser).thenCompose(
+                token -> {
+                    if (token.isDeleted()) {
+                        throw new IllegalArgumentException(
+                                "You can't regenerate the secret of the token scheduled for deletion.");
+                    }
+                    // Pass the creation metadata of the authorized token so that a token recreated
+                    // with the same application ID in the meantime is not rotated.
+                    return mds.regenerateTokenSecret(author, appId, token.creation());
+                });
+    }
+
+    /**
      * PATCH /appIdentities/{appId}/level
      *
      * <p>Updates a level of the app identity of the specified ID.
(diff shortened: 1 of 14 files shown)


ikhoon:
image
image
image

server/src/main/java/com/linecorp/centraldogma/server/metadata/AppIdentityService.java
```diff
@@ -168,6 +168,60 @@ Revision purgeAppIdentity(Author author, String appId) {
...
+                    final Token newToken = new Token(token.appId(), newSecret, token.isSystemAdmin(),
+                                                     token.allowGuestAccess(), token.creation(),
+                                                     token.deactivation(), null);
+                    final Map<String, AppIdentity> newAppIds =
+                            updateMap(registry.appIds(), appId, newToken);
+                    // A deactivated token has no entry in the secret map, so the new secret is not
+                    // added; it is registered when the token is activated. The old secret is removed
+                    // defensively in case a stale entry is left over.
+                    final Map<String, String> newSecrets =
+                            removeFromMap(registry.secrets(), oldSecret);
```
jrhee17: Question) I didn't understand under what situation this is possible. From my understanding the reverse mapping should be removed from `deactivateToken`. Otherwise, deactivated tokens can still be used for auth.

server/src/main/java/com/linecorp/centraldogma/server/internal/api/sysadmin/AppIdentityRegistryService.java
```diff
@@ -248,6 +248,33 @@ public CompletableFuture<AppIdentity> updateAppIdentity(ServiceRequestContext ct
         });
     }
 
+    /**
+     * POST /appIdentities/{appId}/secret
+     *
+     * <p>Regenerates the secret of the deactivated token of the specified {@code appId} and returns the
+     * token with a newly-generated secret. The token must be deactivated first and the new secret does
+     * not authenticate until the token is activated, so that the new secret can be distributed to the
+     * clients before it takes effect.
```
jrhee17, line 257: Note) I understand that a user deactivates the token to call this API first. i.e. it is possible that an unlucky case happens where the token is purged before regenerate can be called.

jrhee17 (approved):
Note) I understood that this API is not intended for zero-downtime rotation.

e.g. Normally users will want to 1) generate a new token 2) migrate tokens in their applications 3) delete the old leaked token

The proposed API assumes at the time of deactivation users have not migrated to a different token

server/src/main/java/com/linecorp/centraldogma/server/internal/api/sysadmin/AppIdentityRegistryService.java
ikhoon: A token has three states - Active -> Inactive(deactivated) -> Deleted
The purge scheduler only purges deleted tokens so deactivated tokens are not purged.
https://github.com/line/centraldogma/blob/384cce05027f37e0717a456a5200cb1eff5c767b/server/src/main/java/com/linecorp/centraldogma/server/internal/storage/PurgeSchedulingService.java#L151-L155

ikhoon:
> Note) I understood that this API is not intended for zero-downtime rotation.

Right. I will handle it in https://github.com/line/centraldogma/issues/1341

ikhoon:
This regeneration flow may be used in an urgent situation where a token has been leaked.

server/src/main/java/com/linecorp/centraldogma/server/metadata/AppIdentityService.java
ikhoon: Good point. It is no longer valid, so it should be removed. Initially, I tried to regenerate an active token but ended up limiting the feature only to deactivated tokens to avoid subtle race conditions.

server/src/main/java/com/linecorp/centraldogma/server/metadata/AppIdentityService.java
```diff
@@ -168,6 +168,53 @@ Revision purgeAppIdentity(Author author, String appId) {
...
+                    if (registry.secrets().containsKey(newSecret)) {
+                        throw new ChangeConflictException("Secret already exists");
+                    }
+
+                    final Token newToken = new Token(token.appId(), newSecret, token.isSystemAdmin(),
+                                                     token.allowGuestAccess(), token.creation(),
+                                                     token.deactivation(), null);
+                    final Map<String, AppIdentity> newAppIds =
+                            updateMap(registry.appIds(), appId, newToken);
+                    return new AppIdentityRegistry(newAppIds, registry.secrets(), registry.certificateIds());
```
minwoox: We need to create new `registry.secrets()`.

server/src/main/java/com/linecorp/centraldogma/server/metadata/AppIdentityService.java
```diff
@@ -168,6 +168,56 @@ Revision purgeAppIdentity(Author author, String appId) {
...
+                    }
+
+                    final Token newToken = new Token(token.appId(), newSecret, token.isSystemAdmin(),
+                                                     token.allowGuestAccess(), token.creation(),
+                                                     token.deactivation(), null);
+                    final Map<String, AppIdentity> newAppIds =
+                            updateMap(registry.appIds(), appId, newToken);
+                    // A deactivated token has no entry in the secret map; create a new map so that
+                    // the new registry does not share the mutable map with the old one.
+                    return new AppIdentityRegistry(newAppIds, ImmutableMap.copyOf(registry.secrets()),
```
minwoox, line 208: Shouldn't we remove the old token string and add new one to the `registry.secrets()`?
ref: https://github.com/line/centraldogma/pull/1340/changes#diff-e5adef6a1ce67aa6ebb68ee9fb85a90d8cd23507acdf33e8a1dac98a23aee844R110-R114

ikhoon: We only allow deactivated tokens to be regenerated. The newly generated token will be added to registry.secrets() when it is activated.
https://github.com/line/centraldogma/blob/a5a7f8300f1c6e114931e623e9ee75487d5fc9d0/server/src/main/java/com/linecorp/centraldogma/server/metadata/AppIdentityService.java#L183-L184

minwoox: Oops, my bad. I missed it. :sweat:

minwoox (approved):
👍 👍 👍

PR #1345, opened 2026-07-30
Keep every stored field when editing a K8s endpoint aggregator

ikhoon:
Motivation:

The K8s endpoint aggregator form modelled only part of the schema and rebuilt the document from that
model on save, so distinctEndpoint, the metadata mappings and the policy were erased. Losing
distinctEndpoint brings duplicate endpoints back on the next rolling restart.

Two more problems sat next to it: overlapping edits both succeeded, the later one rolling the earlier
back unnoticed, and a stored policy never reached the endpoints generated from it.

Modifications:

- The form holds the stored document itself, so a field it does not render can no longer be rebuilt
  away. Distinct endpoint, metadata mappings and the policy are now editable; drop overloads are
  shown read-only, because Envoy applies at most one and rejects the endpoints when it sees more.

- The update endpoint takes the revision the client read the aggregator at and commits on top of it:
  a conflict is 409, an unknown revision 400. Revisions are compared per repository, so any commit in
  the group makes an open editor stale. Omitting the parameter keeps the previous behaviour.

- The policy reaches the generated endpoints and the preview, and is validated on write.

Result:

Editing an aggregator through the console keeps every field it was stored with, a save based on a
stale read is refused, and a policy set on an aggregator takes effect.

An aggregator that already stores a policy Envoy would reject was harmless while it was ignored, and
will now be served. Worth checking before this ships.

### DIFF
--- a/webapp/next.config.js
+++ b/webapp/next.config.js
@@ -11,7 +11,9 @@ const nextConfig = {
   productionBrowserSourceMaps: isDev,
   trailingSlash: false,
   output: isDev ? 'standalone' : 'export',
-  distDir: 'build/web/',
+  // `npm run build` and `next dev` share this directory, so a gradle build (which runs the former as part
+  // of :webapp:runTestServer) wipes a running dev server's manifests. Point the dev server elsewhere.
+  distDir: process.env.NEXT_DIST_DIR || 'build/web/',
   images: {
     unoptimized: true,
   },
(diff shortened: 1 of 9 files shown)


minwoox:
You are going to add `metadataMapping` and `policy` as well. Right?

ikhoon:
> You are going to add `metadataMapping` and `policy` as well. Right?

Right. In addition, I refactored two things:
- I changed `K8sAggregatorEditor` to render k8s endpoint aggregator fields directly instead of re-mapping them. This makes adding new fields easier.
- Since Kubernetes Aggregator has many configuration fields, I grouped related fields together to make the input fields easier to understand.

ikhoon:
image

jrhee17 (approved):
👍 👍

ikhoon:
Empty form:
image

minwoox (approved):
Looks great! Thanks!

PR #1347, opened 2026-08-03
Wait for an in-flight replay before closing the log watcher

ikhoon:
## Motivation

When a replica shuts down, `PathChildrenCache.close()` cancels its in-flight tasks with an interrupt.
If a replay was waiting on `delegate.execute(...)` at that moment, the command had already been applied
to the local data but `lastReplayedRevision` was never advanced, because it is updated only on the
success path. The replica is then left with local data ahead of `<dataDir>/last_revision`, and the next
start-up replays the same revision again. Re-applying a command whose effect is already in the local
data fails, so the replica enters read-only mode and needs a manual re-sync.

```
[INFO ] [command-executor-shutdown] Closing the log watcher
[ERROR] [zookeeper-log-watcher-1-1] Failed to replay a log at revision N; entering read-only mode.
java.lang.InterruptedException
    at java.base/java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:386)
    at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2073)
    at ...ZooKeeperCommandExecutor.replayLogs(ZooKeeperCommandExecutor.java:820)
    at ...ZooKeeperCommandExecutor.childEvent(ZooKeeperCommandExecutor.java:885)
    at ...PathChildrenCache.lambda$callListeners$1(PathChildrenCache.java:529)
```

## Modifications

- Wait for an in-flight replay before closing the log watcher, by acquiring the same monitor that
  `replayLogs()` holds. Releasing it is safe because `listenerInfo` is already null, so a replay that
  starts afterwards returns before executing anything.
- Move the log watcher shutdown ahead of `delegate.stop()`. The barrier waits for a replay that is
  itself waiting on the delegate, so the delegate must still be running. This also matches the drain
  that `shutdown(executor)` already performs for the command executor threads.
  `logWatcher.close()` still runs before `shutdown(logWatcherExecutor)`: reversing the two would let
  `PathChildrenCache` keep submitting to an already shut-down executor, because `submitToExecutor()`
  only guards on its own state, which flips in `close()`.
- Wait uninterruptibly for the replay result, as a safeguard on the line where the failure occurred.
- Log the last replayed revision once the replay is drained.

Note this covers a graceful shutdown only. A `SIGKILL` between the local apply and the progress update
leaves the same divergence; making the replay idempotent is a separate topic.

## Result

- A replica that shuts down while replaying no longer leaves its local data ahead of its recorded
  replication progress, so it no longer enters read-only mode on the next start-up.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
@@ -686,6 +686,22 @@ public final class ZooKeeperCommandExecutor
         logger.info("Stopped the worker threads");
 
         try {
+            if (logWatcher != null) {
+                logger.info("Closing the log watcher");
+                // Wait for an in-flight replay: close() cancels its tasks with an interrupt.
+                // Releasing is safe because listenerInfo is null, so a new replay returns immediately.
+                synchronized (this) {
+                    logger.info("Drained the log watcher; last replayed revision: {}", lastReplayedRevision);
+                }
+                logWatcher.close();
+                interrupted |= shutdown(logWatcherExecutor);
+                logger.info("Closed the log watcher");
+            }
+        } catch (Exception e) {
+            logger.warn("Failed to close the log watcher: {}", e.getMessage(), e);
+        }
+
+        try {
             logger.info("Stopping the delegate command executor");
             delegate.stop();
             logger.info("Stopped the delegate command executor");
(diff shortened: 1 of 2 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -680,11 +680,29 @@ private void stopLater() {
...
                           @Nullable Runnable onReleaseZoneLeadership) throws Exception {
         canReplicate = false;
+        // Stop accepting new replay logs.
         listenerInfo = null;
+
         logger.info("Stopping the worker threads");
         boolean interrupted = shutdown(executor);
         logger.info("Stopped the worker threads");
 
+        try {
```
minwoox, line 690: Could you check whether we need to call `retryPolicy = RETRY_POLICY_NEVER`?

minwoox (approved):
👍 👍 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
```diff
@@ -680,11 +680,29 @@ private void stopLater() {
...
+                interrupted |= shutdown(logWatcherExecutor);
+                logger.info("Closed the log watcher");
+            }
+        } catch (Exception e) {
+            logger.warn("Failed to close the log watcher: {}", e.getMessage(), e);
+        }
+
         try {
             logger.info("Stopping the delegate command executor");
             delegate.stop();
```
minwoox: If we call, `delegate.stop();` in the synchronized block, will the issue be solved as well?

jrhee17 (approved):
👍 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/replication/ZooKeeperCommandExecutor.java
ikhoon: It sounds like a nice idea.

ikhoon: Fixed: https://github.com/line/centraldogma/pull/1347/commits/a96842f5c945e7f29762588bc616316899543df9

PR #1348, opened 2026-08-03
Persist xDS list pagination in the URL so it survives navigation

minwoox:
Motivation:

The xDS group list (and the per-group resource and history lists) kept their pagination only in TanStack Table's in-memory state.

Modifications:

- Add a reusable useUrlPagination hook (webapp/src/dogma/common/components/table/useUrlPagination.ts) that mirrors pageIndex/pageSize into the URL query (`page`, 1-indexed, and `pageSize`) with shallow routing, omits both params while at their defaults, and reads them back through a router.isReady-gated effect so the state is restored on mount and on Back/Forward. The valid page sizes are held in a ref so an inline `pageSizes` array cannot turn the sync effect into a render loop.

Result:

- xDS list pagination now survives navigation.

### DIFF
--- a/webapp/src/dogma/common/components/table/DataTableClientPagination.tsx
+++ b/webapp/src/dogma/common/components/table/DataTableClientPagination.tsx
@@ -6,27 +6,15 @@ import {
   getFilteredRowModel,
   getPaginationRowModel,
   getSortedRowModel,
-  PaginationState,
   SortingState,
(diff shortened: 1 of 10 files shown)


jrhee17 (approved):
👍 👍

ikhoon (approved):
👍 👍

PR #1354, opened 2026-08-11
Serve xDS resources that embed Armeria custom filter configs

minwoox:
Motivation:

The xDS control plane parses every stored Listener/Cluster/Route/Endpoint into a proto with a curioswitch MessageMarshaller, which must resolve every google.protobuf.Any it contains (such as an HttpFilter typed_config). The marshaller only registered types under io.envoyproxy.envoy (from java-control-plane) plus a couple of Central Dogma-specific ones. Armeria's xDS filters (e.g. the Athenz access-token filters) are carried as typed_config Any values whose types live in com.linecorp.armeria.xds.athenz and jp.co.lycorp.ftd.athenz.v1, packaged in armeria-xds-api, which Central Dogma did not depend on.

Modifications:
- Make armeria-xds-api the single provider of io.envoyproxy.envoy types in the :xds module.

Result:
- xDS resources that embed Armeria custom filter configs, such as the Athenz access-token filters, can now be created, stored and served by the control plane.


jrhee17 (approved):
👍👍

xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsResourceManager.java
```diff
@@ -77,24 +75,40 @@ public final class XdsResourceManager {
...
                 MessageMarshaller.builder().omittingInsignificantWhitespace(true);
         // These Central Dogma-specific types are not in the io.envoyproxy.envoy package, so they
-        // must be registered explicitly; envoyExtension() does not pick them up.
+        // must be registered explicitly; the io.envoyproxy.envoy scan does not pick them up.
         builder.register(KubernetesEndpointAggregator.getDefaultInstance())
                .register(LocalityLbEndpoint.getDefaultInstance());
-        envoyExtension(builder);
+        // Envoy types use java_multiple_files, so every message is a top-level class.
+        registerPackage(builder, "io.envoyproxy.envoy", false);
+        // The custom (non-Envoy) protos do not set java_multiple_files, so their messages are nested.
```
ikhoon, line 94: Question) Could you register protos regardless of the `java_multiple_files` setting?

ikhoon (approved):
👍 👍

xds/src/main/java/com/linecorp/centraldogma/xds/internal/XdsResourceManager.java
minwoox: Yeah, but it's efficient not to traverse the nested class if we're sure that the proto has `java_multiple_files`.

PR #1355, opened 2026-08-13
Remove JSON handling from xDS resource storage now that migration is complete

minwoox:
Motivation:

xDS resources were stored as either .json or .yaml while the format migration was in progress, so the code carried dual-format lookups and one-shot .json→.yaml migration commits. The migration is now finished and no .json files remain, leaving this logic dead and misleading.

Modifications:

- XdsResourceManager: drop alternativeFileName() and the dual-extension find; push/update/delete/updateOrDelete now operate solely on the .yaml file. Remove doPush's legacyFileToRemove parameter and change updateOrDelete's callback from Function<String, ...> to Supplier since the resolved filename is always the .yaml file.
- XdsEndpointReadService: read the .yaml file directly and drop the EntryType.JSON branch from the endpoint listing.
- XdsResourceWatchingService: remove the UPSERT_JSON case, the EntryType.JSON init filter, and the atomic-migration skip in the REMOVE case.
- XdsKubernetesEndpointFetchingService: remove the legacyJsonMigrated flag and the atomic-migration commit in pushK8sEndpoints; onFileRemoved now removes the .yaml endpoint file directly.
- XdsEndpointUpdateScheduler: flush() always takes the .yaml transform path; remove the legacy .json read-transform-migrate branch.

Result:
- xDS resources are handled exclusively as YAML. There is no user-facing behavior change; the removed code only executed against .json files, which no longer exist.


ikhoon (approved):
👍 👍

jrhee17 (approved):
👍 👍

PR #1358, opened 2026-08-14
Make the xDS all-resource version deterministic across replicas

minwoox:
Motivation:

A wildcard (empty resource_names) LDS/CDS request from an app-scoped client resolves its version_info through
CentralDogmaSnapshotResources.allResourceVersion(), which hashed versionedResources.values() in map-iteration order. That order is derived from HashMap iteration, so it is not stable: it differs between two maps that hold identical content but were built in a different order.

Because the xDS control plane runs on every replica (ALL_REPLICAS), each replica builds its snapshot from its own HashMaps and therefore hashes the same resources in a different order, producing a different version_info for byte-identical content.

Modifications:

- CentralDogmaSnapshotResources.allResourceVersion(): hash the resources in a canonical, resource-name-sorted order instead of in versionedResources.values() iteration order. This mirrors the name-sorted hashing already used for an explicit set of resource names in resourceVersionResolver.

Result:

- All replicas compute the same version_info for identical resource content, so wildcard (LDS/CDS) clients no longer see version_info flap when their stream moves between replicas.

### DIFF
--- a/xds/src/main/java/com/linecorp/centraldogma/xds/internal/CentralDogmaSnapshotResources.java
+++ b/xds/src/main/java/com/linecorp/centraldogma/xds/internal/CentralDogmaSnapshotResources.java
@@ -120,9 +120,15 @@ final class CentralDogmaSnapshotResources<T extends Message> extends SnapshotRes
         if (allResourceVersion != null) {
             return allResourceVersion;
         }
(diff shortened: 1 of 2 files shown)


jrhee17 (approved):
👍 👍

ikhoon (approved):
👍 👍

PR #1359, opened 2026-08-20
Let plugins create encrypted repositories with custom project roles

minwoox:
Motivation:

Server-side plugins that create repositories call `Command.createRepository()` directly and register the repository metadata in a separate step. That bypasses what `RepositoryServiceUtil.createRepository()` already does for the REST API: generating a wrapped data encryption key so the repository is encrypted at rest, and adding the metadata as part of the same call.

Modifications:

- Expose the server's `EncryptionStorageManager` through `PluginContext` via a new `encryptionStorageManager()` accessor.

Result:

- A plugin can now create a repository through `RepositoryServiceUtil.createRepository()` that is encrypted when the server has encryption at rest enabled, and with the project roles it chooses.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/api/RepositoryServiceUtil.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/api/RepositoryServiceUtil.java
@@ -30,6 +30,7 @@ import com.linecorp.centraldogma.common.Revision;
 import com.linecorp.centraldogma.server.command.Command;
 import com.linecorp.centraldogma.server.command.CommandExecutor;
 import com.linecorp.centraldogma.server.metadata.MetadataService;
+import com.linecorp.centraldogma.server.metadata.ProjectRoles;
 import com.linecorp.centraldogma.server.metadata.RepositoryMetadata;
 import com.linecorp.centraldogma.server.metadata.Roles;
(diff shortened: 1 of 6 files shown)


server/src/main/java/com/linecorp/centraldogma/server/internal/api/RepositoryServiceUtil.java
```diff
@@ -51,6 +52,15 @@ public static CompletableFuture<Revision> createRepository(
             CommandExecutor commandExecutor, MetadataService mds,
             Author author, String projectName, String repoName, boolean assignRoleToAuthor,
             boolean encrypt, @Nullable EncryptionStorageManager encryptionStorageManager) {
+        return createRepository(commandExecutor, mds, author, projectName, repoName, DEFAULT_PROJECT_ROLES,
+                                assignRoleToAuthor, encrypt, encryptionStorageManager);
+    }
+
+    public static CompletableFuture<Revision> createRepository(
```
ikhoon, line 59: > A plugin can now create a repository through RepositoryServiceUtil.createRepository()

Question) Did you intend to use `RepositoryServiceUtil` in a plugin? Since it is an internal API, code that directly calls its methods could break unexpectedly.

jrhee17 (approved):
👍 👍

PR #1360, opened 2026-08-20
Make xDS resource versions deterministic across replicas

minwoox:
Motivation:

A wildcard (empty resource_names) LDS/CDS request whose stream is load-balanced across replicas received a different version_info on essentially every request even when no resource had changed, forcing a needless re-push.

The version ultimately derives from protobuf Message.hashCode(): VersionedResource.create(resource) hashes resource.hashCode() into the per-resource version(), and CentralDogmaSnapshotResources hashed List<VersionedResource>.hashCode() (via AutoValue's resource().hashCode()) for the all-resource and multi-resource versions. But the generated Message.hashCode() folds in getDescriptor().hashCode(), and Descriptors.Descriptor does not override hashCode(), so it resolves to the identity hash of the descriptor singleton. That identity hash is constant within a JVM but differs across JVMs, so byte-identical content produces a different version on every replica.

Modifications:

- CentralDogmaXdsResources: compute each resource's version from the SHA-256 of its deterministically serialized bytes (CodedOutputStream.useDeterministicSerialization()) and pass it to VersionedResource.create(resource, version), instead of relying on the hashCode-based single-argument create(). Deterministic serialization is identical across JVMs and also normalizes protobuf map field ordering (e.g. metadata). Applied to clusters, endpoints, listeners and routes.

Result:

- All replicas compute the same version_info for identical resource content, so wildcard (and named) LDS/CDS clients no longer see version_info flap when their stream moves between replicas.

### DIFF
--- a/xds/src/main/java/com/linecorp/centraldogma/xds/internal/CentralDogmaSnapshotResources.java
+++ b/xds/src/main/java/com/linecorp/centraldogma/xds/internal/CentralDogmaSnapshotResources.java
@@ -18,13 +18,16 @@ package com.linecorp.centraldogma.xds.internal;
 
 import static com.google.common.collect.ImmutableList.toImmutableList;
 
+import java.nio.charset.StandardCharsets;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.stream.Stream;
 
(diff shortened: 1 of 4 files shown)


ikhoon (approved):
👍 👍

PR #1361, opened 2026-08-26
Ensure a mirror can only use a credential of its own repository or project

minwoox:
Motivation:

A mirror references a credential by name. The intended rule is that a mirror may only use a credential scoped to its own repository or to its project. However, this was only enforced when a mirror was created or updated through the dedicated mirroring REST API (in the MirrorRequest constructor). The resolution path performed no scope check, and the generic push API (HTTP content push and Thrift push) allowed mirror and credential files to be written directly to the meta repository, bypassing that validation entirely.

Modifications:

- MirrorConverter.convertToMirror: re-validate the credential scope against the mirror's project and localRepo when a mirror is built from stored config, so an out-of-scope credential is never used even if the config reached storage by some other path (defense-in-depth).
- ContentServiceV1.checkMetaRepoPush: reject any push to the dogma/meta repository that creates or modifies a mirror or credential file, directing callers to the dedicated mirroring/credential REST API. metadata.json remains allowed and other files remain rejected as before. Removed the now unreachable localRepo check and the imports/constant it required.

Result:

- A mirror can no longer use a credential that belongs to a different repository; it is limited to its own repository or its project.
- Mirror and credential files can only be managed through the mirroring and credential REST APIs.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
@@ -76,6 +76,7 @@ import com.linecorp.centraldogma.internal.api.v1.RevertRequest;
 import com.linecorp.centraldogma.internal.api.v1.WatchResultDto;
 import com.linecorp.centraldogma.server.command.Command;
 import com.linecorp.centraldogma.server.command.CommandExecutor;
+import com.linecorp.centraldogma.server.internal.admin.auth.AuthUtil;
 import com.linecorp.centraldogma.server.internal.api.auth.RequiresRepositoryRole;
 import com.linecorp.centraldogma.server.internal.api.converter.ChangesRequestConverter;
 import com.linecorp.centraldogma.server.internal.api.converter.CommitMessageRequestConverter;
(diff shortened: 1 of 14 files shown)


jrhee17 (approved):
👍 👍

server/src/main/java/com/linecorp/centraldogma/server/internal/api/ContentServiceV1.java
```diff
@@ -586,41 +580,22 @@ public <T> CompletableFuture<MergedEntryDto<T>> mergeFiles(
...
-            if (hasChangesOtherThanMetaRepoFiles) {
+            for (Change<?> change : changes) {
+                final String path = change.path();
+                if (METADATA_JSON.equals(path)) {
+                    continue;
+                }
+                // Mirror and credential files must be managed via the dedicated mirroring and credential
+                // REST APIs, which validate the input (e.g. the credential scope). Writing them directly
+                // with the push API bypasses that validation, so it is not allowed.
+                if (isMirrorOrCredentialFile(path)) {
```
ikhoon, line 591: Should we allow administrators to directly manipulate the files? It may be necessary to perform migration tasks or apply workarounds to handle bugs.

minwoox: Fixed. 😉

ikhoon (approved):
Looks good!

ikhoon:
FYI) CI builds are broken.

PR #1362, opened 2026-08-30
Add endsWith/startsWith/under factory methods to PathPattern

jejes323:
Motivation:

As pointed out in #659, creating a `PathPattern` that matches all files with a given extension or all files under a given directory currently requires the caller to hand-write a glob string (e.g. `PathPattern.of("/**/*.json")` or `PathPattern.of("/foo/bar/**")`). Common factory methods for these two everyday cases would make the API easier to discover and use correctly.

Modifications:

- Add `PathPattern.endsWith(String suffix)`, which builds a pattern matching files by extension at any depth (e.g. `PathPattern.endsWith("json")` and `PathPattern.endsWith(".json")` both produce `/**/*.json`).
- Add `PathPattern.startsWith(String prefix)`, which builds a pattern matching all files under the given directory (e.g. `PathPattern.startsWith("/foo/bar")` produces `/foo/bar/**`; a trailing slash on `prefix` is handled).
- Add `PathPattern.under(String directory)` as a more readable alias for `startsWith(String)`.
- Add unit tests for all three new methods in `DefaultPathPatternTest`.

This PR intentionally covers only the `endsWith`/`startsWith`/`under` factory methods requested in #659. The issue's last example (composing multiple patterns so that a file must match *all* of them, e.g. `/foo/bar/**/*.json`) would require the matching engine to support AND-composition in addition to the current OR-only semantics (see `PathPattern`'s class Javadoc and `PathPatternFilter`), which is a larger design change. I'd suggest tracking that separately if there's still interest — happy to help with it as a follow-up once the direction is confirmed.

Result:

- Closes part of #659 (the `endsWith`/`startsWith`/`under` factory methods; AND-composition and `builder()` are left for a follow-up discussion).
- `PathPattern` now provides convenient factory methods for the two most common patterns, without changing any existing behavior.

Test plan:

```
$ ./gradlew :common:test --tests "com.linecorp.centraldogma.common.DefaultPathPatternTest"
BUILD SUCCESSFUL
6 tests completed, 0 failed (3 new: endsWith(), startsWith(), under())

$ ./gradlew :common:checkstyleMain :common:checkstyleTest
BUILD SUCCESSFUL

$ ./gradlew :common:javadoc
BUILD SUCCESSFUL
```

### DIFF
--- a/common/src/main/java/com/linecorp/centraldogma/common/PathPattern.java
+++ b/common/src/main/java/com/linecorp/centraldogma/common/PathPattern.java
@@ -15,9 +15,12 @@
  */
 package com.linecorp.centraldogma.common;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.linecorp.centraldogma.common.DefaultPathPattern.ALL;
 import static com.linecorp.centraldogma.common.DefaultPathPattern.allPattern;
 import static java.util.Objects.requireNonNull;
+
+import java.util.regex.Pattern;
 
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Streams;
@@ -35,6 +38,11 @@ import com.google.common.collect.Streams;
  * </ul>
  */
 public interface PathPattern {
+
+    /**
+     * The pattern that a file extension must match; only alphanumeric characters are allowed.
+     */
+    Pattern EXTENSION_PATTERN = Pattern.compile("^[a-zA-Z0-9]+$");
 
     /**
      * Returns the path pattern that represents all files.
@@ -65,15 +73,17 @@ public interface PathPattern {
     /**
      * Creates a path pattern that matches the files whose extension is the specified {@code extension}.
      * A leading dot in {@code extension} is optional and is added automatically if missing.
+     * The {@code extension} must consist of alphanumeric characters only.
      * For example, {@code PathPattern.ofExtension("json")} matches all JSON files at any depth,
      * which is equivalent to <code>PathPattern.of("/&#42;&#42;/*.json")</code>.
      */
     static PathPattern ofExtension(String extension) {
         requireNonNull(extension, "extension");
-        if (extension.startsWith(".")) {
-            return of("/**/*" + extension);
-        }
-        return of("/**/*." + extension);
+        final String normalized = extension.startsWith(".") ? extension.substring(1) : extension;
+        checkArgument(EXTENSION_PATTERN.matcher(normalized).matches(),
+                      "extension: %s (expected: an alphanumeric extension such as \"json\" or \".json\")",
+                      extension);
+        return of("/**/*." + normalized);
     }
 
     /**
@@ -82,14 +92,14 @@ public interface PathPattern {
      * The match is not restricted to complete path segments; for example,
      * {@code PathPattern.startsWith("/foo/ba")} matches both {@code /foo/bar/a.txt} and {@code /foo/baz.txt},
      * which is equivalent to <code>PathPattern.of("/foo/ba&#42;&#42;")</code>.
+     * The {@code prefix} must not contain a wildcard character ({@code '*'}).
      * Use {@link #under(String)} to match only the files under a directory.
      */
     static PathPattern startsWith(String prefix) {
         requireNonNull(prefix, "prefix");
-        if (prefix.startsWith("/")) {
-            return of(prefix + "**");
-        }
-        return of('/' + prefix + "**");
+        checkArgument(prefix.indexOf('*') < 0, "prefix: %s (must not contain '*')", prefix);
+        final String normalized = prefix.startsWith("/") ? prefix : '/' + prefix;
+        return of(normalized + "**");
     }
 
     /**
@@ -99,9 +109,11 @@ public interface PathPattern {
      * complete path segments; for example, {@code PathPattern.under("/foo/bar")} matches {@code /foo/bar/a.txt}
      * but not {@code /foo/bar-baz.txt}, which is equivalent to
      * <code>PathPattern.of("/foo/bar/&#42;&#42;")</code>.
+     * The {@code directory} must not contain a wildcard character ({@code '*'}).
      */
     static PathPattern under(String directory) {
         requireNonNull(directory, "directory");
+        checkArgument(directory.indexOf('*') < 0, "directory: %s (must not contain '*')", directory);
         String dir = directory;
         if (!dir.startsWith("/")) {
             dir = '/' + dir;
--- a/common/src/main/java/com/linecorp/centraldogma/common/PathPattern.java
+++ b/common/src/main/java/com/linecorp/centraldogma/common/PathPattern.java
@@ -63,39 +63,53 @@ public interface PathPattern {
     }
 
     /**
-     * Creates a path pattern that matches the files whose extension is the specified {@code suffix}.
-     * A leading dot in {@code suffix} is optional and is added automatically if missing.
-     * For example, {@code PathPattern.endsWith("json")} matches all JSON files at any depth,
+     * Creates a path pattern that matches the files whose extension is the specified {@code extension}.
+     * A leading dot in {@code extension} is optional and is added automatically if missing.
+     * For example, {@code PathPattern.ofExtension("json")} matches all JSON files at any depth,
      * which is equivalent to <code>PathPattern.of("/&#42;&#42;/*.json")</code>.
      */
-    static PathPattern endsWith(String suffix) {
-        requireNonNull(suffix, "suffix");
-        if (suffix.startsWith(".")) {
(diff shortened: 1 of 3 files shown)


jejes323:
The `build-windows-latest-jdk-25` failure looks unrelated to this change. The only failing test is a Jest timeout in `webapp/tests/pages/app/settings/app-identities/index.test.tsx`:

> AppIdentityPage › hides the regenerate secret action once the secret is regenerated — thrown: "Exceeded timeout of 5000 ms for a test."

Locally this test (and the full `testWeb` suite) passes consistently in ~900ms, and all Linux/macOS jobs are green. This PR only adds `endsWith`/`startsWith`/`under` factory methods to `PathPattern` and doesn't touch the webapp.

Could you re-run the `build-windows-latest-jdk-25` job? Happy to bump that test's timeout in a separate PR if it keeps flaking on the Windows runner.

common/src/main/java/com/linecorp/centraldogma/common/PathPattern.java
```diff
@@ -62,6 +62,42 @@ static PathPattern of(Iterable<String> patterns) {
...
+        return of("/**/*." + suffix);
+    }
+
+    /**
+     * Creates a path pattern that matches the files under the specified {@code prefix} directory.
+     * For example, {@code PathPattern.startsWith("/foo/bar")} matches all files under
+     * the directory {@code /foo/bar}, which is equivalent to
+     * <code>PathPattern.of("/foo/bar/&#42;&#42;")</code>.
+     */
+    static PathPattern startsWith(String prefix) {
```
jrhee17, line 87: Question) Is the intention of this method to only filter paths that start with `prefix`?

If the path doesn't start with `/`, later on `/**/` will be appended to create `/**/{prefix}` which means prefix doesn't necessarily start at the root

common/src/main/java/com/linecorp/centraldogma/common/PathPattern.java
```diff
@@ -62,6 +62,42 @@ static PathPattern of(Iterable<String> patterns) {
         return new DefaultPathPattern(ImmutableSet.copyOf(patterns));
     }
 
+    /**
+     * Creates a path pattern that matches the files whose extension is the specified {@code suffix}.
+     * A leading dot in {@code suffix} is optional and is added automatically if missing.
+     * For example, {@code PathPattern.endsWith("json")} matches all JSON files at any depth,
+     * which is equivalent to <code>PathPattern.of("/&#42;&#42;/*.json")</code>.
+     */
+    static PathPattern endsWith(String suffix) {
```
jrhee17: This looks like a match on an extension instead of a suffix - perhaps `ofExtension` makes more sense

jejes323: You're right, this is really an extension match, not a general suffix match. I'll rename it to `ofExtension`. If a general suffix match (e.g. `foobar.txt` matching `bar.txt`) turns out to be useful, we can add it as a separate method later.

common/src/main/java/com/linecorp/centraldogma/common/PathPattern.java
jejes323: The intention is a root-relative prefix, so `startsWith("/foo/bar")` should only match `/foo/bar/**`. You're right that a non-`/`-prefixed input would become `/**/{prefix}` and match at any depth, which is not intended. I'll normalize the input to prepend a leading `/` if missing, and clarify this in the Javadoc.

ikhoon: I think there was some ambiguity in the original API design. We should clarify whether `startsWith()` is intended to match only complete path segments or also allow partial path matches. 

What do you think about making `startsWith()` allow partial path matches, while `under()` matches only complete path segments using `<prefix>/**`? This would give the two methods distinct semantics rather than making them behave the same way. 

For endsWith(), I think we can follow the approach suggested by @jrhee17.

jejes323: Sounds good, I'll go with that.

- **`ofExtension(ext)`** (renamed from `endsWith`): `/**/*.{ext}`, normalizes a leading `.`. A general suffix match can be added separately later if needed.
- **`startsWith(prefix)`**: allows partial path matches, i.e. `{prefix}**` (root-relative, no trailing `/`). So `startsWith("/foo/ba")` matches both `/foo/bar/x.txt` and `/foo/baz.txt`.
- **`under(dir)`**: complete path segments only, `{dir}/**`, normalizes a trailing `/`. So `under("/foo/bar")` matches `/foo/bar/x.txt` but not `/foo/bar-baz/x.txt`. No longer delegates to `startsWith`.

One thing to confirm: for `startsWith`, I'm assuming it should still be root-relative (`{prefix}**`), not match-anywhere (`/**/{prefix}**`), so it stays consistent with `under`. Let me know if you'd prefer match-anywhere.

I'll push the changes once this is confirmed.

jrhee17: > I'm assuming it should still be root-relative ({prefix}**), not match-anywhere (/**/{prefix}**)

The method name is `startsWith` - I think it's logical that the path should start with prefix (so root-relative)

common/src/main/java/com/linecorp/centraldogma/common/PathPattern.java
```diff
@@ -62,6 +62,56 @@ static PathPattern of(Iterable<String> patterns) {
...
     }
 
+    /**
+     * Creates a path pattern that matches the files whose extension is the specified {@code extension}.
+     * A leading dot in {@code extension} is optional and is added automatically if missing.
+     * For example, {@code PathPattern.ofExtension("json")} matches all JSON files at any depth,
+     * which is equivalent to <code>PathPattern.of("/&#42;&#42;/*.json")</code>.
+     */
+    static PathPattern ofExtension(String extension) {
+        requireNonNull(extension, "extension");
```
jrhee17, line 72: nit; it may be worth validating extension itself only has `[a-zA-Z0-9]` since it may be odd to have `ofExtension("/foo/bar")`

common/src/main/java/com/linecorp/centraldogma/common/PathPattern.java
```diff
@@ -62,6 +62,56 @@ static PathPattern of(Iterable<String> patterns) {
...
+
+    /**
+     * Creates a path pattern that matches the files under the specified {@code directory}.
+     * The {@code directory} is anchored at the root, so a leading slash is added automatically if missing,
+     * and a trailing slash is optional. Unlike {@link #startsWith(String)}, the match is restricted to
+     * complete path segments; for example, {@code PathPattern.under("/foo/bar")} matches {@code /foo/bar/a.txt}
+     * but not {@code /foo/bar-baz.txt}, which is equivalent to
+     * <code>PathPattern.of("/foo/bar/&#42;&#42;")</code>.
+     */
+    static PathPattern under(String directory) {
```
jrhee17, line 103: Question) What happens if `PathPattern.under("/a/**")`? Will the final path pattern be `/a/****`?
Given that this method is doing regex for the user already, I wonder if we should just validate s.t. asterisks are not included in the input.

Ditto for `startsWith`.

common/src/main/java/com/linecorp/centraldogma/common/PathPattern.java
```diff
@@ -36,6 +39,11 @@
  */
 public interface PathPattern {
 
+    /**
+     * The pattern that a file extension must match; only alphanumeric characters are allowed.
+     */
+    Pattern EXTENSION_PATTERN = Pattern.compile("^[a-zA-Z0-9]+$");
```
jrhee17: Can we move `EXTENSION_PATTERN` s.t. it is not publicly exposed? Perhaps `DefaultPathPattern`

common/src/main/java/com/linecorp/centraldogma/common/PathPattern.java
```diff
@@ -62,6 +70,60 @@ static PathPattern of(Iterable<String> patterns) {
...
     }
 
+    /**
+     * Creates a path pattern that matches the files whose extension is the specified {@code extension}.
+     * A leading dot in {@code extension} is optional and is added automatically if missing.
+     * The {@code extension} must consist of alphanumeric characters only.
+     * For example, {@code PathPattern.ofExtension("json")} matches all JSON files at any depth,
+     * which is equivalent to <code>PathPattern.of("/&#42;&#42;/*.json")</code>.
+     */
+    static PathPattern ofExtension(String extension) {
```
jrhee17, line 74: nit; can we check that extension is not empty either? ditto for the other methods

jrhee17 (approved):
👍 👍

minwoox (approved):
Thanks!

PR #1363, opened 2026-08-31
Do not fetch the remote objects when a remote-to-local mirror is up-to-date

ikhoon:
Motivation:

A remote-to-local Git mirror fetched the remote head before deciding whether it had anything to mirror. The fetch is a shallow depth-2 fetch, so every scheduled run of an up-to-date mirror downloaded the head commit's trees and blobs only to conclude that nothing had changed.

The decision never needed them: shouldRunRemoteToLocal() reads only the remote head's SHA, which the lsRemote() that resolves the head branch already advertises.

Modifications:

- Decide whether to run a remote-to-local mirror from the SHA advertised by lsRemote(), and fetch only when the mirror is going to run.
- Add RemoteToLocalMirrorFetchTest, which asserts that an up-to-date run issues only lsRemote() while a run that mirrors or compares contents also fetches.

Result:

- An up-to-date remote-to-local mirror no longer transfers the remote head's objects on every scheduled run. It issues a single ref advertisement instead.


jrhee17 (approved):
👍

minwoox (approved):
👍 👍 👍

PR #1366, opened 2026-09-08
Speed up the first file view by prefetching and compressing Monaco

minwoox:
Motivation:

Opening a file for the first time was slow because the multi-megabyte Monaco editor bundle was downloaded only when the editor first mounted, and the static assets were served uncompressed.

Modifications:

- Add a shared `loadMonaco()` singleton and prefetch Monaco (core chunk + language workers) during browser idle time via `useMonacoPrefetch`, wired into `_app.tsx` (skipped on the login page).

Result:

- The first file view opens without waiting on the Monaco download.

### DIFF
--- a/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
+++ b/server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
@@ -260,11 +260,10 @@ public class CentralDogma implements AutoCloseable {
 
     private static final int DEFAULT_MAX_FRAME_LENGTH = 1024 * 1024; // 1 MiB
 
-    // The largest pre-compressed web asset (e.g. the ~910 KiB Brotli-compressed Monaco TypeScript worker) is
-    // comfortably under 1 MiB, so we raise the FileService cache-entry limit above Armeria's 64 KiB default.
-    // Otherwise the heavy Monaco chunks and workers would exceed the default and be re-read from the classpath
-    // on every request instead of being served from the in-memory cache.
-    private static final int WEBAPP_MAX_CACHE_ENTRY_SIZE_BYTES = 1024 * 1024; // 1 MiB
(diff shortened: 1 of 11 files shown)


jrhee17 (approved):
👍 👍

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
```diff
@@ -260,6 +260,12 @@ public class CentralDogma implements AutoCloseable {
 
     private static final int DEFAULT_MAX_FRAME_LENGTH = 1024 * 1024; // 1 MiB
 
+    // The largest pre-compressed web asset (e.g. the ~910 KiB Brotli-compressed Monaco TypeScript worker) is
+    // comfortably under 1 MiB, so we raise the FileService cache-entry limit above Armeria's 64 KiB default.
+    // Otherwise the heavy Monaco chunks and workers would exceed the default and be re-read from the classpath
+    // on every request instead of being served from the in-memory cache.
+    private static final int WEBAPP_MAX_CACHE_ENTRY_SIZE_BYTES = 1024 * 1024; // 1 MiB
```
ikhoon: - What do you think of raising the value to 2 MiB to leave some headroom in case the Monaco TypeScript worker grows due to the upstream changes?
- Would you add a test to check whether a compressed file size exceeds this limitation so that we can re-evaluate WEBAPP_MAX_CACHE_ENTRY_SIZE_BYTES if needed?

ikhoon (approved):
Thanks. I left minor comments, but none of them are be blockers.

server/src/main/java/com/linecorp/centraldogma/server/CentralDogma.java
minwoox: Good suggestion. Fixed.

